PR #2, opened 2023-10-04
change denom from stake to upokt

okdas:


config.yml
```diff
@@ -57,12 +57,27 @@ client:
...
   app_state:
+    mint:
+      params:
+        mint_denom: upokt
+    staking:
+      params:
+        bond_denom: upokt
+    crisis:
+      constant_fee:
+          amount: '10000'
```
okdas, line 74: exact numbers might need to change

Olshansk (approved):

Issue #13, opened 2023-10-05
[Miner] Foundation for the Miner submodule


Olshansk:
## Objective

Create a foundation for the `Miner` submodule in the [poktroll](https://github.com/pokt-network/poktroll) repo for the Shannon upgrade.

## Origin Document

The business logic for mining relays from the [Relay Mining](https://arxiv.org/abs/2305.10672) paper.

## Goals

- Enable an E2E HTTP relay with basic session and account management
- Build the foundation for the `Miner` submodule for the `Relayer`
- Identify time-consuming TODOs to be done during the integration and testing stages

## Deliverables

- [ ] Reference structures and types in the [poktroll-alpha repo](https://github.com/pokt-network/poktroll-alpha), the [v1 repo](https://github.com/pokt-network/pocket) and [v0 repo](https://github.com/pokt-network/pocket-core) to understand and use other definitions as a foundation
- [ ] A **single PR** that introduces the `Miner` structure & interface
- [ ] A **single PR** that adds business logic, with the appropriate placeholders & TODOs, to add `Miner` functionality
- [ ] A **single PR** that integrates the miner with other components (e.g. `SessionManager`, `Proxy`, `Relayer`, `Client`, etc...) to enable E2E functionality
- [ ] A **single PR** that integrates with the [SMT](https://github.com/pokt-network/smt) for proof validation & verification
- [x] **Multiple small PRs** that migrate the necessary utilities from the `poktroll-alpha` repo to the `poktroll` repo

## Non-goals / Non-deliverables

- WebSocket support (HTTP only is okay)
- Accounting for all possible edge cases and situations

## General deliverables

- [ ] **Comments**: Add/update TODOs and comments alongside the source code so it is easier to follow.
- [ ] **Testing**: Add new tests (unit and/or E2E) to the test suite.
- [ ] **Makefile**: Add new targets to the Makefile to make the new functionality easier to use.
- [ ] **Documentation**: Update architectural or development READMEs; use [mermaid](https://mermaid-js.github.io/mermaid/) diagrams where appropriate.

---

**Creator**: @Olshansk 
**Co-Owners**: @bryanchriswhite @red-0ne

Olshansk:
@bryanchriswhite Looking at the deliverables of this ticket, I would argue it is not complete yet.

bryanchriswhite:
[...]

PR #31, opened 2023-10-09
[Pkg] feat: add channel observable

bryanchriswhite:
### Summary generated by Reviewpad on 20 Oct 23 08:51 UTC
This pull request contains several file diffs:

1. The diff of the file `observer_test.go` includes the addition of two test functions, `TestObserver_Unsubscribe` and `TestObserver_ConcurrentUnsubscribe`. These test the behavior of the `Unsubscribe` method in the `channelObserver` struct.

2. The diff of the file "observer.go" includes the addition of a new package named "channel", several imports, constants, types, functions, and methods related to an implementation of an observer for channels in the "observable" package.

3. The diff for the file `README.md` includes the addition of comprehensive documentation for the package, including architecture diagrams, installation instructions, features, usage examples, configuration options, API reference, best practices, FAQ, contributing guidelines, changelog, and license information.

4. A new file named "require.go" has been added to the internal/testerrors package. This file contains a package declaration, an import statement, and the definition of a variable named ErrAsync. This variable is used for handling test assertion failures in separate goroutines.

5. The diff adds a new file named "drain.go" to the "internal/testchannel" package. The file contains a function named "DrainChannel" that attempts to receive from a given channel until it is empty. If the channel is not closed by the time it's empty, the function returns an error. Additionally, a TODO_CONSIDERATION comment is included, suggesting the addition of a timeout parameter to the function.

6. The diff shows the addition of a new file `README.md` in the `docs/pkg/observable` directory. This file provides comprehensive documentation for the `pkg/observable` package, including an overview, functionality descriptions, architecture diagrams, usage examples, and considerations for developers.

7. The diff introduces a new file called "errors.go" in the "pkg/observable" directory, which defines a package called "observable" with two global variables and imports another package called "errorsmod" from the "cosmossdk.io/errors" module.

8. A new file called `observable.go` has been added to the `pkg/observable/channel` directory. This file implements the `observable.Observable` interface and provides an implementation for creating and managing observables using channels. It includes functions for creating observables, subscribing and unsubscribing observers, and notifying observers when values are received.

9. The diff in go.mod includes changes related to dependencies, such as adding a new dependency and removing an indirect dependency.

10. The diff introduces a new file called "interface.go" in the "pkg/observable" package, which defines two generic interfaces, `Observable[V]` and `Observer[V]`, for implementing a notifications package.

11. A new file named `observation_test.go` has been added in the `pkg/observable/channel` package. This file contains helper functions and data structures related to observations in the context of channels.

This pull request includes various additions and changes to different files and packages, including tests, documentation, implementation of observers, error handling, and dependency management.

(NOTE: following [the standard go project layout guidelines](https://github.com/golang-standards/project-layout))

## Issue

Related to:
- #13 
- #65 
- #42 

`Observable` is a dependency of the `ServicerClient`, which is a dependency of the `Miner`.

## Type of change

Select one or more:

- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

## Sanity Checklist


pkg/observable/notifiable/observable.go
```diff
@@ -0,0 +1,93 @@
...
+		for _, ch := range obs.subscribers {
+			ch <- v
+		}
+		obs.mu.RUnlock()
+	}
+
+	// Here we know that the notifier has been closed, all subscribers should be closed as well
+	obs.mu.Lock()
+	obs.closed = true
+	for _, ch := range obs.subscribers {
```
red-0ne, line 79: Just noticed that tat this point we are closing subscribers channel but the derived subscriptions should also close properly. They should set their `.closed` flag to false.

It would need storing the subscriptions instead of the channel and call `sub.Unsubscribe` for each of them. WDYT?

pkg/observable/notifiable/observable.go
```diff
@@ -0,0 +1,93 @@
...
+
+	return subscription
+}
+
+// listen to the notifier and notify subscribers when values are received. This
+// function is blocking and should be run in a goroutine.
+func (obs *Observable[V]) listen(notifier <-chan V) {
+	for v := range notifier {
+		// Lock for obs.subscribers slice as it can be modified by subscribers
+		obs.mu.RLock()
```
Olshansk, line 69: NIT: I think we can lock & unlock arpimd the forloop and keep this atomic.

pkg/observable/notifiable/observable.go
```diff
@@ -0,0 +1,93 @@
...
+		for _, ch := range obs.subscribers {
+			ch <- v
+		}
+		obs.mu.RUnlock()
+	}
+
+	// Here we know that the notifier has been closed, all subscribers should be closed as well
+	obs.mu.Lock()
+	obs.closed = true
```
Olshansk, line 78: See my earlier comment around making things simpler by not having this variable.

pkg/observable/notifiable/observable.go
```diff
@@ -0,0 +1,93 @@
...
+	for _, ch := range obs.subscribers {
+		close(ch)
+		obs.subscribers = []chan V{}
```
Olshansk, line 81: This should be done outside the forloop.

pkg/observable/notifiable/observable.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+	obs.mu.Lock()
```
Olshansk, line 83: 1. Unlock
2. Easier to just do a defer unclok where you do the lock above

pkg/observable/notifiable/observable.go
```diff
@@ -0,0 +1,93 @@
...
+				obs.subscribers = append(obs.subscribers[:i], obs.subscribers[i+1:]...)
+				break
+			}
+		}
+	}
+
+	// Subscription gets its closed state from the observable
+	subscription := &Subscription[V]{ch, obs.closed, removeFromObservable}
+
+	go unsubscribeOnDone[V](ctx, subscription)
```
Olshansk, line 59:
```suggestion
	// Asynchronously wait for the context to close to unsubscribe
	go unsubscribeOnDone[V](ctx, subscription)
```

pkg/observable/notifiable/observable_test.go
```diff
@@ -0,0 +1,162 @@
...
+)
+
+func TestNewNotifiableObservable(t *testing.T) {
+	type test struct {
+		name     string
+		notifier chan int
+	}
+
+	input := 123
+	nonEmptyBufferedNotifier := make(chan int, 1)
```
Olshansk, line 28:
```suggestion
	notifierEventBufferSize := 1
	nonEmptyBufferedNotifier := make(chan int, notifierEventBufferSize)
```

Olshansk (changes requested):

pkg/observable/notifiable/observable_test.go
```diff
@@ -0,0 +1,162 @@
...
+			require.NotNil(t, notifier)
+
+			// construct 3 distinct subscriptions, each with its own channel
+			subscriptions := make([]observable.Subscription[int], 3)
+			for i := range subscriptions {
+				subscriptions[i] = notifee.Subscribe(ctx)
+			}
+
+			group := errgroup.Group{}
+			notifiedOrTimedOut := func(subscriptionCh <-chan int) func() error {
```
Olshansk, line 52: OPTIONAL: consider passing in a subscription struct instead of the underlying channel.

pkg/observable/notifiable/observable_test.go
```diff
@@ -0,0 +1,162 @@
...
+	"pocket/pkg/observable"
+	"pocket/pkg/observable/notifiable"
+)
+
+const (
+	notifyTimeout            = 100 * time.Millisecond
+	unsubscribeSleepDuration = notifyTimeout * 2
+)
+
+func TestNewNotifiableObservable(t *testing.T) {
```
Olshansk, line 21:
```suggestion
func TestObservable_NotifySubscribers(t *testing.T) {
```

pkg/observable/notifiable/observable_test.go
```diff
@@ -0,0 +1,162 @@
...
+				case <-subscription.Ch():
+				default:
+					t.Fatal("subscription channel left open")
+				}
+			}
+		})
+	}
+}
+
+func TestSubscription_Unsubscribe(t *testing.T) {
```
Olshansk, line 93:
```suggestion
func TestObservable_UnsubscribeSubscribers(t *testing.T) {
```

pkg/observable/notifiable/observable_test.go
```diff
@@ -0,0 +1,162 @@
...
+	ctx, cancel := context.WithCancel(context.Background())
+	notifee, notifier := notifiable.NewObservable[int](nil)
+	require.NotNil(t, notifee)
+	require.NotNil(t, notifier)
+
+	tests := []struct {
```
Olshansk, line 99: NIT: You're inline the struct type and initialization here but separating the two above. I think we should pick one (either) and be consistent.

My personal preference is to separate the two but don't feel strongly either way.

Olshansk (changes requested):

pkg/observable/interface.go
```diff
@@ -0,0 +1,16 @@
+package observable
+
```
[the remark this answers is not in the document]
bryanchriswhite: @Olshansk could you elaborate on how you see nats factoring in?

pkg/observable/notifiable/observable.go
```diff
@@ -0,0 +1,93 @@
...
+		close(ch)
+		obs.subscribers = []chan V{}
+	}
+	obs.mu.Lock()
+}
+
+// unsubscribeOnDone unsubscribes from the subscription when the context is.
+// It is blocking and intended to be called in a goroutine.
+func unsubscribeOnDone[V any](ctx context.Context, subscription observable.Subscription[V]) {
+	if ctx != nil {
```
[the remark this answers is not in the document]
red-0ne: If `ctx` is `nil` then we don't have any `context` based cancellation. Which means that cancellation could only be done manually by the caller. ie. If we don't pass `ctx` the subscription is still valid and the caller wants to handle unsubscribtions by himself

pkg/observable/notifiable/observable.go
```diff
@@ -0,0 +1,93 @@
...
+	"pocket/pkg/observable"
+)
+
+// Observable implements the observable.Observable interface and can be notified
+// via its corresponding notifier channel.
+type Observable[V any] struct {
+	mu          sync.RWMutex
+	ch          <-chan V // private channel that is used to emit values to subscribers
+	subscribers []chan V // subscribers is a list of channels that are subscribed to the observable
+	closed      bool
```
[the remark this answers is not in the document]
bryanchriswhite: > What are your thoughts on implementing a function like this:
> 
> ```go
> func (o *Observable[v]) isClosed() bool {
> 	_, ok := <-o.ch
> 	return !ok
> }
> ```

I think as you've written it, it would deadlock if the channel never receives after `#isClosed()` is called.

We considered something similar to this which accounted for that case (i.e. the channel is still open and blocking); however, at the time we somehow missed the fact that `ok` would be false when the channel is closed 😅. Tl;dr, I think it should look like this:

```go
func (o *Observable[v]) isClosed() bool {
  select {
    case _, ok := <-o.ch:
      return !ok
    default:
      return false
}
```

pkg/observable/notifiable/observable.go
```diff
@@ -0,0 +1,93 @@
...
+	for _, ch := range obs.subscribers {
+		close(ch)
+		obs.subscribers = []chan V{}
+	}
+	obs.mu.Lock()
+}
+
+// unsubscribeOnDone unsubscribes from the subscription when the context is.
+// It is blocking and intended to be called in a goroutine.
+func unsubscribeOnDone[V any](ctx context.Context, subscription observable.Subscription[V]) {
```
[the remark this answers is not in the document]
bryanchriswhite: We considered/discussed this and decided not to since the functions in question are only being used internally by this package, i.e. not exported. I don't have a strong opinion but wanted  to add that context. Do you think we should add the prefix anyways?

pkg/observable/interface.go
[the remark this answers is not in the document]
Olshansk: Goal was to provide an example. Nats was just the first one that came to mind.

pkg/observable/notifiable/observable.go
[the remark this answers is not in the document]
Olshansk: Cool. So do you think we should go with the `isClosed` function approach?

pkg/observable/notifiable/observable.go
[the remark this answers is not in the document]
Olshansk: Yes, I think we should.

`goFoo` for internal functions and `GoFoo` for external ones.

pkg/observable/notifiable/observable.go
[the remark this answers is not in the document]
Olshansk: Let's add a comment about this in the code.

pkg/observable/notifiable/observable.go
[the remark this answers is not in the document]
bryanchriswhite: I tried really hard but was not able to make this work. The issue I was unable to resolve was the fact that `isClosed()` would consume a value from the channel in the case where it's not closed. After some refactoring though I think I've minimized the amount of state and references  thereto.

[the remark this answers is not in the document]
red-0ne: It seems that `select` as it is a non blocking receive would do the trick. If the channel is closed it will always return. Which makes me think that this would suffice
```go
func (o *Observable[v]) isClosed() bool {
    select {
    case <-o.ch:
        return true
    default:
    }
    return false
}
```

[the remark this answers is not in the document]
Olshansk: Agreed. Let's drop this and leave as is.

pkg/observable/notifiable/observable.go
```diff
@@ -0,0 +1,93 @@
...
+	subscription := &Subscription[V]{ch, obs.closed, removeFromObservable}
+
+	go unsubscribeOnDone[V](ctx, subscription)
+
+	return subscription
+}
+
+// listen to the notifier and notify subscribers when values are received. This
+// function is blocking and should be run in a goroutine.
+func (obs *Observable[V]) listen(notifier <-chan V) {
```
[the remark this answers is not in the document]
bryanchriswhite: I gave it some thought and gave renaming `Subscription` to `Observer` a try. I also had [this conversation w/ chatGPT](https://chat.openai.com/share/a76baee4-537e-4455-8404-2ad0e58163db) and am now considering doing the following as well:

1. rename `notifier` to `producer`
2. rename the `notifier` package to `channel`
3. rename `notifiableObservable` to `channelObservable`
4. rename any references called `notifee` to `observable` (or something similar)

pkg/observable/channel/observable.go
```diff
@@ -0,0 +1,148 @@
...
+// goUnsubscribeOnDone unsubscribes from the subscription when the context is.
+// It is blocking and intended to be called in a goroutine.
+func goUnsubscribeOnDone[V any](ctx context.Context, subscription observable.Observer[V]) {
+	<-ctx.Done()
+	subscription.Unsubscribe()
+}
+
+// onUnsubscribeFactory returns a function that removes a given channelObserver from the
+// observable's list of observers.
+func (obsvbl *channelObservable[V]) onUnsubscribeFactory() UnsubscribeFunc[V] {
```
bryanchriswhite, line 135: Just now noticing that this doesn't need to be a factory, it can just be the function that it currently returns.

pkg/observable/channel/observable.go
```diff
@@ -0,0 +1,148 @@
...
+
+	observer := NewObserver[V](ctx, obsvbl.onUnsubscribeFactory)
+	obsvbl.observers = append(obsvbl.observers, observer)
+
+	// caller can rely on context cancellation or call Close() to unsubscribe
+	// active observers
+	if ctx != nil {
+		// asynchronously wait for the context to close and unsubscribe
+		go goUnsubscribeOnDone[V](ctx, observer)
+	}
```
bryanchriswhite, line 76: This can be moved into `NewObservable()`. Since `channelObservable` has a reference to its `observers`, we can create a single go routine in a higher scope and iterate over the observers to unsubscribe them all at once.

pkg/observable/channel/observable.go
```diff
@@ -0,0 +1,148 @@
...
+// CONSIDERATION: decide whether this should close the producer channel; perhaps
+// only if it was provided.
+func (obsvbl *channelObservable[V]) close() {
+	obsvbl.observersMu.Lock()
+	defer obsvbl.observersMu.Unlock()
+
+	observers := obsvbl.observers
+
+	for _, obsvr := range observers {
+		fmt.Printf("channelObservable#goListen: unsubscribing %p\n", obsvr)
```
bryanchriswhite, line 93: :sweat_smile: still need to remove these print statements

red-0ne (approved):
Left some comments. LGTM :+1:

pkg/observable/channel/observer.go
```diff
@@ -0,0 +1,101 @@
...
+		observerCh:    make(chan V, observerBufferSize),
+		onUnsubscribe: onUnsubscribeFactory(),
+	}
+}
+
+// Unsubscribe closes the subscription channel and removes the subscription from
+// the observable.
+func (obsvr *channelObserver[V]) Unsubscribe() {
+	obsvr.observerMu.Lock()
+	defer func() {
```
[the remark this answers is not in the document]
bryanchriswhite: Good catch, I was debugging

pkg/observable/channel/observer.go
```diff
@@ -0,0 +1,101 @@
...
+	return obsvr.observerCh
+}
+
+// notify is used called by observable to send on the observer channel. Can't
+// use channelObserver#Ch because it's receive-only.
+func (obsvr *channelObserver[V]) notify(value V) {
+	obsvr.observerMu.Lock()
+	defer obsvr.observerMu.Unlock()
+
+	if obsvr.closed {
```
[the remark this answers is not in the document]
bryanchriswhite: Correct

pkg/observable/notifiable/observable.go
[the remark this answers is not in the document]
Olshansk: Are you still contemplating these? I'd personally be in support of not having the `notifier`/`notifiee` naming convention as it's too generic IMO but don't want to block things too much.

Also, read through your thread, for future reference, consider just giving it the code rather than explaining the types & names you have:  https://chat.openai.com/share/d901a206-f152-4230-9c85-59ec08cde8fb

Olshansk (changes requested):
Going to re-review once self-comments are addressed (just marking for now).

bryanchriswhite:
> Going to re-review once self-comments are addressed (just marking for now).

:+1: sounds good. :pray: There will be a few more changes that are coming from the debugging / refactoring I was doing when working on the query client after all. :sweat_smile: A preview of that is currently available in the last commit of #64 (https://github.com/pokt-network/poktroll/pull/64/commits/97432bb3b57d298b168adca25e93cb6cf0fec33e).

pkg/observable/channel/observable.go
bryanchriswhite: I decided against this after all because a consequence of doing so would be either coupling all observers to the same context, which seems wrong, or  having an optional context in both  the constructor function and  the `#Subscribe()` method, which seems unnecessarily complex IMHO.

pkg/observable/channel/observer_test.go
```diff
@@ -0,0 +1,80 @@
...
+
+func TestObserver_ConcurrentUnsubscribe(t *testing.T) {
+	var (
+		onUnsubscribeCalled = false
+		inputCh             = make(chan int, 1)
+	)
+	obsvr := &channelObserver[int]{
+		ctx:        context.Background(),
+		observerMu: &sync.RWMutex{},
+		// using a buffered  channel to keep the test synchronous
```
Olshansk, line 46:
```suggestion
		// using a buffered channel to keep the test synchronous
```

pkg/observable/channel/observer_test.go
```diff
@@ -0,0 +1,80 @@
...
+	go func() {
+		for inputIdx := 0; ; inputIdx++ {
+			select {
+			case <-done:
+				return
+			default:
+			}
+
+			obsvr.notify(inputIdx)
+			//time.Sleep(50 * time.Millisecond)
```
Olshansk, line 66: Remove?

pkg/observable/channel/observer_test.go
```diff
@@ -0,0 +1,80 @@
...
+	done := make(chan struct{}, 1)
+	go func() {
+		for inputIdx := 0; ; inputIdx++ {
+			select {
+			case <-done:
+				return
+			default:
+			}
+
+			obsvr.notify(inputIdx)
```
Olshansk, line 65: It's not really clear to me how this idx value is used. Could you elaborate? #PUC

pkg/observable/channel/observation_test.go
```diff
@@ -0,0 +1,37 @@
+package channel_test
+
```
Olshansk, line 2: 1. Add a manual comment that this is a helper file for observations
2. You can copy-pasta this whole file into chatgpt and get free inline comments ;)

pkg/observable/channel/observable_test.go
```diff
@@ -0,0 +1,366 @@
...
+	type test struct {
+		name            string
+		producer        chan *int
+		inputs          []int
+		expectedOutputs []int
+		setupFn         func(t test)
+	}
+
+	inputs := []int{123, 456, 789}
+	// NB: see INCOMPLETE comment below
```
Olshansk, line 34:
```suggestion
	// NB: see TODO_INCOMPLETE comment below
```

pkg/observable/channel/observable_test.go
```diff
@@ -0,0 +1,366 @@
...
+			name:            "empty buffered len 1 producer",
+			producer:        make(chan *int, 1),
+			inputs:          inputs,
+			expectedOutputs: inputs,
+		},
+		// INCOMPLETE: producer channels which are full are proving harder to test
+		// robustly (no flakiness); perhaps it has to do with the lack of some
+		// kind of guarantee about the receiver order on the consumer side.
+		//
+		// The following scenarios should generally pass but are flaky:
```
Olshansk, line 61: Can you add more details (screenshot is fine) on what the error is?

pkg/observable/channel/observable_test.go
```diff
@@ -0,0 +1,366 @@
...
+			observers := make([]observable.Observer[*int], 1)
+			for i := range observers {
+				observers[i] = obsvbl.Subscribe(ctx)
+			}
+
+			group, ctx := errgroup.WithContext(ctx)
+
+			// ensure all obsvr channels are notified
+			for obsvrIdx, obsvr := range observers {
+				next := func(outputIndex int, output *int) error {
```
Olshansk, line 113: #PUC on what this annonymous function does

pkg/observable/channel/observable_test.go
```diff
@@ -0,0 +1,366 @@
...
+						t, tt.expectedOutputs[outputIndex],
+						*output,
+						"obsvr Idx: %d", obsvrIdx,
+					) {
+						return fmt.Errorf("unexpected output")
+					}
+					return nil
+				}
+
+				done := func(outputs []*int) error {
```
Olshansk, line 125: Ditto

pkg/observable/channel/observable_test.go
```diff
@@ -0,0 +1,366 @@
...
+
+			// construct 3 distinct observers, each with its own channel
+			observers := make([]observable.Observer[*int], 1)
+			for i := range observers {
+				observers[i] = obsvbl.Subscribe(ctx)
+			}
+
+			group, ctx := errgroup.WithContext(ctx)
+
+			// ensure all obsvr channels are notified
```
Olshansk, line 111:
```suggestion
			// ensure all obsvr channels are notified by starting a lot of asynchronous listeners error/success listeners
```

pkg/observable/channel/observable_test.go
```diff
@@ -0,0 +1,366 @@
...
+					(*obsrvn.Notifications)[notificationIdx],
+					"allExpected: %+v, allActual: %+v",
+					expectedNotifications[obsnIdx], *obsrvn.Notifications,
+				)
+			}
+		})
+	}
+}
+
+// TECHDEBT/INCOMPLETE: add coverage for active observers closing when producer closes.
```
Olshansk, line 323:
```suggestion
// TODO_TECHDEBT/TODO_INCOMPLETE: add coverage for active observers closing when the producer closes.
```

pkg/observable/channel/observable_test.go
```diff
@@ -0,0 +1,366 @@
...
+func TestChannelObservable_ObserversCloseOnProducerClose(t *testing.T) {
+	t.Skip("add coverage: all observers should close when producer closes")
+}
+
+func produceWithDelay[V any](producer chan<- V, delay time.Duration) func(value V) {
+	return func(value V) {
+		time.Sleep(delay / 2)
```
Olshansk, line 330: Why do we delay before and after rather than just the full period before? #PUC

pkg/observable/channel/observable_test.go
```diff
@@ -0,0 +1,366 @@
...
+		producer <- value
+		time.Sleep(delay / 2)
+	}
+}
+
+func goNotifiedOrTimedOutFactory[V any](
```
Olshansk, line 336: NIT: Opportunity to get chatgpt to provide some comments.

pkg/observable/channel/observable_test.go
```diff
@@ -0,0 +1,366 @@
...
+				group.Go(goNotifiedOrTimedOutFactory(obsvr, next, done, notifyTimeout))
+			}
+
+			// notify with test input
+			for _, input := range tt.inputs {
+				inputPtr := new(int)
+				*inputPtr = input
+
+				// simulating IO delay in sequential message production
+				produce(inputPtr)
```
Olshansk, line 147: Where is this function defined?

Olshansk, line 147: OPTIONAL NIT
```
inputCopy := input
produce(&inputCopy)
```

pkg/observable/channel/observable_test.go
```diff
@@ -0,0 +1,366 @@
...
+
+			// unsubscribing should close obsvr channel(s)
+			for _, observer := range observers {
+				observer.Unsubscribe()
+
+				// must drain the channel first to ensure it is closed
+				err := testchannel.DrainChannel(observer.Ch())
+				require.NoError(t, err)
+			}
+		})
```
Olshansk, line 163: Should we add a check that the observable is closed?

pkg/observable/channel/observable_test.go
```diff
@@ -0,0 +1,366 @@
...
+				require.Empty(t, value)
+				require.False(t, ok)
+			case <-time.After(notifyTimeout):
+				t.Fatal("observer channel left open")
+			}
+		})
+	}
+}
+
+func TestChannelObservable_ConcurrentSubUnSub(t *testing.T) {
```
Olshansk, line 240:
```suggestion
// TODO_INCOMPLETE/TODO_TECHDEBT: Implement `TestChannelObservable_ConcurrentSubUnSub`
func TestChannelObservable_ConcurrentSubUnSub(t *testing.T) {
```

Olshansk (changes requested):
Few things:

1. Great work! This is really difficult work but it's really clear and easy to follow. 🔥 

2. The current state of things (other than comments almost non of which are blocking) is g2g so we can hopefully merge it in tomorrow. ✅ 

3. We need to focus on functionality before tending to the TODOs here, but I am worried that it might come back to bite us, so please make sure to prioritize this afterwards. ⚠️

pkg/observable/channel/observer.go
```diff
@@ -0,0 +1,125 @@
...
+		observerMu:    new(sync.RWMutex),
+		observerCh:    make(chan V, observerBufferSize),
+		onUnsubscribe: onUnsubscribe,
+	}
+}
+
+// Unsubscribe closes the subscription channel and removes the subscription from
+// the observable.
+func (obsvr *channelObserver[V]) Unsubscribe() {
+	obsvr.unsubscribe()
```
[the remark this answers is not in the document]
bryanchriswhite: When the observer unsubscribes, the observable it's subscribed to needs to update it's `observers` list to remove  this observer. The function below encapsulate that logic using a closure from a scope where that list (and its mutex) are convenient to access. I think the only alternative would be to give each observer an explicit reference to the variables in the closure scope but that seems even more indirect and less readable, in my opinion.

pkg/observable/channel/observable.go
```diff
@@ -0,0 +1,157 @@
...
+	obsvbl.observersMu.Unlock()
+}
+
+// goProduce to the producer and notify observers when values are received. This
+// function is blocking and should be run in a goroutine.
+func (obsvbl *channelObservable[V]) goProduce(producer <-chan V) {
+	var observers []*channelObserver[V]
+	for notification := range producer {
+		for {
+			if !obsvbl.observersMu.TryRLock() {
```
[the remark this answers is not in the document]
bryanchriswhite: This `continue` belongs to a `for` inside the range over `producer` (which you commented on above). This inner loop only `breaks` in the branch where we were able to acquire the lock. Hopefully it becomes more obvious after adding some comments. :sweat_smile:

pkg/observable/interface.go
```diff
@@ -0,0 +1,22 @@
...
```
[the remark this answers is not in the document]
bryanchriswhite: Sneak peek:

image

pkg/observable/channel/observer.go
```diff
@@ -0,0 +1,125 @@
...
+
+	// if observerCh is full, release the lock and try again every sendRetryInterval.
+	sendRetryTicker := time.NewTicker(sendRetryInterval / 2)
+	for {
+		// observerMu must remain read-locked until the value is sent on observerCh
+		if !obsvr.observerMu.TryRLock() {
+			time.Sleep(sendRetryInterval / 2)
+			continue
+		}
+		if obsvr.closed {
```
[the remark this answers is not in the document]
bryanchriswhite: We can't, `obsvr.closed` is one of the  things protected by `obsvr.observerMu`, hence the `TryRLock()` call immediately prior.

pkg/observable/channel/observer_test.go
```diff
@@ -0,0 +1,80 @@
...
+)
+
+func TestObserver_Unsubscribe(t *testing.T) {
+	var (
+		onUnsubscribeCalled = false
+		inputCh             = make(chan int, 1)
+	)
+	obsvr := &channelObserver[int]{
+		observerMu: &sync.RWMutex{},
+		// using a buffered  channel to keep the test synchronous
```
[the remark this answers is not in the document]
bryanchriswhite: It looks like it's about time to replace another keyswitch. :sweat_smile:

pkg/observable/channel/observer.go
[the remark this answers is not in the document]
bryanchriswhite: (see https://github.com/pokt-network/poktroll/pull/31#discussion_r1365171033)

pkg/observable/channel/observer.go
```diff
@@ -0,0 +1,125 @@
...
+	onUnsubscribe func(toRemove *channelObserver[V])
+	// observerMu protects the observerCh and closed fields.
+	observerMu *sync.RWMutex
+	// observerCh is the channel that is used to emit values to the observer.
+	// I.e. on the "N" side of the 1:N relationship between observable and
+	// observer.
+	observerCh chan V
+	// closed indicates whether the observer has been closed. It's set in
+	// unsubscribe; closed observers can't be reused.
+	closed bool
```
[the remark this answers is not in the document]
bryanchriswhite: No reason, thanks for the reminder! :raised_hands:

pkg/observable/channel/observer.go
```diff
@@ -0,0 +1,125 @@
...
+
+	"pocket/pkg/observable"
+)
+
+const (
+	// DISCUSS: what should this be? should it be configurable? It seems to be most
+	// relevant in the context of the behavior of the observable when it has multiple
+	// observers which consume at different rates.
+	// observerBufferSize is the buffer size of a channelObserver's channel.
+	observerBufferSize = 1
```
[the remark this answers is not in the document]
bryanchriswhite: No reason, I :100: agree. I'll try to add some benchmarks to make some kind of informed decision.

pkg/observable/channel/observer.go
```diff
@@ -0,0 +1,125 @@
...
+		}
+
+		select {
+		case <-obsvr.ctx.Done():
+			// if the context is done just release the read-lock (deferred)
+			return
+		case obsvr.observerCh <- value:
+			return
+		case <-sendRetryTicker.C:
+			// CONSIDERATION: repurpose this retry loop into a default path which
```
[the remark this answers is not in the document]
bryanchriswhite: **TL;DR, removing these TODO comments**

It should say `case`.

Thinking about this (and the TODO that follows) again, in the context of the discussion in https://github.com/pokt-network/poktroll/pull/31#discussion_r1364526724, I think that this would be redundant, less idiomatic, and worse for maintenance and debugging than just using the bult-in buffer of the channel. :sweat_smile:

In addition, perhaps it would be prudent to add a logger call in the case of a retry (`sendRetryTicker.C` receives). I think that would make the logger output indicative of  a buffer hitting capacity.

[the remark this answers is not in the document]
bryanchriswhite: I think what I was attempting to articulate here (poorly :sweat_smile:) was a backpressure strategy which would "drain the channel" in the sense that it could receive a value from the publisher and throw it away:

- This would only happen after the channel's buffer  is completely full.
- Will result in the last `n` notifications being available for consumption upon resuming
- It would not interfere with the channel consumer resuming, i.e.: consumer can catch up as fast as it wants and/or even drain the channel completely

There are other strategies of course,  this is just the one that I thought of at the time that seemed like it could be a reasonable choice for a fail-safe mode.

pkg/observable/channel/observer.go
```diff
@@ -0,0 +1,125 @@
...
+			// buffers values so that the producer doesn't block and other observers
+			// can still be notified.
+			// TECHDEBT: add some logic to drain the buffer at some appropriate time
```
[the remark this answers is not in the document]
bryanchriswhite: > Are we sure we want to do this? It would mean this is an "at most once fan-out service"

I'm not sure I followed what you mean by "at most once". Are you referring to the number of subscribers which can receive a given value (A), the number of times any subscriber can receive a given value (B), or something else? 

A) Draining here would only affect this observer's (a single, slow, subscriber's) view of the stream of values. 

B) Subscribers are already only able to read a given value once as once it's received from the channel (regardless of that channel's buffer size).

pkg/observable/channel/observable_test.go
bryanchriswhite: Currently `Observable` doesn't have a "closed" state. Calling `#Close()` actually unsubscribes all observers; perhaps it should be renamed to `#UnsubscribeAll()`. To answer the spirit of your question. There is no public API to test whether the observable has any observers currently subscribed. I think we can (if we're not already) get coverage of that elsewhere, likely in  a smaller unit test.

pkg/observable/channel/observable_test.go
bryanchriswhite: image

pkg/observable/interface.go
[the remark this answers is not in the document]
bryanchriswhite: image

pkg/observable/channel/observable_test.go
bryanchriswhite: image

pkg/observable/channel/observable.go
```diff
@@ -0,0 +1,157 @@
...
+			}
+			obsvbl.observersMu.RUnlock()
+			break
+		}
+
+		// notify observers
+		for _, obsvr := range observers {
+			// CONSIDERATION: perhaps continue trying to avoid making this
+			// notification async as it would effectively use goroutines
+			// in memory as a buffer (with little control surface).
```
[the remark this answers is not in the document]
bryanchriswhite: I suppose I wasn't thinking clearly enough to have imagined a work group type of scenario. That could totally work; although, now I want to write benchmarks to see the diff. :sweat_smile: At the time I was only  imagining the simplest thing you could do which would be to just run this same code in a new go routine every time. :grimacing:

pkg/observable/channel/observable.go
```diff
@@ -0,0 +1,157 @@
...
+			obsvr.notify(notification)
+		}
+	}
+
+	// Here we know that the producer has been closed, all observers should be closed as well
```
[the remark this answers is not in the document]
bryanchriswhite: Can you say more? I'm not quite picking up what you're putting down. :thinking:

pkg/observable/channel/observable_test.go
bryanchriswhite: I've renamed and moved it to be closer.

pkg/observable/channel/observable.go
```diff
@@ -39,112 +43,126 @@ func NewObservable[V any](opts ...option[V]) (observable.Observable[V], chan<- V
...
+			// in memory as a buffer (unbounded).
 			obsvr.notify(notification)
 		}
 	}
 
-	// Here we know that the producer has been closed, all observers should be closed as well
-	obsvbl.close()
+	// Here we know that the publisher channel has been closed.
+	// Unsubscribe all observers as they can no longer receive notifications.
+	obsvbl.unsubscribeAll()
```
Olshansk, line 131: Moving previous convo here: https://github.com/pokt-network/poktroll/pull/31#discussion_r1364521938

Thoughts on doing a `if (len(obsvbl.observers) == 0) pain("Should have no observers after unsubscribing")`

pkg/observable/channel/observable_test.go
```diff
@@ -16,70 +16,70 @@ import (
...
+			publishCh:       make(chan int, 1),
 			inputs:          inputs,
 			expectedOutputs: inputs,
 		},
-		// INCOMPLETE: producer channels which are full are proving harder to test
+		// INCOMPLETE: publisher channels which are full are proving harder to test
 		// robustly (no flakiness); perhaps it has to do with the lack of some
 		// kind of guarantee about the receiver order on the consumer side.
 		//
 		// The following scenarios should generally pass but are flaky:
```
Olshansk, line 61: Can you insert links to the github images from https://github.com/pokt-network/poktroll/pull/31#discussion_r1364559841

pkg/observable/channel/observable_test.go
```diff
@@ -304,39 +307,41 @@ func TestChannelObservable_SequentialProductionAndUnsubscription(t *testing.T) {
...
 				)
 			}
 		})
 	}
 }
 
-// TECHDEBT/INCOMPLETE: add coverage for active observers closing when producer closes.
-func TestChannelObservable_ObserversCloseOnProducerClose(t *testing.T) {
-	t.Skip("add coverage: all observers should close when producer closes")
+// TECHDEBT/INCOMPLETE: add coverage for active observers closing when publishCh closes.
```
Olshansk, line 326:
```suggestion
// TODO_TECHDEBT/TODO_INCOMPLETE: add coverage for active observers closing when publishCh closes.
```

Trying to move to a world where we prefix everything with TODO

Olshansk (approved):
@bryanchriswhite I reviewed all your inline comments and resolved them.

Thanks for clarifying everything with better names & comments. I really think its approachable by anyone now.

A couple small NITs but otherwise LFG!

pkg/observable/channel/observable_test.go
bryanchriswhite: Created https://github.com/pokt-network/poktroll/issues/81 and added `TODO_INCOMPLETE(#81)` in the comment here.

pkg/observable/channel/observable_test.go
bryanchriswhite: Created https://github.com/pokt-network/poktroll/issues/81 and added `TODO_INCOMPLETE(#81)`.

pkg/observable/channel/observable.go
bryanchriswhite: I don't think we want to do this. The observers list is explicitly being set to an empty slice at the end of `unsubscribeAll()`. If we wanted to check that the `obervers` slice is empty here, we would have to acquire a lock on the mutex. This raises some concerns for me with regards to false positives on the timing; although, this change doesn't seem to break tests. Regardless, this could be refactored by moving such a check directly into `unsubscribeAll()`. But then you end up with something that looks like this:

```go
// unsubscribeAll unsubscribes and removes all observers from the observable.
func (obsvbl *channelObservable[V]) unsubscribeAll() {
	...

	// Reset observers to an empty list. This purges any observers which might have
	// subscribed while the observable was closing.
	obsvbl.observersMu.Lock()
	obsvbl.observers = []*channelObserver[V]{}
	if (obsvbl.observers) > 0 {
		panic("impossible")
	}
	obsvbl.observersMu.Unlock()
}
```

PR #44, opened 2023-10-10
[Miner, Supplier] chore: scaffold submit-proof message

bryanchriswhite:
## Summary

This message type is a nested dependency of the `Miner` via the `SupplierClient`.

- Scaffolded `MsgSubmitProof` message types, handlers, etc.:
  ```
  ignite scaffold message submit-proof session_header:SessionHeader proof --signer supplier_address --module supplier
  ```
  (NB: generated a temporary SessionHeader duplicate in the supplier module so the above would work; subsequently, removed and updated protobuf import and usage.)

- Switch `MsgSubmitProof#proof` field type from `string` to `[]byte`  & refactor usages.

## Issue

Related to:
- #13 
- #7 

This message type is a nested dependency of the `Miner` via the `SupplierClient`.

## Type of change

Select one or more:

- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

## Sanity Checklist


proto/pocket/supplier/tx.proto
```diff
@@ -1,7 +1,22 @@
...
 
 option go_package = "pocket/x/supplier/types";
 
 // Msg defines the Msg service.
-service Msg {}
\ No newline at end of file
+service Msg {
+  rpc SubmitProof (MsgSubmitProof) returns (MsgSubmitProofResponse);
+}
+message MsgSubmitProof {
```
red-0ne, line 11: We should revisit `MsgSubmitProof` params to comply with @h5law 's changes with `VerifyClosestProof` which has the following signature [1]: 

```go
func VerifyClosestProof(proof *SparseMerkleClosestProof, root []byte, spec *TreeSpec) (bool, error)
```

And uses `SparseMerkleClosestProof` obtained from the rewritten`ProveClosest` which has the following structure [1]:

```go
type SparseMerkleClosestProof struct {
	Path             []byte             // the path provided to the ProveClosest method
	FlippedBits      []int              // the index of the bits flipped in the path during tree traversal
	Depth            int                // the depth of the tree when tree traversal stopped
	ClosestPath      []byte             // the path of the leaf closest to the path provided
	ClosestValueHash []byte             // the value hash of the leaf closest to the path provided
	ClosestProof     *SparseMerkleProof // the proof of the leaf closest to the path provided
}
```

[1] https://github.com/pokt-network/smt/blob/main/smt_proofs_test.go#L288
[2] https://github.com/pokt-network/smt/blob/main/proofs.go#L166C1-L173C2

Test: https://github.com/pokt-network/smt/blob/main/smt_proofs_test.go#L288

h5law: Yeah @red-0ne is correct, this new struct includes all the values previously returned from the `ProveClosest` method but also links them directly to the path, the `VerifyClosestProof` ensures the proof is structured in a valid way and passes the relevent checks to ensure the path does produce the closestPath found. In addition to this we should, when verifying the proof, ensure that `proof.ClosestPath == {what we expected it to be}` and then the verification of the proof with ensure that path definietly produced the one found and then also validate that proof in turn.

We can continue to embed this entire struct in the protobufs using `proof.Marshal()` and store it as a `[]byte`

red-0ne: @h5law we got confused by this part
> proof.ClosestPath == {what we expected it to be}
Did you mean `proof.Path` ? which is what is provided to `ProveClosest` ?

h5law: @red-0ne yes! sorry my error the `proof.Path` is what will be determined by the function of on-chain data (blockhash etc) **not** the `proof.ClosestPath` <- this is the path we found. Good 👀

Olshansk:
@bryanchriswhite Discussed offline, but do you still plan to scaffold an empty message and add the params in a followup PR?

x/supplier/keeper/msg_server_submit_proof.go
```diff
@@ -0,0 +1,17 @@
...
+	"context"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"pocket/x/supplier/types"
+)
+
+func (k msgServer) SubmitProof(goCtx context.Context, msg *types.MsgSubmitProof) (*types.MsgSubmitProofResponse, error) {
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	// TODO: Handling the message
```
bryanchriswhite, line 13:
```suggestion
	/*
		INCOMPLETE: Handling the message

		## Validation

		### Session validation
		1. claimed session ID == retrieved session ID
		2. this supplier is in the session's suppliers list
		3. proof signer addr == session application addr

		### Msg distribution validation (depends on session validation)
		1. pseudo-randomize earliest block offset
		2. governance-based earliest block offset

		### Proof validation
		1. session validation
		2. msg distribution validation
		3. claim with matching session ID exists
		4. proof path matches last committed block hash at claim height - 1
		5. proof validates with claimed root hash

		## Persistence
		1. submit proof message
			- supplier address
			- session header
			- proof
	*/
```

proto/pocket/supplier/tx.proto
```diff
@@ -20,4 +22,13 @@ message MsgStakeSupplier {
...
 message MsgUnstakeSupplierResponse {}
 
-message MsgStakeSupplierResponse {}
\ No newline at end of file
+message MsgStakeSupplierResponse {}
+
+message MsgSubmitProof {
+  string                       supplier_address = 1;
+  pocket.session.SessionHeader session_header   = 2;
+  bytes                        proof            = 3;
```
Olshansk, line 30: Not a blocker. Did we decide not use an explicit type from the SMT repo?

x/supplier/client/cli/tx_submit_proof.go
```diff
@@ -0,0 +1,55 @@
...
+			if err != nil {
+				return err
+			}
+
+			msg := types.NewMsgSubmitProof(
+				clientCtx.GetFromAddress().String(),
+				argSessionHeader,
+				argSmstProof,
+			)
+			if err := msg.ValidateBasic(); err != nil {
```
Olshansk, line 45: Need to add tests for message validation. Please add `TODO(@bryanchriswhite)` for it.

See #59 for an example

x/supplier/client/cli/tx_submit_proof.go
```diff
@@ -0,0 +1,55 @@
+package cli
+
```
Olshansk, line 2: Need to add tests for executing the CLI command. Please add `TODO(@bryanchriswhite)` for it.

See #59 for an example

x/supplier/types/message_submit_proof_test.go
```diff
@@ -0,0 +1,40 @@
...
+
+import (
+	"testing"
+
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+	"github.com/stretchr/testify/require"
+	"pocket/testutil/sample"
+)
+
+func TestMsgSubmitProof_ValidateBasic(t *testing.T) {
```
Olshansk, line 11: Need to add tests for adding more tests related to all the types. Please add `TODO(@bryanchriswhite)` for it.

See #59 for an example

Olshansk (changes requested):

proto/pocket/supplier/tx.proto
bryanchriswhite: Correct.

> We can continue to embed this entire struct in the protobufs using proof.Marshal() and store it as a []byte

--[@h5law](https://github.com/pokt-network/poktroll/pull/44#discussion_r1352406912)

Olshansk: My only request is to leave a comment explaining that the proof is a serialized version of `smt.XXX`

Olshansk (dismissed):
Approving with a request for one more comment.

proto/pocket/supplier/tx.proto
```diff
@@ -24,10 +25,20 @@ message MsgUnstakeSupplierResponse {}
 
 message MsgStakeSupplierResponse {}
 
+message MsgCreateClaim {
+  string                       supplier_address = 1;
```
Olshansk, line 29:
```suggestion
  option (cosmos.msg.v1.signer) = "address";

  string                       supplier_address = 1  [(cosmos_proto.scalar) = "cosmos.AddressString"];
```

proto/pocket/supplier/tx.proto
```diff
@@ -24,10 +25,20 @@ message MsgUnstakeSupplierResponse {}
...
+  pocket.session.SessionHeader session_header   = 2;
+  bytes                        root_hash        = 3;
```
Olshansk, line 31: Can you add a comment that this is referencing smt.Tree.root

proto/pocket/supplier/tx.proto
Olshansk: You might need to add these comments:

```
import "cosmos_proto/cosmos.proto";
import "cosmos/msg/v1/msg.proto";
```

Olshansk (changes requested):

Olshansk (approved):

PR #64, opened 2023-10-14
[Miner] feat: add events query client

bryanchriswhite:
## @Reviewer
This PR may be more digestible / reviewable on a [commit-by-commit basis](https://github.com/pokt-network/poktroll/pull/64/commits). Commits are organized logically and any given line is only modified in a single commit, with few exceptions*.

*(In the interest of preserving the git-time-continuum :police_officer::rotating_light:, this applies in batches of commits between comments or reviews *by humans*, only once "in review")

## Summary

Adds `EventsQueryClient` interface and implementation.

```mermaid
---
title: Clients Dependency Tree
---
flowchart

sup[SupplierClient]
tx[TxClient]
bl[BlockClient]
evt[EventsQueryClient]

sup --"#SignAndBroadcast()"--> tx
tx --"#CommittedBlocksSequence()"--> bl
tx --"#EventsBytes()"--> evt
bl --"#EventsBytes()"--> evt
```

## Issue

Relates to:
- #13 
- #65

`QueryClient` is a nested dependency of `Miner`.

## Type of change

Select one or more:

- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

- [x] **Run all unit tests**: `make go_test_integration`

## Sanity Checklist


pkg/client/query/client_test.go
```diff
@@ -0,0 +1,55 @@
...
+
+	"pocket/internal/testclient/testquery"
+)
+
+const (
+	msgReceiveTimeout = 5 * time.Second
+	msgReceiveLimit   = 2
+)
+
+func TestQueryClient_SubscribeWithQuery_NewBlock(t *testing.T) {
```
h5law, line 20: For my own knowledge does this test require the chain to be running, I can't see how this wouldn't always timeout without the chain running

Makefile
```diff
@@ -100,7 +100,11 @@ localnet_regenesis: ## Regenerate the localnet genesis file
 
 .PHONY: go_test
 go_test: go_version_check ## Run all go tests
-	go test -v ./...
+	go test -v -race -tags test ./...
+
+.PHONY: go_test_integration
+go_test_integration: go_version_check ## Run all go tests
```
h5law, line 106:
```suggestion
go_test_integration: go_version_check ## Run all go integration tests
```

pkg/client/query/client_test.go
```diff
@@ -0,0 +1,439 @@
...
+			// set up query client
+			dialerOpt := query.WithDialer(dialerMock)
+			queryClient := query.NewQueryClient("", dialerOpt)
+
+			// set up query observer
+			eventObservable, errCh := queryClient.EventsObservable(ctx, testQuery(queryIdx))
+			eventObserver := eventObservable.Subscribe(ctx)
+
+			// concurrently consume events from the observer channel
+			go func() {
```
bryanchriswhite, line 115: Consider refactoring/reusing [`goNotifiedOrTimedOutFactory`](https://github.com/pokt-network/poktroll/pull/31/files#diff-c8136d7c19962dcc168b21e249e0c612450ecc04bc778aa6bfd1946407bcce57R354) from #31.

bryanchriswhite: Alternatively, `goNotifiedOrTimedOutFactory` might be reducible once the `Either` primitive becomes available.

pkg/client/query/client_test.go
bryanchriswhite: This test does not integrate a real server. It mocks the connection's `Receive` function to produce values for the observable's producer to receive.

pkg/client/events_query/client.go
```diff
@@ -0,0 +1,225 @@
...
+	}
+}
+
+// observableConn is a struct which holds an observable & the corresponding
+// connection which produces its inputs. Additionally, it currently holds an
+// error channel for asynchronous error handling.
+// TODO_TECHDEBT(#70): once Either is available, we can remove the error channel and
+// instead use an Observable[Either[any, error]].
+type observableConn struct {
+	observable observable.Observable[[]byte]
```
h5law, line 72: 🔥 This is what I was hoping for <3

pkg/client/interface.go
```diff
@@ -0,0 +1,52 @@
...
+// EventsQueryClient is used to subscribe to chain event messages matching the given query,
+type EventsQueryClient interface {
+	EventsBytes(
+		ctx context.Context,
+		query string,
+	) (EventsBytesObservable, error)
+	//EventsBytes(
+	//	ctx context.Context,
+	//	query string,
+	//) (observable.Observable[either.Either[[]byte]], error)
```
bryanchriswhite: This is cleaned up in 01278b03f8d6214861c57f2e98bf15ba36402879

pkg/client/events_query/client_integration_test.go
```diff
@@ -0,0 +1,62 @@
+//go:build integration
```
bryanchriswhite, line 1: :disappointed: Mistakes were made; this file wasn't meant to be committed until de4defec5dd85d0182ba9cc1457bbc10f16a1e32. There will be modifications in that commit.

pkg/client/events_query/client_test.go
```diff
@@ -0,0 +1,413 @@
+package eventsquery_test
```
bryanchriswhite, line 1: :disappointed: Mistakes were made; this file wasn't meant to be committed until 66fdc790eca790c4173afcb8e7d6adffeff0c6e0. There will be modifications in that commit.

pkg/client/events_query/client_test.go
```diff
@@ -4,84 +4,71 @@ import (
 	"context"
```
bryanchriswhite, line 4: :disappointed: Mistakes were make; this file was already committed in ad0121d63e1a219e551765229bf907731ab28929.

pkg/client/interface.go
```diff
@@ -0,0 +1,52 @@
+//go:generate mockgen -destination=../../internal/mocks/mockclient/query_client_mock.go -package=mockclient . Dialer,Connection
```
bryanchriswhite, line 1: godoc comments added in 01278b03f8d6214861c57f2e98bf15ba36402879

pkg/client/errors.go
```diff
@@ -0,0 +1,11 @@
+package client
+
+import errorsmod "cosmossdk.io/errors"
+
+var (
+	ErrDial       = errorsmod.Register(codespace, 1, "dialing for connection failed")
+	ErrConnClosed = errorsmod.Register(codespace, 2, "connection closed")
+	ErrSubscribe  = errorsmod.Register(codespace, 3, "failed to subscribe to events")
+	ErrReceive    = errorsmod.Register(codespace, 4, "failed to receive event")
+	codespace     = "query_client"
```
bryanchriswhite: This updated to `"events_query_client"` in [ad0121d](https://github.com/pokt-network/poktroll/pull/64/commits/ad0121d63e1a219e551765229bf907731ab28929#diff-a9a023007c887e80b262ad97dfa5092cedfd14e3f02ef3d0662a12dda2925374)

internal/testclient/testeventsquery/client.go
```diff
@@ -0,0 +1,15 @@
...
+
+import (
+	"testing"
+
+	"pocket/internal/testclient"
+	"pocket/pkg/client"
+	eventsquery "pocket/pkg/client/events_query"
+)
+
+func NewLocalnetClient(t *testing.T, opts ...client.Option) client.EventsQueryClient {
```
bryanchriswhite: Updated to `opts ...client.EventsQueryClientOption` in [de4defec](https://github.com/pokt-network/poktroll/pull/64/commits/de4defec5dd85d0182ba9cc1457bbc10f16a1e32#diff-b114b2cac95bfcb1700e8dcbc3de5a1d2e2d4cd7e74c4231d8c4f26a94feb89c)

pkg/client/events_query/dialer.go
```diff
@@ -0,0 +1,35 @@
...
+
+import (
+	"context"
+
+	"github.com/gorilla/websocket"
+
+	"pocket/pkg/client"
+)
+
+var _ client.Dialer = &websocketDialer{}
```
red-0ne, line 11: Should be
```suggestion
var _ client.Dialer = (*websocketDialer)(nil)
```

pkg/client/events_query/connection.go
```diff
@@ -0,0 +1,34 @@
+package eventsquery
+
+import (
+	"github.com/gorilla/websocket"
+
+	"pocket/pkg/client"
+)
+
+var _ client.Connection = &websocketConn{}
```
red-0ne, line 9:
```suggestion
var _ client.Connection = (*websocketConn)(nil)
```

pkg/client/events_query/connection.go
```diff
@@ -0,0 +1,34 @@
...
+
+// websocketConn implements the Connection interface using the gorilla websocket
+// transport implementation.
+type websocketConn struct {
+	conn *websocket.Conn
+}
+
+// Receive implements the respective interface method using the underlying websocket.
+func (wsConn *websocketConn) Receive() ([]byte, error) {
+	_, msg, err := wsConn.conn.ReadMessage()
```
red-0ne, line 19: Shouldn't we guard this method with a `Lock`?

pkg/client/events_query/client.go
```diff
@@ -0,0 +1,289 @@
...
+
+	"pocket/pkg/client"
+	"pocket/pkg/either"
+	"pocket/pkg/observable"
+	"pocket/pkg/observable/channel"
+)
+
+const requestIdFmt = "request_%d"
+
+var _ client.EventsQueryClient = &eventsQueryClient{}
```
red-0ne, line 20:
```suggestion
var _ client.EventsQueryClient = (*eventsQueryClient)(nil)
```

pkg/client/events_query/client.go
```diff
@@ -0,0 +1,289 @@
...
+// event messages matching the given query. It receives an either.Either[[]byte]
+// which is either an error or the event message bytes.
+// (see: https://pkg.go.dev/github.com/cometbft/cometbft/types#pkg-constants)
+// (see: https://docs.cosmos.network/v0.47/core/events#subscribing-to-events)
+func (eqc *eventsQueryClient) EventsBytes(
+	ctx context.Context,
+	query string,
+) (client.EventsBytesObservable, error) {
+	// Check if an event subscription already exists for the given query.
+	if eventsBzConn := eqc.getEventsBytesAndConn(query); eventsBzConn != nil {
```
red-0ne, line 87: I think there should be a lock should wrapping the code from `getEventsBytesAndConn` til `insertEventBytesAndConn` otherwise concurrent calls to `EventsBytes` may both create and insert for the same `query`

red-0ne (changes requested):
Great query client!

Besides some NITs, there may be some locks that need to be added.

pkg/client/events_query/client.go
bryanchriswhite: Both `getEventsBytesAndConn()` and `insertEventsBytesAndConn()` acquire a lock on the mutex before reading  or writing `eventsBytesAndCons`.

pkg/client/events_query/connection.go
bryanchriswhite: To protect what? The connection is assigned by the dialer. If the connection is closed it should return an error but it's location in memory isn't changing.

red-0ne: Was imagining concurrent calls to `Receive`... not sure how `ReadMessage` would react in that case.

bryanchriswhite: 🤔 We shouldn't be calling `ReadMessage()` concurrently on the same connection anywhere, each connection is read in its own for loop. I would assume that concurrent calls to read would result in one caller getting a return and the rest blocking. I could be mistaken, I'll take a look at the code when I get a chance. Regardless, I think it's unlikely that there's a data race because I would expect `ReadMessage` to return a new value or reference each time. Again, I could be mistaken; will have to look at the code

pkg/client/events_query/client.go
```diff
@@ -0,0 +1,273 @@
...
+) (*eventsBytesAndConn, error) {
+	conn, err := eqc.openEventsBytesAndConn(ctx, query)
+	if err != nil {
+		return nil, err
+	}
+
+	// Construct an eventsBytes for the given query.
+	eventsBzObservable, eventsBzPublishCh := channel.NewObservable[either.Either[[]byte]]()
+
+	// TODO_INVESTIGATE: does this require retry on error?
```
Olshansk, line 156: Would the retry happen inside the function given that it's run as a goroutine?

pkg/client/events_query/client.go
```diff
@@ -0,0 +1,273 @@
...
+	}, nil
+}
+
+// openEventsBytesAndConn gets a connection using the configured dialer and sends
+// an event subscription request on it, returning the connection.
+func (eqc *eventsQueryClient) openEventsBytesAndConn(
+	ctx context.Context,
+	query string,
+) (client.Connection, error) {
+	// If no event subscription exists for the given query, create a new one.
```
Olshansk, line 171: > If no event subscription exists for the given query

Where do we do the check?

pkg/client/events_query/connection.go
```diff
@@ -0,0 +1,34 @@
...
+	if err != nil {
+		return nil, ErrReceive.Wrapf("%s", err)
+	}
+	return msg, nil
+}
+
+// Send implements the respective interface method using the underlying websocket.
+func (wsConn *websocketConn) Send(msg []byte) error {
+	return wsConn.conn.WriteMessage(websocket.TextMessage, msg)
```
Olshansk, line 28: Why is this a `TextMessage` and not a `ByteMessage`?

pkg/client/events_query/client.go
```diff
@@ -0,0 +1,273 @@
...
+	ctx context.Context,
+	conn client.Connection,
+	eventsBzPublishCh chan<- either.Either[[]byte],
+) {
+	// Read and handle messages from the websocket. This loop will exit when the
+	// websocket connection is closed and/or returns an error.
+	for {
+		event, err := conn.Receive()
+		if err != nil {
+			// TODO_CONSIDERATION: should we close the publish channel here too?
```
Olshansk, line 207: I think we should add some logging / warning / printing here.

I believe it'll be particularly valuable in this stage of development.

pkg/client/events_query/client.go
```diff
@@ -0,0 +1,273 @@
...
+
+// goUnsubscribeOnDone unsubscribes from the subscription when the context is done.
+// It is intended to be called  in a goroutine.
+func (eqc *eventsQueryClient) goUnsubscribeOnDone(
+	ctx context.Context,
+	query string,
+) {
+	// wait for the context to be done
+	<-ctx.Done()
+	// only close the eventsBytes for the give query
```
Olshansk, line 240:
```suggestion
	// only close the eventsBytes for the given query
```

pkg/client/events_query/client.go
```diff
@@ -0,0 +1,273 @@
...
+	if toClose, ok := eqc.eventsBytesAndConns[query]; ok {
+		toClose.eventsBytes.UnsubscribeAll()
+	}
+	for compareQuery, eventsBzConn := range eqc.eventsBytesAndConns {
+		if query == compareQuery {
+			eventsBzConn.eventsBytes.UnsubscribeAll()
+			return
+		}
+	}
+}
```
Olshansk, line 253: Why is there no here?

```
// remove closed eventsBytesAndConns
delete(eqc.eventsBytesAndConns, query)
```

pkg/client/events_query/client_integration_test.go
```diff
@@ -0,0 +1,55 @@
...
+	"context"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+
+	"pocket/internal/testclient/testeventsquery"
+)
+
+func TestQueryClient_EventsObservable_Integration(t *testing.T) {
```
Olshansk, line 15: Stupid question, but as I'm looking at this test, it's not clear to me where the publishing of NewBlock is coming from.

Can you #PUC a bit so its clearer?

docs/pkg/client/events_query.md
```diff
@@ -0,0 +1,104 @@
...
+The `events_query` package provides a client interface to subscribe to chain event messages. It abstracts the underlying connection mechanisms and offers a clear and easy-to-use way to get events from the chain. Highlights:
+
+- Offers subscription to chain event messages matching a given query.
+- Uses the Gorilla WebSockets package for underlying connection operations.
+- Provides a modular structure with interfaces allowing for mock implementations and testing.
+- Offers considerations for potential improvements and replacements, such as integration with the cometbft RPC client.
+
+## Architecture Diagrams
+
+_TODO(@bryanchriswhite): Add architecture diagrams for the package._
```
Olshansk, line 33: Reminder.

docs/pkg/client/events_query.md
```diff
@@ -0,0 +1,104 @@
...
+
+#### How can I use a different connection mechanism other than WebSockets?
+
+You can implement the `Dialer` and `Connection` interfaces and use the `WithDialer` configuration to provide your custom dialer.
+
+## Contributing
+
+If you're interested in improving the `events_query` package or adding new features, please start by discussing your ideas in the project's issues section. Check our main contributing guide for more details.
+
+## Changelog
```
Olshansk, line 98: I feel like the changelog & license are overkill at this stage of the project. Just requires extra maintenance.

docs/pkg/client/events_query.md
```diff
@@ -0,0 +1,104 @@
...
+
+### Advanced Usage
+
+_TODO(@bryanchriswhite): Add examples of advanced usage_
+
+### Configuration
+
+- **WithDialer**: Configure the client to use a custom dialer for connections.
+
+## API Reference
```
Olshansk, line 69: Personally - I feel like this sort of thing is REALLy hard to maintain.

Engineers always want to put APIs in the .md, but it goes stale right away. https://twitter.com/olshansky/status/1714649292894257332

Olshansk (changes requested):
@bryanchriswhite

1. Reviewed everything other than `pkg/client/events_query/client_test.go`
2. The base branch needs to be updated
3. Still wrapping my head around all the different clients/components but it's slowly coming together
4. I pushed some NITs to the README so please make sure to do a git pull
5. Great job!

pkg/client/events_query/client.go
```diff
@@ -0,0 +1,273 @@
...
+	"pocket/pkg/observable"
+	"pocket/pkg/observable/channel"
+)
+
+const requestIdFmt = "request_%d"
+
+var _ client.EventsQueryClient = (*eventsQueryClient)(nil)
+
+// TODO_CONSIDERATION: the cosmos-sdk CLI code seems to use a cometbft RPC client
+// which includes a `#EventsBytes()` method for a similar purpose. Perhaps we could
```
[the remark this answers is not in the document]
bryanchriswhite: Unfortunately, I was not able to get theirs to work. :confused: 

It's complete speculation on my part at this point but I do recall the behavior expressed when trying to use their was identical to a point where I was debugging ours and it was using binary websocket messages instead of text messages. The symptom was basically that it could connect and wouldn't complain when sending requests, it just would never send anything back.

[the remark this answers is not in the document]
bryanchriswhite: FWIW, I kinda like the relative simplicity of ours. Regardless, I think we have good interfaces now to adapt to whatever. :rocket:

internal/testclient/common.go
```diff
@@ -0,0 +1,3 @@
+package testclient
```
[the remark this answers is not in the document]
bryanchriswhite: :sweat_smile: hehe, no magic, just a file and a convention.

pkg/client/events_query/connection.go
bryanchriswhite: IIRC, this is the gorilla websocket client implementation exposing a websocket degree of freedom which has to do with the encoding (or something along those lines). Since we're working with a JSON-RPC API, the encoding is text, not binary.

bryanchriswhite: (I describe what happens when we try to use binary messages in https://github.com/pokt-network/poktroll/pull/64#discussion_r1372247584)

pkg/client/events_query/client.go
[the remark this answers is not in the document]
Olshansk: > Regardless, I think we have good interfaces now to adapt to whatever. 🚀

Yea, as much as I'd prefer to use theirs and not build our own, I think this is the sort of primitie that will enable us to extend to any RPC service w/o dependency on Cosmos 👌 .

One ask: Link to this comment thread in the code.

pkg/client/events_query/connection.go
Olshansk: #PUC

pkg/client/events_query/client.go
```diff
@@ -0,0 +1,273 @@
...
+
+// eventsBytesAndConn is a struct which holds an eventsBytes observable & the
+// corresponding connection which produces its inputs.
+type eventsBytesAndConn struct {
+	// eventsBytes is an observable which is notified about chain event messages
+	// matching the given query. It receives an either.Either[[]byte] which is
+	// either an error or the event message bytes.
+	eventsBytes observable.Observable[either.Either[[]byte]]
+	conn        client.Connection
+	closed      bool
```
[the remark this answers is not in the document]
bryanchriswhite: :pray: thanks for calling this out! :raised_hands:

pkg/client/events_query/client.go
```diff
@@ -0,0 +1,273 @@
...
+
+		// remove closed eventsBytesAndConns
+		delete(eqc.eventsBytesAndConns, query)
+	}
+}
+
+// getNextRequestId increments and returns the JSON-RPC request ID which should
+// be used for the next request. These IDs are expected to be unique (per request).
+func (eqc *eventsQueryClient) getNextRequestId() string {
+	eqc.nextRequestId++
```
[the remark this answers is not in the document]
bryanchriswhite: I've simplified this out of existence. Instead, we read 8 random bytes and base64 encode it in the same place it was being used  (eventSubscriptionRequest), which I've promoted to a method.

pkg/client/events_query/client.go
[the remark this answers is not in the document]
bryanchriswhite: `#Closed()` is now idempotent so `isClosed` is no longer needed. Best part == no part. :joy:

pkg/client/events_query/client.go
```diff
@@ -0,0 +1,273 @@
...
+	// observableConns map.
+	// The lock must be held for both checking and adding to prevent concurrent
+	// calls to this function from racing.
+	eqc.eventsBytesAndConnsMu.Lock()
+	// Deferred (write) unlock.
+	defer eqc.eventsBytesAndConnsMu.Unlock()
+
+	// Check if an event subscription already exists for the given query.
+	if eventsBzConn := eqc.eventsBytesAndConns[query]; eventsBzConn != nil {
+		// If found it is returned.
```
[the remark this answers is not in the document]
bryanchriswhite: Can you say more about what you're concerned about?

From the perspective of a consumer of this API, it seems totally reasonable to be making calls to `#EventsBytes()` with the same query. In which case, I would not expect to see  a warning.

pkg/client/interface.go
```diff
@@ -0,0 +1,63 @@
...
+
+import (
+	"context"
+
+	"pocket/pkg/either"
+	"pocket/pkg/observable"
+)
+
+// TODO_CONSIDERATION: the cosmos-sdk CLI code seems to use a cometbft RPC client
+// which includes a `#Subscribe()` method for a similar purpose. Perhaps we could
```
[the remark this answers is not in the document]
bryanchriswhite: I don't think so. As I was trying to get at in https://github.com/pokt-network/poktroll/pull/64#discussion_r1372248537, we have interfaces (e.g. `Conection`, `Dialer`, `EventsQueryClient`) which encapsulate and articulate their concerns well, and more importantly, in terms of our concurrency primitives (i.e. `Observable`, etc.). 

Apologies for the sate that #74 ended up in. I've updated it's base to a new branch that won't move and updated its one commit to incorporate recent changes to the interface. It's now +153 LOC diff. :wink:

pkg/client/interface.go
```diff
@@ -0,0 +1,63 @@
...
+// EventsBytesObservable is an observable which is notified with an either
+// value which contains either an error or the event message bytes.
+// TODO_HACK: The purpose of this type is to work around gomock's lack of
+// support for generic types. For the same reason, this type cannot be an
+// alias (i.e. EventsBytesObservable = observable.Observable[either.Either[[]byte]]).
+type EventsBytesObservable observable.Observable[either.Either[[]byte]]
+
+// Connection is a transport agnostic, bi-directional, message-passing interface.
+type Connection interface {
+	// Receive blocks until a message is received or an error occurs.
```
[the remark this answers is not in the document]
bryanchriswhite: That was the  intent behind referencing "message-passing" in the comment above. Do you think that's not clear enough?

pkg/client/interface.go
```diff
@@ -0,0 +1,63 @@
+//go:generate mockgen -destination=../../internal/mocks/mockclient/query_client_mock.go -package=mockclient . Dialer,Connection
```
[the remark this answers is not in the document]
bryanchriswhite: 1. This file will be populated with all client interfaces, not just `EventsQueryClient` (see: #65 & #94).
2. It SHOULD be in a separate package from any implementations to avoid dependency cycles.

pkg/client/events_query/dialer.go
```diff
@@ -0,0 +1,35 @@
+package eventsquery
```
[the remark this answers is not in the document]
bryanchriswhite: Good call! :raised_hands: 

I went with moving `connection.go` and `dialer.go` into `pkg/client/websocket`. :+1: 

This resulted in also having to move `ErrReceive` into a new `pkg/client/websocket/errors.go` file to avoid an import loop.

pkg/client/events_query/client.go
bryanchriswhite: **TL;DR: I don't think we should do this here.**

I was comparing this to the block client implementation where we do a similar thing but are retrying on error. I deleted this TODO in the block client branch because I decided that we shouldn't retry here, I think it should be up to the observer how to handle an async error from the `#EventsBytes()` observable. 

Since the client unsubscribe's all observers (i.e. no more notifications) after encountering an error, observers should only ever see one error before they're unsubscribed. 

An example of this can currently be seen in the block client. In that case, receiving an async error from `#EventsBytes()` likely means that our long-running connection failed at some point, in which case we can simply restart it by calling `#EventsBytes()` again.

pkg/client/events_query/client.go
bryanchriswhite: Where this will be used first, in the block client, there's a `WARN` log which covers this in a select/case that receives from an `errCh` in `retryOrError`. 

I also just fixed a bug that was preventing the async error in the `EventsBytes()` observable from propagating that direction, so now when this :point_up_2: branch is taken, that `WARN` will be logged.

bryanchriswhite:
@Olshansk aside from these three comments:

image

...  and a test I apparently broke :sweat_smile:, it's ready for re-review.

pkg/client/interface.go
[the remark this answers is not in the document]
Olshansk: It's okay

pkg/client/events_query/client.go
```diff
@@ -271,3 +267,19 @@ func eventSubscriptionRequest(requestId, query string) ([]byte, error) {
 	}
 	return requestBz, nil
 }
+
+// randRequestId returns a random 8 byte, base64 request ID which is intended
+// for in JSON-RPC requests to uniquely identify distinct RPC requests.
+// These request IDs only need to be unique to the extent that they are useful
+// to this client for identifying distinct RPC requests.
+// These IDs
+// are expected to be unique (per request). Its size and keyspace are arbitrary.
```
Olshansk, line 276:
```suggestion
// These IDs are expected to be unique (per request). Its size and keyspace are arbitrary.
```

pkg/client/events_query/client_integration_test.go
```diff
@@ -12,6 +12,8 @@ import (
 	"pocket/internal/testclient/testeventsquery"
 )
 
+const committedBlockEventsQuery = "tm.event='NewBlock'"
```
Olshansk, line 15:
```suggestion
// The query use to subscribe for new block events on the websocket endpoint exposed by CometBFT nodes
const committedBlockEventsQuery = "tm.event='NewBlock'"
```

pkg/client/events_query/websocket/connection.go
```diff
@@ -1,7 +1,7 @@
-package eventsquery
+package websocket
 
 import (
-	"github.com/gorilla/websocket"
+	gorillaws "github.com/gorilla/websocket"
```
Olshansk, line 4: I wasn't aware gorrilas have laws 🦍

pkg/client/events_query/client.go
Olshansk: Yup!

pkg/client/events_query/client.go
Olshansk: #PUC with a link to this discussion.

Olshansk (approved):
A few minor nits plus a request to #PUC, but otherwise lgtm!

pkg/client/events_query/websocket/connection.go
bryanchriswhite: gorillaws

docs/pkg/client/events_query.md
bryanchriswhite: Agreed. :100:

PR #65, opened 2023-10-14
[Miner] feat: add block client

bryanchriswhite:
## Summary

Adds `BlockClient` interface and implementation.

```mermaid
---
title: Clients Dependency Tree
---
flowchart

sup[SupplierClient]
tx[TxClient]
bl[BlockClient]
evt[EventsQueryClient]

sup --"#SignAndBroadcast()"--> tx
tx --"#CommittedBlocksSequence()"--> bl
tx --"#EventsBytes()"--> evt
bl --"#EventsBytes()"--> evt
```

## Issue

Relates to:
- #13 
- #94 

`BlockClient` is a nested dependency of `Miner`.

## Type of change

Select one or more:

- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

- [x] **Run all unit tests**: `make go_test_integration`

## Sanity Checklist


pkg/client/block/client_test.go
```diff
@@ -0,0 +1,76 @@
...
+
+func TestBlockClient_LatestBlock(t *testing.T) {
+	ctx := context.Background()
+
+	blockClient := testblock.NewLocalnetClient(ctx, t)
+	require.NotNil(t, blockClient)
+
+	block := blockClient.LatestBlock(ctx)
+	require.NotEmpty(t, block)
+}
```
h5law, line 29:
```suggestion
}

```

docs/pkg/client/events_query.md
```diff
@@ -0,0 +1,87 @@
...
+- Offers considerations for potential improvements and replacements, such as integration with the cometbft RPC client.
+
+## Architecture Diagrams
+
+[Add diagrams here if needed. For the purpose of this mockup, we'll assume none are provided.]
+
+## Installation
+
+```bash
+go get github.com/yourusername/pocket/pkg/client/events_query
```
h5law, line 21:
```suggestion
go get github.com/pokt-network/poktroll/pkg/client/events_query
```
Any reason we are not doing this?

pkg/client/block/client.go
```diff
@@ -0,0 +1,181 @@
...
+	"cosmossdk.io/depinject"
+	cometTypes "github.com/cometbft/cometbft/types"
+
+	"pocket/pkg/client"
+	"pocket/pkg/observable"
+	"pocket/pkg/observable/channel"
+)
+
+var (
+	_              client.BlockClient = &blockClient{}
```
h5law, line 19:
```suggestion
	_              client.BlockClient = (*blockClient)(nil)
```
This is the cosmos-sdk and go standard

pkg/client/block/client_test_integration.go
```diff
@@ -0,0 +1,76 @@
...
+
+func TestBlockClient_LatestBlock(t *testing.T) {
+	ctx := context.Background()
+
+	blockClient := testblock.NewLocalnetClient(ctx, t)
+	require.NotNil(t, blockClient)
+
+	block := blockClient.LatestBlock(ctx)
+	require.NotEmpty(t, block)
+}
```
h5law, line 29:
```suggestion
}

```

pkg/observable/channel/map.go
```diff
@@ -0,0 +1,26 @@
+package channel
+
+import (
+	"context"
+
+	"pocket/pkg/observable"
+)
+
+func Map[S, D any](
```
h5law, line 9: 🔥 This is so cool

docs/pkg/client/events_query.md
bryanchriswhite: This is only necessary for go modules external to `poktroll` who want to consume the `events_query` package as a library.

pkg/client/block/client.go
```diff
@@ -0,0 +1,169 @@
...
+// blockClient implements the BlockClient interface.
+type blockClient struct {
+	// endpointURL is the URL of RPC endpoint which eventsClient subscription
+	// requests will be sent.
+	endpointURL string
+	// eventsClient is the events query client which is used to subscribe to
+	// newly committed block events. It emits an either value which may contain
+	// an error, at most, once and closes immediately after if it does.
+	eventsClient client.EventsQueryClient
+	// latestBlockObsvblsReplay is a replay observable with replay buffer size 1,
```
Olshansk, line 30: Would we be able to make it more than 1 or is there a reason to keep it at that value?

pkg/client/block/client.go
```diff
@@ -0,0 +1,169 @@
...
+
+func NewBlockClient(
+	ctx context.Context,
+	deps depinject.Config,
+	cometWebsocketURL string,
+) (client.BlockClient, error) {
+	// Initialize block client
+	bClient := &blockClient{endpointURL: cometWebsocketURL}
+	bClient.latestBlockObsvblsReplay, bClient.latestBlockObsvblsReplayPublishCh =
+		channel.NewReplayObservable[client.BlocksObservable](ctx, 1)
```
Olshansk, line 51: Move 1 into a local constant

pkg/client/block/client.go
```diff
@@ -0,0 +1,169 @@
...
+
+	// Inject dependencies
+	if err := depinject.Inject(deps, &bClient.eventsClient); err != nil {
```
Olshansk, line 54: Wow. So all we do is just inject an interface into a `depinject.Config` and we can supply it elsewhere?

pkg/client/block/client.go
```diff
@@ -0,0 +1,169 @@
...
+	go bClient.goPublishBlocks(ctx)
+
+	return bClient, nil
+}
+
+// CommittedBlocksSequence returns a ReplayObservable, with a replay buffer size
+// of 1, which is notified when block commit events are received by the events
+// query subscription.
+func (bClient *blockClient) CommittedBlocksSequence(ctx context.Context) client.BlocksObservable {
+	replayedBlocksObservable := bClient.latestBlockObsvblsReplay.Last(ctx, 1)[0]
```
Olshansk, line 68: use local constant here

pkg/client/block/client.go
```diff
@@ -0,0 +1,169 @@
...
+	return replayedBlocksObservable
+}
+
+// LatestBlock returns the latest committed block that's been received by the
+// corresponding events query subscription.
+// It blocks until at least one block event has been received.
+func (bClient *blockClient) LatestBlock(ctx context.Context) (latestBlock client.Block) {
```
Olshansk, line 75: If we increase it to more than 1, can we do "LastNBlocks". I can see us potentially needing it

pkg/client/block/client.go
```diff
@@ -0,0 +1,169 @@
...
+	v := bClient.CommittedBlocksSequence(ctx).Last(ctx, 1)[0]
```
Olshansk, line 76: s/v/block

pkg/client/block/client.go
```diff
@@ -0,0 +1,169 @@
...
+// and closes the events query client.
+func (bClient *blockClient) Close() {
+	// Closing eventsClient will cascade unsubscribe and close downstream observers.
+	bClient.eventsClient.Close()
+}
+
+// goPublishBlocks receives event bytes from the events query client, maps them
+// to block events, and publishes them to the latestBlockObsvblsReplay replay observable.
+func (bClient *blockClient) goPublishBlocks(ctx context.Context) {
+	// NB: cometbft event subscription query
```
Olshansk, line 90: OPTIONAL: Thoughts on moving into a local const?

pkg/client/interface.go
```diff
@@ -9,6 +9,29 @@ import (
 	"pocket/pkg/observable"
 )
 
+type BlockClient interface {
+	// Blocks returns an observable which emits newly committed blocks.
+	CommittedBlocksSequence(context.Context) BlocksObservable
+	//CommittedBlocksSequence() observable.Observable[Block]
```
Olshansk, line 15: delete line 15?

pkg/client/interface.go
```diff
@@ -9,6 +9,29 @@ import (
...
+	// LatestBlock returns the latest block that has been committed.
+	LatestBlock(context.Context) Block
+	// Close unsubscribes all observers of the committed blocks sequence observable
+	// and closes the events query client.
+	Close()
+}
+
+// BlocksObservable is an observable which is notified with an either
+// value which contains either an error or the event message bytes.
+// TODO_HACK: The purpose of this type is to work around gomock's lack of
```
Olshansk, line 25: Fwiw, I kind of like the explicit naming. I would even move it to the top of the file

pkg/client/block/client_test.go
```diff
@@ -0,0 +1,125 @@
...
+	"pocket/internal/testclient/testeventsquery"
+	"pocket/pkg/client"
+	"pocket/pkg/client/block"
+	eventsquery "pocket/pkg/client/events_query"
+)
+
+const blockAssertionLoopTimeout = 100 * time.Millisecond
+
+func main() {
+	fmt.Println("HELLOO!!!")
```
Olshansk, line 25: HELLO TO YOU TOO!

pkg/client/block/client_test.go
```diff
@@ -0,0 +1,125 @@
...
+	for {
+		select {
+		case actualBlock := <-actualBlockCh:
+			require.Equal(t, expectedHeight, actualBlock.Height())
+			require.Equal(t, expectedHash, actualBlock.Hash())
+		case <-done:
+			break blockAssertionLoop
+		case <-time.After(blockAssertionLoopTimeout):
+			t.Fatal("timed out waiting for block event")
+		}
```
Olshansk, line 108: Can you help me understand what happens here.

1. We block on all the cases
2. The first one geta a response (twice)
3. The 2nd one get a response (once)
4. Success

Is that it?

pkg/client/block/client_test.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+
+	// Wait a tick for the observables to be set up.
+	time.Sleep(time.Millisecond)
+
+	blockClient.Close()
+}
+
+// TODO_TECHDEBT/TODO_CONSIDERATION:
+// * we should prefer tests being in their own pkgs (e.g. block_test)
```
Olshansk, line 118: Golang does like to have _test files in the same directory, so we can remove this.

pkg/client/block/client_test.go
```diff
@@ -0,0 +1,125 @@
...
+// * we should prefer to not export types which don't require exporting for API consumption
```
Olshansk, line 119: What are we exporting that we don't want to?

pkg/client/block/client_test.go
```diff
@@ -0,0 +1,125 @@
...
+// * the cometBlockEvent isn't and doesn't need to be exported (except for this test)
+// * TODO_DISCUSS: we could use the //go:build test constraint on a new file which exports it for testing purposes
```
Olshansk, line 121: Can you create a ticket for it?

pkg/client/block/client.go
```diff
@@ -0,0 +1,169 @@
...
+		// Initially set latestBlockObsvblsReplay and update if after retrying on error.
+		bClient.latestBlockObsvblsReplayPublishCh <- blocksObsvbl
+
+		return errCh
+	})
+}
+
+// retryOnError runs the given function, which is expected to return an error
+// channel, and re-runs the function when an error is received.
+// TODO_CONSIDERATION: promote to some shared package (perhaps /internal/concurrency)
```
Olshansk, line 120:
```suggestion
// TODO_TECHDEBT: promote to some shared package (perhaps /internal/concurrency)
```

pkg/client/block/client.go
```diff
@@ -0,0 +1,169 @@
...
+		blocksObsvbl := channel.MapReplay(ctx, 1, eventsBz, blockEventFromEventBz)
+
+		// Initially set latestBlockObsvblsReplay and update if after retrying on error.
+		bClient.latestBlockObsvblsReplayPublishCh <- blocksObsvbl
+
+		return errCh
+	})
+}
+
+// retryOnError runs the given function, which is expected to return an error
```
Olshansk, line 118:
```suggestion
// retryOnError runs the given function, which could return an error
```

Olshansk: Is that correct or no?

pkg/client/block/client.go
```diff
@@ -0,0 +1,169 @@
...
+// channel, and re-runs the function when an error is received.
+// TODO_CONSIDERATION: promote to some shared package (perhaps /internal/concurrency)
+func retryOnError(
```
Olshansk, line 121: COnfirming that the goal here is to retry endlessly. Is that right?

Olshansk (changes requested):
Reviewed everything except for `pkg/client/block/client_integration_test.go` but overall, this PR is pretty clean!

pkg/client/block/client.go
bryanchriswhite: Yessir. :smile:

pkg/client/interface.go
bryanchriswhite: I'll clarity in the comment. The crux of the issue is that we can't make it an alias (i.e. `TypeB = TypeA`) because that is insufficient to work around the gomock incompatibility; however, not using an alias means that we don't  have access to `TypeA`'s methods from `TypeB`. This means that we would either have to cast the type everywhere we need to call a `TypeA` method or implement those methods on `TypeB` which would basically be a thin wrapper that does the same type cast and proxy's the method call.

pkg/client/block/client_test.go
bryanchriswhite: :heavy_check_mark:

pkg/client/block/client.go
bryanchriswhite: Probably not for this observable but perhaps the one it emits if we were to change `#LatestBlock()` to `#LastNBlocks()` https://github.com/pokt-network/poktroll/pull/65#discussion_r1372469509. I think this would also require changing the behavior of `ReplayObservable#Last()` to not block until it has `n` replay notifications. I think that's probably a step in the right direction though for `ReplayObservable` generally, I think it makes more sense in most use cases that I can imagine. There are still some details to resolve in that case but I will continue to think about it. :thinking:

Olshansk: > Probably not for this observable as we're just using it to hold a reactive reference (only ever a single value) to the observable which emits block events

#PUC

> I think this would also require changing the behavior of ReplayObservable#Last() to not block until it has n replay notification

Tbh, I can see this being a boolean param to `Last(n)` to return if `<n` are available or block if we need `=n`.

pkg/client/block/client.go
Olshansk: 🤯

pkg/client/interface.go
Olshansk: Again, I feel like we're doing the right thing here and that a thin wrapper that includes type casting is actually the correct & idiomatic approach, and not a hack :)

https://yourbasic.org/golang/type-alias/

Screenshot 2023-10-26 at 10 45 10 AM

pkg/client/block/client_test.go
Olshansk: #PUC

pkg/client/interface.go
bryanchriswhite: I think I was conflating a generics issue with this. Thanks for helping clear that up. :sweat_smile:

pkg/client/block/client_test.go
bryanchriswhite: Here I'm referring to explicitly writing tests in a "test" package (i.e. `block_test` here instead of `block`). This forces tests to exercise things from their external APIs and often drives better design choices in my opinion.

pkg/client/block/client.go
bryanchriswhite: > I think this would also require changing the behavior of ReplayObservable#Last() to not block until it has n replay notifications. That feels like probably a step in the right direction though for ReplayObservable generally; I think it makes more sense in most use cases that I can imagine. There are still some details to resolve in that case but I will continue to think about it. 🤔

Done.

pkg/client/block/client.go
bryanchriswhite: In fact, on the provide/supply side it doesn't even have to be an object with an interface type. You can provide a concrete type and it will try to figure it out. There is also a mechanism to assist it with this mapping if it struggles.

pkg/client/block/client.go
bryanchriswhite: **TL;DR: I'm less convinced on this one. I #PUC'd it up. :stuck_out_tongue_closed_eyes:**

I agree with moving the `1` to a constant where it's used as the `replayBufferSize` argument but I'm gonna push back on this one. In the former, we're considering "how many observable references should we keep". Here, we're considering "how many observable references should we pull out" in a context where we're expected to return a single observable. I updated the comment:

```go
	// Get the latest block observable from the replay observable. We only ever
	// want the last 1 as any prior latest block observable values are closed.
	// Directly accessing the zeroth index here is safe because the call to Last
	// is guaranteed to return a slice with at least 1 element.
	return bClient.latestBlockObsvblsReplay.Last(ctx, 1)[0]
```

pkg/client/block/client.go
bryanchriswhite: The relevant sentence fragment is "is expected to return an error **channel**"; however, "channel" was wrapped to the next line. I've reformatted to mitigate this confusion going forward as well as adding more detail and adding a branch for `retryOnError` to exit (i.e. instead of run forever :sweat_smile:).

pkg/client/block/client_test.go
h5law: You can achieve this by doing `package XXX_test` instead of `package XXX` at the top and keep in same TLD

bryanchriswhite: updated:

```
/*
TODO_TECHDEBT/TODO_CONSIDERATION(#XXX): this duplicates the unexported block event

type from pkg/client/block/block.go. We seem to have some conflicting preferences
which result in the need for this duplication until a preferred direction is
identified:

  - We should prefer tests being in their own pkgs (e.g. block_test)
  - this would resolve if this test were in the block package instead.
  - We should prefer to not export types which don't require exporting for API
    consumption.
  - This test is the only external (to the block pkg) dependency of cometBlockEvent.
  - We could use the //go:build test constraint on a new file which exports it
    for testing purposes.
  - This would imply that we also add -tags=test to all applicable tooling
    and add a test which fails if the tag is absent.
*/
```

pkg/client/block/client.go
bryanchriswhite: I moved `retryOnError` to its own pkg and added some tests.

pkg/client/block/client.go
bryanchriswhite: Can we make that change when/if the need arises?

pkg/client/block/client_test.go
bryanchriswhite: I still plan on doing this but otherwise, this is ready for review.

pkg/client/block/client.go
Olshansk: Yes

pkg/client/block/client.go
Olshansk: Ty for the explanation!

pkg/retry/retry_test.go
```diff
@@ -0,0 +1,337 @@
+package retry_test
+
+/* TODO_TECHDEBT: improve this test:
```
Olshansk, line 3: #EZ

pkg/client/block/client_test.go
```diff
@@ -0,0 +1,138 @@
...
+	}
+
+	// Wait a tick for the observables to be set up.
+	time.Sleep(time.Millisecond)
+
+	blockClient.Close()
+}
+
+/*
+TODO_TECHDEBT/TODO_CONSIDERATION(#XXX): this duplicates the unexported block event
```
Olshansk, line 120: Noice

Olshansk (approved):
A few last minute NITs but no blockers.

I didn't do a deep dive into `pkg/retry/retry_test.go` since it's really hard to build context into this. When the TODOs are tended to, it'll be easier to compartmentalize on a per-PR basis.

pkg/retry/retry_test.go
bryanchriswhite: :sweat_smile:

bryanchriswhite:
Waiting to merge until CI is working and green:

- #116

okdas:
Closing this as ignite.com now works. Will add a comment linking to this PR in case a similar issue arises.

Olshansk:
@okdas SHouldn't we close #116 but keep this one open?

okdas:
> @okdas Shouldn't we close #116 but keep this one open?

LOL, sorry. I pushed the comment on the wrong tab. :)

Olshansk:
@bryanchriswhite Try to merge with main to see if CI passes and please take not of #121 as well

okdas:
There's a test failure unfortunately: https://github.com/pokt-network/poktroll/actions/runs/6712974558/job/18243662555?pr=65

PR #69, opened 2023-10-16
[Service] Ran ignite scaffold type service --module service --yes

Olshansk:
## Summary

Created the service type by running the following command:

```bash
ignite scaffold type service --module service --yes
```

### Human Summary

- Added and defined the `Service` type plus supporting structures
- Move the `Service` type from `service` to `shared`

### Summary generated by Reviewpad on 17 Oct 23 14:56 UTC
This pull request introduces new protobuf files and modifies existing ones. It seems to resolve circular dependencies and add new definitions related to service configuration and identifiers. There are also some comments suggesting further research and future improvements.

## Issue

- #10

## Type of change

Select one or more:

- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

- [x] **Run all unit tests**: `make test_all_unit`

## Sanity Checklist


proto/pocket/shared/service.proto
```diff
@@ -0,0 +1,54 @@
...
+// NOTE that the `shared` package is not a Cosmos module,
+// but rather a manually created package to resolve circular dependencies.
+
+option go_package = "pocket/x/shared/types";
+
+// Service is the main service message describing all configurations for a service
+// TODO_RESEARCH(@olshansk): Should these be permissioned or permissionless?
+message Service {
+    ServiceId id = 1; // Unique and semantic identifier for the service
+    ServiceConfig config = 2; // Configuration settings for the service
```
red-0ne, line 13: If the `Service` is a standalone struct. I feel like `ServiceConfig` should not contain `Endpoint`s, which is more of a `Supplier` specific thing. It should be enough to have `RPCType`s there which describe the supported transports of that service.

I would suggest we either replace `ServiceConfig config` with `repated RPCType rpc_types` or if we chose to keep `ServiceConfig`, this later should contain the `repeated RPCType rpc_types` entry instead of `repeated Endpoint endpoints`.

If anyway we want to describe `Endpoints` . It would be better to have a `SupplierService` which will be in terms of `Service` and `Supplier` but we still need the standalone variant describing a service and its possible RPC types
```proto
message Service {
  ServiceId id = 1;
  repeated RPCType rpc_types = 2;
}
```
and optionnaly
```proto
message SupplierService {
  Service service = 1;
  ServiceConfig config = 2; // config.enpoints[].rpc_type must be a subset of Service.rpc_types
}
```

Olshansk: @red-0ne PTAL at the changes and LMK WYT

red-0ne (approved):
Looks good to me!

Just a note regarding `RPCType`. In that setup we have no way to prevent a `Supplier` to declare an `RPCType` on its `SupplierServiceConfig` that does not exist for a given `Service`. 
But this could also be the desired behavior if we assume that a `Service` should not enforce this (eg. Ethereum does not have an "official" GRPC `RPCType` but a `Supplier` could provide an implementation that does :brain: )

Olshansk:
@red-0ne That is indeed the desired behaviour!

PR #70, opened 2023-10-16
feat: add Either type

red-0ne:
## Summary

Bring the Either type which is helpful to handle errors on observables.

## Issue

Related to #31 

Either[T] is a type with generic that wraps a possible value or error. It works around tuples that are not available on channel emissions.

## Type of change

Select one or more:

- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

- [ ] **Run all unit tests**: `make test_all_unit`

## Sanity Checklist


pkg/maybe/maybe.go
```diff
@@ -0,0 +1,28 @@
+package types
```
Olshansk, line 1: I want to us to make sure to understand if we really need this or if a an error pointer would suffice.

Alternatively, this is really reminding me of optional. There's no native implementation in go, but packages are available: https://github.com/markphelps/optional

red-0ne: I don't fell like it's a good solution for our use case.

`optional` package seem to do more thinks like serialization which we don't need in our case.

Also this Maybe type is actually misnamed. It does not represent the presence or absence of a value but rather the presence of a value or an error which translates in functional programming to the `Either` type. The closest external implementation would be this one https://github.com/olehmushka/golang-toolkit/blob/main/either/either.go

Will update it to reflect this.

Olshansk: `Either` makes it clearer. Will re-review accordingly.

pkg/either/either.go
```diff
@@ -0,0 +1,28 @@
+package either
```
Olshansk, line 1: There are two big issues with this package:

1. Lack of comments
2. Using reflections in prod.

I had a conversation with ChatGPT and came up with this solution:

```go
package either

// Either represents a type that can hold either a value of type T or an error.
// It's commonly used in functional programming to encapsulate functions that might fail.
// This way, instead of returning a value and an error separately, both are combined into a single type.
type Either[T any] struct {
	value T    // holds the success value
	err   error // holds the error, if any
}

// NewEither creates a new Either instance with the given value and error.
// This function is generally not needed as Success and Error should cover most use cases.
func NewEither[T any](value T, err error) Either[T] {
	return Either[T]{value: value, err: err}
}

// Success creates a successful Either value.
func Success[T any](value T) Either[T] {
	return Either[T]{value: value, err: nil}
}

// Error creates an error Either value.
func Error[T any](err error) Either[T] {
	return Either[T]{value: default(T), err: err}
}

// IsSuccess checks if the Either contains a success value.
func (m Either[T]) IsSuccess() bool {
	return m.err == nil
}

// IsError checks if the Either contains an error.
func (m Either[T]) IsError() bool {
	return m.err != nil
}

// ValueOrError unpacks the Either value.
func (m Either[T]) ValueOrError() (T, error) {
	return m.value, m.err
}

// Helper function to get the zero value for any type.
func default[T any]() T {
	var t T
	return t
}
```

Wdyt of using this instead?

Olshansk (changes requested):

pkg/either/either.go
red-0ne: Really like this API!

We introduced reflection API to extract the zero value of the generic type which seems to be handled nicely by the `default` function. I tried it with some types (primitives, structs and pointers) and it seems to do the job pretty well.

```go
package main

import "fmt"

func Default[T any]() T {
	var t T
	return t
}

type structType struct { foo int }

func main() {
	typeInt := Default[int]()
	typeBool := Default[bool]()
	typeStruct := Default[structType]()
	typeStructPointer := Default[*structType]()

	fmt.Printf("type %T, value %v\n", typeInt, typeInt)
	fmt.Printf("type %T, value %v\n", typeBool, typeBool)
	fmt.Printf("type %T, value %v\n", typeStruct, typeStruct)
	fmt.Printf("type %T, value %v\n", typeStructPointer, typeStructPointer)
}
```
output
```
type int, value 0
type bool, value false
type main.structType, value {0}
type *main.structType, value <nil>
```

I can't remember how we ended up with the reflection solution and if it handles additional cases. cc @bryanchriswhite

bryanchriswhite: My recollection is that in a previous version there was a method that was trying to check if the value was empty, which turned out not to be as simple a comparing against nil. Here the zero value is needed for a slightly different reason and I think this is a nice way to get it using generics. :ok_hand:

bryanchriswhite: My recollection is that in a previous version there was a method that was trying to check if the value was empty, which turned out not to be as simple a comparing against nil. Here the zero value is needed for a slightly different reason and I think this is a nice way to get it using generics. :ok_hand: 

I see a couple of bugs:

1. `default` is a reserved word and not a valid function name.
2. it's currently trying to pass `T`, which is a generic type parameter, as an argument to the function. It should invoke the function like `default[T]()` instead.

bryanchriswhite: Wdyt about the following:

```go
package either

// Either represents a type that can hold either a value of type T or an error.
// It's commonly used in functional programming to encapsulate functions that might fail.
// This way, instead of returning a value and an error separately, both are combined into a single type.
type Either[R, L any] struct {
	right R    // conventionally holds the success value
	left  L // conventionally holds the error, if any
}

type EitherErr[T any]  Either[T, error]

func NewEither[R, L any](right R, left L) Either[R, L] {
	return Either[R, L]{right: right, left: left}
}

// NewEither creates a new Either instance with the given value and error.
// This function is generally not needed as Success and Error should cover most use cases.
func NewEitherErr[T any](value T, err error) EitherErr[T] {
	return EitherErr[T](NewEither[T, error](value, err))
}

// Success creates a successful Either value.
func Success[T any](value T) EitherErr[T] {
	return EitherErr[T]{right: value, left: nil}
}

// Error creates an error Either value.
func Error[T any](err error) EitherErr[T] {
	return EitherErr[T]{right: anyDefault[T](), left: err}
}

// IsSuccess checks if the Either contains a success value.
func (m EitherErr[T]) IsSuccess() bool {
	return m.left == nil
}

// IsError checks if the Either contains an error.
func (m EitherErr[T]) IsError() bool {
	return m.left != nil
}

// ValueOrError unpacks the Either value.
func (m EitherErr[T]) ValueOrError() (T, error) {
	return m.right, m.left
}

// Helper function to get the zero value for any type.
func anyDefault[T any]() T {
	var t T
	return t
}
```

Olshansk: @bryanchriswhite I think a few things:

1. Some comments need to be updated. E.g. `// conventionally holds the success value`
2. LGTM overall, let's go with it.
3. I don't think it's possible for any group of engineers to over-engineering a generic type as much as this 😅

pkg/either/either.go
```diff
@@ -0,0 +1,43 @@
...
+
+// Either represents a type that can hold either a value of type T or an error.
+// It's commonly used in functional programming to encapsulate functions that might fail.
+// This way, instead of returning a value and an error separately, both are combined into a single type.
+type Either[R any] struct {
+	right R     // conventionally holds the success value
+	left  error // conventionally holds the error, if any
```
Olshansk, line 8:
```suggestion
	left  error // holds the error, if any
```

pkg/either/either.go
```diff
@@ -0,0 +1,43 @@
+package either
+
+// Either represents a type that can hold either a value of type T or an error.
+// It's commonly used in functional programming to encapsulate functions that might fail.
+// This way, instead of returning a value and an error separately, both are combined into a single type.
+type Either[R any] struct {
```
Olshansk, line 6: What is the thinking behind Not having an `EitherErr`?

It feels kind of weird that EIther is not a generic `R, L` with a helper to define a subset called `R, error`

Olshansk (changes requested):

pkg/either/either.go
red-0ne: Because `L` would always be of `error` type. We would be adding a superfluous generic type describing it.

Also, if we go with the `EitherErr`, the definition in the comments abobe `Either` struct would apply to `EitherErr` and not `Either` which makes `EitherErr` be the `Either` itself (by definition).

Olshansk (approved):

Issue #14, opened 2023-10-05
[Proxy] Foundation for the Proxy submodule


Olshansk:
## Objective

Create a foundation for the `Proxy` submodule in the [poktroll](https://github.com/pokt-network/poktroll) repo for the Shannon upgrade.

## Origin Document

The business logic for proxying relays as it relates to the [Pocket Network](pokt.network) from the [Relay Mining](https://arxiv.org/abs/2305.10672) paper.

## Goals

- Enable an E2E HTTP relay with basic session and account management
- Build the foundation for the `Proxy` submodule for the `Relayer`
- Identify time-consuming TODOs to be done during the integration and testing stages

## Deliverables

- [ ] Reference structures and types in the [poktroll-alpha repo](https://github.com/pokt-network/poktroll-alpha), the [v1 repo](https://github.com/pokt-network/pocket) and [v0 repo](https://github.com/pokt-network/pocket-core) to understand and use other definitions as a foundation
- [ ] A **single PR** that introduces the `Proxy` structure & interface
- [ ] A **single PR** that adds business logic, with the appropriate placeholders & TODOs, to add `Proxy` functionality
- [ ] A **single PR** that integrates the proxy with other components (e.g. `SessionManager`, `Miner`, `Relayer`, `Client`, etc...) to enable E2E functionality
- [ ] **Multiple small PRs** that migrate the necessary utilities from the `poktroll-alpha` repo to the `poktroll` repo

## Non-goals / Non-deliverables

- WebSocket support (HTTP only is okay)
- Accounting for all possible edge cases and situations

## General deliverables

- [ ] **Comments**: Add/update TODOs and comments alongside the source code so it is easier to follow.
- [ ] **Testing**: Add new tests (unit and/or E2E) to the test suite.
- [ ] **Makefile**: Add new targets to the Makefile to make the new functionality easier to use.
- [ ] **Documentation**: Update architectural or development READMEs; use [mermaid](https://mermaid-js.github.io/mermaid/) diagrams where appropriate.

---

**Creator**: @Olshansk 
**Co-Owners**: @bryanchriswhite @red-0ne

PR #80, opened 2023-10-19
[RelayerProxy] feat: add RelayerProxy interface

red-0ne:
### Human Summary

Add interface for the RelayerProxy

### Summary generated by Reviewpad on 20 Oct 23 17:13 UTC
This pull request adds a new file `interface.go` to the `pkg/relayer/proxy` directory. This file defines an interface `RelayerProxy` that serves relays to the application. It includes methods for starting and stopping supported proxies, as well as notifying the miner about the relays that have been served.

## Issue

- #14 

## Type of change

Select one or more:

- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

## Sanity Checklist


pkg/proxy/interface.go
```diff
@@ -0,0 +1,26 @@
...
+	Start(ctx context.Context) error
+
+	// Stop stops all the supported proxies or returns an error if any of them fails.
+	Stop() error
+
+	// ServedRelays returns an observable that notifies the miner about the relays that has been served.
+	// A served relay is a relay that its RelayRequest's signature and session has been verified
+	// and its RelayResponse has been signed and successfully sent to the client.
+	ServedRelays() observable.Observable[*types.Relay]
+}
```
Olshansk, line 26:
```suggestion
package proxy

import (
	"context"

	"pocket/pkg/observable"
	"pocket/x/service/types"
)

// RelayerProxy is the interface for the proxy that serves relays to the application.
// It is responsible for starting and stopping all supported proxies.
// While handling requests and responding in a closed loop, it also notifies
// the miner about the relays that have been served.
type RelayerProxy interface {

	// Start starts all supported proxies and returns an error if any of them fail to start.
	Start(ctx context.Context) error

	// Stop stops all supported proxies and returns an error if any of them fail.
	Stop() error

	// ServedRelays returns an observable that notifies the miner about the relays that have been served.
	// A served relay is one whose RelayRequest's signature and session have been verified,
	// and its RelayResponse has been signed and successfully sent to the client.
	ServedRelays() observable.Observable[*types.Relay]
}
```

Olshansk: Screenshot 2023-10-19 at 7 18 50 PM

Olshansk (changes requested):

Olshansk:
@red-0ne Two things:
1. Please make sure to update PR metadata (see screenshot below)
2. When you do a `squash and merge`, make sure to update the base branch first.

Screenshot 2023-10-20 at 9 47 52 AM

pkg/relayerproxy/interface.go
```diff
@@ -0,0 +1,26 @@
+package relayerproxy
```
Olshansk, line 1: I noticed that the package name is `relayerproxy`.

Not a blocker but curious to understand the intuition of this instead of 

`pkg/relayer/proxy/interface.go`?

Olshansk (changes requested):
Just want to understand the intuition/plan behind the package naming.

pkg/relayerproxy/interface.go
red-0ne: Indeed `pkg/relayer/proxy/` is better. I first had `pkg/proxy/` and didn't want to have `pkg/proxy/proxy`.

Will update the code and PR naming :+1:

red-0ne: This would also mean that the `Miner` module would live in `pkg/relayer/miner` right?

Olshansk: Yup!

pkg/relayer/proxy/interface.go
```diff
@@ -0,0 +1,26 @@
...
+	// Start starts all supported proxies and returns an error if any of them fail to start.
+	Start(ctx context.Context) error
+
+	// Stop stops all supported proxies and returns an error if any of them fail.
+	Stop() error
+
+	// ServedRelays returns an observable that notifies the miner about the relays that have been served.
+	// A served relay is one whose RelayRequest's signature and session have been verified,
+	// and its RelayResponse has been signed and successfully sent to the client.
+	ServedRelays() observable.Observable[*types.Relay]
```
Olshansk, line 25: When we get to the implementation, I want to see how this is used and might ask to rename it to `RelayServer` since it does not serve/return static data.

Not a blocker for this PR (merge as is), but will follow up.

Olshansk (approved):
Thanks for keeping the PR small!

pkg/relayer/proxy/interface.go
red-0ne: Basically `RelayerProxy.ServedRelays` would be ranged over from the miner to update the corresponding `SMST`.

It would be very similar to what we have in the alpha repo (simplified pseudo go version):
```go
// goHandleRelays is called as a go routine when the `Miner` starts
func (m *Miner) goHandleRelays() {
  ch := m.relayerProxy.ServedRelays().Subscribe().Ch()
  for relay := range ch {
    go m.goHandleSingleRelay(relay)
  }
}

func (m *Miner) goHandleSingleRelay(relay *types.Relay) {
    smst := m.getTreeForSession()
    smst.Update(relay.Hash(), relay.Marshal(), 1)
}
```

Will document this in the next PR.

Olshansk: This explanation makes me think even more that we should rename it.

You don't subscribe to `ServedRelays`, but rather to a `RelayServer` or `RelayObservable` or other.

@bryanchriswhite  wdyt?

PR #82, opened 2023-10-20
[RelayerProxy] feat: implement relayerProxy struct

red-0ne:
### Human Summary

Initial implementation of the relayerProxy struct. It complies with the RelayerProxy interface, but none of its methods are actually implemented.

The only non-existing client assumed to be present is the `BlockClient`.

### Summary generated by Reviewpad on 23 Oct 23 15:24 UTC
This pull request adds a new file called `proxy.go` in the `pkg/relayer/proxy` directory. The file contains code for a `relayerProxy` struct, which implements the `RelayerProxy` interface. The `relayerProxy` struct has several fields and methods. It also introduces a `ProvidedService` struct and a `buildProvidedServices` function. There are some TODOs marked within the code indicating incomplete parts that need to be implemented.

## Issue

- #14 

## Type of change

Select one or more:

- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

## Sanity Checklist


Olshansk:
@red-0ne Please make sure to update these.
Screenshot 2023-10-20 at 9 54 49 AM

pkg/relayer/proxy/proxy.go
```diff
@@ -0,0 +1,110 @@
...
+	sessiontypes "pocket/x/session/types"
+	suppliertypes "pocket/x/supplier/types"
+)
+
+var (
+	_ RelayerProxy = &relayerProxy{}
+)
+
+type relayerProxy struct {
+	// keyName is the supplier's key name in the keyring. It is used along with the keyring to
```
Olshansk, line 23:
```suggestion
	// keyName is the supplier's key name in the keyring, Cosmos' keybase. It is used along with the keyring to
```

pkg/relayer/proxy/proxy.go
```diff
@@ -0,0 +1,110 @@
...
+// It loops over the retrieved `SupplierServiceConfig` and, for each `SupplierEndpoint`, it creates the necessary
+// server and client to populate the corresponding `ProvidedService` struct in the map.
+func buildProvidedServices(
+	ctx context.Context,
+	supplierQuerier suppliertypes.QueryClient,
+) map[string]ProvidedService {
+	panic("not implemented")
+}
+
+// TODO_INCOMPLETE: Add the appropriate server and client interfaces to be implemented by each RPC type.
```
Olshansk, line 105:
```suggestion
// TODO_INCOMPLETE(@red-0ne)_: Add the appropriate server and client interfaces to be implemented by eachRPC type.
```

pkg/relayer/proxy/proxy.go
```diff
@@ -0,0 +1,110 @@
...
+	_ RelayerProxy = &relayerProxy{}
+)
+
+type relayerProxy struct {
+	// keyName is the supplier's key name in the keyring. It is used along with the keyring to
+	// get the supplier address and sign the relay responses.
+	keyName string
+	keyring keyring.Keyring
+
+	// blocksClient is the client used to get the latest block height from the blockchain
```
Olshansk, line 28:
```suggestion
	// blocksClient is the client used to get the block at the latest height from the blockchain
```

pkg/relayer/proxy/proxy.go
```diff
@@ -0,0 +1,110 @@
...
+	// and be notified about new blocks to update the current session when needed.
```
Olshansk, line 29:
```suggestion
	// and be notified of new incoming blocks. One of its uses is to update the current session data.
```

pkg/relayer/proxy/proxy.go
```diff
@@ -0,0 +1,110 @@
...
+	blocksClient blocktypes.BlockClient
+
+	// accountsQuerier is the querier used to get the application public key from the blockchain,
```
Olshansk, line 32:
```suggestion
	// accountsQuerier is the querier used to get account data (e.g. app publicKey) from the blockchain,
```

pkg/relayer/proxy/proxy.go
```diff
@@ -0,0 +1,110 @@
...
+	// which is used to verify the relay request signatures.
```
Olshansk, line 33:
```suggestion
	// which is used for things like verifying the relay request signatures.
```

pkg/relayer/proxy/proxy.go
```diff
@@ -0,0 +1,110 @@
...
+	accountsQuerier accounttypes.QueryClient
+
+	// supplierQuerier is the querier used to get the supplier's advertised information from the blockchain,
+	// which contains the supported services, the RPC types, and endpoints needed to query the services.
```
Olshansk, line 37:
```suggestion
	// which contains the supported services, RPC types, endpoints, etc...
```

pkg/relayer/proxy/proxy.go
```diff
@@ -0,0 +1,110 @@
...
+	// which is needed to check if the relay proxy should be serving an incoming relay request.
+	sessionQuerier sessiontypes.QueryClient
+
+	// providedServices is a map of the services provided by the relayer proxy. Each provided service
+	// has the necessary information to start the server that listens for incoming relay requests and
+	// the client that proxies the request to the supported native service.
+	providedServices map[string]ProvidedService
+
+	// servedRelays is an observable that notifies the miner about the relays that have been served.
+	servedRelays observable.Observable[*types.Relay]
```
Olshansk, line 50: Wdyt of renaming this to `relayServer` or `relayObservable`?

`servedRelays` to me feels like a `read operation` but this is a `subscribe & listen operation`.

Olshansk: @bryanchriswhite Curious for your opinion on naming here.

pkg/relayer/proxy/proxy.go
```diff
@@ -0,0 +1,110 @@
...
+
+	// servedRelaysProducer is a channel that emits the relays that have been served so that the
+	// servedRelays observable can fan out the notifications to its subscribers.
```
Olshansk, line 53: Make sure to update the comment if we rename the above

pkg/relayer/proxy/proxy.go
```diff
@@ -0,0 +1,110 @@
...
+	servedRelaysProducer chan<- *types.Relay
+}
+
+func NewRelayerProxy(
```
Olshansk, line 57: Really clean!

pkg/relayer/proxy/proxy.go
```diff
@@ -0,0 +1,110 @@
...
+		accountsQuerier:      accountQuerier,
+		supplierQuerier:      supplierQuerier,
+		sessionQuerier:       sessionQuerier,
+		providedServices:     providedServices,
+		servedRelays:         servedRelays,
+		servedRelaysProducer: servedRelaysProducer,
+	}
+}
+
+func (rp *relayerProxy) Start(ctx context.Context) error {
```
Olshansk, line 83: Consider adding godoc comments here and elsewhere. I bet chatgpt could do it for you ;)

pkg/relayer/proxy/proxy.go
```diff
@@ -0,0 +1,110 @@
...
+	panic("not implemented")
```
Olshansk, line 84:
```suggestion
	panic("TODO: implement relayerProxy.start")
```

Ditto below please

Olshansk (approved):
Overall feedback: including specific examples of how something is used is very valuable but it can get outdated really fast. Mention it as an example to avoid the reader thinking it's THE ONLY thing it's used for.

Other than that, REALLLY great work!

Olshansk (changes requested):
Overall feedback: including specific examples of how something is used is very valuable but it can get outdated really fast. Mention it as an example to avoid the reader thinking it's THE ONLY thing it's used for.

Other than that, REALLLY great work!

Olshansk (changes requested):
Overall feedback: including specific examples of how something is used is very valuable but it can get outdated really fast. Mention it as an example to avoid the reader thinking it's THE ONLY thing it's used for.

Other than that, REALLLY great work!

pkg/relayerproxy/proxy/proxy.go
```diff
@@ -0,0 +1,111 @@
+package proxy
+
+import (
+	"context"
+
+	sdkclient "github.com/cosmos/cosmos-sdk/client"
+	"github.com/cosmos/cosmos-sdk/crypto/keyring"
+	accounttypes "github.com/cosmos/cosmos-sdk/x/auth/types"
+
+	blocktypes "pocket/pkg/client"
```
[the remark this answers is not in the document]
red-0ne: It is actually because the package does not exist yet. This will definitely break the build; we could TODO-comment it or have a temporary interface for it, until it gets merged. WDYT

pkg/relayer/proxy/proxy.go
red-0ne: Did you mean this instead?:
```
// keyName is the supplier's key name in the Cosmos' keybase. It is used along with the keyring to...
```

pkg/relayer/proxy/proxy.go
red-0ne: It is exactly its usage in the context of the `RelayerProxy`. IMO, we should stick to `It is used to ...` in this comment and will definitely update it when its usage evolves (probably with ring signatures)

pkg/relayer/proxy/proxy.go
red-0ne: It is indeed a read operation. The caller is responsible of subscribing (creating the observer) and listening to its channel.

I should have implemented `ServedRelays` since it's a one-liner which is only returning this property.

red-0ne:
@Olshansk , Please check my comments above before I make more changes.

pkg/relayer/proxy/proxy.go
Olshansk: > It and will definitely update it when its usage evolves (probably with ring signatures)

Sounds good. I won't push hard on it.

Personally, I've never people do a good job at updating documentation unless someone (e.g. the reviewer) remembers to tell the implementor to change documentation.

As long as you remember to enforce that as well, I'm okay with a 🤝

pkg/relayer/proxy/proxy.go
Olshansk: I still feel like `relayServer` or `relayObservable` is more accurate. Again, not going to push hard on it, but would like another POV @bryanchriswhite

Olshansk (changes requested):

pkg/relayer/proxy/proxy.go
bryanchriswhite: **TL;DR, personally, I think `servedRelays` is a good name. :shrug:**

I don't feel like adding the word "observer" adds any real value as it's clear from its type that it is an observable. "Server" doesn't feel right to me either because the observable itself has nothing to do with any network or request handling which is what I would expect from a "server" kind of thing.

My intuition for  naming these has been to put the focus on the thing that the observable is returning as the observable is merely an async primitive, like a promise or future from other languages. I think the name "servedRelays" makes it pretty unambiguously clear what kind of data you're going to get from that observable.

Olshansk: Thanks for the response. Let's go with `servedRelays`

red-0ne:
Commented blockClient occurrences until available.

pkg/relayerproxy/proxy/proxy.go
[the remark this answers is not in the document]
red-0ne: Resolved now by commenting the BlockClient import and occurrences in the code.

Olshansk (approved):

Issue #16, opened 2023-10-05
[Testing] E2E Testing Framework


Olshansk:
## Objective

Migrate the previously built E2E Testing framework to this repo.

## Origin Document

The [pocket repo](https://github.com/pokt-network/pocket) implements an [E2E testing framework](https://github.com/pokt-network/pocket/tree/main/e2e) that needs to be migrated to the poktroll repo.

## Goals

- Enable E2E regression testing
- Enable automation that'll be used in load testing

## Deliverables

- [x] A single PR that adapts [this directory from the v1 repo](https://github.com/pokt-network/pocket/tree/main/e2e) to this repository
- [x] A single PR that introduces an E2E test for staking/unstaking
- [ ] A single PR that introduces an E2E test for a relay that verifies on-chain claims & proofs

## Non-goals / Non-deliverables

- Provide the tooling for every type of E2E Test that we may need
- Automated E2E tests in our CI/CD pipeline

## General deliverables

- [ ] **Comments**: Add/update TODOs and comments alongside the source code so it is easier to follow.
- [ ] **Testing**: Add new tests (unit and/or E2E) to the test suite.
- [ ] **Makefile**: Add new targets to the Makefile to make the new functionality easier to use.
- [ ] **Documentation**: Update architectural or development READMEs; use [mermaid](https://mermaid-js.github.io/mermaid/) diagrams where appropriate.

---

**Creator**: @Olshansk 
**Co-Owners**: @bryanchriswhite @red-0ne @h5law

Olshansk:
@h5law As I was working on #59, I saw that there is a lot of infra & utils to do things like (for example)

1. Run simulation ([advanced simulation link](https://docs.cosmos.network/main/learn/advanced/simulation))
2. Startup a mock network in tests for a CLI context (e.g. see `x/application/client/cli/helpers_test.go` in #59 and code below)

My added ask/requirement is if you could research & look into those as well (as you're enabling a cucamber/gherking development framework) to make sure we leverage as much as possible of Cosmos' infrastructure and don't recreate the wheel.

---

```go
func networkWithApplicationObjects(t *testing.T, n int) (*network.Network, []types.Application) {
	t.Helper()
	cfg := network.DefaultConfig()
	state := applicationModuleGenesis(n)
	buf, err := cfg.Codec.MarshalJSON(state)
	require.NoError(t, err)
	cfg.GenesisState[types.ModuleName] = buf
	return network.New(t, cfg), state.ApplicationList
```

[...]

PR #88, opened 2023-10-21
[E2E] Add (Un)Stake Tests

h5law:
### Human Summary

This PR adds a generic e2e test for staking and unstaking a gateway. It assumes the gateway is not previously staked. This can be used for any staking actor that doesn't also stake for services in its current form.

### Summary generated by Reviewpad on 27 Oct 23 20:35 UTC
This pull request includes changes to the e2e/tests/help.feature, e2e/tests/init_test.go, e2e/tests/send.feature, and e2e/tests/stake.feature files. The changes involve adding and modifying scenarios related to user interactions with the pocketd binary, including help commands, account staking and unstaking, and sending uPOKT between accounts.

## Issue

Fixes in part [16](https://github.com/pokt-network/poktroll/issues/16)

## Type of change

Select one or more:

- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

## Sanity Checklist


e2e/tests/init_test.go
```diff
@@ -5,26 +5,37 @@ package e2e
...
+	addrRe           *regexp.Regexp
+	amountRe         *regexp.Regexp
+	accNameToAddrMap = make(map[string]string)
+	keyRingFlag      = "--keyring-backend=test"
+)
 
 func init() {
-	addrRe = regexp.MustCompile(`address:\s+(pokt1\w+)`)
+	addrRe = regexp.MustCompile(`address: (\S+)\s+name: (\S+)`)
+	amountRe = regexp.MustCompile(`amount: "(.+?)"\s+denom: upokt`)
```
bryanchriswhite, line 26: It looks like previous version was expressing whitespace in terms of `\s+`,  my assumption would be to mitigate some shell formatting thing from causing a false negative match due to multiple spaces, a tab, or other whitespace character(s) instead of ` `. Wdyt about the following?

```suggestion
	addrRe = regexp.MustCompile(`address:\s+(\S+)\s+name:\s+(\S+)`)
	amountRe = regexp.MustCompile(`amount:\s+"(.+?)"\s+denom:\s+upokt`)
```

e2e/tests/init_test.go
```diff
@@ -123,6 +123,85 @@ func (s *suite) TheUserShouldWaitForSeconds(dur int64) {
...
+	if err != nil {
+		s.Fatalf("error unstaking %s: %s", actorType, err)
+	}
+	s.pocketd.result = res
+}
+
+func (s *suite) TheForAccountIsNotStaked(actorType, accName string) {
+	found, _ := s.getStakedAmount(actorType, accName)
+	if found {
+		s.Fatalf("account %s is staked", accName)
```
Olshansk, line 164:
```suggestion
		s.Fatalf("account %s should not be staked", accName)
```

e2e/tests/init_test.go
```diff
@@ -123,6 +123,85 @@ func (s *suite) TheUserShouldWaitForSeconds(dur int64) {
...
+	}
+}
+
+func (s *suite) TheForAccountIsStakedWithUpokt(actorType, accName string, amount int64) {
+	found, stakeAmount := s.getStakedAmount(actorType, accName)
+	if !found {
+		s.Fatalf("account %s is not staked", accName)
```
Olshansk, line 171:
```suggestion
		s.Fatalf("account %s should be staked", accName)
```

e2e/tests/init_test.go
```diff
@@ -123,6 +123,85 @@ func (s *suite) TheUserShouldWaitForSeconds(dur int64) {
...
+		s.Fatalf("account %s stake amount is not %d", accName, amount)
+	}
+}
+
+func (s *suite) getStakedAmount(actorType, accName string) (bool, int) {
+	s.Helper()
+	args := []string{
+		"query",
+		actorType,
+		fmt.Sprintf("list-%s", actorType),
```
Olshansk, line 183: Have you checked if using `show` instead of `list` is easier?

Screenshot 2023-10-26 at 3 26 44 PM

e2e/tests/stake.feature
```diff
@@ -0,0 +1,25 @@
...
+    Given the user has the pocketd binary installed
+    And the "gateway" for account "gateway1" is not staked
+    And the account "gateway1" has a balance greater than "1000" uPOKT
+    When the user stakes a "gateway" with "1000" uPOKT from the account "gateway1"
+    Then the user should be able to see standard output containing "txhash:"
+    And the user should be able to see standard output containing "code: 0"
+    And the pocketd binary should exit without error
+    And the user should wait for "5" seconds
+    And the "gateway" for account "gateway1" is staked with "1000" uPOKT
+    And the account balance of "gateway1" should be "1000" uPOKT "less" than before
```
Olshansk, line 13:
```suggestion
    Scenario: User can stake a Gateway
	    Given the user has the pocketd binary installed
	    And the "gateway" for account "gateway1" is not staked
	    And the account "gateway1" has a balance greater than "1000" uPOKT
	    When the user stakes a "gateway" with "1000" uPOKT from the account "gateway1"
	    Then the user should be able to see standard output containing "txhash:"
	    And the user should be able to see standard output containing "code: 0"
	    And the pocketd binary should exit without error
	    And the user should wait for "5" seconds
	    And the "gateway" for account "gateway1" is staked with "1000" uPOKT
	    And the account balance of "gateway1" should be "1000" uPOKT "less" than before
```

e2e/tests/stake.feature
```diff
@@ -0,0 +1,25 @@
...
+
+    Scenario: User can unstake a Gateway
```
Olshansk, line 15: Make sure the indent above works and ditto here.

Olshansk (changes requested):
These tests are absolute 🔥 !

bryanchriswhite (approved):
+1 to @Olshansk's indentation recommendation - I agree this is :fire:

e2e/tests/init_test.go
h5law: So the reason I am not using `show-gateway` is because it errors if the key isn't found (aka not staked) which causes the test to fail. We could instead check the combined output for this error message but it overcomplicates what we are trying to do here.

Plus these tests should never result in a paginated result so I think this is fine.

Olshansk (approved):
LFG!

PR #92, opened 2023-10-23
[Miner] feat: add the map channel observable operator

bryanchriswhite:
### Human Summary

Adds `Map`, a function which transforms a "source" observable into a new, "destination", observable via the provided `transformFn`.

### Summary generated by Reviewpad on 24 Oct 23 12:39 UTC
This pull request introduces the map channel observable operator. It includes the implementation of the `Map` function in the `pkg/observable/channel/map.go` file, as well as tests in the `pkg/observable/channel/map_test.go` file. The `Map` function takes an observable and applies a transform function to each notification received from the observable. If the transform function returns a `skip` boolean of true, the notification is skipped and not emitted to the resulting observable. The tests cover various scenarios of mapping bytes to palindromes and checking their validity. The patch also includes improvements to comments and test names for better readability and clarity.

## Issue

Relates to:
- #13 
- #64

`Map` enables transforming observables, including their generic types, which is a dependency of `EventsQueryClient` (#64).

## Type of change

Select one or more:

- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

## Sanity Checklist

### DIFF
--- a/pkg/observable/channel/map.go
+++ b/pkg/observable/channel/map.go
@@ -1,0 +1,30 @@
+package channel
+
+import (
+	"context"
+
+	"pocket/pkg/observable"
+)
+
+func Map[S, D any](
+	ctx context.Context,
+	srcObservable observable.Observable[S],
+	// TODO_CONSIDERATION: if this were variadic, it could simplify serial transformations.
+	transformFn func(src S) (dst D, skip bool),
+) observable.Observable[D] {
+	dstObservable, dstProducer := NewObservable[D]()
+	srcObserver := srcObservable.Subscribe(ctx)
+
+	go func() {
+		for srcNotification := range srcObserver.Ch() {
+			dstNotification, skip := transformFn(srcNotification)
+			if skip {
+				continue
+			}
+
+			dstProducer <- dstNotification
+		}
+	}()
+
+	return dstObservable
+}
--- a/pkg/observable/channel/map.go
+++ b/pkg/observable/channel/map.go
@@ -6,6 +6,10 @@ import (
 	"pocket/pkg/observable"
 )
 
+// Map transforms the given observable by applying the given transformFn to each
+// notification received from the observable. If the transformFn returns a skip
+// bool of true, the notification is skipped and not emitted to the resulting
+// observable.
 func Map[S, D any](
 	ctx context.Context,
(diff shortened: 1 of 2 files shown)


pkg/observable/channel/map.go
```diff
@@ -0,0 +1,30 @@
+package channel
+
+import (
+	"context"
+
+	"pocket/pkg/observable"
+)
+
+func Map[S, D any](
```
Olshansk, line 9: Why do we need this? #PUC

Also, consider `MapObservable` but not a strong stance.

pkg/observable/channel/map_test.go
```diff
@@ -0,0 +1,86 @@
...
+						require.NotEmptyf(t, string(tt.wordBz), word.backwards)
+						require.Falsef(t, word.IsValid(), "palindrome should be invalid")
+					}
+				}
+			}()
+		})
+	}
+}
+
+// palindrome is a word that is spelled the same forwards and backwards.
```
Olshansk, line 62: It would also help to understand why need such a struct in `map_test`. #PUC

pkg/observable/channel/map_test.go
```diff
@@ -0,0 +1,86 @@
...
+import (
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"pocket/pkg/observable/channel"
+)
+
+func TestMapWord_BzToPalindrome(t *testing.T) {
```
Olshansk, line 12:
```suggestion
func TestMap_Word_BzToPalindrome(t *testing.T) {
```

Olshansk (changes requested):

pkg/observable/channel/map.go
bryanchriswhite: :pray: godoc comment added. :raised_hands: 

I think adding "observable" would be inconsistent as this isn't an observable implementation itself. It uses the `channel.NewObservable()` constructor for the new observable that it returns.

Olshansk: Sgtm!

Olshansk (approved):
Everything about this PR (size, comments/docs, scope) is absolute 🔥

PR #93, opened 2023-10-23
[Miner] feat: add replay observable

bryanchriswhite:
### Human Summary

Adds a new generic observable type, `replayObservable`, which extends `channelObservable` to buffer the last `n` notifications  and replay them to new subscribers prior to sending new notifications.

### Summary generated by Reviewpad on 30 Oct 23 10:26 UTC
This pull request modifies multiple files in the codebase. Here is a summary of the changes:

1. The file `pkg/observable/channel/observable.go` has been refactored to use the `observerManager` type to handle concurrent-safe read/write access to observers. Several import statements, variable declarations, and method implementations have been modified or removed.

2. The test case `TestChannelObservable_ObserversCloseOnPublishChannelClose` in the file `observable_test.go` has been updated with a modified skipped test message.

3. A new file named `replay.go` has been introduced in the `pkg/observable/channel` package. This file implements a `replayObservable` type that allows the replaying of the last `replayBufferSize` number of values published to the source observable to new observers. It also includes functions for subscribing, unsubscribing, and accumulating replay values.

4. The file `observer.go` has undergone changes related to the onUnsubscribe field, the UnsubscribeFunc type, and the NewObserver function.

5. The file `pkg/observable/interface.go` has changes that introduce a new type called `ReplayObservable` and modify the existing `Observable` type. The `ReplayObservable` type replays the last `n` values for new observers, and the `Observable` type no longer includes the `Next` method.

6. The file `observer_test.go` has undergone various changes related to import statements, function modifications, and assertion updates.

7. A new file named `replay_test.go` has been added in the `pkg/observable/channel` package. This file contains test functions for the replay functionality of the observable package. The tests cover the basic functionality of the replay feature, behavior when the replay buffer is full, and blocking and timing out scenarios.

Overall, these changes involve refactoring, introducing new functionality, and updating tests to ensure the correct behavior of the codebase.

## Issue

Relates to:
- #13 
- #64

`Replay` enables support for async return values to be read multiple times, concurrently. This enables simplification around storing and referencing async return values, especially in concurrent scenarios.

## Type of change

Select one or more:

- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

## Sanity Checklist


pkg/observable/channel/replay.go
```diff
@@ -0,0 +1,109 @@
...
+
+	srcObserver := srcObsvbl.Subscribe(ctx)
+	go replayObsvbl.goBufferReplayNotifications(srcObserver)
+
+	return replayObsvbl
+}
+
+// Next synchronously returns the next value from the observable. This will always
+// return the first value in the replay buffer, if it exists.
+func (ro *replayObservable[V]) Next(ctx context.Context) V {
```
red-0ne, line 52: I think what we'd need is a `Last(ctx, n)` which better serves the LastBlock & LastNBlock like scenarios. WDYT?

red-0ne (commented):
This is great and useful to many scenarios.

pkg/observable/channel/replay.go
bryanchriswhite: Yes, thank you! Not sure why I couldn't see it for myself. :sweat_smile:

bryanchriswhite: I think the reason I went with `Next` was because we're using `replayObservable[V]` as an `Observable[V]` interface, which meant that whatever method we add had to be included in the interface. 

Thinking about it some more, I feel like we should:

1. Remove this method from `Observable[V]`
2. Add a `ReplayObservable[V]` interface type which embeds `Observable[V]` and includes `Last()`
3. Change the return type of `Replay()` to `ReplayObservable[V]` so that it's explicit everywhere that we're using a replay observable. This should improve readability.

pkg/observable/channel/replay.go
```diff
@@ -0,0 +1,134 @@
...
+	replayBufferSize int,
+) (observable.ReplayObservable[V], chan<- V) {
+	obsvbl, publishCh := NewObservable[V]()
+	return Replay[V](ctx, replayBufferSize, obsvbl), publishCh
+}
+
+// Replay returns an observable which replays the last replayBufferSize number of
+// values published to the source observable to new observers, before publishing
+// new values.
+func Replay[V any](
```
Olshansk, line 41: I don't think this function should be publically exposed. I was a bit confused if I should be using `NewReplayObservable` or `Replay` based on the structures here.

pkg/observable/channel/replay_test.go
```diff
@@ -0,0 +1,87 @@
...
+	"github.com/stretchr/testify/require"
+	"testing"
+	"time"
+
+	"pocket/pkg/observable/channel"
+)
+
+func TestReplayObservable(t *testing.T) {
+	var (
+		n, values   = 3, []int{1, 2, 3, 4, 5}
```
Olshansk, line 14: NIT: more semantic name for n (e.g. replayBufferSize)

Olshansk (changes requested):

pkg/observable/channel/replay.go
```diff
@@ -0,0 +1,134 @@
...
+	"pocket/pkg/observable"
+)
+
+const replayNotificationTimeout = 1 * time.Second
+
+var _ observable.ReplayObservable[any] = &replayObservable[any]{}
+
+type replayObservable[V any] struct {
+	*channelObservable[V]
+	// replayBufferSize is  the number of replayBuffer to buffer so that they
```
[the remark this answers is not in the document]
bryanchriswhite: Yes, thank you. I think this was my IDE getting carried away with a variable rename. Nice catch. 🙏

pkg/observable/channel/replay.go
```diff
@@ -0,0 +1,134 @@
...
+	// can be replayed to new observers.
+	replayBufferSize int
+	// replayBufferMu protects replayBuffer from concurrent access/updates.
+	replayBufferMu sync.RWMutex
+	// replayBuffer is the buffer of notifications into which new notifications
```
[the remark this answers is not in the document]
bryanchriswhite: A channel would only allow reading any given value once whereas what we need here is a buffer that will allow us to read those same value each time a new observer subscribes.

pkg/observable/interface.go
```diff
@@ -7,12 +7,18 @@ import "context"
 // grow, other packages (e.g. https://github.com/ReactiveX/RxGo) can be considered.
 // (see: https://github.com/ReactiveX/RxGo/pull/377)
 
+// ReplayObservable is an observable which replays the last n values published
```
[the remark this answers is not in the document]
bryanchriswhite: This is used in the block client to replay the latest block. It simplified a few things and removes the need for additional synchronization.

pkg/observable/channel/replay.go
```diff
@@ -0,0 +1,134 @@
...
+	srcObserver := srcObsvbl.Subscribe(ctx)
+	go replayObsvbl.goBufferReplayNotifications(srcObserver)
+
+	return replayObsvbl
+}
+
+// Last synchronously returns the last n values from the replay buffer. This will always
+// return the first value in the replay buffer, if it exists.
+func (ro *replayObservable[V]) Last(ctx context.Context, n int) []V {
+	tempObserver := ro.Subscribe(ctx)
```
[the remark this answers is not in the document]
bryanchriswhite: 1. We want `Last()` to block until it has `n` values to return.
2. It simplifies synchronization by leveraging existing (and more idiomatic) mechanisms.

pkg/observable/channel/replay.go
```diff
@@ -0,0 +1,134 @@
...
+	// to hold all replay (buffered) replayBuffer.
+	for _, notification := range ro.replayBuffer {
+		observer.notify(notification)
+	}
+
+	// must (write) lock observersMu so that we can safely append to the observers list
+	ro.observersMu.Lock()
+	defer ro.observersMu.Unlock()
+
+	ro.observers = append(ro.observers, observer)
```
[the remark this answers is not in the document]
bryanchriswhite: **TL;DR, we could move the lock up but not the assignment to `ro.observers` and I think it makes more sense to keep them together.**

The fact that the observer isn't added to the observers slice until after the replay notifications have been sent is what ensures that replayed  notifications don't get re-replayed. I will add a comment.

pkg/observable/channel/replay.go
bryanchriswhite: I think there are valid use cases for both. In some scenarios you might already have an observable that you need to turn into a replay observable.

pkg/observable/channel/replay_test.go
bryanchriswhite: `n` is the value that is passed to `#Last()` as the `n` arg. I've updated it to `lastN`.

red-0ne (approved):
LGTM :+1:

pkg/observable/channel/replay.go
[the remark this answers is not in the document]
Olshansk: #PUC :)

pkg/observable/channel/replay_test.go
```diff
@@ -42,46 +56,135 @@ func TestReplayObservable(t *testing.T) {
...
+			actualValues := replayObsvbl.Last(ctx, tt.lastN)
+			require.ElementsMatch(t, tt.expectedValues, actualValues)
+		})
+	}
+}
+
+func TestReplayObservable_Last_Blocks_Goroutine(t *testing.T) {
+	var (
+		lastN    = 5
+		splitIdx = 3
```
Olshansk, line 139: #PUC on what this is for

pkg/observable/channel/replay_test.go
```diff
@@ -42,46 +56,135 @@ func TestReplayObservable(t *testing.T) {
...
+	require.ElementsMatch(t, []int{1, 2, 3}, replayObsvbl.Last(ctx, 3))
+
+	// Concurrently call Last with a value greater than the replay buffer size.
+	lastValues := make(chan []int, 1)
+	go func() {
+		// Last should block until lastN values have been published.
+		lastValues <- replayObsvbl.Last(ctx, lastN)
+	}()
+
+	select {
```
Olshansk, line 163: // Since `lastN` is greater than than number of values we have published to `publishCh`, the
// call to `Last(ctx, LastN)` above in the goroutine is expected to block, and therefore this select
// statement should time out.

pkg/observable/channel/replay_test.go
```diff
@@ -42,46 +56,135 @@ func TestReplayObservable(t *testing.T) {
...
+	case actualValues := <-lastValues:
+		t.Fatalf(
+			"Last should block until the replay buffer is full. Actual values: %v",
+			actualValues,
+		)
+	case <-time.After(10 * time.Millisecond):
+	}
```
Olshansk, line 170:
```suggestion
		log.Println("OK: `lastValues` did not return any values becausethe `replayObcvbl.Last(lastN)` has fewer than `lastN` values.")
	}
```

pkg/observable/channel/replay_test.go
```diff
@@ -42,46 +56,135 @@ func TestReplayObservable(t *testing.T) {
...
+
+	// Publish values after splitIdx.
+	for _, value := range values[splitIdx:] {
+		publishCh <- value
+		time.Sleep(time.Millisecond)
+	}
+
+	select {
+	case actualValues := <-lastValues:
+		require.ElementsMatch(t, values, actualValues)
```
Olshansk, line 180: I'd also add a `require.Len(t, lastN, actualValues)` for readability purposes.

Olshansk (changes requested):

pkg/observable/channel/replay.go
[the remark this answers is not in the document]
bryanchriswhite: This should read cleaner now, I moved some things into an `observableInternals` interface, including `#addObserver()`.

pkg/observable/channel/replay_test.go
bryanchriswhite: I rearranged some things but didn't end up using this verbatim. LMK if you would like me to take another pass at it.

Olshansk: Resolving this thread and will re-review.

pkg/observable/interface.go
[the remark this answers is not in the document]
Olshansk: Makes sense to me.

pkg/observable/channel/replay.go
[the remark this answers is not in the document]
Olshansk: > I will add a comment. 

Ty!

pkg/observable/channel/internals.go
```diff
@@ -0,0 +1,111 @@
...
+func (coi *channelObservableInternals[V]) onUnsubscribe(
+	toRemove observable.Observer[V],
+) {
+	// must (write) lock to iterate over and modify the observers list
+	coi.observersMu.Lock()
+	defer coi.observersMu.Unlock()
+
+	for i, observer := range coi.observers {
+		if observer == toRemove {
+			coi.observers = append((coi.observers)[:i], (coi.observers)[i+1:]...)
```
Olshansk, line 63: Double checking that we don't need to `Clean/Close/Shutdown/Unsubscribe` the `toRemove` observer here. Correct?

pkg/observable/channel/internals.go
```diff
@@ -0,0 +1,111 @@
...
+	// immediately (if they receive any at all).
+	currentObservers := coi.copyObservers()
+	for _, observer := range currentObservers {
+		observer.Unsubscribe()
+	}
+
+	// Reset observers to an empty list. This purges any observers which might have
+	// subscribed while the observable was closing.
+	coi.observersMu.Lock()
+	coi.observers = []*channelObserver[V]{}
```
Olshansk, line 88: NIT: pretty sure this is more idiomatic.

```suggestion
	coi.observers = make([]*channelObserver[V], 0)
```

pkg/observable/channel/internals.go
```diff
@@ -0,0 +1,111 @@
...
+	coi.observersMu.Unlock()
+}
+
+// copyObservers returns a copy of the current observers list. It is safe to
+// call concurrently.
+func (coi *channelObservableInternals[V]) copyObservers() (observers []*channelObserver[V]) {
+	defer coi.observersMu.RUnlock()
+
+	// This loop blocks on acquiring a read lock on observersMu. If TryRLock
```
Olshansk, line 97: If we're going to be in a while loop calling `TryRLock`, why not just call `RLock`?

pkg/observable/channel/observable.go
```diff
@@ -13,31 +11,30 @@ import (
...
+	_ observableInternals[any]   = (*channelObservable[any])(nil)
+)
 
 // option is a function which receives and can modify the channelObservable state.
 type option[V any] func(obs *channelObservable[V])
 
 // channelObservable implements the observable.Observable interface and can be notified
 // by sending on its corresponding publishCh channel.
 type channelObservable[V any] struct {
+	//observableInternals[V]
```
Olshansk, line 25: rm?

pkg/observable/channel/observer_test.go
```diff
@@ -7,20 +7,23 @@ import (
...
 	"github.com/stretchr/testify/require"
+
+	"pocket/pkg/observable"
 )
 
 func TestObserver_Unsubscribe(t *testing.T) {
 	var (
-		onUnsubscribeCalled = false
 		publishCh           = make(chan int, 1)
+		onUnsubscribeCalled = false
```
Olshansk, line 17: Very clear setup!

pkg/observable/channel/replay.go
```diff
@@ -0,0 +1,209 @@
...
+	"context"
+	"log"
+	"sync"
+	"time"
+
+	"pocket/pkg/observable"
+)
+
+// TODO_CONSIDERATION: perhaps this should be parameterized.
+
```
Olshansk, line 13:
```suggestion
```

pkg/observable/channel/replay.go
```diff
@@ -0,0 +1,209 @@
...
+// replayPartialBufferTimeout is the duration to wait for the replay buffer to
+// accumulate at least 1 value before returning the accumulated values.
+const replayPartialBufferTimeout = 100 * time.Millisecond
```
Olshansk, line 16: Do you think we need the word `Partial` in this varname? I was going to suggest `Single` after reading the comment, but feel like removing altogether also works

pkg/observable/channel/replay.go
```diff
@@ -0,0 +1,209 @@
...
+
+var _ observable.ReplayObservable[any] = (*replayObservable[any])(nil)
+
+type replayObservable[V any] struct {
+	// observableInternals[V] is embedded to provide safe read and write access
```
Olshansk, line 21: This is making me `observableInternals` -> `observableSubscriberList` or something along those lines?

pkg/observable/channel/replay.go
```diff
@@ -0,0 +1,209 @@
...
+	replayBufferSize int,
+) (observable.ReplayObservable[V], chan<- V) {
+	obsvbl, publishCh := NewObservable[V]()
+	return ToReplayObservable[V](ctx, replayBufferSize, obsvbl), publishCh
+}
+
+// ToReplayObservable returns an observable which replays the last replayBufferSize
+// number of values published to the source observable to new observers, before
+// publishing new values.
+// It panics if srcObservable does not implement the observableInternals interface.
```
Olshansk, line 48: If we end up renaming `observableInternals`, make sure to update the comments too.

pkg/observable/channel/replay.go
```diff
@@ -0,0 +1,209 @@
...
+
+// Subscribe returns an observer which is notified when the publishCh channel
+// receives a value.
+func (ro *replayObservable[V]) Subscribe(ctx context.Context) observable.Observer[V] {
+	ro.replayBufferMu.RLock()
+	defer ro.replayBufferMu.RUnlock()
+
+	observer := NewObserver[V](ctx, ro.onUnsubscribe)
+
+	// ToReplayObservable all buffered replayBuffer to the observer channel buffer before
```
Olshansk, line 126: Is this comment out of place or outdated? Can't understand what `ToReplayObservable` is intended for here.

pkg/observable/channel/replay.go
```diff
@@ -0,0 +1,209 @@
...
+// cancellation function, if the context isn't done, before using the accumulated
+// values.
+func accumulateReplayValues[V any](
+	observer observable.Observer[V],
+	n int, accValues *[]V,
+) (context.Context, context.CancelFunc) {
+	ctx, cancel := context.WithCancel(context.Background())
+	go func() {
+		for {
+			// The context was cancelled since the last iteration.
```
Olshansk, line 176:
```suggestion
			// The context was canceled since the last iteration.
```

Olshansk (changes requested):
My head hurts a bit after reviewing so much concurrent logic, but it's impressive work!

pkg/observable/channel/internals.go
```diff
@@ -0,0 +1,111 @@
...
+
+import (
+	"sync"
+
+	"pocket/pkg/observable"
+)
+
+// observableInternals is an interface intended to be used between an observable and some
+// higher-level abstraction and/or observable implementation which would embed it.
+// Embedding this interface rather than a channelObservable directly allows for
```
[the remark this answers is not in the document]
bryanchriswhite: I don't think so. The relationship is the other way around. The rationale is that higher-level objects that want to closely couple to observable behavior (e.g. `ReplayObservable`) can assert that `observable.Observable[V]` implements some `ObserverManager` interface which gives it sufficient access, in this case, to the observers list  for read/write and to call their `#notify()` method.

pkg/observable/channel/internals.go
bryanchriswhite: Correct, that's outside the scope of this object.

pkg/observable/channel/internals.go
```diff
@@ -0,0 +1,111 @@
...
+// more transparency and flexibility in higher-level code.
+// NOTE: this interface MUST be used with a common concrete Observer type.
+type observableInternals[V any] interface {
```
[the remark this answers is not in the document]
bryanchriswhite: I like the direction, thanks! I went with `observerManager`.

pkg/observable/channel/internals.go
```diff
@@ -0,0 +1,111 @@
...
+	addObserver(toAdd observable.Observer[V])
+	onUnsubscribe(toRemove observable.Observer[V])
```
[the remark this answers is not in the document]
bryanchriswhite: :ok_hand:

pkg/observable/channel/internals.go
bryanchriswhite: Nice catch! I guess we don't need to do this after all / any more. :raised_hands:

pkg/observable/channel/replay.go
bryanchriswhite: :pray:

pkg/observable/channel/replay.go
bryanchriswhite: Nice catch! I think this was my IDE getting carried away when renaming `Replay` to `ToReplayObservable`.

pkg/observable/channel/replay.go
bryanchriswhite: I suppose my bias is exposed: https://www.grammarly.com/blog/canceled-vs-cancelled/

pkg/observable/channel/replay.go
bryanchriswhite: The rationale behind including the word `Partial` was to serve as a reminder that this timeout only applies in cases where the replay buffer is not full. If it is the timeout is ignored.

pkg/observable/channel/observer_manager.go
```diff
@@ -0,0 +1,154 @@
...
+	}
+}
+
+// addObserver implements the respective member of observerManager. It is used
+// by the channelObservable implementation as well as embedders of observerManager
+// (e.g. replayObservable).
+// It panics if toAdd is not a channelObserver.
+func (com *channelObserverManager[V]) add(
+	toAdd observable.Observer[V],
+) {
```
Olshansk, line 77:
```suggestion
func (com *channelObserverManager[V]) add(toAdd observable.Observer[V]) {
```

pkg/observable/channel/observer_manager.go
```diff
@@ -0,0 +1,154 @@
...
+	com.observers = append(com.observers, toAdd.(*channelObserver[V]))
+}
+
+// remove removes a given observer from the observable's list of observers.
+// It implements the respective member of observerManager and is used by
+// the channelObservable implementation as well as embedders of observerManager
+// (e.g. replayObservable).
+func (com *channelObserverManager[V]) remove(
+	toRemove observable.Observer[V],
+) {
```
Olshansk, line 91:
```suggestion
func (com *channelObserverManager[V]) remove(toRemove observable.Observer[V]) {
```

Olshansk (approved):
Left a couple NITs and a comment but g2g as is.

Even though I reviewed all of the observable code, I've started losing a bit of site of the big picture. Synchronous vs asynchronous, which interfaces I should be using in which case, etc...

I know you're already working on documentation, but if you can think of consolidation opportunities in opportunities, let's revisit it at some point. Fwiw, given the naming of interfaces/types/structs and comments throughout, I think anyone should be able to get a full understanding with a bit of time.

pkg/observable/channel/observer_manager.go
```diff
@@ -0,0 +1,154 @@
...
+	goUnsubscribeOnDone(ctx context.Context, observer observable.Observer[V])
+}
+
+// TODO_CONSIDERATION: if this were a generic implementation, we wouldn't need
+// to cast `toAdd` to a channelObserver in add. There are two things
+// currently preventing a generic observerManager implementation:
+// 1. channelObserver#notify() is not part of the observable.Observer interface
+// 	  and is therefore not accessible here. If we move everything into the
+//	  `observable` pkg so that the unexported member is in scope, then the channel
+//	  pkg can't implement it for the same reason, it's an unexported method defined
```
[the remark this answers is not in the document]
bryanchriswhite: I don't think it makes sense for it to be part of the public interface. It seems wrong to me to expose a mechanism to notify a single observer outside the context of its observable.

pkg/observable/channel/replay.go
bryanchriswhite: Well, this is akward:

image

PR #96, opened 2023-10-24
[Proxy] feat: add general proxy logic and server builder

red-0ne:
### Summary generated by Reviewpad on 28 Oct 23 01:29 UTC
- The file `supplier.proto` had a deletion of an empty line at the end of the file.
- The file `errors.go` was added to the `pkg/relayer/proxy` directory. It defines a package `proxy` and imports `cosmossdk.io/errors` as `sdkerrors`. It declares a variable `codespace` as "relayer/proxy" and an error `ErrUnsupportedRPCType` with the code 1 and message "unsupported rpc type".
- The file contains interface definitions for the RelayerProxy and RelayServer. The RelayerProxy interface is responsible for starting and stopping all advertised relay servers and notifying the miner about the served relays. It also includes methods for verifying relay request signatures and sessions, as well as signing relay responses. The RelayServer interface represents the advertised relay servers provided by the RelayerProxy and includes methods for starting and stopping the service server, as well as getting the service ID.
- The file diff also shows the addition of a new file called "jsonrpc.go" in the "pkg/relayer/proxy" directory. This file contains the implementation of a JSON-RPC server that listens for incoming relay requests and forwards them to a proxied service endpoint. It includes methods to start, stop, and serve HTTP requests. The implementation is currently incomplete.
- The file `server_builder.go` includes several changes, such as adding imports for `pocket/x/shared/types` and `pocket/x/supplier/types`, implementing a function `BuildProvidedServices` to build advertised relay servers based on on-chain advertised services, retrieving the supplier address and advertised information from the blockchain, creating relay servers for each service, switching to the appropriate RPC type to create the RelayServer, and updating the `advertisedRelayServers` map.

Overall, this pull request introduces changes related to the supplier, error handling, interface definitions, JSON-RPC server implementation, and building advertised relay servers based on on-chain data.

## Issue

- #14 

## Type of change

Select one or more:

- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

## Sanity Checklist

### DIFF
--- a/pkg/relayer/proxy/interface.go
+++ b/pkg/relayer/proxy/interface.go
@@ -24,3 +24,15 @@ type RelayerProxy interface {
 	// and its RelayResponse has been signed and successfully sent to the client.
 	ServedRelays() observable.Observable[*types.Relay]
 }
+
+// RelayServer is the interface of the proxy servers provided by the RelayerProxy.
+type RelayServer interface {
+	// Start starts the service server and returns an error if it fails.
+	Start(ctx context.Context) error
+
+	// Stop terminates the service server and returns an error if it fails.
+	Stop(ctx context.Context) error
+
+	// Name returns the name of the service.
+	Name() string
+}
--- a/pkg/relayer/proxy/interface.go
+++ b/pkg/relayer/proxy/interface.go
@@ -8,15 +8,15 @@ import (
 )
 
 // RelayerProxy is the interface for the proxy that serves relays to the application.
-// It is responsible for starting and stopping all supported proxies.
+// It is responsible for starting and stopping all supported RelayServers.
 // While handling requests and responding in a closed loop, it also notifies
 // the miner about the relays that have been served.
 type RelayerProxy interface {
 
-	// Start starts all supported proxies and returns an error if any of them fail to start.
+	// Start starts all advertised relay servers and returns an error if any of them fail to start.
 	Start(ctx context.Context) error
 
-	// Stop stops all supported proxies and returns an error if any of them fail.
+	// Stop stops all advertised relay servers and returns an error if any of them fail.
 	Stop(ctx context.Context) error
 
 	// ServedRelays returns an observable that notifies the miner about the relays that have been served.
@@ -25,7 +25,7 @@ type RelayerProxy interface {
 	ServedRelays() observable.Observable[*types.Relay]
 }
 
-// RelayServer is the interface of the proxy servers provided by the RelayerProxy.
+// RelayServer is the interface of the advertised relay servers provided by the RelayerProxy.
 type RelayServer interface {
 	// Start starts the service server and returns an error if it fails.
 	Start(ctx context.Context) error
@@ -33,6 +33,6 @@ type RelayServer interface {
 	// Stop terminates the service server and returns an error if it fails.
 	Stop(ctx context.Context) error
 
-	// Name returns the name of the service.
-	Name() string
+	// ServiceId returns the serviceId of the service.
+	ServiceId() string
 }
--- a/pkg/relayer/proxy/interface.go
+++ b/pkg/relayer/proxy/interface.go
@@ -5,6 +5,7 @@ import (
 
 	"pocket/pkg/observable"
 	"pocket/x/service/types"
+	sharedtypes "pocket/x/shared/types"
 )
 
 // RelayerProxy is the interface for the proxy that serves relays to the application.
@@ -41,5 +42,5 @@ type RelayServer interface {
 	Stop(ctx context.Context) error
 
 	// ServiceId returns the serviceId of the service.
-	ServiceId() string
+	ServiceId() *sharedtypes.ServiceId
 }
--- a/pkg/relayer/proxy/interface.go
+++ b/pkg/relayer/proxy/interface.go
@@ -23,6 +23,13 @@ type RelayerProxy interface {
 	// A served relay is one whose RelayRequest's signature and session have been verified,
 	// and its RelayResponse has been signed and successfully sent to the client.
 	ServedRelays() observable.Observable[*types.Relay]
+
+	// VerifyRelayRequest is a shared method used by RelayServers to check the
+	// relay request signature and session validity.
+	VerifyRelayRequest(relayRequest *types.RelayRequest) (isValid bool, err error)
+
+	// SignRelayResponse is a shared method used by RelayServers to sign the relay response.
+	SignRelayResponse(relayResponse *types.RelayResponse) ([]byte, error)
 }
 
 // RelayServer is the interface of the advertised relay servers provided by the RelayerProxy.
--- a/pkg/relayer/proxy/jsonrpc.go
+++ b/pkg/relayer/proxy/jsonrpc.go
@@ -3,6 +3,8 @@ package proxy
 import (
 	"context"
 	"net/http"
+
+	"pocket/x/service/types"
 )
 
 var _ RelayServer = &jsonRPCServer{}
@@ -14,23 +16,38 @@ type jsonRPCServer struct {
 	// endpointUrl is the url that the server listens to for incoming relay requests.
 	endpointUrl string
 
+	// nativeServiceListenAddress is the address of the native service that the server relays requests to.
+	nativeServiceListenAddress string
+
 	// server is the http server that listens for incoming relay requests.
 	server *http.Server
 
 	// relayerProxy is the main relayer proxy that the server uses to perform its operations.
 	relayerProxy RelayerProxy
+
+	// servedRelaysProducer is a channel that emits the relays that have been served so that the
+	// servedRelays observable can fan out the notifications to its subscribers.
+	servedRelaysProducer chan<- *types.Relay
 }
 
 // NewJSONRPCServer creates a new HTTP server that listens for incoming relay requests
 // and proxies them to the supported native service.
 // It takes the serviceId, endpointUrl, and the main RelayerProxy as arguments and returns
 // a RelayServer that listens to incoming RelayRequests
-func NewJSONRPCServer(serviceId string, endpointUrl string, proxy RelayerProxy) RelayServer {
+func NewJSONRPCServer(
+	serviceId string,
+	endpointUrl string,
+	nativeServiceListenAddress string,
+	servedRelaysProducer chan<- *types.Relay,
+	proxy RelayerProxy,
+) RelayServer {
 	return &jsonRPCServer{
-		serviceId:    serviceId,
-		endpointUrl:  endpointUrl,
-		server:       &http.Server{Addr: endpointUrl},
-		relayerProxy: proxy,
+		serviceId:                  serviceId,
+		endpointUrl:                endpointUrl,
+		server:                     &http.Server{Addr: endpointUrl},
+		relayerProxy:               proxy,
+		nativeServiceListenAddress: nativeServiceListenAddress,
+		servedRelaysProducer:       servedRelaysProducer,
 	}
 }
 
--- a/pkg/relayer/proxy/jsonrpc.go
+++ b/pkg/relayer/proxy/jsonrpc.go
@@ -3,21 +3,24 @@ package proxy
 import (
 	"context"
 	"net/http"
+	"net/url"
 
 	"pocket/x/service/types"
+	sharedtypes "pocket/x/shared/types"
 )
 
 var _ RelayServer = (*jsonRPCServer)(nil)
 
 type jsonRPCServer struct {
 	// serviceId is the id of the service that the server is responsible for.
-	serviceId string
+	serviceId *sharedtypes.ServiceId
 
-	// endpointUrl is the url that the server listens to for incoming relay requests.
-	endpointUrl string
+	// serverEndpoint is the advertised endpoint configuration that the server uses to
+	// listen for incoming relay requests.
+	serverEndpoint *sharedtypes.SupplierEndpoint
 
-	// nativeServiceListenAddress is the address of the native service that the server relays requests to.
-	nativeServiceListenAddress string
+	// proxiedServiceEndpoint is the address of the proxied service that the server relays requests to.
+	proxiedServiceEndpoint url.URL
 
 	// server is the http server that listens for incoming relay requests.
 	server *http.Server
@@ -31,28 +34,28 @@ type jsonRPCServer struct {
 }
 
 // NewJSONRPCServer creates a new HTTP server that listens for incoming relay requests
-// and proxies them to the supported native service.
+// and forwards them to the supported proxied service endpoint.
 // It takes the serviceId, endpointUrl, and the main RelayerProxy as arguments and returns
 // a RelayServer that listens to incoming RelayRequests
 func NewJSONRPCServer(
-	serviceId string,
-	endpointUrl string,
-	nativeServiceListenAddress string,
+	serviceId *sharedtypes.ServiceId,
+	supplierEndpoint *sharedtypes.SupplierEndpoint,
+	proxiedServiceEndpoint url.URL,
 	servedRelaysProducer chan<- *types.Relay,
 	proxy RelayerProxy,
 ) RelayServer {
 	return &jsonRPCServer{
-		serviceId:                  serviceId,
-		endpointUrl:                endpointUrl,
-		server:                     &http.Server{Addr: endpointUrl},
-		relayerProxy:               proxy,
-		nativeServiceListenAddress: nativeServiceListenAddress,
-		servedRelaysProducer:       servedRelaysProducer,
+		serviceId:              serviceId,
+		serverEndpoint:         supplierEndpoint,
+		server:                 &http.Server{Addr: supplierEndpoint.Url},
+		relayerProxy:           proxy,
+		proxiedServiceEndpoint: proxiedServiceEndpoint,
+		servedRelaysProducer:   servedRelaysProducer,
 	}
 }
 
 // Start starts the service server and returns an error if it fails.
-// It also waits for the passed in context to be done to shut down.
+// It also waits for the passed in context to end before shutting down.
 // This method is blocking and should be called in a goroutine.
 func (j *jsonRPCServer) Start(ctx context.Context) error {
 	go func() {
@@ -68,8 +71,8 @@ func (j *jsonRPCServer) Stop(ctx context.Context) error {
 	return j.server.Shutdown(ctx)
 }
 
-// ServiceId returns the serviceId of the service.
-func (j *jsonRPCServer) ServiceId() string {
+// ServiceId returns the serviceId of the JSON-RPC service.
+func (j *jsonRPCServer) ServiceId() *sharedtypes.ServiceId {
 	return j.serviceId
 }
 
@@ -78,4 +81,5 @@ func (j *jsonRPCServer) ServiceId() string {
 // when jsonRPCServer is used as an http.Handler with an http.Server.
 // (see https://pkg.go.dev/net/http#Handler)
 func (j *jsonRPCServer) ServeHTTP(writer http.ResponseWriter, request *http.Request) {
+	panic("TODO: implement jsonRPCServer.ServeHTTP")
 }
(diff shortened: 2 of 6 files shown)


pkg/relayer/proxy/interface.go
```diff
@@ -24,3 +24,15 @@ type RelayerProxy interface {
 	// and its RelayResponse has been signed and successfully sent to the client.
 	ServedRelays() observable.Observable[*types.Relay]
 }
+
+// RelayServer is the interface of the proxy servers provided by the RelayerProxy.
```
bryanchriswhite, line 28: Wdyt about updating the last line of the `RelayerProxy` comment, from:
```go
// It is responsible for starting and stopping all supported proxies.
```

To:
```go
// It is responsible for starting and stopping all supported RelayServers.
```

This comment references `RelayerProxy`, so I feel having `RelayerProxy` like referencing `RelayServer` adds clarity to the relationship.

Perhaps we could also add a bit of detail here to give the reader an idea about which axes across which `RelayServer` implementations may be distributed (i.e. serialization, transport, etc.)

pkg/relayer/proxy/jsonrpc.go
```diff
@@ -0,0 +1,62 @@
...
+	endpointUrl string
+
+	// server is the http server that listens for incoming relay requests.
+	server *http.Server
+
+	// relayerProxy is the main relayer proxy that the server uses to perform its operations.
+	relayerProxy RelayerProxy
+}
+
+// NewHTTPServer creates a new HTTP server that listens for incoming relay requests
```
bryanchriswhite, line 24:
```suggestion
// NewJSONRPCServer creates a new HTTP server that listens for incoming relay requests
```

pkg/relayer/proxy/jsonrpc.go
```diff
@@ -0,0 +1,62 @@
...
+	return j.server.ListenAndServe()
+}
+
+// Stop terminates the service server and returns an error if it fails.
+func (j *jsonRPCServer) Stop(ctx context.Context) error {
+	return j.server.Shutdown(ctx)
+}
+
+// Name returns the name of the service.
+func (j *jsonRPCServer) Name() string {
```
bryanchriswhite, line 55: Should this be "serviceId" instead of "name"?

```suggestion
// Name returns the serviceId of the service.
func (j *jsonRPCServer) ServiceId() string {
```

pkg/relayer/proxy/jsonrpc.go
```diff
@@ -0,0 +1,62 @@
...
+	return j.serviceId
+}
+
+// ServeHTTP is the http handler that listens for incoming relay requests.
```
bryanchriswhite, line 59:
```suggestion
// ServeHTTP listens for incoming relay requests. It implements  the respective
// method of the `http.Handler` interface. It is called by http.ListenAndServe()
// when jsonRPCServer is used as an http.Handler with an http.Server.
// (see: https://pkg.go.dev/net/http@go1.21.3#Handler)
```

pkg/relayer/proxy/proxy.go
```diff
@@ -44,7 +45,7 @@ type relayerProxy struct {
 	// providedServices is a map of the services provided by the relayer proxy. Each provided service
 	// has the necessary information to start the server that listens for incoming relay requests and
 	// the client that proxies the request to the supported native service.
-	providedServices map[string][]*ProvidedService
+	providedServices map[string][]RelayServer
```
bryanchriswhite, line 48: Should  this name be updated as well?

```suggestion
	relayServers map[string][]RelayServer
```

pkg/relayer/proxy/proxy.go
```diff
@@ -66,53 +67,61 @@ func NewRelayerProxy(
...
-		providedServices:     providedServices,
 		servedRelays:         servedRelays,
 		servedRelaysProducer: servedRelaysProducer,
 	}
 }
 
 // Start starts all supported proxies and returns an error if any of them fail to start.
 func (rp *relayerProxy) Start(ctx context.Context) error {
-	panic("TODO: implement relayerProxy.Start")
+	// The provided services map is built from the supplier's on-chain advertised information,
```
bryanchriswhite, line 87:
```suggestion
	// The provided services map is built from the supplier's on-chain advertised information,
```

pkg/relayer/proxy/proxy.go
```diff
@@ -66,53 +67,61 @@ func NewRelayerProxy(
...
+	// which is a runtime parameter that can be changed by the supplier.
+	// Build the provided services map at Start instead of NewRelayerProxy to avoid having to
+	// return an error from the constructor.
+	err := rp.BuildProvidedServices(ctx)
+	if err != nil {
+		return err
+	}
+
+	eg, gctx := errgroup.WithContext(ctx)
```
bryanchriswhite, line 96: NIT: this name is too short and doesn't convey any information about what it is. Regardless of the fact that there is only a single reference, unless the reader is familiar with this code, they will have to manually look it up before being able to reason about it. My personal preference is to only use really short abbreviations like this for receiver names in method definitions.

```suggestion
	errGroup, gctx := errgroup.WithContext(ctx)
```

pkg/relayer/proxy/proxy.go
```diff
@@ -66,53 +67,61 @@ func NewRelayerProxy(
...
+
+	for _, providedService := range rp.providedServices {
+		for _, svr := range providedService {
```
bryanchriswhite, line 99: Here's another example of the inconsistency between "server" and "service" nomeclature.

pkg/relayer/proxy/server_builder.go
```diff
@@ -0,0 +1,57 @@
...
+
+	sharedtypes "pocket/x/shared/types"
+	suppliertypes "pocket/x/supplier/types"
+)
+
+type RelayServersMap map[string][]RelayServer
+
+// BuildProvidedServices builds the provided services from the supplier's on-chain advertised services.
+// It populates the relayerProxy's `providedServices` map of servers for each service, where each server
+// is responsible for listening for incoming relay requests and poxying them to the supported native service.
```
bryanchriswhite, line 14: Similar concerns about consistent naming. We have "advertised services", "providedServices", and "server" referring to the same object(s), as well as "proxying" in that same context; whereas, it seems like it should be "relaying" to avoid ambiguity between `RelayerProxy` and `RelayServer`.

pkg/relayer/proxy/server_builder.go
```diff
@@ -0,0 +1,57 @@
+package proxy
+
+import (
+	"context"
+
+	sharedtypes "pocket/x/shared/types"
+	suppliertypes "pocket/x/supplier/types"
+)
+
+type RelayServersMap map[string][]RelayServer
```
bryanchriswhite, line 10: NIT: wdyt about something like this to be more explicit about what the key is expected to be?:

```suggestion
type (
  serviceId = string
  RelayServersMap map[serviceId][]RelayServer
)
```

pkg/relayer/proxy/server_builder.go
```diff
@@ -0,0 +1,57 @@
...
+	// Get the supplier's advertised information from the blockchain
+	supplierQuery := &suppliertypes.QueryGetSupplierRequest{Address: supplierAddress.String()}
+	supplierQueryResponse, err := rp.supplierQuerier.Supplier(ctx, supplierQuery)
+	if err != nil {
+		return err
+	}
+
+	services := supplierQueryResponse.Supplier.Services
+
+	// Build the provided services map. For each service's endpoint, create the appropriate server.
```
bryanchriswhite, line 31:
```suggestion
	// Build the provided services map. For each service's endpoint, create the appropriate relay server.
```

bryanchriswhite (changes requested):
Looking good! :raised_hands: 

I left some questions and suggestions in the comments. I think there's some low-hanging fruit on the naming and comment fronts. Overall, very clean and well commented though. :ok_hand:

pkg/relayer/proxy/jsonrpc.go
bryanchriswhite: UPDATE: I'm noticing that there is a `name`  field  on the service ID protobuf type, is this intended to return that?

red-0ne: It was intended to return the `serviceId` string (serviceId.Id) so it should be as you suggested.

pkg/relayer/proxy/proxy.go
```diff
@@ -41,10 +47,14 @@ type relayerProxy struct {
...
 
-	// providedServices is a map of the services provided by the relayer proxy. Each provided service
+	// advertisedRelayServers is a map of the services provided by the relayer proxy. Each provided service
 	// has the necessary information to start the server that listens for incoming relay requests and
-	// the client that proxies the request to the supported native service.
-	providedServices map[string][]*ProvidedService
+	// the client that relays the request to the supported native service.
+	advertisedRelayServers RelayServersMap
+
+	// nativeServiceListenAddresses is a map of the native services server's listen addresses
```
Olshansk, line 55: Ditto one of my previous comments. Are these the proxy addresses?

pkg/relayer/proxy/proxy.go
```diff
@@ -18,6 +19,11 @@ import (
 
 var _ RelayerProxy = (*relayerProxy)(nil)
 
+type (
+	serviceId       = string
+	RelayServersMap = map[serviceId][]RelayServer
```
Olshansk, line 24: Does this need to be publically exposed?

pkg/relayer/proxy/proxy.go
```diff
@@ -41,10 +47,14 @@ type relayerProxy struct {
...
+	// advertisedRelayServers is a map of the services provided by the relayer proxy. Each provided service
 	// has the necessary information to start the server that listens for incoming relay requests and
-	// the client that proxies the request to the supported native service.
-	providedServices map[string][]*ProvidedService
+	// the client that relays the request to the supported native service.
+	advertisedRelayServers RelayServersMap
+
+	// nativeServiceListenAddresses is a map of the native services server's listen addresses
+	// that are supported by the relayer proxy.
+	nativeServicesListenAddress map[serviceId]string
```
Olshansk, line 57: Why did you create a custom type for the map above but not this one?

pkg/relayer/proxy/proxy.go
```diff
@@ -59,60 +69,79 @@ func NewRelayerProxy(
...
+		sessionQuerier:              sessionQuerier,
+		nativeServicesListenAddress: nativeServicesListenAddress,
+		servedRelays:                servedRelays,
+		servedRelaysProducer:        servedRelaysProducer,
 	}
 }
 
-// Start starts all supported proxies and returns an error if any of them fail to start.
+// Start concurrently starts all advertised relay servers and returns an error if any of them fails to start.
 func (rp *relayerProxy) Start(ctx context.Context) error {
```
Olshansk, line 96: The other `Start` function has a comment that it's a blocking function that needs to be run in a goroutine. Does the same thing apply here?

pkg/relayer/proxy/proxy.go
```diff
@@ -59,60 +69,79 @@ func NewRelayerProxy(
...
-	panic("TODO: implement relayerProxy.Start")
+	// The provided services map is built from the supplier's on-chain advertised information,
+	// which is a runtime parameter that can be changed by the supplier.
+	// Build the provided services map at Start instead of NewRelayerProxy to avoid having to
```
Olshansk, line 99:
```suggestion
	// NOTE: We build the provided services map at Start instead of NewRelayerProxy to avoid having to
```

pkg/relayer/proxy/proxy.go
```diff
@@ -59,60 +69,79 @@ func NewRelayerProxy(
...
+	// return an error from the constructor.
+	err := rp.BuildProvidedServices(ctx)
+	if err != nil {
+		return err
```
Olshansk, line 103:
```suggestion
	if err := rp.BuildProvidedServices(ctx); err != nil {
		return err
```

pkg/relayer/proxy/proxy.go
```diff
@@ -59,60 +69,79 @@ func NewRelayerProxy(
...
+	}
+
+	startGroup, gctx := errgroup.WithContext(ctx)
+
+	for _, relayServer := range rp.advertisedRelayServers {
+		for _, svr := range relayServer {
+			server := svr // create a new variable scoped to the anonymous function
```
Olshansk, line 110: Does this not work given that we reinitiliaze `svr` on every iteration of the forloop?

pkg/relayer/proxy/proxy.go
```diff
@@ -59,60 +69,79 @@ func NewRelayerProxy(
...
+	// The provided services map is built from the supplier's on-chain advertised information,
+	// which is a runtime parameter that can be changed by the supplier.
+	// Build the provided services map at Start instead of NewRelayerProxy to avoid having to
+	// return an error from the constructor.
+	err := rp.BuildProvidedServices(ctx)
+	if err != nil {
+		return err
+	}
+
+	startGroup, gctx := errgroup.WithContext(ctx)
```
Olshansk, line 106: I was looking in https://pkg.go.dev/golang.org/x/sync/errgroup and they seem to reuse the `ctx` varname.

Consider either renaming to `relayServerCtx` or just overriding `ctx`. Ditto below.

```
func MD5All(ctx context.Context, root string) (map[string][md5.Size]byte, error) {
	// ctx is canceled when g.Wait() returns. When this version of MD5All returns
	// - even in case of error! - we know that all of the goroutines have finished
	// and the memory they were using can be garbage-collected.
	g, ctx := errgroup.WithContext(ctx)
```

pkg/relayer/proxy/proxy.go
```diff
@@ -59,60 +69,79 @@ func NewRelayerProxy(
...
+			server := svr // create a new variable scoped to the anonymous function
+			startGroup.Go(func() error { return server.Start(gctx) })
+		}
+	}
+
+	return startGroup.Wait()
 }
 
-// Stop stops all supported proxies and returns an error if any of them fail.
+// Stop concurrently stops all advertised relay servers and returns an error if any of them fails.
```
Olshansk, line 118: Make a note about blocking or no.

pkg/relayer/proxy/server_builder.go
```diff
@@ -0,0 +1,62 @@
...
+import (
+	"context"
+
+	sharedtypes "pocket/x/shared/types"
+	suppliertypes "pocket/x/supplier/types"
+)
+
+// BuildProvidedServices builds the advertised relay servers from the supplier's on-chain advertised services.
+// It populates the relayerProxy's `advertisedRelayServers` map of servers for each service, where each server
+// is responsible for listening for incoming relay requests and relaying them to the supported native service.
```
Olshansk, line 12:
```suggestion
// is responsible for listening for incoming relay requests and relaying them to the supported proxied service.
```

I feel like `native` is a new term, but this whole package is called `proxy`

pkg/relayer/proxy/server_builder.go
```diff
@@ -0,0 +1,62 @@
...
+func (rp *relayerProxy) BuildProvidedServices(ctx context.Context) error {
```
Olshansk, line 13: Nice function!

proto/pocket/shared/supplier.proto
```diff
@@ -8,12 +8,11 @@ option go_package = "pocket/x/shared/types";
...
 import "cosmos/base/v1beta1/coin.proto";
+import "pocket/shared/service.proto";
 
 // Supplier is the type defining the actor in Pocket Network that provides RPC services.
 message Supplier {
   string address = 1 [(cosmos_proto.scalar) = "cosmos.AddressString"]; // The Bech32 address of the supplier using cosmos' ScalarDescriptor to ensure deterministic deterministic encoding using cosmos' ScalarDescriptor to ensure deterministic deterministic encoding
   cosmos.base.v1beta1.Coin stake = 2; // The total amount of uPOKT the supplier has staked
-  // TODO(@Olshansk): Uncomment the line below once the `ServiceId` proto is defined
-  // repeated service.SupplierServiceConfig services = 3; // The service configs this supplier can support
+  repeated SupplierServiceConfig services = 3; // The service configs this supplier can support
```
Olshansk, line 17:
```suggestion
 // TODO(@olshansk): This is uncommented for integration purposes but business logic is not yet implemented.
  repeated SupplierServiceConfig services = 3; // The service configs this supplier can support
```

Olshansk (changes requested):

pkg/relayer/proxy/interface.go
```diff
@@ -8,19 +8,38 @@ import (
...
 	Stop(ctx context.Context) error
 
 	// ServedRelays returns an observable that notifies the miner about the relays that have been served.
 	// A served relay is one whose RelayRequest's signature and session have been verified,
 	// and its RelayResponse has been signed and successfully sent to the client.
 	ServedRelays() observable.Observable[*types.Relay]
+
+	// VerifyRelayRequest is a shared method used by RelayServers to check the
+	// relay request signature and session validity.
+	VerifyRelayRequest(relayRequest *types.RelayRequest) (isValid bool, err error)
```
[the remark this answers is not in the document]
red-0ne: It is not static as it needs the `accountQuerier`, `sessionQuerier` and `BlockClient` see PR 101 [pkg/relayer/proxy/relay_verifier.go](https://github.com/pokt-network/poktroll/pull/101/files#diff-db9c7f03567a68094434fdd144e0ad4e7fba82323732505d1d7a7aad014cd59f)

pkg/relayer/proxy/jsonrpc.go
```diff
@@ -0,0 +1,81 @@
...
+	endpointUrl string
+
+	// nativeServiceListenAddress is the address of the native service that the server relays requests to.
+	nativeServiceListenAddress string
+
+	// server is the http server that listens for incoming relay requests.
+	server *http.Server
+
+	// relayerProxy is the main relayer proxy that the server uses to perform its operations.
+	relayerProxy RelayerProxy
```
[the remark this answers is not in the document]
red-0ne: This is the full `RelayerProxy` that is aggregating all the `RelayServer`s and this is a back reference to it.

`RelayerProxy` t does not have an address on its own and is not the `nativeServiceListenAddress` which is the address of the native service being serviced.

pkg/relayer/proxy/jsonrpc.go
```diff
@@ -0,0 +1,81 @@
...
+func (j *jsonRPCServer) ServiceId() string {
+	return j.serviceId
+}
+
+// ServeHTTP listens for incoming relay requests. It implements the respective
+// method of the http.Handler interface. It is called by http.ListenAndServe()
+// when jsonRPCServer is used as an http.Handler with an http.Server.
+// (see https://pkg.go.dev/net/http#Handler)
+func (j *jsonRPCServer) ServeHTTP(writer http.ResponseWriter, request *http.Request) {
+}
```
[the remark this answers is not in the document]
red-0ne: Yes, a `panic("TODO: implement jsonRPCServer.ServeHTTP")` :smile:

pkg/relayer/proxy/proxy.go
red-0ne: When mentioning `native services` it's referring to the "Anvil" like server address, I apologize if the naming is totally misleading :/

pkg/relayer/proxy/proxy.go
red-0ne: Not really, it's only used by the proxy package :+1:

pkg/relayer/proxy/proxy.go
red-0ne: This on is a simple map, which is only a `map[string][string]` while `RelayServersMap` is in the order of `map[string][]struct{}`. I'll have an alias for the former for consistency :+1:

pkg/relayer/proxy/proxy.go
red-0ne: If we use `svr` directly we have a `loop variable svr captured by func literal` warning
image

Olshansk (approved):
Awesome! So much cleaner after renaming a few things too 🚢

PR #108, opened 2023-10-28
[AppGate] Implement the MVP AppGateServer

red-0ne:
### Human Summary

`AppGateServer` implementation that supports JSONRPC over HTTP. Used to sign and send `RelayRequests`, verify the corresponding `RelayResponse`'s supplier signature and reply to the client with the original service response.

The `appGateServer` will build rings for the application and cache them in order to send relays to the network signed. This PR also implements signature verification for the `RelayerProxy`.

This PR adds the cobra cmd to start the AppGateServer from the CLI.

Outstanding TODOs:

- Move the rign creation and caching logic into a shared package
- Emit delegation events from the App module and listen for these to know when to invalidate a cached ring

### Summary generated by Reviewpad on 10 Nov 23 22:45 UTC
This pull request introduces various changes to several files. Here is a summary of the changes:

1. root.go: Adds support for an appgate server command to the existing root command.

2. go.sum: Adds new dependencies to the project.

3. simple_signer.go: Adds the implementation of the SimpleSigner type, which provides functionality for signing messages using a keyring and key name.

4. appgateserver.go: Provides logic related to creating and managing rings for signing relays in the appgateserver package.

5. relay_verifier.go: Makes changes related to verifying relay request and response signatures.

6. endpoint_selector.go: Adds functionality for selecting a relayer endpoint based on service ID and JSON RPC type.

7. errors.go: Defines error variables registered using sdkerrors.Register for the appgateserver and relayer packages.

8. interface.go: Introduces the Signer interface, which abstracts the signing of a message.

9. options.go: Adds functions for setting signing information and listening URL for the appgate server.

10. session.go: Adds functionality for retrieving the current session for a given service in the appgateserver package.

11. relay.go: Adds methods for generating signable bytes for relay requests and responses.

12. proxy package: Adds functions related to creating and managing rings for signing relays in the proxy package.

13. options.go: Adds functions for setting signing information and listening URL for the appgate server.

14. session.go: Adds functionality for retrieving the current session for a given service in the appgateserver package.

15. relay.go: Adds methods for generating signable bytes for relay requests and responses.

16. errors.go: Makes changes to error variables in the relayer/proxy package.

17. go.mod: Adds new dependencies to the project.

18. pkg/client/block/client.go: Updates the block client functionality and error handling.

19. prog.go: Deletes the file containing code for generating a package using GoMock reflections.

20. proxy.go: Adds new import statements and additional fields related to the application module and ring caching.

21. appGateServer.go: Contains the implementation of the appGateServer struct, which handles listening to application requests, signing and relaying them, and verifying response signatures.

Please review these changes for your pull request.

## Type of change

Select one or more:

- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

## Sanity Checklist


pkg/appclient/appclient.go
```diff
@@ -0,0 +1,159 @@
...
+	"github.com/cosmos/cosmos-sdk/crypto/keyring"
+	accounttypes "github.com/cosmos/cosmos-sdk/x/auth/types"
+
+	blocktypes "pocket/pkg/client"
+	"pocket/x/service/types"
+	sessiontypes "pocket/x/session/types"
+)
+
+type appClient struct {
+	// keyName is the name of the key in the keyring that will be used to sign relay requests.
```
h5law, line 20:
```suggestion
	// TODO(@h5law): Replace with ring signatures
	// keyName is the name of the key in the keyring that will be used to sign relay requests.
```

pkg/appclient/appclient.go
```diff
@@ -0,0 +1,159 @@
...
+// Stop stops the application server and returns any error that occurred.
+func (app *appClient) Stop(ctx context.Context) error {
+	return app.server.Shutdown(ctx)
+}
+
+// ServeHTTP is the HTTP handler for the application server.
+// It captures the application request, signs it, and sends it to the supplier.
+// After receiving the response from the supplier, it verifies the response signature
+// before returning the response to the application.
+// The serviceId is extracted from the request path.
```
h5law, line 105: Can we add an example of what the path would look like here

pkg/appclient/endpoint_selector.go
```diff
@@ -0,0 +1,38 @@
...
+	"context"
+	"net/url"
+
+	sessiontypes "pocket/x/session/types"
+	sharedtypes "pocket/x/shared/types"
+)
+
+// getRelayerUrl gets the URL of the relayer for the given service.
+// It gets the suppliers list from the current session and returns
+// the first relayer URL that matches the JSON RPC RpcType.
```
h5law, line 13:
```suggestion
// the first relayer URL that matches the JSON RPC RpcType.
// TODO(@h5law): Look into different endpoint selection algorithms depending on their suitability
```

pkg/appclient/jsonrpc.go
```diff
@@ -0,0 +1,116 @@
...
+	// Get the current block height to query for the current session.
+	currentBlock := app.blockClient.LatestBlock(ctx)
+
+	// Query for the current session.
+	sessionQueryReq := sessiontypes.QueryGetSessionRequest{
+		ApplicationAddress: app.appAddress,
+		ServiceId:          &sharedtypes.ServiceId{Id: serviceId},
+		BlockHeight:        currentBlock.Height(),
+	}
+	sessionQueryRes, err := app.sessionQuerier.GetSession(ctx, &sessionQueryReq)
```
h5law, line 48: As applications can be in a single session at a time we may want to look into a `SessionClient` similar to the `BlockClient` that removes the need to do this query on every request.

pkg/appclient/relay_verifier.go
```diff
@@ -0,0 +1,47 @@
...
+
+// verifyResponse verifies the relay response signature.
+func (app *appClient) verifyResponse(
+	ctx context.Context,
+	supplierAddress string,
+	relayResponse *types.RelayResponse,
+) error {
+	// Query for the supplier account to get the application's public key to verify the relay request signature.
+	accQueryReq := &accounttypes.QueryAccountRequest{Address: supplierAddress}
+	accQueryRes, err := app.accountQuerier.Account(ctx, accQueryReq)
```
h5law, line 20: Here we may want to maintain a cache of the node's pubkeys (those in the session) against their addresses, or include the pubkey with the signature (the node would just need to store there own). The idea is that this would reduce latency and reduce the number of on chain queries we have to do, instead of each verification only just when cache is missing.

Consider adding a `TODO` here

pkg/appclient/relay_verifier.go
```diff
@@ -0,0 +1,47 @@
...
+	var payloadBz []byte
+	_, err = relayResponse.Payload.MarshalTo(payloadBz)
+	if err != nil {
+		return err
+	}
+	hash := crypto.Sha256(payloadBz)
+
+	// accQueryRes.Account.Value is a protobuf Any type that should be unmarshaled into an AccountI interface.
+	// TODO_TECHDEBT: Make sure our `AccountI`/`any` unmarshalling is correct.
+	// See https://github.com/pokt-network/poktroll/pull/101/files/edbd628e9146e232ef58c71cfa8f4be2135cdb50..fbba10626df79f6bf6e2218513dfdeb40a629790#r1372464439
```
h5law, line 35:
```go
	// get application public key
	appPubKeyReq := &authTypes.QueryAccountRequest{Address: address}
	appPubKeyRes, err := relayerHandler.accountQueryClient.Account(relayerHandler.ctx, appPubKeyReq)
	if err != nil {
		return nil, fmt.Errorf("unable to get applications account: %s [%w]", address, err)
	}
	acc := new(authTypes.BaseAccount)
	if err := acc.Unmarshal(appPubKeyRes.Account.Value); err != nil {
		return nil, fmt.Errorf("unable to deserialise applications account: %s [%w]", address, err)
	}
```

Unsure if this is any better but this is how I was doing this in the alpha repo

h5law (changes requested):
Left a few comments on adding some `TODO`s for me to pick up. But this is great work bringing over the MVP 10/10

pkg/appclient/appclient.go
```diff
@@ -0,0 +1,159 @@
...
+func (app *appClient) ServeHTTP(writer http.ResponseWriter, request *http.Request) {
```
Olshansk, line 106: Confirming that this needs to be publically exposed?

pkg/appclient/endpoint_selector.go
Olshansk, line 13: Why not make `JSON_RPC` a parameter and make this auto-customizable right away?

Otherwise, it should probably be part of the `jsonrpc` file.

pkg/appclient/endpoint_selector.go
```diff
@@ -0,0 +1,38 @@
...
+		for _, service := range supplier.Services {
+			// Skip services that don't match the requested serviceId.
+			if service.ServiceId.Id != serviceId {
+				continue
+			}
+
+			for _, endpoint := range service.Endpoints {
+				// Return the first endpoint url that matches the JSON RPC RpcType.
+				if endpoint.RpcType == sharedtypes.RPCType_JSON_RPC {
+					supplierUrl, err := url.Parse(endpoint.Url)
```
Olshansk, line 29: If there's an error:
1. Log it
2. Continue looking for a successful one

pkg/appclient/endpoint_selector.go
```diff
@@ -0,0 +1,38 @@
...
+
+import (
+	"context"
+	"net/url"
+
+	sessiontypes "pocket/x/session/types"
+	sharedtypes "pocket/x/shared/types"
+)
+
+// getRelayerUrl gets the URL of the relayer for the given service.
```
Olshansk, line 11:
```suggestion
// TODO_IMPROVE: This implements a naive greedy approach that defaults to the 
// first available URL but future optimizations can be introduced.
// getRelayerUrl gets the URL of the relayer for the given service.
```

pkg/appclient/jsonrpc.go
```diff
@@ -0,0 +1,116 @@
...
+
+	"github.com/cometbft/cometbft/crypto"
+
+	"pocket/x/service/types"
+	sessiontypes "pocket/x/session/types"
+	sharedtypes "pocket/x/shared/types"
+)
+
+// handleJSONRPCRelays handles JSON RPC relay requests.
+func (app *appClient) handleJSONRPCRelays(
```
Olshansk, line 17:
```suggestion
// handleJSONRPCRelay handles JSON RPC relay requests.
func (app *appClient) handleJSONRPCRelay(
```

pkg/appclient/jsonrpc.go
```diff
@@ -0,0 +1,116 @@
...
+) (responseBz []byte, err error) {
+	// Read the request body bytes.
+	payloadBz, err := io.ReadAll(request.Body)
+	if err != nil {
+		return nil, err
+	}
+
+	// Hash and sign the request payload.
+	hash := crypto.Sha256(payloadBz)
+	signature, _, err := app.keyring.Sign(app.keyName, hash)
```
Olshansk, line 30: MAJOR BLOCKER!

This feels incorrect.

We need to:
1. Create the `RelayRequest` protobuf
2. Sign it (without the signature)
3. Insert the signature

TIPS:
- RelayRequest should have a function (where it is a receiver) that signs things
- The RelayRequest object itself will be responsible for removing the fields that don't need to be part of the signature.

This is how we can enforce consistent signature addition & verification.

Otherwise, it feels like we're signing an unstructured hash, which is hard to track & verify.

pkg/appclient/jsonrpc.go
```diff
@@ -0,0 +1,116 @@
...
+	// Unmarshal the response bytes into a RelayResponse.
+	relayResponse := &types.RelayResponse{}
+	if err := relayResponse.Unmarshal(relayResponseBz); err != nil {
+		return nil, err
+	}
+
+	// Verify the response signature. We use the supplier address that we got from
+	// the getRelayerUrl function since this is the address we are expecting to sign the response.
+	// TODO_TECHDEBT: if the RelayResponse is an internal error response, we should not verify the signature
+	// as in some relayer early failures, it may not be signed by the supplier.
```
Olshansk, line 106: Wdyt of making `RelayResponse` have a function (where it is the receiver) that verifies the signature?

pkg/appclient/jsonrpc.go
```diff
@@ -0,0 +1,116 @@
...
+	if err := app.verifyResponse(ctx, supplierAddress, relayResponse); err != nil {
+		return nil, err
+	}
+
+	// Marshal the response payload to bytes to be sent back to the application.
+	var responsePayloadBz []byte
+	_, err = relayResponse.Payload.MarshalTo(responsePayloadBz)
+
+	return responsePayloadBz, err
```
Olshansk, line 115: It feels inconsistent that we take in an `http.Request` and return `Bz`. 

Is the caller of this responsible for converting this to json/yaml?

pkg/appclient/relay_verifier.go
```diff
@@ -0,0 +1,47 @@
...
+	accQueryReq := &accounttypes.QueryAccountRequest{Address: supplierAddress}
+	accQueryRes, err := app.accountQuerier.Account(ctx, accQueryReq)
+	if err != nil {
+		return err
+	}
+
+	// Marshal the relay response payload to bytes and get the hash.
+	var payloadBz []byte
+	_, err = relayResponse.Payload.MarshalTo(payloadBz)
+	if err != nil {
```
Olshansk, line 28:
```suggestion
	if _, err = relayResponse.Payload.MarshalTo(payloadBz); err != nil {
```

pkg/appclient/relay_verifier.go
```diff
@@ -0,0 +1,47 @@
...
+		return err
+	}
+	hash := crypto.Sha256(payloadBz)
```
Olshansk, line 31: I think there should be a small help on `RelayResponse` that removes the necessary fields and takes in helpers supplied by the appclient to do the actual signature verification.

Other than that, the compartmentalization is great

Olshansk (changes requested):

pkg/appclient/appclient.go
```diff
@@ -0,0 +1,159 @@
...
+	// blockClient is the client for the block module.
+	// It is used to get the current block height to query for the current session.
+	blockClient blocktypes.BlockClient
+
+	// server is the HTTP server that will be used capture application requests
+	// so that they can be signed and relayed to the supplier.
+	server *http.Server
+}
+
+func NewAppClient(
```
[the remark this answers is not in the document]
h5law: `appClientProxy` is probably the better term. It is intended to listen for requests and handle the chain interaction

pkg/appclient/jsonrpc.go
h5law: I think as I didn't get to work on this last week when I bring this into main I will just be adding rings in straight away instead of a separate PR. This is true tho. We need the method `GetSignableBytes()` on the request and response and to do it in this fashion 👍🏼

h5law:
@red-0ne @Olshansk I will be picking this up tmrw 20231107 and bringing it into main for the E2E relay and will address all the comments when I do <3 exciting times!!!!

pkg/appclient/appclient.go
red-0ne: We are complying to the `http.Server` interface.

pkg/appclient/jsonrpc.go
red-0ne: I added a session cacher but didn't use the SessionClient. Feel free to change if you think it would be better

pkg/appclient/appclient.go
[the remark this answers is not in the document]
red-0ne: Renamed to appServer :+1:

red-0ne:
> @red-0ne @Olshansk I will be picking this up tmrw 20231107 and bringing it into main for the E2E relay and will address all the comments when I do <3 exciting times!!!!

I addressed most of the comments, but haven't touched the signature parts yet. As per @Olshansk comments signatures should go into `Relay(Request/Reply)` methods.

pkg/appclient/jsonrpc.go
h5law: Spoke offline: I think having a `Signer` interface that is used by both the app/gateway client and the supplier with seperate implementations would be better in conjunction with the `GetSignableBytes()` method on the relay parts

Olshansk: I'd bias towards a single implementino (or collocated) but either use an "overloading" like technique or params to generalize. 

If it's not too much effort, wold love to see it in an independent PR.

h5law: Now the signer interface is in I think it will be easier to abstract it away and replace with the reciever method you think of. Will check it out tmrw

pkg/appgateserver/cmd/cmd.go
```diff
@@ -0,0 +1,199 @@
...
+behave as a gateway and will sign relays on behalf of any application sending it relays provided
+that the address recieved in the query parameters of a request has been delegated to by the
+gateway, this is so that it can sign relays using the ring of the application with the
+key associated with the --signing-key-name flag.
+
+If an application doesn't provide the --self-signing flag, it will be able to send relays
+to the AppGate server and it will still function as an application, however each request
+will have to contain the "?senderAddress=[address]" query parameter, where [address] is
+the address of the application that is sending the request. This is so that the server
+can generate the correct ring for the application and sign the request.`,
```
Olshansk, line 55:
```suggestion
		Long: `Starts the AppGate server that listens for incoming relay requests and handles the necessary on-chain interactions (sessions, suppliers, etc) to receive the respective relay response.

-- App Mode (Flag)- - 
If the server is started with a defined `--signing-key-name` flag, it will behave
as an Application. Any incoming requests will be signed with the Application's private key.

-- Gateway Mode (Flag)-- 
If the `--signing-key-name` flag is not provided, the server will behave as a Gateway. It will
sign relays on behalf of any Application sending it relays, provided that the address associated
with `--signing-key-name` has been delegated to. This is necessary for the application<->gateway
ring signature to function.

-- App Mode (HTTP) --
If an application doesn't provide the `--signing-key-name` flag, it can stillsend relays
to the AppGate server and function as an Application, provided that:
1. Each request contains the `?senderAddress=[address]` query parameter
2. The keyring on the AppGate server contains the privateKey corresponding to the address provided`
```

pkg/appgateserver/cmd/cmd.go
```diff
@@ -0,0 +1,199 @@
...
+		Args: cobra.NoArgs,
+		RunE: runAppGateServer,
+	}
+
+	cmd.Flags().StringVar(&signingKeyName, "signing-key-name", "", "The name of the key that will be used to sign relays")
+	cmd.Flags().StringVar(&listeningEndpoint, "listening-endpoint", "http://localhost:42069", "The host and port that the server will listen on")
```
Olshansk, line 61:
```suggestion
	cmd.Flags().StringVar(&listeningEndpoint, "listening-endpoint", "http://localhost:42069", "The host and port that the appgate server will listen on")
```

pkg/appgateserver/cmd/cmd.go
```diff
@@ -0,0 +1,199 @@
...
+	cmd.Flags().StringVar(&cometWebsocketUrl, "comet-websocket-url", "ws://localhost:36657/websocket", "The URL of the tendermint websocket endpoint to interact with the chain")
```
Olshansk, line 62:
```suggestion
	cmd.Flags().StringVar(&cometWebsocketUrl, "comet-websocket-url", "ws://localhost:36657/websocket", "The URL of the comet websocket endpoint to communicate with the pocket blockchain")
```

pkg/appgateserver/cmd/cmd.go
```diff
@@ -0,0 +1,199 @@
...
+	cmd.Flags().BoolVar(&selfSigning, "self-signing", false, "Whether the server should sign all incoming requests with its own ring (for applications)")
```
Olshansk, line 63: @h5law My understanding was that the presence of `signing-key-name` is what determines this, no?

pkg/appgateserver/cmd/cmd.go
```diff
@@ -0,0 +1,199 @@
...
+	cryptotypes "github.com/cosmos/cosmos-sdk/crypto/types"
+	"github.com/spf13/cobra"
+
+	"github.com/pokt-network/poktroll/pkg/appgateserver"
+	blockclient "github.com/pokt-network/poktroll/pkg/client/block"
+	eventsquery "github.com/pokt-network/poktroll/pkg/client/events_query"
+)
+
+var (
+	signingKeyName    string
```
Olshansk, line 29: NIT: Can we follow the same naming pattern as in cosmos` and prefix these with `flagXXX`

pkg/appgateserver/cmd/cmd.go
```diff
@@ -0,0 +1,199 @@
...
+		RunE: runAppGateServer,
+	}
+
+	cmd.Flags().StringVar(&signingKeyName, "signing-key-name", "", "The name of the key that will be used to sign relays")
+	cmd.Flags().StringVar(&listeningEndpoint, "listening-endpoint", "http://localhost:42069", "The host and port that the server will listen on")
+	cmd.Flags().StringVar(&cometWebsocketUrl, "comet-websocket-url", "ws://localhost:36657/websocket", "The URL of the tendermint websocket endpoint to interact with the chain")
+	cmd.Flags().BoolVar(&selfSigning, "self-signing", false, "Whether the server should sign all incoming requests with its own ring (for applications)")
+
+	cmd.Flags().String(flags.FlagKeyringBackend, "", "Select keyring's backend (os|file|kwallet|pass|test)")
+	cmd.Flags().String(flags.FlagNode, "tcp://localhost:36657", "tcp://<host>:<port> to tendermint rpc interface for this chain")
```
Olshansk, line 66:
```suggestion
	cmd.Flags().String(flags.FlagNode, "tcp://localhost:36657", "The URL of the comet tcp endpoint to communicate with the pocket blockchain")
```

pkg/appgateserver/cmd/cmd.go
```diff
@@ -0,0 +1,199 @@
...
+
+	return cmd
+}
+
+func runAppGateServer(cmd *cobra.Command, _ []string) error {
+	// Create a context that is cancelled when the command is interrupted
```
Olshansk, line 72:
```suggestion
	// Create a context that is canceled when the command is interrupted
```

pkg/appgateserver/cmd/cmd.go
```diff
@@ -0,0 +1,199 @@
...
+	clientCtx := cosmosclient.GetClientContextFromCmd(cmd)
+
+	// Parse the listening endpoint.
+	listeningUrl, err := url.Parse(listeningEndpoint)
+	if err != nil {
+		cancelCtx()
+		return fmt.Errorf("failed to parse listening endpoint: %w", err)
+	}
+
+	// Obtain the tendermint websocket endpoint from the client context.
```
Olshansk, line 85:
```suggestion
	// Obtain the comet websocket endpoint from the client context.
```

pkg/appgateserver/cmd/cmd.go
```diff
@@ -0,0 +1,199 @@
...
+	cometWSUrl, err := url.Parse(clientCtx.NodeURI + "/websocket")
```
Olshansk, line 86: The default value is `ws://localhost:36657/websocket`

Are we sure we need/want to append `/websocket` here

pkg/appgateserver/cmd/cmd.go
```diff
@@ -0,0 +1,199 @@
...
+	if err != nil {
+		cancelCtx()
+		return fmt.Errorf("failed to parse block query URL: %w", err)
+	}
+	cometWSUrl.Scheme = "ws"
```
Olshansk, line 90: Does `url.Parse` not populate this piece?

pkg/appgateserver/cmd/cmd.go
```diff
@@ -0,0 +1,199 @@
...
+	// If the comet websocket URL is not provided, use the one from the client context.
+	if cometWebsocketUrl == "" {
+		cometWebsocketUrl = cometWSUrl.String()
+	}
+
+	log.Printf("INFO: Creating block client, using websocket URL: %s...", cometWebsocketUrl)
```
Olshansk, line 96:
```suggestion
	log.Printf("INFO: Creating block client using comet websocket URL: %s", cometWebsocketUrl)
```

pkg/appgateserver/cmd/cmd.go
```diff
@@ -0,0 +1,199 @@
...
+	deps := depinject.Supply(eventsQueryClient)
+	blockClient, err := blockclient.NewBlockClient(ctx, deps, cometWebsocketUrl)
+	if err != nil {
+		cancelCtx()
+		return fmt.Errorf("failed to create block client: %w", err)
+	}
+
+	log.Println("INFO: Creating AppGate server...")
+
+	key, err := clientCtx.Keyring.Key(signingKeyName)
```
Olshansk, line 108: s/key/signingKey or s/key/privateKey?

pkg/appgateserver/cmd/cmd.go
```diff
@@ -0,0 +1,199 @@
...
+	if selfSigning {
+		signingAddress = appAddress.String()
+	}
+
+	signingKey, err := recordLocalToScalar(key.GetLocal())
+	if err != nil {
+		cancelCtx()
+		return fmt.Errorf("failed to convert private key to scalar: %w", err)
+	}
+	signingKey = signingKey
```
Olshansk:
```suggestion

```

pkg/appgateserver/cmd/cmd.go
```diff
@@ -0,0 +1,199 @@
...
+	if err != nil {
+		cancelCtx()
+		return fmt.Errorf("failed to get address from key: %w", err)
+	}
+	signingAddress := ""
+	if selfSigning {
+		signingAddress = appAddress.String()
+	}
+
+	signingKey, err := recordLocalToScalar(key.GetLocal())
```
Olshansk, line 122: Can you add a comment on what `recordLocalToScalar` is doing above (just one sentence)

Olshansk (changes requested):
Going to review this in pieces to move this along.

Just finished looking at `pkg/appgateserver/cmd/cmd.go` and will move on to the next file.

pkg/appgateserver/endpoint_selector.go
```diff
@@ -0,0 +1,45 @@
...
+	"context"
+	"log"
+	"net/url"
+
+	sessiontypes "github.com/pokt-network/poktroll/x/session/types"
+	sharedtypes "github.com/pokt-network/poktroll/x/shared/types"
+)
+
+// TODO_IMPROVE: This implements a naive greedy approach that defaults to the
+// first available URL but future optimizations can be introduced.
```
Olshansk, line 13:
```suggestion
// first available supplier. Future optimizations (e.g. Quality-of-Service) can be introduced here.
```

pkg/appgateserver/endpoint_selector.go
```diff
@@ -0,0 +1,45 @@
...
+// TODO(@h5law): Look into different endpoint selection depending on their suitability.
+// getRelayerUrl gets the URL of the relayer for the given service.
+func (app *appGateServer) getRelayerUrl(
```
Olshansk, line 16: Nice, big fan of how the first version of this turned out (looping through suppliers -> services -> endpoints ..)

pkg/appgateserver/jsonrpc.go
```diff
@@ -0,0 +1,127 @@
...
+	supplierUrl, supplierAddress, err := app.getRelayerUrl(ctx, serviceId, sharedtypes.RPCType_JSON_RPC, session)
+	if err != nil {
+		return err
+	}
+
+	// Create the relay request.
+	relayRequest := &types.RelayRequest{
+		Meta: &types.RelayRequestMetadata{
+			SessionHeader: session.Header,
+			Signature:     nil,
```
Olshansk, line 49:
```suggestion
			Signature:     nil, // signature is added below
```

pkg/appgateserver/jsonrpc.go
```diff
@@ -0,0 +1,127 @@
...
+	"net/http"
+
+	"github.com/cometbft/cometbft/crypto"
+
+	"github.com/pokt-network/poktroll/x/service/types"
+	sharedtypes "github.com/pokt-network/poktroll/x/shared/types"
+)
+
+// handleJSONRPCRelay handles JSON RPC relay requests.
+func (app *appGateServer) handleJSONRPCRelay(
```
Olshansk, line 17:
```suggestion
// It does everything from preparing, signing and sending the request.
// It then blocks on the response to come back and forward it to the provided writer.
func (app *appGateServer) handleJSONRPCRelay(
```

pkg/appgateserver/jsonrpc.go
```diff
@@ -0,0 +1,127 @@
...
+	relayResponse := &types.RelayResponse{}
+	if err := relayResponse.Unmarshal(relayResponseBz); err != nil {
+		return err
+	}
+
+	// Verify the response signature. We use the supplier address that we got from
+	// the getRelayerUrl function since this is the address we are expecting to sign the response.
+	// TODO_TECHDEBT: if the RelayResponse is an internal error response, we should not verify the signature
+	// as in some relayer early failures, it may not be signed by the supplier.
+	if err := app.verifyResponse(ctx, supplierAddress, relayResponse); err != nil {
```
Olshansk, line 111:
```suggestion
	// TODO_IMPROVE: Add more logging & telemetry so we can get visibility and signal into
	// failed responses.
	if err := app.verifyResponse(ctx, supplierAddress, relayResponse); err != nil {
```

pkg/appgateserver/jsonrpc.go
```diff
@@ -0,0 +1,127 @@
...
+	if _, err = relayResponse.Payload.MarshalTo(responsePayloadBz); err != nil {
+		return err
+	}
+
+	// Reply with the RelayResponse payload.
+	if _, err := writer.Write(relayRequestBz); err != nil {
+		return err
+	}
+
+	return nil
```
Olshansk, line 126: This function is FRIKKING 🔥

Olshansk (changes requested):
Finished of `pkg/appgateserver/jsonrpc.go` and contunuing along. Looking 🔥 so far!

pkg/appgateserver/cmd/cmd.go
h5law: Updated the comment to match the logic but kept the formatting

pkg/appgateserver/cmd/cmd.go
h5law: So this is not the case, in fact for both gateways and applications they need a key to sign with: `signing-key` (removed the name)

However, gateways do not "self-sign" instead they "others-sign" but applications do "self-sign". This means that if an appgate server is started and declares it will "self-sign" for every request it will fetch its own ring and sign it that way.

If it doesnt opt to "self sign" for every request it will fetch the ring of the application address it is provided, **but** it will still have to sign the ring with its `signing-key`.

Does that clear it up?

pkg/appgateserver/rings.go
```diff
@@ -0,0 +1,135 @@
...
+		// if the ring is not in the cache, get it from the application module
+		log.Printf("DEBUG: No ring cached for address: %s", appAddress)
+		ring, err = app.getRingForAppAddress(ctx, appAddress)
+	} else {
+		// if the ring is in the cache, create it from the points
+		log.Printf("DEBUG: Ring cached for address: %s", appAddress)
+		ring, err = newRingFromPoints(points)
+	}
+	if err != nil {
+		log.Printf("ERROR: Unable to get ring for address: %s [%w]", appAddress, err)
```
Olshansk, line 44: Looks like we might need to use `%v` instead of `%w`

pkg/appgateserver/rings.go
```diff
@@ -0,0 +1,135 @@
...
+	ring "github.com/noot/ring-go"
+
+	"github.com/pokt-network/poktroll/pkg/signer"
+	apptypes "github.com/pokt-network/poktroll/x/application/types"
+)
+
+// getRingSingerForAppAddress returns the RingSinger used to sign relays.
+// This method first attempts to get the points of the ring from the cache, if it
+// fails it queries the application module for the points and creates the ring.
+func (app *appGateServer) getRingSingerForAppAddress(ctx context.Context, appAddress string) (*signer.RingSigner, error) {
```
Olshansk, line 24: Well designed function!

pkg/appgateserver/cmd/cmd.go
h5law: Where are you seeing this default value? Could you link it so I can include in this

pkg/appgateserver/cmd/cmd.go
h5law: I think I didn't understand what the default was and made sure it kept to websocket

pkg/appgateserver/rings.go
```diff
@@ -0,0 +1,135 @@
...
+	return newRingFromPoints(points)
+}
+
+// newRingFromPoints creates a new ring from a slice of points on the secp256k1 curve
+func newRingFromPoints(points []ringtypes.Point) (*ring.Ring, error) {
+	return ring.NewFixedKeyRingFromPublicKeys(ring_secp256k1.NewCurve(), points)
+}
+
+// getDelegatedPubKeysForAddress returns the ring used to sign a message for the given application
+// address, by querying the portal module for it's delegated pubkeys
```
Olshansk, line 69:
```suggestion
// address, by querying the gateway module for it's delegated pubkeys
```

pkg/appgateserver/rings.go
```diff
@@ -0,0 +1,135 @@
...
+	req := &apptypes.QueryGetApplicationRequest{Address: appAddress}
+	res, err := app.applicationQuerier.Application(ctx, req)
+	if err != nil {
+		return nil, fmt.Errorf("unable to retrieve application for address: %s [%w]", appAddress, err)
+	}
+
+	// create a slice of addresses for the ring
+	ringAddresses := make([]string, 0)
+	ringAddresses = append(ringAddresses, appAddress) // app address is index 0
+	ringAddresses = append(ringAddresses, appAddress) // add app address twice to make the ring size of mininmum 2
```
Olshansk, line 87:
```suggestion
	// TODO_TECHDEBT: We are adding the appAddress twice because a ring signature requires AT LEAST
	// two pubKeys. When the Application acts independently, we use the same key twice, but need to research
	// if ring signatuers can be create with just one key.
	ringAddresses = append(ringAddresses, appAddress) // add app address twice to make the ring size of mininmum 2
```

pkg/appgateserver/rings.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+	return newRingFromPoints(points)
+}
+
+// newRingFromPoints creates a new ring from a slice of points on the secp256k1 curve
+func newRingFromPoints(points []ringtypes.Point) (*ring.Ring, error) {
+	return ring.NewFixedKeyRingFromPublicKeys(ring_secp256k1.NewCurve(), points)
+}
+
+// getDelegatedPubKeysForAddress returns the ring used to sign a message for the given application
```
Olshansk, line 68: The name / comment of this function is a bit confusing because we're not doing any delegation. It's just the app and itself.

@h5law  Can you expand or update appropriately?

pkg/appgateserver/cmd/cmd.go
Olshansk: #PUC

pkg/appgateserver/cmd/cmd.go
Olshansk: Line 66

Olshansk (changes requested):

pkg/appgateserver/rings.go
h5law: We are getting the delegated pub keys for the application. We add the app twice incase the app is not delegated to anyone but ultimately this will build the ring with all the delegated gateways addresses too

pkg/appgateserver/cmd/cmd.go
h5law: image

pkg/appgateserver/cmd/cmd.go
h5law: Removed entirely as we set a default 😅

pkg/appgateserver/server.go
```diff
@@ -0,0 +1,231 @@
...
+	sessiontypes "github.com/pokt-network/poktroll/x/session/types"
+)
+
+type SigningInformation struct {
+	// SigningKey is the scalar point on the appropriate curve corresponding to the
+	// signer's private key, and is used to sign relay requests via a ring signature
+	SigningKey ringtypes.Scalar
+
+	// AppAddress is the address of the application that the server is serving if
+	// it is nil then the application address must be included in each request
```
Olshansk, line 29:
```suggestion
	// AppAddress is the address of the application that the server is serving.
	// If it is nil, then the application address must be included in each request via a query parameter.
```

pkg/appgateserver/server.go
```diff
@@ -0,0 +1,231 @@
...
+	AppAddress string
+}
+
+// appGateServer is the server that listens for application requests and relays them to the supplier.
+// it is responsible for maintaining the current session for the application, signing the requests,
```
Olshansk, line 34:
```suggestion
// It is responsible for maintaining the current session for the application, signing the requests,
```

pkg/appgateserver/rings.go
h5law: They cannot this is part of the ring library updated comment with the rest tho

pkg/appgateserver/server.go
```diff
@@ -0,0 +1,231 @@
...
+// and verifying the response signatures.
+// The appGateServer is the basis for both applications and gateways, depending on whether the application
+// is running their own instance of the appGateServer or they are sending requests to a gateway running an
+// instance of the appGateServer, they will need to either include the application address in the request or not.
+type appGateServer struct {
```
Olshansk, line 39: This struct is amazing!

pkg/appgateserver/server.go
```diff
@@ -0,0 +1,231 @@
...
+	// Start the HTTP server.
+	return app.server.ListenAndServe()
+}
+
+// Stop stops the application server and returns any error that occurred.
+func (app *appGateServer) Stop(ctx context.Context) error {
+	return app.server.Shutdown(ctx)
+}
+
+// ServeHTTP is the HTTP handler for the application server.
```
Olshansk, line 145:
```suggestion
// ServeHTTP is the HTTP handler for the appgate server.
```

pkg/appgateserver/server.go
```diff
@@ -0,0 +1,231 @@
...
+// After receiving the response from the supplier, it verifies the response signature
+// before returning the response to the application.
+// The serviceId is extracted from the request path.
+// The request's path should be of the form:
+//
+//	"<protocol>://host:port/serviceId[/other/path/segments]?senderAddr=<senderAddr>"
+//
+// where the serviceId is the id of the service that the application is requesting
+// and the other (possible) path segments are the JSON RPC request path.
+func (app *appGateServer) ServeHTTP(writer http.ResponseWriter, request *http.Request) {
```
Olshansk, line 156:
```suggestion
// TODO_TECHDEBT: Revisit the requestPath above based on the SDK that'll be exposed in the future.
func (app *appGateServer) ServeHTTP(writer http.ResponseWriter, request *http.Request) {
```

pkg/appgateserver/server.go
```diff
@@ -0,0 +1,231 @@
...
+	// Extract the serviceId from the request path.
+	path := request.URL.Path
+	serviceId := strings.Split(path, "/")[1]
+	var appAddress string
+
+	if app.signingInformation.AppAddress == "" {
+		appAddress = request.URL.Query().Get("senderAddr")
+	} else {
+		appAddress = app.signingInformation.AppAddress
+	}
```
Olshansk, line 168:
```suggestion

	appAddress := app.signingInformation.AppAddress
	if appAddress == "" {
		appAddress = request.URL.Query().Get("senderAddr")
	}
```

pkg/appgateserver/server.go
```diff
@@ -0,0 +1,231 @@
...
+
+	if appAddress == "" {
+		app.replyWithError(writer, ErrAppGateMissingAppAddress)
```
Olshansk, line 171: Nice

pkg/relayer/proxy/rings.go
```diff
@@ -0,0 +1,108 @@
...
+}
+
+// newRingFromPoints creates a new ring from a slice of points on the secp256k1 curve
+func newRingFromPoints(points []ringtypes.Point) (*ring.Ring, error) {
+	return ring.NewFixedKeyRingFromPublicKeys(ring_secp256k1.NewCurve(), points)
+}
+
+// getDelegatedPubKeysForAddress returns the ring used to sign a message for the given application
+// address, by querying the portal module for it's delegated pubkeys
+func (rp *relayerProxy) getDelegatedPubKeysForAddress(
```
Olshansk, line 46: This file seems to be a complete copy-paste of `pkg/appgateserver/rings.go`. What's up with that?

Olshansk (changes requested):

pkg/relayer/proxy/rings.go
h5law: See top line, this will be refactored into its on `pkg` subdir as this logic should be encapsulated but I think thats not work for today copy-pasta is better as the caching logic etc will be more complex and require more time to implement

pkg/appgateserver/rings.go
```diff
@@ -84,9 +84,17 @@ func (app *appGateServer) getDelegatedPubKeysForAddress(
 	// create a slice of addresses for the ring
 	ringAddresses := make([]string, 0)
 	ringAddresses = append(ringAddresses, appAddress) // app address is index 0
-	ringAddresses = append(ringAddresses, appAddress) // add app address twice to make the ring size of mininmum 2
-	if len(res.Application.DelegateeGatewayAddresses) > 0 {
-		ringAddresses = append(ringAddresses, res.Application.DelegateeGatewayAddresses...) // delegatee addresses are index 1+
+	if len(res.Application.DelegateeGatewayAddresses) < 1 {
+		// add app address twice to make the ring size of mininmum 2
+		// TODO_TECHDEBT: We are adding the appAddress twice because a ring
```
Olshansk, line 89:
```suggestion
		// TODO_HACK: We are adding the appAddress twice because a ring
```

pkg/appgateserver/cmd/cmd.go
```diff
@@ -0,0 +1,186 @@
...
+	// Parse the listening endpoint.
+	listeningUrl, err := url.Parse(flagListeningEndpoint)
+	if err != nil {
+		return fmt.Errorf("failed to parse listening endpoint: %w", err)
+	}
+
+	log.Printf("INFO: Creating block client, using comet websocket URL: %s...", flagCometWebsocketUrl)
+
+	// Create the block client with its dependency on the events client.
+	eventsQueryClient := eventsquery.NewEventsQueryClient(flagCometWebsocketUrl)
```
red-0ne, line 92: We should follow the dependency construction used in PR #177 [1] and have a function that is responsible for that

x/service/types/relay.go
```diff
@@ -0,0 +1,23 @@
+package types
```
Olshansk, line 1: OPTIONAL - LMK WYT

I can see you documented the side effect of nullifying the signature when retrieving the signable bytes but....

Wdyt of storing the signature, marshaling and avoiding side-effect altogether?

e.g.

```go
sigTemp = req.Meta.Signature
req.Meta.Signature = nil
signableBz := req.Marshal()
req.Meta.Signature = sigTemp
return signableBz
```

pkg/relayer/proxy/rings.go
```diff
@@ -0,0 +1,116 @@
...
+	return newRingFromPoints(points)
+}
+
+// newRingFromPoints creates a new ring from a slice of points on the secp256k1 curve
+func newRingFromPoints(points []ringtypes.Point) (*ring.Ring, error) {
+	return ring.NewFixedKeyRingFromPublicKeys(ring_secp256k1.NewCurve(), points)
+}
+
+// getDelegatedPubKeysForAddress returns the ring used to sign a message for the given application
+// address, by querying the portal module for it's delegated pubkeys
```
Olshansk, line 45:
```suggestion
// address, by querying the gateway module for it's delegated pubkeys
```

pkg/appgateserver/cmd/cmd.go
```diff
@@ -0,0 +1,186 @@
...
+	}
+
+	// Convert the key record to a private key and return the scalar
+	// point on the secp256k1 curve that it corresponds to.
+	// If the key is not a secp256k1 key, this will return an error.
+	signingKey, err := recordLocalToScalar(keyRecord.GetLocal())
+	if err != nil {
+		return fmt.Errorf("failed to convert private key to scalar: %w", err)
+	}
+	signingInfo := appgateserver.SigningInformation{
```
red-0ne, line 122: I still feel that the signing key construction should be done in the constructor. We should pass in the `signing-key` and `self-sign` flag values and off-load the cmd from any `appGateServer` knowledge except the dependencies.

WDYT?

pkg/relayer/proxy/relay_verifier.go
```diff
@@ -3,8 +3,10 @@ package proxy
 import (
 	"context"
 
+	sdkerrors "cosmossdk.io/errors"
+	ring_secp256k1 "github.com/athanorlabs/go-dleq/secp256k1"
 	"github.com/cometbft/cometbft/crypto"
-	accounttypes "github.com/cosmos/cosmos-sdk/x/auth/types"
+	"github.com/noot/ring-go"
```
Olshansk, line 9: Just FYI that this is finally happening @noot  :)

x/service/types/relay.go
h5law: There is no side affect as this is not a pointer reciever and so this is ultimately added steps for something that doesnt change the signature field anyway

pkg/appgateserver/cmd/cmd.go
h5law: 👍🏼

pkg/relayer/proxy/relay_verifier.go
```diff
@@ -17,33 +19,63 @@ func (rp *relayerProxy) VerifyRelayRequest(
...
+			ErrRelayerProxyInvalidRelayRequest,
+			"missing signature from relay request: %v", relayRequest,
+		)
+	}
+
+	ringSig := new(ring.RingSig)
+	if err := ringSig.Deserialize(ring_secp256k1.NewCurve(), signature); err != nil {
+		return sdkerrors.Wrapf(
+			ErrRelayerProxyInvalidRelayRequestSignature,
+			"error deserializing signature: %v", err,
```
Olshansk, line 35:
```suggestion
			"error deserializing ring signature: %v", err,
```

pkg/client/block/client.go
```diff
@@ -141,7 +140,9 @@ func (bClient *blockClient) goPublishBlocks(ctx context.Context) {
 	// If we get here, the retry limit was reached and the retry loop exited.
 	// Since this function runs in a goroutine, we can't return the error to the
 	// caller. Instead, we panic.
-	panic(fmt.Errorf("BlockClient.goPublishBlocks shold never reach this spot: %w", publishErr))
+	if publishErr != nil {
```
Olshansk, line 143: lol

pkg/appgateserver/session.go
```diff
@@ -0,0 +1,47 @@
...
+	appAddress, serviceId string,
+) (*sessiontypes.Session, error) {
+	app.sessionMu.RLock()
+	defer app.sessionMu.RUnlock()
+
+	latestBlock := app.blockClient.LatestBlock(ctx)
+	if currentSession, ok := app.currentSessions[serviceId]; ok {
+		sessionEndBlockHeight := currentSession.Header.SessionStartBlockHeight + currentSession.NumBlocksPerSession
+
+		// Return the current session if it is still valid.
```
Olshansk, line 24: DO we want to return an error in the else case?

Olshansk (changes requested):

pkg/relayer/proxy/relay_verifier.go
h5law: Big up @noot loving the ring library <3

x/service/types/relay.go
Olshansk: giphy (3)

pkg/relayer/proxy/rings.go
```diff
@@ -0,0 +1,108 @@
+// TODO(@h5law): Move all this logic out into a shared package
```
Olshansk, line 1:
```suggestion
// TODO_BLOCKER(@h5law): Move all this logic out into a shared package to avoid
// the duplication of core business logic between `pkg/relayer/proxy/rings.go`
// and `pkg/appgateserver/rings.go`
```

pkg/appgateserver/cmd/cmd.go
Olshansk: Yup, I'd say #PUC right next to the flag definitions!

pkg/appgateserver/session.go
h5law: No the else case is querying the chain for the current session and updating the cache

pkg/appgateserver/cmd/cmd.go
```diff
@@ -0,0 +1,147 @@
...
+	if err := appGateServer.Start(ctx); err != nil && !errors.Is(err, http.ErrServerClosed) {
+		return fmt.Errorf("failed to start app gate server: %w", err)
+	} else if errors.Is(err, http.ErrServerClosed) {
+		log.Println("INFO: AppGate server stopped")
+	}
+
+	return nil
+}
+
+func setupAppGateServerDependencies(cmd *cobra.Command, ctx context.Context, cometWebsocketUrl string) (depinject.Config, error) {
```
h5law, line 130: @bryanchriswhite @red-0ne is this what you guys think is better?

pkg/appgateserver/server.go
```diff
@@ -0,0 +1,281 @@
...
+
+	for _, opt := range opts {
+		opt(app)
+	}
+
+	if err := app.validateConfig(); err != nil {
+		return nil, err
+	}
+
+	keyRecord, err := app.clientCtx.Keyring.Key(app.signingInformation.SigningKeyName)
```
h5law, line 127: @red-0ne is this more in line with your thinking?

pkg/appgateserver/cmd/cmd.go
red-0ne: Yes, thank you.

Olshansk (approved):
DALL·E 2023-11-10 15 06 47 - Impressionist style mixed media painting crafted with metallic ink and watercolor, depicting a captivating scene symbolic of the Pocket Network's AppG

Issue #39, opened 2023-10-09
[AppGate] Enable Application -> Gateway Delegation


Olshansk:
## Objective

Create a foundation for the `Gateway` module in the [poktroll](https://github.com/pokt-network/poktroll) repo for the Shannon upgrade.

## Origin Document

[Gateway actor in the Pocket Network Utility Specification](https://github.com/pokt-network/pocket-network-protocol/tree/main/utility#35-gateway-protocol)

## Goals

- Enable an E2E HTTP relay
- Build the foundation for the `Gateway` actor in [Pocket Network](https://pokt.network/)
- Build the foundation for the `gateway` module using the [Cosmos SDK](https://v1.cosmos.network/sdk) in Pocket
- Identify time-consuming TODOs to be done during the integration and testing stages
- Enable implementing business logic atop of a well-documented and well-tested codebase in downstream PRs

## Deliverables

- [x] A **single PR** that uses [ignite](https://ignite.com/) to scaffold the `DelegateToGateway` and `UndelegateFromGateway` messages in the `application` module
- [x] A **single PR** that adds a parameter for the maximum number of gateway
- [x] A **single PR** that adds the business logic and test utils for this functionality
- [ ] #111
- [x] Unit tests & validation across the board

## Non-goals / Non-deliverables

- Support for `unbonding` periods after undlegating
- Allow-listing or other "scope creep" like features

## General deliverables

- [x] **Comments**: Add/update TODOs and comments alongside the source code so it is easier to follow.
- [x] **Testing**: Add new tests (unit and/or E2E) to the test suite.
- [x] **Makefile**: Add new targets to the Makefile to make the new functionality easier to use.
- [x] **Documentation**: Update architectural or development READMEs; use [mermaid](https://mermaid-js.github.io/mermaid/) diagrams where appropriate.

---

**Creator**: @Olshansk 
**Co-owners**: @h5law

RossiNYC:
https://hackmd.io/@blockscience/HkxbDcWZ6

h5law:
Closing this as its done 🔥

Olshansk:
Fuck yeah

PR #109, opened 2023-10-28
[AppGate] Add the MaxDelegatedGateways parameter

h5law:
### Human Summary

- Add max_delegated_gateways param
- Add check in delegation to enforce the param
- Add unit test covering the usage

### Summary generated by Reviewpad on 31 Oct 23 21:44 UTC
1. In the `config.yml` file, a new configuration parameter `maxDelegatedGateways` has been added under the `params` section of the `application` block.
2. The `msg_server_delegate_to_gateway.go` file has changes to the `DelegateToGateway` function, including adding a check to ensure that the application is not already delegated to the maximum number of gateways before delegating it to a new gateway.
3. The `x/application/types/errors.go` file now includes two new error variables, `ErrAppMaxDelegatedGateways` and `ErrAppInvalidMaxDelegatedGateways`.
4. The `genesis_test.go` file has new test cases for the validation of the `GenesisState` struct, including the `MaxDelegatedGateways` field.
5. Changes have been made to the `params.go` file in the `x/application/types` package, including adding a default value for `MaxDelegatedGateways` and validating its value.
6. The `application.go` file in the `testutil/keeper` directory has import and comment changes.
7. The `openapi.yml` file now includes a new property for the maximum number of gateways an application can delegate trust to.
8. The `msg_server_delegate_to_gateway_test.go` file has changes to test cases, including assertions and a new test case for maximum number of gateways reached.
9. The `params.proto` file has a new field for the maximum number of gateways an application can delegate trust to.

## Issue

Fixes (in part) [39](https://github.com/pokt-network/poktroll/issues/39) 

## Type of change

Select one or more:

- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

## Sanity Checklist


x/application/keeper/msg_server_delegate_to_gateway.go
```diff
@@ -36,6 +36,13 @@ func (k msgServer) DelegateToGateway(goCtx context.Context, msg *types.MsgDelega
 		return nil, sdkerrors.Wrapf(types.ErrAppGatewayNotFound, "gateway not found with address: %s", msg.GatewayAddress)
 	}
 
+	// Ensure the application is not already delegated to the maximum number of gateways
+	maxDelegatedParam := k.GetParams(ctx).MaxDelegatedGateways
+	if int64(len(app.DelegateeGatewayPubKeys)) >= maxDelegatedParam {
+		logger.Info("Application already delegated to maximum number of gateways: %d", maxDelegatedParam)
```
bryanchriswhite, line 42: Am I understanding correctly that this log would print on the poktrolld node? If so, are we sure we want that? As a sequencer or supplier (i.e. poktrolld node operator) I wouldn't expect to see logs from message handler functions which are effectively user error.

x/application/keeper/msg_server_delegate_to_gateway.go
```diff
@@ -36,6 +36,13 @@ func (k msgServer) DelegateToGateway(goCtx context.Context, msg *types.MsgDelega
...
+		return nil, sdkerrors.Wrapf(types.ErrAppMaxDelegatedGateways, "application already delegated to maximum number of gateways: %d", maxDelegatedParam)
```
bryanchriswhite, line 43: 1. I was under the impression (link missing - the need for the GH comment search tool returns :sweat_smile:) that the canonical way to import cosmossdk.io/errors was using the `errorsmod` alias.

2. The combination of this `Wrapf()` call and the base error results in quite a verbose error message:
    ```
    application already delegated to maximum number of gateways: 7: maximum number of delegated gateways reached
    ```
    Wdyt reword this one to something like?:

    ```suggestion
		return nil, errorsmod.Wrapf(types.ErrAppMaxDelegatedGateways, "application already delegated to %d gateways", maxDelegatedParam)
    ```
    Resulting in:
    ```
    application already delegated to 7 gateways: maximum number of delegated gateways reached
    ```

x/application/types/params.go
```diff
@@ -7,14 +7,20 @@ import (
 
 var _ paramtypes.ParamSet = (*Params)(nil)
 
+const (
+	DefaultMaxDelegatedGateways int64 = 7
+)
```
bryanchriswhite, line 12: NIT: unless we plan on adding more constants immediately after, I think we should prefer one-line `const` and `var` declarations.

```suggestion
const DefaultMaxDelegatedGateways int64 = 7
```

bryanchriswhite: NIT: I would also move the `const` above the `var`.

x/application/types/params.go
```diff
@@ -7,14 +7,20 @@ import (
...
 func ParamKeyTable() paramtypes.KeyTable {
 	return paramtypes.NewKeyTable().RegisterParamSet(&Params{})
 }
 
 // NewParams creates a new Params instance
 func NewParams() Params {
-	return Params{}
+	return Params{
+		MaxDelegatedGateways: DefaultMaxDelegatedGateways,
+	}
```
bryanchriswhite, line 23: NIT:
```suggestion
	return Params{MaxDelegatedGateways: DefaultMaxDelegatedGateways}
```

bryanchriswhite (dismissed):
Nice small PR! :sunglasses::+1:

x/application/keeper/msg_server_delegate_to_gateway_test.go
```diff
@@ -191,3 +191,61 @@ func TestMsgServer_DelegateToGateway_FailGatewayNotStaked(t *testing.T) {
...
+func TestMsgServer_DelegateToGateway_FailMaxReached(t *testing.T) {
+	k, ctx := keepertest.ApplicationKeeper(t)
+	srv := keeper.NewMsgServerImpl(*k)
+	wctx := sdk.WrapSDKContext(ctx)
+
+	// Generate an address for the application and gateway
+	appAddr := sample.AccAddress()
+	gatewayAddr, gatewayPubKey := sample.AddrAndPubKey()
+	keepertest.AddrToPubKeyMap[gatewayAddr] = gatewayPubKey
+	defer delete(keepertest.AddrToPubKeyMap, gatewayAddr)
```
bryanchriswhite, line 204: #PUC, why do we need to do this?

x/application/keeper/msg_server_delegate_to_gateway_test.go
```diff
@@ -191,3 +191,61 @@ func TestMsgServer_DelegateToGateway_FailGatewayNotStaked(t *testing.T) {
...
+		GatewayAddress: gatewayAddr,
+	}
+
+	// Delegate the application to the max number of gateways
+	maxDelegatedParam := k.GetParams(ctx).MaxDelegatedGateways
+	for i := int64(0); i < k.GetParams(ctx).MaxDelegatedGateways; i++ {
+		// Prepare the delegation message
+		gatewayAddr, gatewayPubKey := sample.AddrAndPubKey()
+		keepertest.AddrToPubKeyMap[gatewayAddr] = gatewayPubKey
+		defer delete(keepertest.AddrToPubKeyMap, gatewayAddr)
```
bryanchriswhite, line 235: #PUC (same as above)

x/application/keeper/msg_server_delegate_to_gateway_test.go
```diff
@@ -191,3 +191,61 @@ func TestMsgServer_DelegateToGateway_FailGatewayNotStaked(t *testing.T) {
...
+			GatewayAddress: gatewayAddr,
+		}
+		// Delegate the application to the gateway
+		_, err = srv.DelegateToGateway(wctx, delegateMsg)
+		require.NoError(t, err)
+	}
+
+	// Attempt to delegate the application when the max is already reached
+	_, err = srv.DelegateToGateway(wctx, delegateMsg)
+	require.Error(t, err)
```
bryanchriswhite, line 247:
```suggestion
	require.ErrorIs(t, err, types.ErrAppMaxDelegatedGateways)
```

x/application/keeper/msg_server_delegate_to_gateway.go
h5law: This is errorsmod aliased to sdkerrors this was a decision olshansky and I had previously. He updated all occurances to be sdkerrors

x/application/keeper/msg_server_delegate_to_gateway.go
Olshansk: **tl;dr I think it's okay to leave as is but open to w/e approach you think is best.**

THEORETICALLY this SHOULD be an event that anyone can capture.

However, in practice, looking at validator/sequencer nodes is how things actually get debugged (from personal experience), so I think it's reasonable.

For example, the tx error would be what we return below and the log line is some telimitary the sequencer operator can use to understand that someone is either missing or doesn't know how to interact with the protocol.

x/application/keeper/msg_server_delegate_to_gateway.go
Olshansk: +1

I think we agreed to use sdkerrors instead of errorsmod from a comment @bryanchriswhite left on one of my PRs :)

x/application/keeper/msg_server_delegate_to_gateway_test.go
Olshansk: Consider using `t.Cleanup` instead.

Olshansk (changes requested):
@bryanchriswhite Left some great comments. In addition to those, @h5law, can you confirm that there is or isn't an opportunity for some `*genesis*` tests?

I'm not 100% sure if there is but my gut is telling me there might be. I'll take a look tomorrow.

h5law:
@Olshansk have added a less than 1 check in the params validate that is called during genesis validation

x/application/types/params.go
```diff
@@ -1,10 +1,13 @@
 package types
 
 import (
+	"cosmossdk.io/errors"
```
Olshansk, line 4:
```suggestion
	sdkerrors "cosmossdk.io/errors"
```

x/application/types/params.go
```diff
@@ -1,10 +1,13 @@
...
 	paramtypes "github.com/cosmos/cosmos-sdk/x/params/types"
 	"gopkg.in/yaml.v2"
 )
 
+const DefaultMaxDelegatedGateways int64 = 7
```
Olshansk, line 9:
```suggestion
// TODO: Revisit default param values
const DefaultMaxDelegatedGateways int64 = 7
```

x/application/keeper/msg_server_delegate_to_gateway_test.go
```diff
@@ -177,8 +177,72 @@ func TestMsgServer_DelegateToGateway_FailGatewayNotStaked(t *testing.T) {
...
+func TestMsgServer_DelegateToGateway_FailMaxReached(t *testing.T) {
+	k, ctx := keepertest.ApplicationKeeper(t)
+	srv := keeper.NewMsgServerImpl(*k)
+	wctx := sdk.WrapSDKContext(ctx)
+
+	// Generate an address for the application and gateway
+	appAddr := sample.AccAddress()
+	gatewayAddr := sample.AccAddress()
+	// Mock the gateway being staked via the staked gateway map
+	keepertest.StakedGatewayMap[gatewayAddr] = struct{}{}
```
Olshansk, line 195: Please add a comment about `StakedGatewayMap` (where we define it) with a `// WARNING: Using this map may cause issues if running tests with parallelization > ``

x/application/keeper/msg_server_delegate_to_gateway_test.go
```diff
@@ -177,8 +177,72 @@ func TestMsgServer_DelegateToGateway_FailGatewayNotStaked(t *testing.T) {
...
+		_, err = srv.DelegateToGateway(wctx, delegateMsg)
+		require.NoError(t, err)
+	}
+
+	// Attempt to delegate the application when the max is already reached
+	_, err = srv.DelegateToGateway(wctx, delegateMsg)
+	require.ErrorIs(t, err, types.ErrAppMaxDelegatedGateways)
+	foundApp, isAppFound := k.GetApplication(ctx, appAddr)
+	require.True(t, isAppFound)
+	require.Equal(t, maxDelegatedParam, int64(len(foundApp.DelegateeGatewayAddresses)))
```
Olshansk, line 247: Wdyt of also checking the number of delegated gateway addresses inside the forloop? As an extra sanity check

Olshansk (changes requested):
A couple minor NITs

x/application/keeper/msg_server_delegate_to_gateway_test.go
h5law: I will add parallisability in a seperate PR probably useful

Olshansk (approved):

PR #116, opened 2023-10-30
[CI] Quick change: install ignite from github releases

okdas:
Quick PR to address an ongoing issue ignite experiences with CloudFlare.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -14,7 +14,9 @@ jobs:
     steps:
       - name: install ignite
         run: |
-          curl https://get.ignite.com/cli! | bash
+          wget https://github.com/ignite/cli/releases/download/v0.27.1/ignite_0.27.1_linux_amd64.tar.gz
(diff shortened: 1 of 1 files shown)


.github/workflows/go.yml
```diff
@@ -14,7 +14,9 @@ jobs:
     steps:
       - name: install ignite
         run: |
-          curl https://get.ignite.com/cli! | bash
+          wget https://github.com/ignite/cli/releases/download/v0.27.1/ignite_0.27.1_linux_amd64.tar.gz
```
okdas, line 17: Also pins to the version we currently use locally.

Olshansk: Can you #PUC that this pins the version and we can use `curl https://get.ignite.com/cli! | bash` (when it works) for the latest & greatest?

Olshansk (approved):
@okdas #PUC (Please Update Comment) with details and alternatives, but other than that, 🚢

okdas:
Closing this, as ignite.com now works. Will add a comment linking to this PR in case a similar issue arises.

PR #94, opened 2023-10-23
[Miner] feat: add `TxClient`

bryanchriswhite:
### Human Summary

```mermaid
---
title: TxClient Dependency Graph
---


flowchart

    subgraph tclient[TxClient]
        tctx[TxContext]
        builder[TxBuilder]
        keyring[Keyring]

        bclient[BlockClient]
        eclient[EventsQueryClient]

        tclient_internal((_))
    end

    chain[Blockchain]

    tctx --"#GetKeyring()"--> keyring
tctx --"#SignTx()"--> builder
tctx --"#EncodeTx()"--> builder
tctx -."#BroadcastTx()".-> chain
tctx -."#QueryTx()".-> chain

eclient -."websocket connection".-> chain
bclient --"committed  block subscription"--> eclient
tclient_internal --"own tx subscription"--> eclient
```

### Summary generated by Reviewpad on 07 Nov 23 07:56 UTC
This pull request includes several changes across multiple files. 

- A new file `godoc.go` has been added to the `pkg/client` directory. This file defines a package called `client` that provides interfaces and types for interacting with blockchain functionalities.

- The `go.mod` file has been modified to update dependencies and remove unused dependencies.

- A new file `services.go` has been added to the `pkg/client` package. This file contains a function `NewTestApplicationServiceConfig` that returns a slice of pointers to `sharedtypes.ApplicationServiceConfig` structs.

- The `context.go` file has been modified to update comments and function names related to the cosmos-sdk transaction handling.

- A new file `README.md` has been added, which provides an overview of the `client` package, installation instructions, usage examples, and other sections.

- The `errors.go` file has been added to the `pkg/client/tx` package. It defines error variables related to transaction handling.

- The `README.md` file has been modified to remove certain sections.

- Several files have been modified or added to the `pkg/client/tx` package, including the addition of a new mock client for the `BlockClient` interface and changes to the `TxClient` and `TxContext` interfaces.

- The `go.sum` file has been updated to include changes to the versions of certain dependencies.

- Files related to test client functionality have been modified or added.

Overall, this pull request includes various additions, modifications, and updates across multiple files to enhance the functionality and documentation of the codebase.

## Issue

Relates to:
- #13 

`TxClient` & `TxContext` are used to build, encode, sign, and  broadcast transactions from an off-chain context.

## Type of change

Select one or more:

- [x] New feature, functionality or library
- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

[...]

PR #118, opened 2023-10-30
[Miner] feat: add `TxContext`

bryanchriswhite:
### Human Summary

Adds `TxContext` interface & `cosmosTxContext` implementation which abstracts cosmos-specific API into discrete concerns.

### Summary generated by Reviewpad on 02 Nov 23 11:30 UTC
This pull request includes various changes across multiple files. 

In the `internal/testclient` package:
- The file `common.go` has been deleted, which contained the `CometLocalWebsocketURL` constant set to `"ws://localhost:36657/websocket"`.
- A new file `keyring.go` has been added, which defines the `NewKey` function for generating a new key and mnemonic for testing purposes.

In the package `testclient`:
- The file `localnet.go` has been added, which includes various imports, functions, and constants related to localnet testing. Some notable additions include the `CometLocalWebsocketURL` constant and the `EncodingConfig` variable.

In the `either` package:
- The `types.go` file has been modified to include changes to the `AsyncError` type and the introduction of the `Bytes` type.

In the `pkg/client/tx` package:
- The `context.go` file has been added, which defines the `cosmosTxContext` struct implementing the `client.TxContext` interface. This file also includes various methods related to transaction context within the Cosmos SDK.

In the `interface.go` file:
- Changes have been made to include `go:generate` directives, import packages, and introduce new interfaces for the client, keyring, and types.

These changes add functionality related to localnet testing, key generation, transaction context, and types across multiple packages.

## Issue

Relates to:
- #13
- #94 

`TxContext`

## Type of change

Select one or more:

- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

## Sanity Checklist


internal/testclient/testtx/context.go
```diff
@@ -0,0 +1,96 @@
...
+				Tx:        nil,
+				Timestamp: "",
+				Events:    nil,
+			}, nil
+		}).Times(1)
+
+	return txCtxMock
+}
+
+func NewAnyTimesTxTxContext(
```
bryanchriswhite, line 65: @h5law PTAL re: mocking cosmos' client context object...

pkg/client/tx/context.go
```diff
@@ -0,0 +1,74 @@
...
+)
+
+var _ client.TxContext = (*cosmosTxContext)(nil)
+
+type cosmosTxContext struct {
+	clientCtx cosmosclient.Context
+	txFactory cosmostx.Factory
+}
+
+func NewTxContext(deps depinject.Config) (client.TxContext, error) {
```
Olshansk, line 24: NIT: godoc comments

pkg/client/tx/context.go
```diff
@@ -0,0 +1,74 @@
...
+	txCtx := cosmosTxContext{}
+
+	if err := depinject.Inject(
```
Olshansk, line 27: ❤️ it

pkg/client/interface.go
```diff
@@ -1,14 +1,42 @@
-//go:generate mockgen -destination=../../internal/mocks/mockclient/query_client_mock.go -package=mockclient . Dialer,Connection
+//go:generate mockgen -destination=../../internal/mocks/mockclient/events_query_client_mock.go -package=mockclient . Dialer,Connection
```
Olshansk, line 1: 😮

internal/testclient/testtx/context.go
```diff
@@ -0,0 +1,350 @@
...
+//
+// The function operates as follows:
+//  1. Retrieves the signer's cryptographic key from the provided keyring based on
+//     the signingKeyName.
+//  2. Determines the corresponding address of the signer's key.
+//  3. Composes an error message suggesting that the fee payer's address is unrecognized.
+//  4. Creates a base mock transaction context using the NewBaseTxContext function.
+//  5. Sets up the mock behavior for the BroadcastTxSync method to return a specific
+//     error response related to the check phase of the transaction.
+func NewOneTimeErrCheckTxTxContext(
```
Olshansk, line 136: Question: Why not have the `Times(1)` be `Times(N)`, remove `OneTime` from the function name and provide `N` as a parameter.

Not a strong preference but curious to hear what you think.

internal/testclient/testtx/context.go
```diff
@@ -0,0 +1,350 @@
...
+	*expectedErrMsg = fmt.Sprintf(
+		"fee payer address: %s does not exist: unknown address",
+		signerAddr.String(),
+	)
+
+	txCtxMock := NewBaseTxContext(
+		t, signingKeyName,
+		keyring,
+		expectedTx,
+		expectedTxHash,
```
Olshansk, line 72: Not actionable, just sharing an idea for the future.

Rather than providing `expectedTxHash` and `expectedTx` as separate params, we could provide a function that transforms `expectedTx` to `expectedTxHas`. Probably overkill for now but just planting the seed.

internal/testclient/testtx/context.go
```diff
@@ -0,0 +1,350 @@
...
+	txCtxMock.EXPECT().NewTxBuilder().
+		DoAndReturn(txCtx.NewTxBuilder).
+		Times(1)
+	txCtxMock.EXPECT().SignTx(
+		gomock.Eq(signingKeyName),
+		gomock.AssignableToTypeOf(txCtx.NewTxBuilder()),
+		gomock.Eq(false), gomock.Eq(false),
+	).DoAndReturn(txCtx.SignTx).Times(1)
+	txCtxMock.EXPECT().EncodeTx(gomock.Any()).
+		DoAndReturn(func(txBuilder cosmosclient.TxBuilder) (_ []byte, err error) {
```
Olshansk, line 280: NIT: Move `func...` to a new line

internal/testclient/testtx/context.go
```diff
@@ -0,0 +1,350 @@
...
+			// intercept cosmosTxContext#EncodeTx to get the encoded tx cometbytes
+			*expectedTx, err = txCtx.EncodeTx(txBuilder)
+			*expectedTxHash = comettypes.Tx(expectedTx.Bytes()).Hash()
+			require.NoError(t, err)
+			return expectedTx.Bytes(), nil
+		}).Times(1)
```
Olshansk, line 286: Given that the function name does not include `OneTime`, is it fair to assume `.Times(1`) everywhere?

internal/testclient/testtx/context.go
```diff
@@ -0,0 +1,350 @@
...
+
+	return txCtxMock
+}
+
+// NewBaseTxContext establishes a foundational mock transaction context with
+// predefined behaviors suitable for a broad range of testing scenarios. It ensures
+// that when interactions like transaction building, signing, and encoding occur
+// in the test environment, they produce predictable and controlled outcomes.
+//
+// Parameters:
```
Olshansk, line 240: Did ChatGPT help out with these?

Olshansk (changes requested):
No blockers. Just a couple NITs & questions, but otherwise g2g!

pkg/client/interface.go
bryanchriswhite: :pray:

internal/testclient/testtx/context.go
bryanchriswhite: Yessss

internal/testclient/common.go
```diff
@@ -1,3 +1,86 @@
...
+// Returns:
+// - A string representation of the default home directory path.
+func DefaultHomeDir() string {
+	userHomeDir, err := os.UserHomeDir()
+	if err != nil {
+		panic(err)
+	}
+
+	return filepath.Join(userHomeDir, "."+app.Name)
+}
```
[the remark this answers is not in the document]
bryanchriswhite: I'm realizing that we don't actually need this at all. We can just use `app.DefaultHomeDir` as the canonical default, it just threw me off because it's a var that's assigned in `init()`, as opposed to a function. :thinking: All that code was generated by `ignite` though.

image

internal/testclient/testtx/context.go
bryanchriswhite: That's a reasonable suggestion IMO. I've mostly been letting the tests drive the shape of the helpers. We don't have  any tests which need to produce an error like this more than once. As it's a pretty straightforward refactor in this case, I would prefer to defer it to when the need arises.

internal/testclient/testtx/context.go
bryanchriswhite: :ok_hand:

internal/testclient/testtx/context.go
bryanchriswhite: **TL;DR: will note in a comment but deferring for now.**

I think that could work but we should be careful about how we set that precedent. There can be multiple mock objects to call `Times()` on which may not necessarily have a constant correlation. In some cases we may even prefer to call `MinTimes()` or `MaxTimes()`,  in which case, it gets more  complicated. I think the unambiguous version of this would end up looking something like:

```go
package  mockthing

func NewThing(t *testing.T, startTimes, doSomethingTimes, repeatMinTimes int) {
  thingMock :=  mockpkg.NewMockThing()
  thingMock.EXPECT().Start(...).Times(startTimes)
  thingMock.EXPECT().DoSomething(...).Times(doSomethingTimes)
  thingMock.EXPECT().Repeat(...).MinTimes(repreatMinTimes)
}
```

Which is probably fine, but unless there's a need for it,  I think I would prefer to wait until such a time  as there is to do that refactor.

internal/testclient/testtx/context.go
bryanchriswhite: Nice catch! :raised_hands:  I think this can be `AnyTimes()` for now.

internal/testclient/testtx/context.go
Olshansk: Makes sense to me!

pkg/client/interface.go
```diff
@@ -1,4 +1,4 @@
-//go:generate mockgen -destination=../../internal/mocks/mockclient/events_query_client_mock.go -package=mockclient . Dialer,Connection
+//go:generate mockgen -destination=../../internal/mocks/mockclient/events_query_client_mock.go -package=mockclient . Dialer,Connection,EventsQueryClient
```
Olshansk, line 1: Seeing we have different packages but keep them defined in `interface.go`, it'll be really nice to get a sense of all the different interfaces we have available.

In fact, this is how I went down the #128 🐇 hole

pkg/client/interface.go
```diff
@@ -19,24 +19,57 @@ import (
 )
 
 type TxContext interface {
```
Olshansk, line 21: #PUC for the interface

pkg/client/interface.go
```diff
@@ -19,24 +19,57 @@ import (
...
 }
 
+// BlocksObservable is an observable which is notified with an either
+// value which contains either an error or the event message bytes.
+// TODO_HACK: The purpose of this type is to work around gomock's lack of
+// support for generic types. For the same reason, this type cannot be an
+// alias (i.e. EventsBytesObservable = observable.Observable[either.Either[[]byte]]).
+type BlocksObservable observable.ReplayObservable[Block]
+
+type BlockClient interface {
```
Olshansk, line 58: Ditto

Olshansk (approved):
A couple NITs but LGTM otherwise!

pkg/client/interface.go
bryanchriswhite: Thanks for calling this out! It left it on the wrong branch. :sweat_smile:

PR #168, opened 2023-11-09
[Miner] feat: add `Miner` component

bryanchriswhite:
### Human Summary

Ports and refactors the `Miner` component from poktroll-alpha.

### Summary generated by Reviewpad on 10 Nov 23 20:30 UTC
This pull request includes the following changes:

- The `interface.go` file in the `pkg/relayer` package has been modified to add a new import statement and several interface definitions and updates. These changes are related to the implementation of the relayer's session lifecycles and the introduction of a new miner interface.

- A new file named "miner.go" has been added in the "pkg/relayer/miner" directory. This file defines a package called "miner" and includes the implementation of the `relayer.Miner` interface. It also contains some utility functions and a constructor for creating a new miner.

- The `relayerSessionsManager` struct, defined in a separate file, has been modified to include a new field, `relayObs`, of type `observable.Observable[*relayer.MinedRelay]`. The struct also includes methods for managing the relayer's session lifecycles, such as starting and stopping sessions, inserting relays, and ensuring the session tree.

- Another file named "claim.go" has been added in the `pkg/relayer/session` package. This file defines the `relayerSessionsManager` struct and includes methods for creating claims for relayer sessions. These methods use observables to handle asynchronous operations and error handling.

- A new file named "miner_test.go" has been added in the `pkg/relayer/miner` package. This file includes a single test function that is currently skipped, indicating that more test coverage needs to be added for the miner implementation.

- The `types.go` file in the `pkg/relayer` package has been modified to introduce a new struct called `MinedRelay`, which is a wrapper around a serialized and hashed relay. This struct contains two fields, `Bytes` and `Hash`, which are of type `[]byte`.

- The `types.go` file in the `pkg/either` package has been modified to add a new import statement and update the `SessionTree` type, which is now an `Either` type wrapping `relayer.SessionTree`. Some type aliases have also been reformatted for clarity.

- A new file named "difficulty.go" has been added in the `pkg/relayer/protocol` directory. This file includes the implementation of a function that checks if a given byte slice exceeds a specified difficulty. This function is used to determine if a relay is volume applicable.

These changes are related to the implementation of the relayer's session lifecycles, the introduction of a new miner interface, and the handling of relay claims and proofs. Some areas, such as test coverage and algorithm optimization, have been identified for future improvements.

Please let me know if you need any further assistance!

## Issue

- #13 

Adds the miner component.

## Type of change

Select one or more:

- [x] New feature, functionality or library
- [ ] Bug fix
- [ ] Code health or cleanup
[...]

PR #172, opened 2023-11-09
[Relayer] feat: Add Relayer struct

red-0ne:
### Summary generated by Reviewpad on 10 Nov 23 22:33 UTC

- In `pkg/relayer/interface.go`, the import statement for `github.com/pokt-network/poktroll/x/service/types` has been removed and replaced with `types`.
- In `pkg/relayer/proxy/proxy.go`, the comment for the `Start` method has been updated to clarify that it is blocking as long as all `RelayServers` are running.
- A new file `pkg/relayer/relayminer.go` has been added, which contains a new struct `relayMiner` and its corresponding methods.
- In `pkg/relayer/session/session.go`, there is a typo fix in a comment, changing "attampting" to "attempting".

## Type of change

Select one or more:

- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

## Sanity Checklist


pkg/relayer/relayer.go
```diff
@@ -0,0 +1,49 @@
...
+
+import (
+	"context"
+
+	"cosmossdk.io/depinject"
+)
+
+type RelayerOption func(*Relayer)
+
+// Relayer is the main struct that encapsulates the relayer's responsibilities.
```
Olshansk, line 11:
```suggestion
// Relayer is the main struct that encapsulates the relayer's responsibilities (i.e. Relay Mining).
```

pkg/relayer/relayer.go
```diff
@@ -0,0 +1,49 @@
...
+// It starts and stops the RelayerProxy and provide the served relays observable to them miner.
+type Relayer struct {
```
Olshansk, line 13: Thoughts on calling this struct (and the file) `RelayMiner`? @bryanchriswhite

pkg/relayer/relayer.go
```diff
@@ -0,0 +1,49 @@
...
+	}
+
+	return rel, nil
+}
+
+// Start provides the miner with the served relays observable and starts the relayer proxy.
+// This method is blocking while the relayer proxy is running and returns when Stop is called
+// or when the relayer proxy fails to start.
+func (rel *Relayer) Start(ctx context.Context) error {
+	rel.miner.MineRelays(ctx, rel.relayerProxy.ServedRelays())
```
Olshansk, line 41: Not clear if this is a background go routine or not. I think it means we either have to add comments and/or rename downstream functions.

maybe `rel.miner.StartMining`?

pkg/relayer/relayer.go
```diff
@@ -0,0 +1,49 @@
...
+	return rel.relayerProxy.Start(ctx)
+}
+
+// Stop stops the relayer proxy which in turn stops all advertised relay servers
+// and unsubscribes the miner from the served relays observable.
+func (rel *Relayer) Stop(ctx context.Context) error {
+	return rel.relayerProxy.Stop(ctx)
```
Olshansk, line 48: Do we have to stop mining?

pkg/relayer/relayer.go
```diff
@@ -0,0 +1,49 @@
...
+
+	return rel, nil
+}
+
+// Start provides the miner with the served relays observable and starts the relayer proxy.
+// This method is blocking while the relayer proxy is running and returns when Stop is called
+// or when the relayer proxy fails to start.
+func (rel *Relayer) Start(ctx context.Context) error {
+	rel.miner.MineRelays(ctx, rel.relayerProxy.ServedRelays())
+	return rel.relayerProxy.Start(ctx)
```
Olshansk, line 42: This might involve some downstream dependencies, but I think it'd be "c

Olshansk (changes requested):

pkg/relayer/relayer.go
red-0ne: Not a background go routine.
Add comments: Yes.
Rename to `StartMining[Relays]`: Yes
:+1:

pkg/relayer/relayer.go
red-0ne: I need to double check, but the idea is that stopping the proxy would unsubscribe so mining stops automatically.

pkg/relayer/relayer.go
red-0ne: Calling `Start` should be enough

pkg/relayer/relayer.go
bryanchriswhite: :joy: I would love that - haha. That's what we originally wanted to call it but IIRC there was some sort of decree that was made and "it shall be called 'the relayer'". 

To be clear on file naming, I would propose `relay_miner.go`.

pkg/relayer/relayer.go
bryanchriswhite: `#MineRelays()` will not block:

```
// MineRelays kicks off relay mining by mapping the servedRelays ovservable through
// a pipeline which hashes the relay, checks if it's above the mining difficulty,
// adds it to the session tree, and then maps any errors to a  new observable.
// It also starts the claim and proof pipelines which are subsequently driven by
// mapping over RelayerSessionsManager's SessionsToClaim return observable.
// It does not block as map operations run in their own goroutines.
```

red-0ne:
We'll wait for #168 to be merge to main before this could be merged

pkg/relayer/relayminer.go
```diff
@@ -0,0 +1,64 @@
...
+
+import (
+	"context"
+	"log"
+
+	"cosmossdk.io/depinject"
+)
+
+// relayMiner is the main struct that encapsulates the relayer's responsibilities (i.e. Relay Mining).
+// It starts and stops the RelayerProxy and provide the served relays observable to them miner.
```
Olshansk, line 11:
```suggestion
// It starts and stops the RelayerProxy and provide the served relays observable to the miner.
```

pkg/relayer/relayminer.go
```diff
@@ -0,0 +1,64 @@
...
+// This method is blocking while the relayer proxy is running and returns when Stop is called
+// or when the relayer proxy fails to start.
+func (rel *relayMiner) Start(ctx context.Context) error {
+	// relayerSessionsManager.Start does not block.
+	// Set up the session (proof/claim) lifecycle pipeline.
+	log.Println("INFO: Starting relayer sessions manager...")
+	rel.relayerSessionsManager.Start(ctx)
+
+	// Start the flow of relays by starting relayer proxy.
+	// This is a blocking call as it waits for the waitgroup to be done.
```
Olshansk, line 50: Which waitgroup are you referring to?

pkg/relayer/relayminer.go
```diff
@@ -0,0 +1,64 @@
...
+	log.Println("INFO: Starting relayer proxy...")
+	if err := rel.relayerProxy.Start(ctx); err != nil {
```
Olshansk, line 52: Is blocking inside of `Start` something we want to keep around long-term or just a current solution? Fine either way but just asking to understand where your mind is at.

Might be a `TODO_REVISIT` if you think it's worth it but not a blocker!

Olshansk (approved):
A couple small comments & questions, but no blockers!

pkg/relayer/relayminer.go
red-0ne: Silly copy past from #177 :facepalm:

pkg/relayer/relayminer.go
red-0ne: I think we want to keep `Start` blocking as it's starting the relayer proxies. The passed context (from cmd) could close it anyway. And this is the top level function that is called from cmd. cc @bryanchriswhite

Olshansk: Sgtm!

Olshansk (approved):

PR #178, opened 2023-11-10
[Test] Updating `relay.feature` to run curl command to enable E2E Relay Test

Olshansk:
## Summary

- Adding `APPGATE_SERVER` to the Makefile
- Adding the ability to run a `curl` command in the `e2e-test framework`
- Particl update to `relay.feature` to start sending a curl command as part of the E2E test

Testing the new test with:

```bash
export APPGATE_SERVER=http://localhost:42069 POCKET_NODE=tcp://127.0.0.1:36657 && \
POKTROLLD_HOME=./localnet/poktrolld && \
go test -run TestFeatures/Relay_Namespace/App_can_send_relay_to_Supplier -v ./e2e/tests/... -tags=e2e
```

## Type of change

Select one or more:

- [x] New feature, functionality or library
- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

- [ ] **Run all unit tests**: `make go_develop_and_test`
- [ ] **Verify Localnet manually**: See the instructions [here](TODO: add link to instructions)

## Sanity Checklist

- [ ] I have tested my changes using the available tooling
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, updated documentation and left TODOs throughout the codebase

PR #177, opened 2023-11-10
[Relay] E2E Relay Gaps

bryanchriswhite:
## Summary

Spike to close the E2E relay & claim/proof loop.

## Issue

- #126 

## Type of change

Select one or more:

- [x] New feature, functionality or library
- [x] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

- [ ] **Run all unit tests**: `make go_develop_and_test`
- [ ] **Verify Localnet manually**: See the instructions [here](TODO: add link to instructions)

## Sanity Checklist

- [ ] I have tested my changes using the available tooling
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, updated documentation and left TODOs throughout the codebase

Issue #126, opened 2023-11-01
[☂️ DRAFT] Umbrella Ticket for automated E2E Relay


Olshansk:
## Objective

[ What? Describe the issue in 1-2 sentences]

## Origin Document

[ Why? Justify the issue in 1-2 sentences. Consider adding a link or a screenshot.]

## Goals

- ...

## Deliverables

- ...

## Non-goals / Non-deliverables

- ...

## General deliverables

- [ ] **Comments**: Add/update TODOs and comments alongside the source code so it is easier to follow.
- [ ] **Testing**: Add new tests (unit and/or E2E) to the test suite.
- [ ] **Makefile**: Add new targets to the Makefile to make the new functionality easier to use.
- [ ] **Documentation**: Update architectural or development READMEs; use [mermaid](https://mermaid-js.github.io/mermaid/) diagrams where appropriate.

---

**Creator**: [GitHub handle of issue owner]
**Co-Owners**: [OPTIONAL - GitHub handle of co-owner(s)]

PR #184, opened 2023-11-14
[E2E relay test] Update to #178

bryanchriswhite:
### Human Summary

This PR is based on a branch that merges #178, `main`, and #177 and builds on top to get an E2E relay test passing (and failing) for the right reasons.

- Adds `-features-path` flag to the E2E go test such that we can select (via glob pattern) which features we want to run on invocation.
- Implements `TheApplicationReceivesASuccessfulRelayResponseSignedBy`.
- Improves error E2E messaging.

#### To run the relay test (outside of tilt):

```bash
# Start the appgate-server (blocks terminal)
poktrolld --home ./localnet/poktrolld appgate-server --signing-key app1 --self-signing --keyring-backend test --node tcp://127.0.0.1:36657

# Start the relayminer (blocks terminal)
poktrolld --home ./localnet/poktrolld/ relayminer --signing-key supplier1 --keyring-backend test --node tcp://127.0.0.1:36657

# Stake supplier1 & app1 so their keys are on-chain
make supplier1_stake && make app1_stake

# Run just the relay feature file
go test -v -tags=e2e ./e2e/tests  -features-path=./relay.feature
```

### Summary generated by Reviewpad on 14 Nov 23 16:09 UTC
- Added `flag` package import
- Added `featuresPathFlag` and `keyRingFlag` variables
- Added `features-path` flag in `TestMain` function
- Renamed `TheApplicationSendsTheSupplierARequestForServiceWithData` function to `TheApplicationSendsTheSupplierARequestForServiceWithData` and added code comments
- Added `getRelayKey` function
- Added `TheApplicationReceivesASuccessfulRelayResponseSignedBy` function
- Added `runCurlPostCmd` function in `pocketdBin` struct
- Modified error message in `ErrAppGateEmptyRelayResponse` constant
- Modified error handling in `verifyResponse` function in `appGateServer` struct
- Modified error handling in `VerifyRelayRequest` function in `relayerProxy` struct

## Issue

Relates to:
- #178 
- #126 

It would be really useful to have the E2E relay test working already such that we can quickly and easily tell if the integration has regressed to a broken state. This should close that loop.

## Type of change

Select one or more:

- [ ] Bug fix
- [x] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

## Sanity Checklist


h5law (approved):
Short and sweet, nice changes nothing stands out lets get this merged 🔥

e2e/tests/init_test.go
```diff
@@ -31,12 +33,21 @@ var (
 	accNameToAppMap      = make(map[string]apptypes.Application)
 	accNameToSupplierMap = make(map[string]sharedtypes.Supplier)
 
-	keyRingFlag = "--keyring-backend=test"
+	featuresPathFlag string
```
Olshansk, line 36: NIT: Let's go with `flagFeaturesPath` to stay consistent with cosmos. https://pkg.go.dev/github.com/cosmos/cosmos-sdk/client/flags

e2e/tests/init_test.go
```diff
@@ -31,12 +33,21 @@ var (
...
+	keyRingFlag      = "--keyring-backend=test"
```
Olshansk, line 37: flagKeyRingDefault?

e2e/tests/init_test.go
```diff
@@ -243,15 +254,30 @@ func (s *suite) TheSessionForApplicationAndServiceContainsTheSupplier(appName st
 }
 
 func (s *suite) TheApplicationSendsTheSupplierARequestForServiceWithData(appName, supplierName, serviceId, requestData string) {
-	res, err := s.pocketd.RunCurl("", serviceId, requestData)
+	// TODO_TECHDEBT(#179): Once relayminer and appgateserver are running in tilt,
+	// use their respective in-tilt hostnames and run E2E tests in tilt. This
+	// should match the on-chain advertised endpoint for the service with the
+	// given serviceId.
+	res, err := s.pocketd.RunCurl("http://localhost:42069", serviceId, requestData)
```
Olshansk, line 261: Thoughts on moving `localhost` into a local const and moving the comment along with it?

That way we're decoupling the "constant test state" and test parameters.

e2e/tests/init_test.go
```diff
@@ -243,15 +254,30 @@ func (s *suite) TheSessionForApplicationAndServiceContainsTheSupplier(appName st
...
 		s.Fatalf("error sending relay request from app %s to supplier %s for service %s: %v", appName, supplierName, serviceId, err)
 	}
-	fmt.Println("OLSH Res", res.Stdout)
+
+	relayKey := getRelayKey(appName, supplierName)
+	s.scenarioState[relayKey] = res.Stdout
+}
+
+// TODO_IN_THIS_COMMIT: move
+func getRelayKey(appName, supplierName string) string {
```
Olshansk, line 271: 1. Consider renaming to `relayReferenceKey` + godoc comment.
2. `TODO_IMPROVE` to use (`sessionId`, `supplierName`) since those are the two values used to create the primary composite key on-chain to uniquely distinguish relays

pkg/appgateserver/relay_verifier.go
```diff
@@ -26,7 +26,16 @@ func (app *appGateServer) verifyResponse(
 
 	// Extract the supplier's signature
 	if relayResponse.Meta == nil {
-		return ErrAppGateEmptyRelayResponse
+		payload := relayResponse.GetPayload()
+		payloadBz := make([]byte, payload.Size())
+		if _, err := payload.MarshalTo(payloadBz); err != nil {
+			return ErrAppGateEmptyRelayResponseSignature.Wrapf(
+				"unable to unmarshal relay response payload: %s", err,
```
Olshansk, line 33:
```suggestion
				"unable to marshal relay response payload: %s", err,
```

pkg/appgateserver/relay_verifier.go
```diff
@@ -26,7 +26,16 @@ func (app *appGateServer) verifyResponse(
 
 	// Extract the supplier's signature
 	if relayResponse.Meta == nil {
-		return ErrAppGateEmptyRelayResponse
+		payload := relayResponse.GetPayload()
+		payloadBz := make([]byte, payload.Size())
+		if _, err := payload.MarshalTo(payloadBz); err != nil {
+			return ErrAppGateEmptyRelayResponseSignature.Wrapf(
```
Olshansk, line 32: NIT: Should we be using `Signature` or `Metadata` in the error name?

pkg/appgateserver/errors.go
```diff
@@ -10,6 +10,6 @@ var (
 	ErrAppGateMissingAppAddress             = sdkerrors.Register(codespace, 4, "missing application address")
 	ErrAppGateMissingSigningInformation     = sdkerrors.Register(codespace, 5, "missing app client signing information")
 	ErrAppGateMissingListeningEndpoint      = sdkerrors.Register(codespace, 6, "missing app client listening endpoint")
-	ErrAppGateEmptyRelayResponse            = sdkerrors.Register(codespace, 7, "empty relay response")
+	ErrAppGateEmptyRelayResponseSignature   = sdkerrors.Register(codespace, 7, "empty relay response signature")
```
Olshansk, line 13: Consider creating another error called metadata

Olshansk (changes requested):

e2e/tests/init_test.go
bryanchriswhite: I'm reluctant to align the naming between these two "flag vars". This is categorically different than the features path flag, it is not registered as a go flag but is instead both the key and value of a flag which is passed to binaries called by the go command API.

pkg/appgateserver/relay_verifier.go
bryanchriswhite: I think I misinterpreted a comment.

e2e/tests/init_test.go
bryanchriswhite: **TL;DR, adding.**

I think we would need to alter the test suite design a bit if we really need to distinguish between individual relays in this E2E suite as it would also require tracking an index (or some unique field) on the relays themselves. I was assuming that we won't be needing that level of discrimination, at least not in this suite and/or any time soon. My focus was to close the E2E relay test loop.

e2e/tests/init_test.go
```diff
@@ -22,6 +24,12 @@ import (
 	suppliertypes "github.com/pokt-network/poktroll/x/supplier/types"
 )
 
+// TODO_TECHDEBT(#179): Once relayminer and appgateserver are running in tilt,
```
Olshansk, line 27:
```suggestion
// TODO_TECHDEBT(@okdas): Once relayminer and appgateserver are running in tilt,
```

Olshansk (approved):
A minor NIT and requires a merge with main but approving pre-emptively!

It's broken on main as is, so we'll fix things afterwards.

PR #186, opened 2023-11-14
[Proto] Enable both maps of params and lists of params in the `JSONRPCRequestPayload`

h5law:
### Human Summary

- Add oneof to JSONRPCRequestPayload proto
- Add interim unmarshalling step
- Handle both maps and list types of params

### Summary generated by Reviewpad on 14 Nov 23 22:53 UTC
This pull request introduces a feature that generalizes the parameters in the JSON RPC request to be either maps or lists. It modifies multiple files to implement this feature.

## Issue

- Fixes N/A

## Type of change

Select one or more:

- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

## Sanity Checklist

### DIFF
--- a/pkg/appgateserver/cmd/cmd.go
+++ b/pkg/appgateserver/cmd/cmd.go
@@ -165,12 +165,7 @@ func newSupplyQueryClientContextFn(pocketQueryClientUrl string) config.SupplierF
 			return nil, err
 		}
 
-		// NB: Currently, the implementations of GetClientTxContext() and
-		// GetClientQueryContext() are identical, allowing for their interchangeable
-		// use in both querying and transaction operations. However, in order to support
-		// independent configuration of client contexts for distinct querying and
-		// transacting purposes. E.g.: transactions are dispatched to the sequencer
-		// while queries are handled by a trusted full-node.
+		// Get the client context from the command.
 		queryClientCtx, err := cosmosclient.GetClientQueryContext(cmd)
 		if err != nil {
 			return nil, err


pkg/appgateserver/jsonrpc.go
```diff
@@ -13,6 +14,15 @@ import (
 	sharedtypes "github.com/pokt-network/poktroll/x/shared/types"
 )
 
+// InterimJSONRPCRequestPayload is a partial JSON RPC request payload that
+// excludes the params field, which is unmarshaled sperately.
```
Olshansk, line 18:
```suggestion
// excludes the params field, which is unmarshaled separately.
// This is used because the proto object cannot be unmarshalled directly, so this is a
// temporary object to be used interim.
```

pkg/appgateserver/jsonrpc.go
```diff
@@ -13,6 +14,15 @@ import (
...
+type InterimJSONRPCRequestPayload struct {
```
Olshansk, line 19: Does the struct and its types need to be exposed publically?

proto/pocket/service/relay.proto
```diff
@@ -40,10 +41,22 @@ message JSONRPCRequestPayload {
...
-    //map<string, string> params = 4; // Parameters for the method. https://www.jsonrpc.org/specification#parameter_structures
-    repeated string params = 4; // Parameters for the method. https://www.jsonrpc.org/specification#parameter_structures
+    oneof params {
+        JSONRPCRequestPayloadParamsList list_params = 4 [(gogoproto.jsontag) = "params"];
+        JSONRPCRequestPayloadParamsMap map_params = 5 [(gogoproto.jsontag) = "params"];
+    }
+}
+
+// JSONRPCRequestPayloadParamsList contains the list of parameters for a JSON-RPC request.
+// TODO_RESEARCH: Will this always be a string list? Or should we use a more generic type?
```
Olshansk, line 51: I thought about it and realized this could be a string of objects: `[obj1, obj2, etc..]`.

Wdyt of just making `prams` an `Any` like you suggested in discord?

Olshansk (changes requested):
Asking a pretty big structural question before moving on.

proto/pocket/service/relay.proto
h5law: @Olshansk I think an approach could be storing the payload as its bytes. And just passing these directly. The downside of this is what if we wanted to do compute units?

Answer: we interim unmarshal without the params as compute units **should** probably only depend on the method being called (params shouldn't affect the method as the method's "weighting" would factor in the types of params it receives) and assign the weighting that way.

The reason I am against using `Any` in the protobuf is that the protobuf `Any` isn't 1-1 correspondance with the go `any` ie we cannot unmarhsal any type into an `Any` an any needs to be an `Any` type in go which the unmarhsaling step would again require some interim step of unmarshaling without the params and then a longer type checking.

Overall I think us trying to proto-ify the payload is probably the wrong path. What do we need it to be a protobuf for? Why cant be just directly pass the payload from the request to the service? Personally I think making the payload a protobuf is not for us to deal with - as we dont do anything with it, unlike the other field of the request structure.

WDYT?

PR #129, opened 2023-11-02
[Tooling] add `go_lint` & `go_imports` make targets & CI step

bryanchriswhite:
### Human Summary

- Updates `install_ci_deps` make target to install `golangci-lint` and `goimports`
- Adds .golangci.yml config file; poised to progressively enable linters
- Adds  `go_lint` make target to run `golangci-lint`
- Adds  `go_imports` make target to run `goimports` via a go script: /tools/scripts/goimports.go
- Fixes non-excluded goimport issues
- Add stubs for mock packages that were causing CI to fail & cleanup .gitkeep files

## Issue

Related to:
- #128 

#128 surfaces some goimports issues and touches basically all imports already; might as well fix them all.

## Type of change

Select one or more:

- [ ] New feature, functionality or library
- [ ] Bug fix
- [x] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

- [ ] **Run all unit tests**: `make go_develop_and_test`
- [ ] **Verify Localnet manually**: See the instructions [here](TODO: add link to instructions)

## Sanity Checklist

- [x] I have tested my changes using the available tooling
- [ ] I have performed a self-review of my own code
- [x] I have commented my code, updated documentation and left TODOs throughout the codebase

PR #187, opened 2023-11-15
[Rings] Create a shared `RingCache` package

h5law:
### Human Summary

This PR creates a shared `RingCache` package responsible for fetching, caching, and invalidating rings used for signing and verification of relay requests. It is used by both the `AppGateServer` and `RelayMiner`

### AI Summary

reviewpad:summary

## Issue

- Fixes N/A (#126)

## Type of change

Select one or more:

- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

## Sanity Checklist


pkg/crypto/rings/cache.go
```diff
@@ -0,0 +1,201 @@
...
+
+	clientCtx := cosmosclient.Context(rc.clientCtx)
+
+	rc.accountQuerier = accounttypes.NewQueryClient(clientCtx)
+	rc.applicationQuerier = apptypes.NewQueryClient(clientCtx)
+
+	return rc, nil
+}
+
+// GetRingForAddress returns the ring for the address provided. If it does not
```
Olshansk, line 67: >  If it does not exist in the cache, it will be created by querying the application module

Congrats, you've rediscovered memorization: https://en.wikipedia.org/wiki/Memoization

pkg/crypto/rings/cache.go
```diff
@@ -0,0 +1,201 @@
...
+	}
+	return newRingFromPoints(points)
+}
+
+// newRingFromPoints creates a new ring from points on the secp256k1 curve
+func newRingFromPoints(points []ringtypes.Point) (*ring.Ring, error) {
+	return ring.NewFixedKeyRingFromPublicKeys(ring_secp256k1.NewCurve(), points)
+}
+
+// getDelegatedPubKeysForAddress returns the ring used to sign a message for
```
Olshansk, line 121: I feel like this function is doing a lot more than `getDelegatedPubKeysForAddress` and doesn't actually return pub keys but a ring.

We should either:
1. Break it out into 2 smaller helpers
2. Rename it

My preference is for (1) but not a stong opinion.

pkg/crypto/rings/cache.go
```diff
@@ -0,0 +1,201 @@
...
+	appAddress string,
+) ([]ringtypes.Point, error) {
+	rc.ringPointsMu.Lock()
+	defer rc.ringPointsMu.Unlock()
+
+	// get the application's on chain state
+	req := &apptypes.QueryGetApplicationRequest{Address: appAddress}
+	res, err := rc.applicationQuerier.Application(ctx, req)
+	if err != nil {
+		return nil, ErrRingsAccountNotFound.Wrapf("address: %s [%v]", appAddress, err)
```
Olshansk, line 136:
```suggestion
		return nil, ErrRingsAccountNotFound.Wrapf("app address: %s [%v]", appAddress, err)
```

pkg/crypto/rings/cache.go
```diff
@@ -0,0 +1,201 @@
...
+	ringAddresses = append(ringAddresses, appAddress) // app address is index 0
+	if len(res.Application.DelegateeGatewayAddresses) < 1 {
+		// add app address twice to make the ring size of mininmum 2
+		// TODO_HACK: We are adding the appAddress twice because a ring
+		// signature requires AT LEAST two pubKeys. When the Application has
+		// not delegated to any gateways, we add the application's own address
+		// twice. This is a HACK and should be investigated as to what is the
+		// best approach to take in this situation.
+		ringAddresses = append(ringAddresses, appAddress)
+	} else if len(res.Application.DelegateeGatewayAddresses) > 0 {
```
Olshansk, line 150: I think this if structure would read easier if you did:

```go
if len(res.Application.DelegateeGatewayAddresses)  == 0 {...}
else {...}
```

pkg/crypto/rings/cache.go
```diff
@@ -0,0 +1,201 @@
...
+
+	// get the points on the secp256k1 curve for the addresses
+	points, err := rc.addressesToPoints(ctx, ringAddresses)
+	if err != nil {
+		return nil, err
+	}
+
+	// update the cache overwriting the previous value
+	rc.ringPointsCache[appAddress] = points
+
```
Olshansk, line 163: Good opportunity for a log line that you're updating the cache for `appAddress`

pkg/crypto/rings/interface.go
```diff
@@ -0,0 +1,15 @@
...
+
+import (
+	"context"
+
+	"github.com/noot/ring-go"
+)
+
+// RingCache is used to store rings used for signing and verifying relay requests.
+// It will cache rings for future use after querying the application module for
+// the addresses of the gateways the application is delegated to, and converting
```
Olshansk, line 11:
```suggestion
// the pubKeys of the gateways the application is delegated to, and converting
```

pkg/deps/config/config.go
```diff
@@ -71,3 +74,55 @@ func NewSupplyBlockClientFn(pocketNodeWebsocketUrl string) SupplierFn {
...
+		if err != nil {
+			return nil, err
+		}
+
+		// NB: Currently, the implementations of GetClientTxContext() and
+		// GetClientQueryContext() are identical, allowing for their interchangeable
+		// use in both querying and transaction operations. However, in order to support
+		// independent configuration of client contexts for distinct querying and
+		// transacting purposes. E.g.: transactions are dispatched to the sequencer
+		// while queries are handled by a trusted full-node.
```
Olshansk, line 98: Tbh, I think we might end up using the pocket network as a full node for doing this. 

It's going to get meta

pkg/crypto/rings/cache.go
```diff
@@ -0,0 +1,201 @@
...
+	points := make([]ringtypes.Point, len(addresses))
+	for i, addr := range addresses {
+		pubKeyReq := &accounttypes.QueryAccountRequest{Address: addr}
+		pubKeyRes, err := rc.accountQuerier.Account(ctx, pubKeyReq)
+		if err != nil {
+			return nil, ErrRingsAccountNotFound.Wrapf("address: %s [%v]", addr, err)
+		}
+		var acc accounttypes.AccountI
+		reg := codectypes.NewInterfaceRegistry()
+		accounttypes.RegisterInterfaces(reg)
```
Olshansk, line 185: OPTIONAL NIT: Decoupling business logic that does "codec and math" away from code that does "network interfacing" will really help readability, maintainability & testing.

pkg/crypto/rings/cache.go
```diff
@@ -0,0 +1,201 @@
...
+	}
+
+	// update the cache overwriting the previous value
+	rc.ringPointsCache[appAddress] = points
+
+	// return the public key points on the secp256k1 curve
+	return points, nil
+}
+
+// addressesToPoints converts a slice of addresses to a slice of points on the
```
Olshansk, line 168: Do you plan to add unit tests in this PR or a followup? If it's a follow up, please add a `TODO_TEST(@h5law)`

Olshansk (changes requested):
Really like how the codebase is evolving!

giphy

pkg/crypto/rings/cache.go
h5law: I actually did this one in class recently lmao (recursion and backtracking) 🤓

pkg/crypto/rings/cache.go
h5law: It does return pubkeys - just in the format we desire `Points` these are then used to build the ring - they are not the ring itself

pkg/crypto/rings/cache.go
h5law: No unit tests in this one: Potentially 2 (definietly 1) follow up PR for this pkg:

1. Listen for application module `EventDelegation` events
  - create the event and emit on delegation and undelegations
  - add events query client as dependency and use accordingly to invalidate address in cache
2. Potentially invalidate cache dependiing on a specified time period after inactivity
  - add block listener
  - invalidate inactive addresses in cache
    - track inactivity
    - 3 maps? points, access time, expirary time?

pkg/crypto/rings/interface.go
h5law: We are actually querying for addresses tho, thats how we store them in the application

pkg/crypto/rings/cache.go
Olshansk: First one sounds good. 

Second one might be a bit overkill since it'll be a lot of work to test cache features in a cache. I'd leave a note about it as an idea and move on.

pkg/deps/config/config.go
```diff
@@ -37,36 +42,129 @@ func SupplyConfig(
...
 		}
 
 		return depinject.Configs(deps, depinject.Supply(blockClient)), nil
 	}
 }
+
+// NewSupplyQueryClientContextFn returns a function with constructs a ClientContext
+// instance with the given cmd and returns a new depinject.Config which is
+// supplied with the given deps and the new ClientContext.
+func NewSupplyQueryClientContextFn(pocketQueryNodeURL string) SupplierFn {
```
bryanchriswhite, line 82: :ok_hand: solid refactor here!

pkg/deps/config/config.go
```diff
@@ -37,36 +42,129 @@ func SupplyConfig(
...
+	return func(_ context.Context,
+		deps depinject.Config,
+		cmd *cobra.Command,
+	) (depinject.Config, error) {
+		// Set --node flag to the --pocket-node for the client context
+		// This flag is read by cosmosclient.GetClientQueryContext.
+		err := cmd.Flags().Set(cosmosflags.FlagNode, pocketQueryNodeURL)
```
bryanchriswhite, line 89: Given the apparent design of parts of the cosmos-sdk and that they seem to effectively rely on the shared mutable state of the command flags, perhaps we should consider either resetting this value to empty string or whatever it was before we modified it. Wdyt?

pkg/deps/config/config.go
```diff
@@ -37,36 +42,129 @@ func SupplyConfig(
 }
 
 // NewSupplyEventsQueryClientFn returns a new function which constructs an
-// EventsQueryClient instance and returns a new depinject.Config which is supplied
-// with the given deps and the new EventsQueryClient.
+// EventsQueryClient instance and returns a new depinject.Config which is
+// supplied with the given deps and the new EventsQueryClient.
 func NewSupplyEventsQueryClientFn(
-	pocketNodeWebsocketUrl string,
+	pocketNodeWebsocketURL string,
```
bryanchriswhite, line 48: NIT: **TL;DR, this is technically, conventionally correct; however, I think we should prefer (dromedary) camel-casing acronyms for consistency for now but we should correct later.**

Thanks for taking notice of this @h5law! :raised_hands: While it would indeed be more conventional to use all caps for variable names (or parts thereof) consisting of acronyms, we seem to have gotten into a habit of ignoring this exception to the go variable naming convention extremely consistently thus far.

When I was pulling #129, I was reminded of our unconventional approach and figured that we'd have the opportunity to rectify it when we start [turning on linters](https://github.com/pokt-network/poktroll/pull/129/files#diff-6179837f7df53a6f05c522b6b7bb566d484d5465d9894fb04910dd08bb40dcc9), specifically `govet` IIRC.

pkg/deps/types/accquerier.go
bryanchriswhite, line 1: I :heart: that you've introduced an interface here! This is great because it more concisely represents the **shared/common** needs of **our** pkgs. :mechanical_arm:  It also, of course, provides better decoupling from cosmos-sdk, which I think is something we should prefer, especially when it's convenient in the moment. This is an early (and compounding) investment in our pkgs being able to support non-cosmos chains. :rocket:

pkg/deps/types/accquerier.go
```diff
@@ -0,0 +1,53 @@
...
+	"context"
+
+	"cosmossdk.io/depinject"
+	"github.com/cosmos/cosmos-sdk/client"
+	accounttypes "github.com/cosmos/cosmos-sdk/x/auth/types"
+
+	"github.com/pokt-network/poktroll/pkg/relayer"
+)
+
+type AccountQuerier interface {
```
bryanchriswhite, line 13: I :heart: that you've introduced an interface here! This is great because it more concisely represents the **shared/common** needs of **our** pkgs. :mechanical_arm:  It also, of course, provides better decoupling from cosmos-sdk, which I think is something we should prefer, especially when it's convenient in the moment. This is an early (and compounding) investment in our pkgs being able to support non-cosmos chains. :rocket:

That being said, I feel like this interface probably makes more sense to live in pkg/client/interface.go. This is a real thing describing real behavior, let's categorize it as such. :sunglasses:

pkg/deps/types/accquerier.go
```diff
@@ -0,0 +1,53 @@
...
+	GetAccount(ctx context.Context, address string) (accounttypes.AccountI, error)
+}
+
+type accQuerier struct {
```
bryanchriswhite, line 17: Following-on from the comment above, I would suggest moving this to a `pkg/client/query` package.

In general, we should (strongly) prefer putting interfaces in distinct packages from their implementations so as to avoid any potential for import cycles between them.

E.g. because the `Notification` interface is defined next to its implementation, `user` pkg has to import it, creating a cycle:
```go
// user/user.go
package user

import (
    "example/notification"
)

// User struct
type User struct {
    Name string
}

// SendNotification function that takes a Notification interface
func SendNotification(u User, n notification.Notification) {
    n.Notify(u)
}
```

```go
// notification/notification.go
package notification

import (
    "example/user"
    "fmt"
)

// Notification interface
type Notification interface {
    Notify(u user.User)
}

// UserNotification struct implementing Notification
type UserNotification struct{}

// Notify method for UserNotification
func (n UserNotification) Notify(u user.User) {
    fmt.Printf("Notifying user %s\n", u.Name)
}
```

pkg/deps/types/appquerier.go
```diff
@@ -0,0 +1,49 @@
...
+
+	"github.com/pokt-network/poktroll/pkg/relayer"
+	apptypes "github.com/pokt-network/poktroll/x/application/types"
+)
+
+type ApplicationQuerier interface {
+	GetApplication(ctx context.Context, appAddress string) (apptypes.Application, error)
+}
+
+type appQuerier struct {
```
bryanchriswhite, line 17: same as https://github.com/pokt-network/poktroll/pull/187#discussion_r1398799602; let's move this to `pkg/client/query`

pkg/deps/types/appquerier.go
```diff
@@ -0,0 +1,49 @@
...
+	"context"
+
+	"cosmossdk.io/depinject"
+	"github.com/cosmos/cosmos-sdk/client"
+
+	"github.com/pokt-network/poktroll/pkg/relayer"
+	apptypes "github.com/pokt-network/poktroll/x/application/types"
+)
+
+type ApplicationQuerier interface {
```
bryanchriswhite, line 13: same as https://github.com/pokt-network/poktroll/pull/187#discussion_r1398782145; let's move this to `pkg/client/interface.go`.

pkg/appgateserver/server.go
```diff
@@ -51,16 +52,13 @@ type appGateServer struct {
...
-	ringCache      map[string][]ringtypes.Point
-	ringCacheMutex *sync.RWMutex
+	// ringCache is used to obtain and store the ring for the application.
+	ringCache rings.RingCache
 
 	// clientCtx is the client context for the application.
 	// It is used to query for the application's account to unmarshal the supplier's account
 	// and get the public key to verify the relay response signature.
-	clientCtx sdkclient.Context
+	clientCtx relayer.QueryClientContext
```
bryanchriswhite, line 61: I think it may make sense to move `relayer.QueryClientContext` to `pkg/client/query` as well. If it's not too much trouble, I think it would make this PR easier to review if the query client refactor gets moved out to its own PR.

(Bonus points for moving `relayer.TxClientContext` as well.  :sweat_smile:)

pkg/deps/config/config.go
h5law: Yeah my linter started complaining here lmao

pkg/client/interface.go
```diff
@@ -176,3 +178,23 @@ type TxClientOption func(TxClient)
 
 // SupplierClientOption defines a function type that modifies the SupplierClient.
 type SupplierClientOption func(SupplierClient)
+
+// QueryClientContext is used to distinguish a cosmosclient.Context intended
+// for use in queries from others.
+// This type is intentionally not an alias in order to make this distinction
+// clear to the dependency injector
+type QueryClientContext cosmosclient.Context
```
bryanchriswhite, line 186: Wdyt about renaming this to just  `QueryContext`? I feel like `Client` is redundant now that it's in the `client ` pkg.

pkg/client/query/appquerier.go
```diff
@@ -0,0 +1,52 @@
...
+
+// GetApplication returns an apptypes.Application interface for a given address
+func (aq *appQuerier) GetApplication(
+	ctx context.Context,
+	appAddress string,
+) (apptypes.Application, error) {
+	req := apptypes.QueryGetApplicationRequest{Address: appAddress}
+	res, err := aq.applicationQuerier.Application(ctx, &req)
+	if err != nil {
+		return apptypes.Application{}, ErrQueryAccountNotFound.Wrapf("app address: %s [%v]", appAddress, err)
```
Olshansk, line 49: Should this be an app error rather than a `QueryAccount` error?

testutil/testclient/testqueryclients/appquerier.go
```diff
@@ -0,0 +1,86 @@
...
+	"github.com/pokt-network/poktroll/x/application/types"
+	apptypes "github.com/pokt-network/poktroll/x/application/types"
+	sharedtypes "github.com/pokt-network/poktroll/x/shared/types"
+)
+
+// addressApplicationMap is a map of application addresses that are deemed to
+// exist on chain, if an address is not in this map an error will be returned
+// from the mock ApplicationQueryClient's GetApplication method.
+// The array of strings is the addresses of the delegated gateways that the
+// application is supposed to have.
```
Olshansk, line 21:
```suggestion
// addressApplicationMap is a map of:
// 	applicationAddresses -> list of delegated gateways by app.
// All addresses are assumed to exist on chain.
```

Olshansk: 1. I think the error piece is good but doesn't relate to the map
2. Sometimes, less words and more pseudo code in the comments is easier for future developers to read.

testutil/testclient/testqueryclients/appquerier.go
```diff
@@ -0,0 +1,86 @@
...
+var addressApplicationMap map[string][]string
```
Olshansk, line 22: Why not `appToGatewaysMap`?

testutil/testclient/testqueryclients/appquerier.go
```diff
@@ -0,0 +1,86 @@
...
+
+func init() {
+	addressApplicationMap = make(map[string][]string)
+}
+
+// NewTestApplicationQueryClient creates a mock of the ApplicationQueryClient
+// which allows the caller to call GetApplication any times and will return
+// an application with the given address.
+// The delegateeNumber parameter is used to determine how many delegated
```
Olshansk, line 31: There is no `delegateeNumber` parameter

pkg/client/query/codec.go
```diff
@@ -0,0 +1,18 @@
+package query
+
+import (
+	"github.com/cosmos/cosmos-sdk/codec"
+	codectypes "github.com/cosmos/cosmos-sdk/codec/types"
+	accounttypes "github.com/cosmos/cosmos-sdk/x/auth/types"
+)
+
+// queryCodec is a codec used to unmarshal the account interface returned by the
```
Olshansk, line 9: 1. I feel like this comment is very specific to a specific use case, but can actually be applicable to so much more.
2. Can you add a `TODO_TEHDEBT` to make sure we use this everywhere?

cc @bryanchriswhite

pkg/client/query/types/context.go
```diff
@@ -0,0 +1,11 @@
+package types
+
+import (
+	cosmosclient "github.com/cosmos/cosmos-sdk/client"
+)
+
+// Context is used to distinguish a cosmosclient.Context intended
```
Olshansk, line 7: It's still unclear to me why we need this even after reading the comment.

#PUC?

testutil/mockclient/application_query_client.go
```diff
@@ -0,0 +1,51 @@
+// Code generated by MockGen. DO NOT EDIT.
```
Olshansk, line 1: We shouldn't be committing this.

testutil/mockclient/account_query_client.go
```diff
@@ -0,0 +1,51 @@
+// Code generated by MockGen. DO NOT EDIT.
```
Olshansk, line 1: We shuoldn't be commiting this.

Olshansk (changes requested):
Still have a few files left for review so I'm just posting my comments up until now.

Screenshot 2023-11-22 at 8 29 48 PM

pkg/appgateserver/relay_verifier.go
```diff
@@ -60,21 +57,11 @@ func (app *appGateServer) getSupplierPubKeyFromAddress(
 
 	// Query for the supplier account to get the application's public key
 	// to verify the relay request signature.
-	accQueryReq := &accounttypes.QueryAccountRequest{Address: supplierAddress}
-	accQueryRes, err := app.accountQuerier.Account(ctx, accQueryReq)
+	acc, err := app.accountQuerier.GetAccount(ctx, supplierAddress)
```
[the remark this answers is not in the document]
h5law: the beauty of abstraction

pkg/client/query/accquerier.go
```diff
@@ -0,0 +1,56 @@
...
+// querying of on-chain account information through a single exposed method
+// which returns an accounttypes.AccountI interface
+type accQuerier struct {
+	clientCtx      types.Context
+	accountQuerier accounttypes.QueryClient
+}
+
+// NewAccountQuerier returns a new instance of a client.AccountQueryClient by
+// injecting the dependecies provided by the depinject.Config
+func NewAccountQuerier(
```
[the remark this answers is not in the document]
h5law: This one was probably while I was wrestling with my linters 😅

pkg/client/query/accquerier.go
```diff
@@ -0,0 +1,56 @@
...
+	ctx context.Context,
+	address string,
+) (accounttypes.AccountI, error) {
+	req := &accounttypes.QueryAccountRequest{Address: address}
+	res, err := aq.accountQuerier.Account(ctx, req)
+	if err != nil {
+		return nil, ErrQueryAccountNotFound.Wrapf("address: %s [%v]", address, err)
+	}
+	var acc accounttypes.AccountI
+	if err = queryCodec.UnpackAny(res.Account, &acc); err != nil {
```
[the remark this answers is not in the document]
h5law: in `codec.go`

testutil/testclient/testqueryclients/appquerier.go
h5law: This was previously there I refactored and forgot the comments

pkg/client/query/codec.go
h5law: I think the idea of using this codec here is that it will only be used here. We only need it when we decode an account from the chain. Thats what this interface provides and should only be used via the interface not the codec.

Previously we had this codec defined in multiple places but now this codec registering this stuff from the auth module allows it to be only used via the interface which is what should be used everyhwere.

WDYT?

pkg/appgateserver/cmd/cmd.go
```diff
@@ -62,7 +59,8 @@ provided that:
 
 	// Cosmos flags
 	cmd.Flags().String(cosmosflags.FlagKeyringBackend, "", "Select keyring's backend (os|file|kwallet|pass|test)")
-	cmd.Flags().String(cosmosflags.FlagNode, omittedDefaultFlagValue, "registering the default cosmos node flag; needed to initialize the cosmostx and query contexts correctly and uses flagQueryNodeUrl underneath")
+	cmd.Flags().
+		String(cosmosflags.FlagNode, omittedDefaultFlagValue, "registering the default cosmos node flag; needed to initialize the cosmostx and query contexts correctly and uses flagQueryNodeUrl underneath")
```
bryanchriswhite, line 63: I think the part about "uses flagQueryNodeUrl underneath" is ambiguous and therefore, potentially misleading. If we're going to mention this connection, let's provide more detail on "how" it uses the flag; e.g.: "sets the value of flagQueryNodeUrl internally for use by cosmos-sdk API".

pkg/relayer/cmd/cmd.go
```diff
@@ -53,7 +54,8 @@ for such operations.`,
 	}
 
 	cmd.Flags().String(cosmosflags.FlagKeyringBackend, "", "Select keyring's backend (os|file|kwallet|pass|test)")
-	cmd.Flags().String(cosmosflags.FlagNode, omittedDefaultFlagValue, "registering the default cosmos node flag; needed to initialize the cosmostx and query contexts correctly and uses flagQueryNodeUrl underneath")
+	cmd.Flags().
+		String(cosmosflags.FlagNode, omittedDefaultFlagValue, "registering the default cosmos node flag; needed to initialize the cosmostx and query contexts correctly and uses flagQueryNodeUrl underneath")
```
bryanchriswhite, line 58: (same as https://github.com/pokt-network/poktroll/pull/187#discussion_r1405228316)

pkg/relayer/miner/relay_fixtures_test.go
```diff
@@ -11,11 +11,11 @@ var (
 	// populating the signature with random bytes. It is intended for use
 	// in tests.
 	marshaledMinableRelaysHex = []string{
-		"0a140a12121019ab64049c78f6662406f248767173b0",
-		"0a140a12121028d1bc278d1b7fd6d7eec37d2a837a14",
-		"0a140a1212105aef8beb695d875371673c3d9388cd33",
-		"0a140a121210dde4ceef11a75a2f62ea59dae9558cbc",
-		"0a140a121210549b0164fe5179fa0529a7c070ad35bd",
+		"0a140a1212104a812fad82147600bcf3f7e29e148ec8",
```
bryanchriswhite, line 14: NIT: This can be reverted.

pkg/relayer/miner/relay_fixtures_test.go
bryanchriswhite, line 1: NIT: This can be reverted.

bryanchriswhite: For reference, what's going on here is that `make go_mockgen` currently runs `go generate ./pkg/...`, which now  includes the fixture generation script for the miner pkg test. The solution is to substitute that go generate line with the following ([as in #219](https://github.com/pokt-network/poktroll/pull/219/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52)):

```bash
go generate ./pkg/client/interface.go
go generate ./pkg/miner/interface.go
go generate ./pkg/relayer/interface.go
```

pkg/appgateserver/cmd/cmd.go
```diff
@@ -62,7 +59,8 @@ provided that:
 
 	// Cosmos flags
 	cmd.Flags().String(cosmosflags.FlagKeyringBackend, "", "Select keyring's backend (os|file|kwallet|pass|test)")
-	cmd.Flags().String(cosmosflags.FlagNode, omittedDefaultFlagValue, "registering the default cosmos node flag; needed to initialize the cosmostx and query contexts correctly and uses flagQueryNodeUrl underneath")
+	cmd.Flags().
+		String(cosmosflags.FlagNode, omittedDefaultFlagValue, "This flag is present to register the cosmos node flag, which is needed to initialise the comsos query context correctly. Ultimately the QueryNodeUrl field in the config file is used to build this context and this flag's value isn't used.")
```
h5law, line 63: @bryanchriswhite wdyt of this?

pkg/appgateserver/server.go
```diff
@@ -146,9 +138,9 @@ func NewAppGateServer(
 	}
 	app.signingInformation.SigningKey = signingKey
 
-	app.sessionQuerier = sessiontypes.NewQueryClient(app.clientCtx)
-	app.accountQuerier = accounttypes.NewQueryClient(app.clientCtx)
-	app.applicationQuerier = apptypes.NewQueryClient(app.clientCtx)
+	clientCtx := cosmosclient.Context(app.clientCtx)
+
+	app.sessionQuerier = sessiontypes.NewQueryClient(clientCtx)
```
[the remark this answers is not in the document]
h5law: I didnt make one for the session querier as it wasnt needed for the RingCache. However when it comes to testing the AppGateServer and your work on the RelayerProxy we should make one of the interfaces and dep inject it yeah

testutil/mockclient/application_query_client.go
Olshansk: This was resolved without a comment

Olshansk (changes requested):
Didn't have time to finish re-reviewing everything today but just leaving comments on what I had and we'll aim to merge this by EOD tomorrow.

pkg/client/interface.go
```diff
@@ -177,3 +181,17 @@ type TxClientOption func(TxClient)
...
+type AccountQueryClient interface {
+	// GetAccount queries the chain for the details of the account provided
+	GetAccount(ctx context.Context, address string) (accounttypes.AccountI, error)
+}
+
+// ApplicationQueryClient defines an interface that enables the querying of the
+// on-chain application information
+type ApplicationQueryClient interface {
+	// GetApplication queries the chain for the details of the application provided
+	GetApplication(ctx context.Context, appAddress string) (apptypes.Application, error)
```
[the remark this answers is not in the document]
h5law: No the actual method returns a non pointer this caused a huge debugging problem, the querier returns a non pointer so this should too

testutil/testcrypto/rings/godoc.go
```diff
@@ -0,0 +1,4 @@
+// Package testrings provides helper functions for constructing real (e.g. localnet)
+// and mock RingCache objects with pre-configured and/or parameterized call
+// arguments, return value(s), and/or expectations thereof. Intended for use in tests.
```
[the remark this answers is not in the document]
h5law: I think as this is an actual package for using tests this requires a normal godoc.go @Bryan might correct me but this is the standard IIRC

pkg/deps/config/config.go
```diff
@@ -70,3 +76,146 @@ func NewSupplyBlockClientFn(pocketNodeWebsocketUrl string) SupplierFn {
...
+		}
+
+		return deps, nil
+	}
+}
+
+// NewSupplyAccountQuerierFn returns a function with constructs an AccountQuerier
+// instance with the required dependencies and returns a new depinject.Config which
+// is supplied with the given deps and the new AccountQuerier.
+func NewAccountQuerierFn() SupplierFn {
```
red-0ne, line 165:
```suggestion
func NewSupplyAccountQuerierFn() SupplierFn {
```

pkg/deps/config/config.go
```diff
@@ -70,3 +76,146 @@ func NewSupplyBlockClientFn(pocketNodeWebsocketUrl string) SupplierFn {
...
+		// Supply the account querier to the provided deps
+		return depinject.Configs(deps, depinject.Supply(accountQuerier)), nil
+	}
+}
+
+// NewSupplyApplicationQuerierFn returns a function with constructs an
+// ApplicationQuerier instance with the required dependencies and returns a new
+// instance with the required dependencies and returns a new depinject.Config
+// which is supplied with the given deps and the new ApplicationQuerier.
+func NewApplicationQuerierFn() SupplierFn {
```
red-0ne, line 186:
```suggestion
func NewSupplyApplicationQuerierFn() SupplierFn {
```

red-0ne (dismissed):

pkg/appgateserver/cmd/cmd.go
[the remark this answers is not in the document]
h5law: This wasnt the case but have now added this functionality

pkg/client/interface.go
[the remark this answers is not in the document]
h5law: Ultimately we are abstracting the reply of the response even though the query returns a pointer the response is a copied value and if we remove the abstractiion we receive a value not a pointer.

From this the abstractiion should be doing the same and we can just return the same type as without the abstractiion

pkg/crypto/rings/cache_test.go
```diff
@@ -0,0 +1,173 @@
...
+				for i := 0; i < test.expectedSize-1; i++ {
+					gatewayAcc := newAccount("secp256k1")
+					accMap[gatewayAcc.address] = gatewayAcc.pubKey
+				}
+			}
+			// add the application's account and the accounts of all its
+			// delegated gateways to the testing state
+			testqueryclients.AddAddressToApplicationMap(t, test.appAccount.address, test.appAccount.pubKey, accMap)
+			// Attempt to retrieve the ring for the address and cache it
+			ring1, err := rc.GetRingForAddress(ctx, test.appAccount.address)
```
Olshansk, line 150: Can you add another assertion statement on the ring size BEFORE we cache it to make sure that it increases w/ the operation below.

pkg/crypto/rings/errors.go
```diff
@@ -0,0 +1,10 @@
+package rings
+
+import (
+	sdkerrors "cosmossdk.io/errors"
+)
+
+var (
+	codespace          = "rings"
+	ErrRingsWrongCurve = sdkerrors.Register(codespace, 1, "key is not a secp256k1 public key")
```
Olshansk, line 9: s/ErrRingsWrongCurve/ErrRingsNotSecp256

The message is very specific but the type could be misused.

pkg/deps/config/config.go
```diff
@@ -37,36 +52,181 @@ func SupplyConfig(
...
+		// to use the flag as expected.
+		if err := cmd.Flags().Set(cosmosflags.FlagNode, tmp); err != nil {
+			return nil, err
+		}
+
+		return deps, nil
+	}
+}
+
+// NewSupplyTxClientContextFn returns a function with constructs a ClientContext
```
Olshansk, line 132: thoughts on moving all the `NewSupply...` functions into `config/suppliers.go`?

Olshansk (approved):
Going to approve & trust the last set of comments are addressed and tested (E2E CI is green) before merging. 

Please make sure the following is done as well:

Screenshot 2023-11-29 at 3 23 05 PM

pkg/crypto/rings/cache_test.go
Olshansk: You have a `require.Equal(t, test.expectedSize, ring1.Size())` below, so there must be some sort of `require.GreaterThan(t, test.expectedSize, ring1.Size())`

h5law: We cant test this until cache invalidation is in this just checks the cache caches the ring correctly

Olshansk: `the cache caches the ring correctly` - Why not?

PR #189, opened 2023-11-15
[Relays] Remove Payload Types Replace with `[]byte`

h5law:
### Human Summary

- Remove all payload types
- Use payload `[]byte` in place
- Avoid unmarshalling payloads

### Summary generated by Reviewpad on 23 Nov 23 19:12 UTC
This pull request includes the following summaries:
1. This is a new file called "rest.go" in the "pkg/partials/payloads" package. It contains a struct called "PartialRESTPayload" with fields for headers. The file also includes methods for unmarshaling the payload, validating the required fields, getting the request type, generating an error payload, and getting the compute units for the RPC request. However, most of the methods are not yet implemented.
2. This file diff includes the following changes:
   - Added the import statement for the "io" package.
   - Added new code to read the request body and handle possible errors.
   - Added logging statements to print debug and error messages.
   - Updated the error handling logic to include the payload in error responses.
   - Added new code to handle synchronous relays and commented out a TODO for future improvements.
   - Removed the import statement for the "github.com/pokt-network/poktroll/x/service/types" package.
   - Updated the replyWithError() method to use the new types and error handling logic.
   - Removed the replyWithError() method and the comments associated with it.
   - Updated the serveHTTP() method to include the new code for handling relays and error responses.
   - Removed the validateConfig() method and the comments associated with it.
3. This file diff adds and modifies the RPC types in the openapi.yml file to include support for REST-RPC.
4. This diff renames the file `pkg/appgateserver/jsonrpc.go` to `pkg/appgateserver/synchronous.go`. The code changes involve renaming the function `handleJSONRPCRelay` to `handleSynchronousRelay`. The function now handles relay requests for synchronous protocols and includes additional logic related to request types and payload handling. There are also some changes related to error handling and logging.
5. Changes are made to the BuildProvidedServices function in the relayerProxy struct. These changes involve adding a comment about switching to the RPC type to create the appropriate RelayServer, adding a TODO comment for implementing different types of RPC, and replacing the creation of a JSONRPCServer with a SynchronousServer when the RPC type is JSON_RPC.
6. A new file named `jsonrpc.go` is added to the `pkg/partials/payloads` directory. This file contains a definition for `PartialJSONPayload` struct, used for partially unmarshalling JSON-RPC request payloads and extracting minimal fields for responses and compute units.
7. Changes have been made to the endpoint selection logic in the appgateserver package. The previous implementation used a naive greedy approach that defaulted to the first available supplier, while the new implementation retains this default behavior but removes a comment about potential future optimizations. Additionally, there is a comment change that clarifies the matching criteria for selecting an endpoint URL based on the request's RpcType.
8. Changes related to error handling in the `proxy` package of the codebase. The diff includes the addition of an import statement, the addition of new error variables, and updates to existing error codes.
9. Changes made to the `relay_builders.go` file, including updating method names and comments, and improving code efficiency.
10. Significant changes to the `relay.proto` file, including the repositioning of the `go_package` option, proper indentation of message fields, updated comments, removal of specific payload formats, and more.
11. Changes made to verify and handle the relay request signature in the specified file.
12. Refactoring of code to remove a block related to extracting the supplier's signature from the relay response payload, and removal of error handling for marshaling the payload.
13. Addition of a new file called `interface.go` in the `pkg/partials` package, defining an interface for partial RPC payloads and methods related to error responses, request types, compute units, and field validation.
14. Introduction of a new file in the `pkg/partials/payloads` package, containing partial unmarshalling of RPC payloads, error replies, and request compute unit calculation methods.
15. The `error_reply.go` file has been added, which includes a method called `replyWithError` in the `appgateserver` package. This method is used to reply with an error response and write it to the provided writer.
16. Introduction of a new file called `partial.go` in the `pkg/partials` package, which contains functions for handling partial requests, generating error replies, and extracting request types and compute units.
17. A new enum value "REST" has been added to the `RPCType` enum in the `service.proto` file, representing different types of remote procedure calls.
18. A renaming of the `jsonrpc.go` file to `synchronous.go`, along with changes in struct and function names, comments, and log messages to reflect the renaming.

Please let me know if you need more details on any specific summary.

## Issue

- Fixes N/A

[Explanation Document](https://github.com/pokt-network/poktroll/blob/DONT_DELETE_ME/pkg/appgateserver/JSONRPC.md)

## Type of change

Select one or more:

- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

## Sanity Checklist


red-0ne (approved):
This is how we'll scale in terms of supported protocols. I totally approve this move :+1: 
Thanks for spotting it

proto/pocket/service/relay.proto
```diff
@@ -24,82 +24,23 @@ message RelayRequestMetadata {
...
     RelayResponseMetadata meta = 1;
-    oneof payload {
-        JSONRPCResponsePayload json_rpc_payload = 2;
-        RESTResponsePayload rest_payload = 3;
-        // Future possible relay types:
-        // WebSocketsResponsePayload websockets_payload = 4;
-        // GRPCResponsePayload grpc_payload = 5;
-        // GraphQLResponsePayload graphql_payload = 6;
-    }
+    // payload is the serialised payload for the response.
```
Olshansk, line 36: Someone call the 🚒

pkg/relayer/proxy/error_reply.go
```diff
@@ -13,20 +14,22 @@ import (
 // TODO_TECHDEBT: This method should be aware of the nature of the error to use the appropriate JSONRPC
```
Olshansk, line 14: Should we remove this TECHDEBT?

pkg/relayer/proxy/error_reply.go
```diff
@@ -13,20 +14,22 @@ import (
...
-				Id:      0,
-				Jsonrpc: "2.0",
-				Error: &types.JSONRPCResponseError{
-					// Using conventional error code indicating internal server error.
-					Code:    -32000,
-					Message: err.Error(),
-					Data:    nil,
-				},
-			},
+	response := map[string]interface{}{
```
Olshansk, line 17: Interesting. 

Can you update the comment (and potentially doc) that even though we are making `payload` general and transparent, we are still in the process of determining if we should have protocol-aware errors. 

I want to think about this more, but it's not a bad design.

pkg/relayer/proxy/jsonrpc.go
```diff
@@ -138,14 +138,8 @@ func (jsrv *jsonRPCServer) serveHTTP(ctx context.Context, request *http.Request)
 	// Get the relayRequest payload's `io.ReadCloser` to add it to the http.Request
```
Olshansk, line 138: Should this be `NopCloser`?

pkg/appgateserver/server.go
```diff
@@ -222,20 +223,22 @@ func (app *appGateServer) ServeHTTP(writer http.ResponseWriter, request *http.Re
...
-				Id:      0,
-				Jsonrpc: "2.0",
-				Error: &types.JSONRPCResponseError{
-					// Using conventional error code indicating internal server error.
-					Code:    -32000,
-					Message: err.Error(),
-					Data:    nil,
-				},
-			},
+	response := map[string]interface{}{
```
Olshansk, line 226: It makes sense to have the `jsonServer` respond with a general purpose json rpc error, but it doesn't make sense in the context of the `appGateServer`.

This should be just a plain error string or something pocket specific.

Olshansk (changes requested):
A few minor comments but otherwise a big fan of this change!

proto/pocket/service/relay.proto
```diff
@@ -24,82 +24,23 @@ message RelayRequestMetadata {
...
+    // The payload is passed directly from the service and as such can be any
+    // format the the service responds with: JSON-RPC, REST, gRPC, etc.
```
red-0ne, line 38: @h5law , @Olshansk , @bryanchriswhite
I think this raises very pertinent questions regarding how we categorize our services.

Except the need to know the RPC type for:
- Accounting: Know the methods of the JSON-RPC calls and HTTP headers and paths for REST...
- Internal error replies: e.g. We reply from our internal errors with a 500 status code in the header for HTTP, stick an Id to a JSON-RPC reply for this one...

But workflow-wise we could categorize them into:
- Symmetric protocols: To each request corresponds a reply (1-1). These almost exclusively rely on TCP/HTTP transport (? At least we'll be sticking to it for a some time). This could include JSON-RPC, REST and gRPC
- Asymmetric protocols: Where we could have many replies for a single request (e.g. subscriptions) or many requests for a single reply (e.g. chunked file uploads) these are n-m req/res protocols. They may transport over TCP (messaging web apps...) or UDP/WebRTC (video streaming, gaming, etc...). This could include Websocket, gRPC and JSON-RPC

If we adhere to just passing the bytes of the original request, it doesn't matter which RPC type it is but knowing if we're gonna handle stateful or stateless communication matters a lot and will dictate which part of the code is going to handle that request.

Our current jsonrpc.go servers could support all the symmetric protocols with very low effort and both the appgateserver and the relayerproxy could be agnostic of the RPC type as long as its symmetric.

We could then implement the asymmetric ones (starting with Websockets then gRPC) but it would be the same logic/flow handling them IMO.

We could couple that to partial decoders given the RPC type is known (regardless of its flow) to extract the relevant info. for accounting and error handling.

WDYT?

pkg/relayer/proxy/error_reply.go
h5law: Do you suggest just replying directly with the payload response bytes we receive no matter what? That would make the errors specific to the service which is probably ideal for the end user

pkg/relayer/proxy/jsonrpc.go
h5law: `NopCloser` == turn my reader into a no op closer. It returns an `io.ReadCloser` with a no-op `Close` method

pkg/appgateserver/server.go
h5law: I think errors should be returned in the form of the request being sent, if ur code sending the requests is expecting a json reply lets give them a json reply or else the decoding will get weird on the app end

proto/pocket/service/relay.proto
h5law: Will read this through more later and come back with a full response @red-0ne however, we could look into adding some specifics to the config file:

As an application I want to send json-rpc requests to anvil over a websocket connection (this would be in the app's config.yaml file on staking)

The protocol can then read this and treat any incoming requests according to this on chain service config on behalf of the application - avoiding reading interim unmarshal steps. This would enforce correct config setups as we would start erroring on invalid configs but thats a user error.

Will come back with an actual response sooon ™️

Olshansk: @red-0ne Thanks for taking the initiative for asking the right questions. There's A LOT to what you asked and I think it will expand much further than just a github thread. 

Wdyt of starting a discussion [1] with what you shared above? We can use it as a starting point to discuss some of our goals, user requirements, features, etc, and then turn that into a proper ADR or design doc before moving forward with more concrete features?

I also believe the spike with the backend team will help shed some light on it.

[1] https://github.com/pokt-network/poktroll/discussions

h5law:
@Olshansk @red-0ne I have addressed the comments regarding error reply. This PR is going to have a followup PR that does the following:

1. Change the naming of the `jsonrpc` files and structures and references to `JSON-RPC` specific code to `SymmetricProtocol` related stuff (as it is now generalised)
2. Introduce documentation related to the `AppGateServer`, payloads and rings

And another follow-up PR will include:
1. REST payload interim unmarshaling
2. Unit tests for REST payload interim unmarshaling

pkg/partials/partial.go
```diff
@@ -0,0 +1,34 @@
...
+	if er != nil {
+		return nil, er
+	}
+	return partialRequest.GenerateErrorPayload(err)
+}
+
+// partiallyUnmarshalRequest unmarshals the payload into a partial request
+// that contains only the fields necessary to generate an error response and
+// handle accounting for the request's method.
+func partiallyUnmarshalRequest(payloadBz []byte) (PartialPayload, error) {
```
red-0ne, line 22: How are we going to know form `payloadBz` to which type we are going to unmarshal when we'll support more than one RpcType?

h5law: Here will just do a trial and error, try to unmarshal into JSON-RPC, if its empty or we get an error then try REST same situation.

proto/pocket/service/relay.proto
Olshansk: @red-0ne Bump

pkg/partials/interface.go
```diff
@@ -0,0 +1,11 @@
+package partials
+
+// PartialPayload is an interface that is implemented by each of the partial
+// payload types that allows for error messages to be created using the provided
+// error and request payload, that matches the correct format required by the
+// request type. As well as for accounting the weight of the request payload,
+// which is determined by the request's method field.
+type PartialPayload interface {
```
Olshansk, line 8: Not the biggest fan of `PartialPayload`. Why don't we just call it `RPCPayload`?

I think our goal is to the payload type transparent (i.e. bytes, proxyable, etc) to us, but it's not partial. Might need to update the comment too.

pkg/partials/interface.go
```diff
@@ -0,0 +1,11 @@
...
+	GenerateErrorPayload(err error) ([]byte, error)
```
Olshansk, line 9: Please add godoc comments next to the functions in the interface a well. It was unclear to me what the are.

pkg/partials/interface.go
```diff
@@ -0,0 +1,11 @@
...
+	GetMethodWeighting() (uint64, error)
```
Olshansk, line 10: I think I know what you have in mind when you say `Method` (i.e. JSON RPC), and you're thinking about tree weights, but need to take a step back and understand what we want here.

I think`GetRPCCost` or `GetRPCComputeUnits` is less implementation/protocol specific and gives us what we need.

pkg/partials/partial.go
```diff
@@ -0,0 +1,34 @@
...
+func GetErrorReply(payloadBz []byte, err error) ([]byte, error) {
+	log.Printf("DEBUG: Partially Unmarshalling request: %s", string(payloadBz))
+	partialRequest, er := partiallyUnmarshalRequest(payloadBz)
+	if er != nil {
+		return nil, er
+	}
+	return partialRequest.GenerateErrorPayload(err)
+}
+
+// partiallyUnmarshalRequest unmarshals the payload into a partial request
```
Olshansk, line 19:
```suggestion
// TODO_BLOCKER(@h5law): This function currently only supports JSON-RPC and must
// be extended to other request types.
// partiallyUnmarshalRequest unmarshals the payload into a partial request
```

pkg/partials/partial.go
```diff
@@ -0,0 +1,34 @@
...
+// that contains only the fields necessary to generate an error response and
+// handle accounting for the request's method.
+func partiallyUnmarshalRequest(payloadBz []byte) (PartialPayload, error) {
+	// First attempt to unmarshal the payload into a partial JSON-RPC request
+	var jsonReq partialJSONPayload
```
Olshansk, line 24: Can you move json marshal / unmarshal into a `partiallyUnmarshalJsonRPCRequest` helper so it'll set a pattern to add other types?

pkg/partials/partials_test.go
```diff
@@ -0,0 +1,72 @@
...
+	"errors"
+	"testing"
+
+	sdkerror "cosmossdk.io/errors"
+	"github.com/stretchr/testify/require"
+)
+
+// TODO(@h5law): Expand coverage with more test cases when more request types
+// are implemented in the partials package.
+func TestPartials_GetErrorReply(t *testing.T) {
```
Olshansk, line 14: Make sure to rename `Partials` depending on how we call it per the discussion above

pkg/partials/partials_test.go
```diff
@@ -0,0 +1,72 @@
...
+	tests := []struct {
+		name        string
+		err         error
+		payload     map[string]any
```
Olshansk, line 18: Payload should be `[]byte` if we're really making it json-rpc agnostic.

proto/pocket/service/relay.proto
```diff
@@ -1,46 +1,46 @@
...
 }
 
 // RelayRequestMetadata contains the metadata for a RelayRequest.
 message RelayRequestMetadata {
-    session.SessionHeader session_header = 1; // Session header associated with the relay.
-    // TODO_COMMENT(@h5law): Add link or more details to how this is related to ring signatures once implemented.
-    bytes signature = 2; // The request signature. This may be the application signature, or any gateway it delegated to.
+  session.SessionHeader session_header = 1; // Session header associated with the relay.
+  // TODO_COMMENT(@h5law): Add link or more details to how this is related to ring signatures once implemented.
+  bytes signature = 2; // The request signature. This may be the application signature, or any gateway it delegated to.
```
Olshansk, line 21:
```suggestion
  bytes signature = 2; // The request signature. This may be the application signature, or any gateway it is delegated to.
```

pkg/partials/structs.go
```diff
@@ -0,0 +1,60 @@
+package partials
+
+import (
+	"encoding/json"
+)
+
+var (
+	_ PartialPayload = (*partialJSONPayload)(nil)
```
Olshansk, line 8: I like where this design is headed, but instead of a single `partials/structs.go` file, I think we should have `pkg/payload/json.go` and `pkg/payload/rest.go`

pkg/partials/structs.go
```diff
@@ -0,0 +1,60 @@
...
+	_ PartialPayload = (*partialRESTPayload)(nil)
+)
+
+// partialJSONPayload is a partial representation of a JSON-RPC request payload
+// that contains only the fields necessary to generate an error response and
```
Olshansk, line 13:
```suggestion
// that contains the minimal fields necessary to generate an error response and
```

pkg/relayer/proxy/errors.go
```diff
@@ -1,15 +1,54 @@
...
-	codespace                                        = "relayer_proxy"
-	ErrRelayerProxyUnsupportedRPCType                = sdkerrors.Register(codespace, 1, "unsupported relayer proxy rpc type")
-	ErrRelayerProxyInvalidRelayRequestSignature      = sdkerrors.Register(codespace, 2, "invalid relay request signature")
-	ErrRelayerProxyInvalidSession                    = sdkerrors.Register(codespace, 3, "invalid session in relayer request")
-	ErrRelayerProxyInvalidSupplier                   = sdkerrors.Register(codespace, 4, "invalid relayer proxy supplier")
-	ErrRelayerProxyUndefinedSigningKeyName           = sdkerrors.Register(codespace, 5, "undefined relayer proxy signing key name")
-	ErrRelayerProxyUndefinedProxiedServicesEndpoints = sdkerrors.Register(codespace, 6, "undefined proxied services endpoints for relayer proxy")
-	ErrRelayerProxyInvalidRelayRequest               = sdkerrors.Register(codespace, 7, "invalid relay request")
-	ErrRelayerProxyInvalidRelayResponse              = sdkerrors.Register(codespace, 8, "invalid relay response")
+	codespace                         = "relayer_proxy"
```
Olshansk, line 8: 😭

Olshansk (changes requested):

proto/pocket/service/relay.proto
red-0ne: Created a discussion for it https://github.com/pokt-network/poktroll/discussions/207

Olshansk: @red-0ne As discussed offline, few requests for the discussion:

1. Please update formatting
2. Please reference the PRs harry has put together (as a starting point)

We want to make it easy for others to understand and join in on the discussion.

pkg/partials/interface.go
h5law: I think it is partial, it contains only some of the parts of the request - its parts-ial lmao. But I can change it if you think thatll make it clearer

pkg/relayer/proxy/errors.go
h5law: ?

Olshansk: I prefer the spacing before

h5law: I modified my golines formatter to have max length of 150 this is reverted now I didnt notice the change

pkg/partials/partial.go
```diff
@@ -0,0 +1,66 @@
...
+func GetErrorReply(payloadBz []byte, err error) ([]byte, error) {
+	partialRequest, er := partiallyUnmarshalRequest(payloadBz)
+	if er != nil {
+		return nil, er
+	}
+	return partialRequest.GenerateErrorPayload(err)
+}
+
+// GetComputeUnits returns the compute units for the RPC request provided
+func GetComputeUnits(payloadBz []byte) (uint64, error) {
```
Olshansk, line 36: I'm pretty 💉💪 for this

pkg/partials/payloads/godoc.go
```diff
@@ -0,0 +1,6 @@
+// payloads contains the different types of RPC payloads the partials package
+// supports. The structs defined here are used to partially unmarshal the
+// payload and extract the minimal fields required to: generate error responses,
+// get the RPC request type and determine compute units. This is done through
+// the partial unmarshaling of the payload into the types defined here.
```
Olshansk, line 5:
```suggestion
// payloads contains the different types of RPC payloads the partials package
// supports. The structs defined here are used to partially unmarshal the
// payload and extract the minimal fields required to: generate error responses,
// retrieve RPC request type, determine request compute units, etc...
// This is done through partially unmarshalling the payload into the minimum
// required set of pre-defined fields that need to be explicitly determined for
// each RPC type supported.
```

pkg/relayer/proxy/error_reply.go
```diff
@@ -4,32 +4,23 @@ import (
...
-				Jsonrpc: "2.0",
-				Error: &types.JSONRPCResponseError{
-					// Using conventional error code indicating internal server error.
-					Code:    -32000,
-					Message: err.Error(),
-					Data:    nil,
-				},
-			},
-		},
+// replyWithError builds the appropirate error format according to the payload
```
Olshansk, line 11:
```suggestion
// replyWithError builds the appropriate error format according to the payload
```

pkg/partials/partials_test.go
```diff
@@ -0,0 +1,64 @@
...
+				unmarshalledReply := map[string]any{}
+				require.NoError(t, json.Unmarshal(replyBz, &unmarshalledReply))
+				require.Equal(t, unmarshalledReply["jsonrpc"], testPayload["jsonrpc"])
+				require.Equal(t, unmarshalledReply["id"], testPayload["id"])
+				require.Equal(
+					t,
+					unmarshalledReply["error"].(map[string]any)["message"],
+					test.err.Error(),
+				)
+			}
```
Olshansk, line 61: Can you add a default that panic?

pkg/partials/payloads/jsonrpc.go
```diff
@@ -0,0 +1,79 @@
...
+	"encoding/json"
+	"errors"
+	"log"
+
+	"github.com/pokt-network/poktroll/x/shared/types"
+)
+
+// PartialJSONPayload is a partial representation of a JSON-RPC request payload
+// that contains the minimal fields necessary to generate an error response and
+// handle accounting for the request's method field.
```
Olshansk, line 13:
```suggestion
// PartialJSONPayload is a partial representation of a JSON-RPC request payload
// that contains the minimal fields necessary for basic business logic such as
// generating an  error response, handling request account, etc...
```

pkg/partials/payloads/rest.go
```diff
@@ -0,0 +1,49 @@
+package payloads
+
+import (
+	"github.com/pokt-network/poktroll/x/shared/types"
+)
+
+// PartialRESTPayload is a partial representation of a REST request payload
+// that contains the minimal fields necessary to generate an error response and
+// handle accounting for the request's method as determined by its headers.
```
Olshansk, line 9:
```suggestion
// PartialRESTPayload is a partial representation of a REST request payload
// that contains the minimal fields necessary for basic business logic such as
// generating an  error response, handling request account, etc...
```

pkg/partials/payloads/rest.go
```diff
@@ -0,0 +1,49 @@
...
+	// TODO(@h5law): Implement this function
+	return nil, false
+}
+
+// ValidateBasic ensures that all the required fields are set in the partial
+// REST payload.
+// It uses a non-pointer receiver to ensure the default values of unset fields
+// are present
+func (r PartialRESTPayload) ValidateBasic() error {
+	// @h5law: Implement this function
```
Olshansk, line 28:
```suggestion
	// TODO(@h5law): Implement this function
```

Olshansk (approved):
@h5law Please don't miss my last set of suggestions - but otherwise lgtm!

PR #195, opened 2023-11-16
[Draft] Update `Name` app.go

Olshansk:
Creating a draft for this as a reminder to investigate what it should be

Screenshot 2023-11-16 at 11 41 34 AM

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -136,7 +136,7 @@ import (
 
 const (
 	AccountAddressPrefix = "pokt"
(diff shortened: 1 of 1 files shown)


Olshansk:
@okdas Do you see any reason why we shouldn't do this?

okdas:
> @okdas Do you see any reason why we shouldn't do this?

Not really, we can always change back if something breaks

okdas (approved):

Issue #159, opened 2023-11-07
[☂️ Documentation] Umbrella ticket to update all-the-docs


Olshansk:
## Objective

Improve the poktroll docs to enable external contributors and 3rd party gateways.

## Origin Document

The documentation in [pokt-network/poktroll](https://github.com/pokt-network/poktroll) fell behind as development kicked off. The bar in [pokt-network/pocket](pokt-network/pocket), and we want to get at least halfway there.

## Goals

- Create a better onboarding experience for anyone who comes by the [pokt-network/poktroll](https://github.com/pokt-network/poktroll) repo.
- Achieve "close to parity" with the prior "v1" repo
- Enable 3rd party gateways to easily understand & integrate with Shannon TestNet
- Share the development, work & research we have done with other projects in the ecosystem

## Deliverables

- [ ] [Everyone] Add missing architectural diagrams for this repo, where appropriate (new, migrated, etc...)
- [ ] [Everyone] Move over diagrams from [alpha repo](https://github.com/pokt-network/poktroll-alpha)
- [ ] [@bryanchriswhite] Move over our best practices (code reviews, coding guidelines, etc..) from the [v1](https://github.com/pokt-network/pocket)
- [ ] [@okdas] Make sure documentation related to DevNet, Telemetry & Observability is easily accessible
- [ ] [@h5law] Add documentation related to the usage & integration of the SMT
- [ ] [@red-0ne ] Provide documentation on the design and use of `AppGate`
- [ ] [@RossiNYC] Identify which tool (e.g. [GitBook](https://www.gitbook.com/), [Docusaurus](https://docusaurus.io/)) we want to use to mirror our documentation in this repo.
- [ ] [@Olshansk] Update the main entrypoint README

---

**Creator**: @Olshansk 
**Co-Owners**: @red-0ne @bryanchriswhite @okdas @h5law @RossiNYC

Olshansk:
Sharing notes from a team meeting today o how we'll distribute some of the work on documentation:
- [ ] @bryanchriswhite @Olshansk: choose a platform for docs to make them more discoverable (gitbook, docusaurus, etc...) and decide how it should be organized
- [ ] @Olshansk: Update the main README, developer resources, onboarding, education, etc...
- [ ] @Olshansk: High-level architecture (i.e. component diagram) showing the interaction between pocket actors, the DA layer, sequencers, etc...
- [ ] @Olshansk: E2E Claim & Proof lifecycle (docs & sequence diagram)
[...]

Issue #137, opened 2023-11-03
[Supplier] Provide config/spec file when staking Supplier


Olshansk:
## Objective

Productionize the interface via which supplier will be staking for numerous complex services

## Origin Document

- Should be somewhat similar to how we enabled specifying requests via a flag in the [v1 repo](https://github.com/pokt-network/pocket)
Related to #138
- The following snippet from [https://github.com/pokt-network/poktroll/blob/main/x/supplier/client/cli/tx_stake_supplier.go](x/supplier/client/cli/tx_stake_supplier.go) captures it:

```go
func CmdStakeSupplier() *cobra.Command {
	// fromAddress & signature is retrieved via `flags.FlagFrom` in the `clientCtx`
	cmd := &cobra.Command{
		// TODO_HACK: For now we are only specifying the service IDs as a list of of strings separated by commas.
		// This needs to be expand to specify the full SupplierServiceConfig. Furthermore, providing a flag to
		// a file where SupplierServiceConfig specifying full service configurations in the CLI by providing a flag that accepts a JSON string
		Use:   "stake-supplier [amount] [svcId1;url1,svcId2;url2,...,svcIdN;urlN]",
		Short: "Stake a supplier",
		Long: `Stake an supplier with the provided parameters. This is a broadcast operation that
will stake the tokens and associate them with the supplier specified by the 'from' address.
TODO_HACK: Until proper service configuration files are supported, suppliers must specify the services as a single string
of comma-separated values of the form 'service;url' where 'service' is the service ID and 'url' is the service URL.
For example, an application that stakes for 'anvil' could be matched with a supplier staking for 'anvil;http://anvil:8547'.
```

## Goals

- Set the foundation for how complex specification & configuration files for supplier staking will be provided
- Enable node runners to stake for repeatedly staking for complex services

## Deliverables

- [ ] A PR that modifies how `supplier stake` functions by providing a path to a file on disk
- [ ] An ability to provide the complete complex configuration via an inline json file
- [ ] Cleanup of all downstream code (e.g. message preparation) when staking for the service
- [ ] Unit tests & validation for all existing fields & parametesr

## Non-goals / Non-deliverables

- Adding/removing configurations to the `SupplierServiceConfig`

## General deliverables

- [ ] **Comments**: Add/update TODOs and comments alongside the source code so it is easier to follow.
- [ ] **Testing**: Add new tests (unit and/or E2E) to the test suite.
[...]

PR #198, opened 2023-11-17
[CI/E2E] Allow to change endpoint for e2e tests

okdas:
### Human Summary

TBD, I needed to open a PR so I could provision DevNet for it.

### AI Summary

reviewpad:summary

## Issue

- #{ISSUE_NUMBER}

[Explain the reasoning for the PR in 1-2 sentences. Consider adding a link or a screenshot.]

## Type of change

Select one or more:

- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

## Sanity Checklist

### DIFF
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -5,8 +5,8 @@ FROM golang:1.20 as base
 
 RUN apt update && \
     apt-get install -y \
-        ca-certificates \
-        curl jq make vim less
+        ca-certificates net-tools \
+        curl jq make vim less dnsutils
 
 # enable faster module downloading.
 ENV GOPROXY https://proxy.golang.org
--- a/e2e/tests/init_test.go
+++ b/e2e/tests/init_test.go
@@ -6,6 +6,7 @@ import (
 	"flag"
 	"fmt"
 	"log"
+	"os"
 	"regexp"
 	"strconv"
 	"strings"
@@ -24,12 +25,6 @@ import (
 	suppliertypes "github.com/pokt-network/poktroll/x/supplier/types"
 )
 
-// TODO_TECHDEBT(@okdas): Once relayminer and appgateserver are running in tilt,
-// use their respective in-tilt hostnames and run E2E tests in tilt. This
-// should match the on-chain advertised endpoint for the service with the
-// given serviceId.
-const localnetAppGateServerUrl = "http://localhost:42069"
-
 var (
 	addrRe   *regexp.Regexp
 	amountRe *regexp.Regexp
@@ -41,6 +36,7 @@ var (
 
 	flagFeaturesPath string
 	keyRingFlag      = "--keyring-backend=test"
+	appGateServerUrl = "http://localhost:42069"
 )
 
 func init() {
@@ -48,6 +44,11 @@ func init() {
 	amountRe = regexp.MustCompile(`amount:\s+"(.+?)"\s+denom:\s+upokt`)
 
 	flag.StringVar(&flagFeaturesPath, "features-path", "*.feature", "Specifies glob paths for the runner to look up .feature files")
+
+	// If "APPGATE_SERVER_URL" envar is present, use it for appGateServerUrl
+	if url := os.Getenv("APPGATE_SERVER_URL"); url != "" {
+		appGateServerUrl = url
+	}
 }
 
 func TestMain(m *testing.M) {
@@ -260,7 +261,7 @@ func (s *suite) TheSessionForApplicationAndServiceContainsTheSupplier(appName st
 }
 
 func (s *suite) TheApplicationSendsTheSupplierARequestForServiceWithData(appName, supplierName, serviceId, requestData string) {
-	res, err := s.pocketd.RunCurl(localnetAppGateServerUrl, serviceId, requestData)
+	res, err := s.pocketd.RunCurl(appGateServerUrl, serviceId, requestData)
 	if err != nil {
 		s.Fatalf("error sending relay request from app %s to supplier %s for service %s: %v", appName, supplierName, serviceId, err)
 	}


Makefile
```diff
@@ -366,20 +366,14 @@ supplier_stake: ## Stake tokens for the supplier specified (must specify the SUP
 # TODO_IMPROVE(#180): Make sure genesis-staked actors are available via AccountKeeper
 .PHONY: supplier1_stake
 supplier1_stake: ## Stake supplier1 (also staked in genesis)
-	# TODO_UPNEXT(@okdas): once `relayminer` service is added to tilt, this hostname should point to that service.
```
okdas, line 369: I checked configs @red-0ne introduced and didn't find a need to change anything to pass e2e tests.

docs/static/openapi.yml
```diff
@@ -46480,7 +46480,7 @@ paths:
                           service:
                             title: >-
                               The Service for which the application is
-                              configured for
+                              configured
```
okdas, line 46483: This file always changes for me - am I the only one? I usually just `stash` it.

Olshansk: tl;dr Don't worry too much about it. 

It gets regenerated with `ignite generate openapi --yes` and we don't have anything in place to keep good track of it atm.

Makefile
Olshansk: There's so much in flight that TODOs are going to get outdated quick. If you don't think it's relevant, remove it and we'll double check in code review!

pkg/relayer/proxy/server_builder.go
```diff
@@ -42,17 +41,16 @@ func (rp *relayerProxy) BuildProvidedServices(ctx context.Context) error {
 		var serviceEndpoints []relayer.RelayServer
 
 		for _, endpoint := range serviceConfig.Endpoints {
-			url, err := url.Parse(endpoint.Url)
-			if err != nil {
-				return err
-			}
-			supplierEndpointHost := url.Host
+			// TODO: Move `supplierEndpointHost` into a separate config file (could be a part of chains.json if we
```
okdas, line 44: Should this TODO mention #137? I'm confused because #137 seems to be addressing stake, but I am not sure about the configuration of the service itself.

Olshansk: > Should this TODO mention #137?

Yes. 

> I'm confused because #137 seems to be addressing stake, but I am not sure about the configuration of the service itself.

When you run stake you say: `I, supplier #1, am staking for 100 POKT for service X with endpoint XX and service Y with endpoint YY, etc...

Here's an example:

Screenshot 2023-11-22 at 12 15 03 PM

okdas: After reviewing the situation, I've realized there are several issues with our current configuration setup that I initially overlooked, leading to some confusion. Here's a summary of my observations and suggestions:

The configuration file we're using is named 'stake' config, which seems misleading. In the mainnet environment, 'stake' typically refers to components like the node's discoverable endpoint (the HTTPS endpoint for /v1/relay) and the supported services (chains). However, this 'stake' config we have is more of an internal setup specific to supplier1 and doesn't necessarily reflect the services staked on the network (e.g. current chains.json). Therefore, I believe we should consider renaming it to avoid confusion. Perhaps, there could be a separate file that can specify the services supported by the node, and the URL the node can be reached at?

I encountered a problem where the endpoint defined in this config file was being used to bind or open a port. For instance, if the config file specifies an 'anvil' hostname, the supplier attempts to open a port on this hostname, which is impractical unless there's a network interface corresponding to the 'anvil' DNS. My current workaround, setting a hardcoded '0.0.0.0:8545', is a temporary fix and won't support multiple endpoints. We should consider relocating this endpoint to either another config file or a command-line argument to make it more dynamic. I can explore this further in a separate pull request to address the immediate issue.

I admit that the overall configuration, port management, and service interactions are a bit confusing to me. To tackle this, I've started compiling information on a [Notion page](https://www.notion.so/buildwithgrove/Actors-configs-dependencies-etc-ab9a93322b8744a2aaa089d5b4758b75?pvs=4) which outlines our system's actors, configurations, and dependencies. This effort should not only clarify things for me but also serve as a valuable resource for #159 .

Olshansk (approved):
Seems like we have a bit of followup work as we integrate with the other configs (cc @red-0ne), but other than that this looks great!

Thanks for working through all the kinks @okdas 🙌 

:shipit:

pkg/relayer/proxy/server_builder.go
Olshansk: @okdas I believe this relates to the [comment](
https://github.com/pokt-network/poktroll/issues/205#issuecomment-1822094055) I left on @red-0ne's PR in #205:

Staking configs:
- `App` config -> for `staking` applications
- `Gateway` config -> for `staking` gateways
- `Supplier` config -> for `staking` suppliers

Operations configs:
- `AppGate` config -> for `operating` a gateway or an application
- `RelayMiner` config -> for `operating` a supplier

Olshansk: > After reviewing the situation, I've realized there are several issues with our current configuration setup that I initially overlooked, leading to some confusion. Here's a summary of my observations and suggestions:

1. Is this a blocker for this PR or subsequent ones?

> I admit that the overall configuration, port management, and service interactions are a bit confusing to me. To tackle this, I've started compiling information on a [Notion page](https://www.notion.so/buildwithgrove/Actors-configs-dependencies-etc-ab9a93322b8744a2aaa089d5b4758b75?pvs=4) which outlines our system's actors, configurations, and dependencies. This effort should not only clarify things for me but also serve as a valuable resource for https://github.com/pokt-network/poktroll/issues/159 .

2. I agree and have discussed it with @red-0ne. Documenting them is a top priority.

> The configuration file we're using is named 'stake' config, which seems misleading. In the mainnet environment, 'stake' typically refers to components like the node's discoverable endpoint (the HTTPS endpoint for /v1/relay) and the supported services (chains). However, this 'stake' config we have is more of an internal setup specific to supplier1 and doesn't necessarily reflect the services staked on the network (e.g. current chains.json). Therefore, I believe we should consider renaming it to avoid confusion. Perhaps, there could be a separate file that can specify the services supported by the node, and the URL the node can be reached at?

3. 

We `STAKE` the `App`, `Gateway` and `Supplier.
We `DEPLOY` the `AppGate` and `RelayMiner`

As of writing this comment, the state on main is (I manually added space and comments to the output):

```bash
$ find . -name '*.yaml' | grep "config" | sort

./localnet/poktrolld/config/appgate_server_config.yaml # Deploy Appgate (internal, off-chain)

./localnet/poktrolld/config/application1_stake_config.yaml # Stake App (external, on-chain)
./localnet/poktrolld/config/application2_stake_config.yaml # Stake App (external, on-chain)
./localnet/poktrolld/config/application3_stake_config.yaml # Stake App (external, on-chain)

./localnet/poktrolld/config/relayminer_config.yaml  deploy RelayMiner (internal, off-chain)

./localnet/poktrolld/config/supplier1_stake_config.yaml # stake supplier (external, on-chain)
./localnet/poktrolld/config/supplier2_stake_config.yaml # stake supplier (external, on-chain)
./localnet/poktrolld/config/supplier3_stake_config.yaml # stake supplier (external, on-chain)

./localnet_config.yaml
```

As you can see the only thing we're missing is `Stake Gateway`

> I encountered a problem where the endpoint defined in this config file was being used to bind or open a port. For instance, if the config file specifies an 'anvil' hostname, the supplier attempts to open a port on this hostname, which is impractical unless there's a network interface corresponding to the 'anvil' DNS. My current workaround, setting a hardcoded '0.0.0.0:8545', is a temporary fix and won't support multiple endpoints. We should consider relocating this endpoint to either another config file or a command-line argument to make it more dynamic. I can explore this further in a separate pull request to address the immediate issue.

4. 

If you look at `localnet/poktrolld/config/supplier1_stake_config.yaml` for example. The expectation here is that when you (as an app) get paired with `supplier1` during a session, you'd be using `http://anvil:8547` to request service  for `anvil`. In practice, this would be a domain like `anvil.supplier1.com` behind a load balancer. 

```yaml
- service_id: anvil
  endpoints:
  - url: http://anvil:8547
    rpc_type: json_rpc
- service_id: svc1
  endpoints:
  - url: http://lovalhost:8081
    rpc_type: json_rpc 
```

@okdas I'd defer to you for the best way to handle this right now, but I'm assuming we should (in theory) something like the following:

Replace `http://anvil:8547` with `anvil.supplier1.poktroll.com` and update DNS configs for `poktroll.com` to be different in localnet/devnet.

_Side note: Looks like we have a typo of `lovalhost` to `localhost`._

red-0ne: @okdas , `supplier1_stake_config.yaml` should be used to stake and advertise endpoints on-chain.

The fact that we have `anvil` is definitely misleading. Since what the supplier should be advertising is the service and not a specific implementation of it.

I propose having `supplier1_stake_config.yaml` with the following:

```yaml
- service_id: ethereum
  endpoints:
  - url: http://ethereum.supplier1.poktroll.com:<whatever_port>
    rpc_type: json_rpc
```

This file will be used to stake and have the previous config available on-chain. So, for bootstrapping the chain, I would recommend issuing a `poktrolld tx supplier-stake 1000upokt --config supplier1_stake_config.yaml` instead of having hardcoded values in genesis configuration `config.yaml`.

So, when starting the relay miner, the `RelayerProxy component` would query its on-chain supplier entry and make the relevant services available.

Just like it's done here:
```go
for _, endpoint := range serviceConfig.Endpoints {
	url, err := url.Parse(endpoint.Url)
	if err != nil {
		return err
	}
	supplierEndpointHost := url.Host

	var server relayer.RelayServer

	switch endpoint.RpcType {
	case sharedtypes.RPCType_JSON_RPC:
		server = NewSynchronousServer(
			service,
			supplierEndpointHost,
			proxiedServicesEndpoints,
			rp.servedRelaysPublishCh,
			rp,
		)
	default:
		return ErrRelayerProxyUnsupportedRPCType
	}
	serviceEndpoints = append(serviceEndpoints, server)
}
```

Which should be reverted in the current commit BTW https://github.com/pokt-network/poktroll/pull/198/files#diff-360f9ba4c864cfdbf1bd894500d507361d5c04c9d78d064d2062140dcdbaf62bR47

pkg/relayer/proxy/server_builder.go
```diff
@@ -42,17 +41,16 @@ func (rp *relayerProxy) BuildProvidedServices(ctx context.Context) error {
...
 		for _, endpoint := range serviceConfig.Endpoints {
-			url, err := url.Parse(endpoint.Url)
-			if err != nil {
-				return err
-			}
-			supplierEndpointHost := url.Host
+			// TODO(#137): Move `supplierEndpointHost` into a separate config file (could be a part of chains.json if we
+			// should bind different ports for different chains, or just one port as a part of the main config/CLI argument)
+			// `endpoint.Url` should only be used to advertise the endpoint to the internet as a part of stake.
+			supplierEndpointHost := "0.0.0.0:8545"
```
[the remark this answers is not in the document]
okdas: @red-0ne We should not do that. I understand this is counterintuitive, but nodes are almost never going to be exposed to the internet directly. Most often, they are going to be used with a reverse proxy - load balancer, nginx, cloudflare, etc. - so there are many options. We should provide a way to change the network interface we bind a port to. Because if we do not, here's what happens:

```
2023/11/27 02:03:35 INFO: starting relay server for service anvil at endpoint http://relayminers:8545
Error: listen tcp 10.96.6.176:8545: bind: cannot assign requested address
2023/11/27 02:03:35 WARN: redundant unsubscribe: observer is closed
```

The node makes a DNS request on the address from the on-chain data, resolves it to an IP address, and tries to bind to that address. Whatever the hostname resolves to should not be used to determine the network interface to bind.

In meanwhile, I add this code back to be able to run e2e tests. We should put it in a config file! I can look into how to do that later.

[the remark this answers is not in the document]
red-0ne: @okdas Makes sense thanks for the explanation. As you stated, a config entry would make sense then. Do you think we should add this in the relay miner config file? Should it be a 1-1 configuration entry w.r.t the hosts exposed on-chain? or just one config entry that handles all of them?

Keep in mind that we probably need one server for each service.

Issue #181, opened 2023-11-14
[Logging] Adopt a proper logging library in our off-chain `pkg`


Olshansk:
## Objective

Standardize the logging tooling in poktroll's `off-chain` (i.e. `/pkg`) source code.

## Origin Document

In the prior [v1 repo](https://github.com/pokt-network/pocket), we built on top of a good login package that is found and documented at [pokt-network/pocket/tree/main/logger](https://github.com/pokt-network/pocket/tree/main/logger) after an extensive discussion in [pokt-network/discussions/214](https://github.com/orgs/pokt-network/discussions/214).

Poktroll's on-chain logging uses standard cosmos logging (see discussion in [cosmos-sdk/issues/15341](https://github.com/cosmos/cosmos-sdk/issues/15341)), but #177 using go's standard logging libraries: 

![Screenshot 2023-11-13 at 4 35 15 PM](https://github.com/pokt-network/poktroll/assets/1892194/98dd1b86-7b16-497d-9296-9cd41e5b297b)

Though sufficient, it is limited in having different sources/syncs, logging levels, etc... Given that we have already. 

## Goals

- Set a standard for how logging should be done under `/pkg` - our off-chain codebase
- Improve logging & telemetry in our off-chain services
- Use consistent logging libraries

## Deliverables

- [x] A decision on which logging library to use (why and how)
- [x] A PR that updates all the logs under `/pkg` to use the newly selected library

## Non-goals / Non-deliverables

- Building a new logging library
- Adding new logs
- Updates to dashboards or other to easily inspect the logs

## General deliverables

- [x] **Comments**: Add/update TODOs and comments alongside the source code so it is easier to follow.
- [x] **Testing**: Add new tests (unit and/or E2E) to the test suite.
- [ ] **Makefile**: Add new targets to the Makefile to make the new functionality easier to use.
- [ ] **Documentation**: Update architectural or development READMEs; use [mermaid](https://mermaid-js.github.io/mermaid/) diagrams where appropriate.

---

**Creator**: @Olshansk 
**Co-Owners**: @bryanchriswhite @h5law  @okdas

PR #209, opened 2023-11-22
[Off-chain, Polylog] test: add testutils

bryanchriswhite:
### Human Summary

This PR adds a test helper to reduce the verbosity of executing test cases across all levels and method permutations.

### Summary generated by Reviewpad on 05 Dec 23 09:31 UTC
This pull request includes changes to the `pkg/polylog/interface.go` and `testutil/testpolylog/event.go` files. 

In the `pkg/polylog/interface.go` file, a new interface called `Level` has been added. This interface defines two methods, `String()` and `Int()`, which return the string and integer representations of a log level, respectively. Additionally, the `WithLevel()` method now takes an input of type `Level` instead of an integer.

In the `Logger` interface, the `WithLevel()` method signature has been updated to accept an input of type `Level` instead of an integer.

In the `testutil/testpolylog/event.go` file, a new package called `testpolylog` has been added. This package includes various test helper functions for testing the `polylog` package. 

Overall, these changes aim to improve the logging capabilities and testing of the `polylog` package.

## Issue

Relates to:
- #181
- #210
- #211
- #212
- #213

Test code would contain lots of duplication around this logic were it not factored out.

## Type of change

Select one or more:

- [ ] Bug fix
- [x] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

## Sanity Checklist


pkg/polylog/interface.go
```diff
@@ -11,6 +11,11 @@ import (
 // TODO_INVESTIGATE: check whether the pkg dependency tree includes all logging
 // libraries.
 
+type Level interface {
```
bryanchriswhite, line 14: godoc comment

testutil/testpolylog/event.go
```diff
@@ -0,0 +1,164 @@
...
+		// Construct a spy which implements a #Fn() method which we can use to
+		// assert that the function passed to polylog.Event#Func() is called with
+		// the expected arg(s).
+		funcSpy := FnMethodSpy{}
+		funcSpy.On("Fn", mock.AnythingOfType(funcMethodEventTypeName)).Return()
+
+		logger.Debug().Func(funcSpy.Fn).Send()
+
+		// Assert that the Func field contains the expected value.
+		// asserting that `Fn` is not called!
```
bryanchriswhite, line 161:
```suggestion
		// Asserting that `Fn` was called with an event whose type name is the funcMethodEventTypeName string.
```

testutil/testpolylog/event.go
```diff
@@ -0,0 +1,166 @@
...
+// Fn is a mock method which can be asserted on via the mock.Mock API.
+// See: https://pkg.go.dev/github.com/stretchr/testify@v1.8.4/mock#Mock.
+func (m *FnMethodSpy) Fn(event polylog.Event) {
+	m.Called(event)
+}
+
+// EventMethodsTest is a test case for expressing and exercising polylog.Event
+// methods in a concise way.
+type EventMethodsTest struct {
+	Msg                    string
```
Olshansk, line 51: Confirming these should be exposed?

testutil/testpolylog/event.go
```diff
@@ -0,0 +1,166 @@
...
+	MsgFmt                 string
+	MsgFmtArgs             []any
+	Key                    string
+	Value                  any
+	EventMethodName        string
+	ExpectedOutputContains string
```
Olshansk, line 57: Can you add short comments on the params? It's not evident without diving into the code so I think it would help if it becomes a primitive.

testutil/testpolylog/event.go
```diff
@@ -0,0 +1,166 @@
...
+
+// Fn is a mock method which can be asserted on via the mock.Mock API.
+// See: https://pkg.go.dev/github.com/stretchr/testify@v1.8.4/mock#Mock.
+func (m *FnMethodSpy) Fn(event polylog.Event) {
+	m.Called(event)
+}
+
+// EventMethodsTest is a test case for expressing and exercising polylog.Event
+// methods in a concise way.
+type EventMethodsTest struct {
```
Olshansk, line 50: Optional: thoughts on `EventMethodsTestCase`?

testutil/testpolylog/event.go
```diff
@@ -0,0 +1,166 @@
...
+	Value                  any
+	EventMethodName        string
+	ExpectedOutputContains string
+}
+
+// RunEventMethodTests runs a set of tests for a given level.
+func RunEventMethodTests(
+	t *testing.T,
+	level polylog.Level,
+	tests []EventMethodsTest,
```
Olshansk, line 64: Rename to `testCases` if we rename the above

testutil/testpolylog/event.go
```diff
@@ -0,0 +1,166 @@
...
+) {
+	t.Helper()
+
+	// Title-case level string so that it can be used as the name of the
+	// method to call on the logger using reflect and for the sub-test
+	// descriptions.
+	//
+	// TODO_TECHDEBT/TODO_COMMUNITY: `strings.Title()` is deprecated. Follow
+	// migration guidance in godocs: https://pkg.go.dev/strings@go1.21.4#Title.
+	levelMethodStr := strings.Title(level.String())
```
Olshansk, line 78: s/levelMethodStr/levelStr for consistency

testutil/testpolylog/event.go
```diff
@@ -0,0 +1,166 @@
...
+
+	for _, tt := range tests {
+		// If the test case does not specify an event method name, use the test
+		// case's key as the event method name. This is done for convenience only.
```
Olshansk, line 82: `This is done for convenience only.`

Is it hard to avoid? I want to try to avoid easy things this early in the design of a new testing primitive.

testutil/testpolylog/event.go
```diff
@@ -0,0 +1,166 @@
...
+		var (
+			eventMethodArgs []reflect.Value
+			doneMethodName  string
+			doneMethodArgs  []reflect.Value
+		)
+
+		// Ensure that calls to #Msg(), #Msgf(), and #Send() are mutually exclusive.
+		switch {
+		case tt.Msg != "":
+			// Set up call args for polylog.Event#Msg() if tt.msg is not emtpy.
```
Olshansk, line 96:
```suggestion
			// Set up call args for polylog.Event#Msg() if tt.msg is not empty.
```

testutil/testpolylog/event.go
```diff
@@ -0,0 +1,166 @@
...
+			tt.EventMethodName = tt.Key
+		}
+
+		var (
+			eventMethodArgs []reflect.Value
+			doneMethodName  string
+			doneMethodArgs  []reflect.Value
+		)
+
+		// Ensure that calls to #Msg(), #Msgf(), and #Send() are mutually exclusive.
```
Olshansk, line 93: 1. I think we should leave a comment explaining this in the struct (no need for an overkill with specialized "one of" structs")

2. We should through an error if `Msg != ""` and one of `MsgF` and `MsgArgs` is non-empty. It'll save some developer time one day in the future.

testutil/testpolylog/event.go
```diff
@@ -0,0 +1,166 @@
...
+		switch {
+		case tt.Msg != "":
+			// Set up call args for polylog.Event#Msg() if tt.msg is not emtpy.
+			doneMethodName = "Msg"
+			doneMethodArgs = append(doneMethodArgs, reflect.ValueOf(tt.Msg))
+		case tt.MsgFmt != "":
+			// Set up call args for polylog.Event#Msgf() if tt.msgFmt is not emtpy.
```
Olshansk, line 100:
```suggestion
			// Set up call args for polylog.Event#Msgf() if tt.msgFmt is not empty.
```

testutil/testpolylog/event.go
```diff
@@ -0,0 +1,166 @@
...
+				reflect.ValueOf(tt.MsgFmt),
+				reflect.ValueOf(tt.MsgFmtArgs),
+			)
+		default:
+			// Default to calling polylog.Event#Send() if tt.msg and tt.msgFmt are
+			// both empty.
+			doneMethodName = "Send"
+		}
+
+		// If tt.EventMethodName and tt.Key are both empty, then use the done
```
Olshansk, line 113: Need to either add lots of comments in the struct or rethink this because the line of thinking isn't clear to me.

testutil/testpolylog/event.go
```diff
@@ -0,0 +1,166 @@
...
+			logger, logOutput := newLoggerAndOutput(t, level)
+
+			// Need to use reflection in order to minimize the test code necessary
+			// to exercise all the permutations of logger level and event type methods.
+			logEvent := newEventWithLevel(t, logger, level)
+			logEventValue := reflect.ValueOf(logEvent)
+
+			// If tt.EventMethodName is not empty, build the args and call it.
+			if tt.EventMethodName != "" {
+				// Append tt.key to polylog.Event#<level>() call args.
```
Olshansk, line 132: Given that `Key` and `Value` are used very similarly, why not just have an `[]Args`?

testutil/testpolylog/event.go
```diff
@@ -0,0 +1,166 @@
...
+				// E.g.: logEvent := logger.Debug().Str("str", "str_value")
+				//   or: logEvent := logger.Debug().Bool("bool", true)
+				logEventValue.
+					MethodByName(tt.EventMethodName).
+					Call(eventMethodArgs)
+			}
+
+			// E.g.: logEvent.Send()
+			//   or: logEvent.Msg("msg")
+			//   or: logEvent.Msgf("meaning of life: %d", 42)
```
Olshansk, line 150: 190724-hitchhikers-guide-to-the-galaxy-news

Olshansk (changes requested):
I asked some big questions because I want to set a solid foundation. Lmkwyt (let me know what you think) and if a call/discussion might be easier than typing everything out.

testutil/testpolylog/event.go
bryanchriswhite: I would prefer moving in a direction like `levelMethodName` to make it clear that this is something categorically different from other usages of "level". The actual level strings are all lowercase which makes them different by definition from their corresponding logger level method (i.e. "debug" != "Debug"). Do you have a suggestion for improving the comment above?

testutil/testpolylog/event.go
bryanchriswhite: The type of an `args` slice would have to be `[]any` which is largely antithetical to the test suite as it is exercising a bunch of type-specific methods (at each level). This would require lots of additional casting and create additional indirection.

testutil/testpolylog/event.go
bryanchriswhite: This seems reasonable.

testutil/testpolylog/event.go
bryanchriswhite: How strongly do you feel about  this? I recently switch my personal convention **from** `testCases` to `tests` (and have been doing so elsewhere in this codebase). I take your point that if the struct name includes the work "case" then that's clearly what we're talking about here. However,  I feel like it's unnecessarily verbose and I've since seen multiple examples of code in the wild (incl. cosmos-sdk) that uses the name `tests` conventionally for this purpose.

testutil/testpolylog/event.go
```diff
@@ -0,0 +1,166 @@
...
+	// configured with the given level and options. It returns the logger and
+	// the buffer to which the logger writes. It is useful for decoupling test
+	// helpers from a specific logger implementation.
+	NewLoggerAndOutputFn func(
+		*testing.T,
+		polylog.Level,
+		...polylog.LoggerOption,
+	) (polylog.Logger, *bytes.Buffer)
+
+	// NewEventWithLevelFn is called in the test helper to create a new event
```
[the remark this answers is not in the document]
bryanchriswhite: This is a way to call a `logger.<level method>()` without knowing the logger concrete type nor the level. This is necessary in order for `RunEventMethodTests()` to be reusable between different logger implementations and log levels.

testutil/testpolylog/event.go
```diff
@@ -0,0 +1,199 @@
...
+	"github.com/stretchr/testify/require"
+
+	"github.com/pokt-network/poktroll/pkg/polylog"
+)
+
+type (
+	// NewLoggerAndOutputFn is called in the test helper to create a new logger
+	// configured with the given level and options. It returns the logger and
+	// the buffer to which the logger writes. It is useful for decoupling test
+	// helpers from a specific logger implementation.
```
Olshansk, line 20:
```suggestion
	// the buffer to which the logger writes. It is useful for decoupling test
	// helpers from a specific logger implementation.
```

testutil/testpolylog/event.go
Olshansk: I'll push back on this. I think the added verbosity needing to explicitly specify this will remove potential confusion/indirection from test readers/authors.

testutil/testpolylog/event.go
```diff
@@ -0,0 +1,199 @@
...
+	NewLoggerAndOutputFn func(
+		*testing.T,
+		polylog.Level,
+		...polylog.LoggerOption,
+	) (polylog.Logger, *bytes.Buffer)
+
+	// NewEventWithLevelFn is called in the test helper to create a new event
+	// at the given level from the given logger. It is useful for decoupling
+	// test helpers from a specific logger implementation.
```
Olshansk, line 29:
```suggestion
	// test helpers from a specific logger implementation so that we can
	// call `logger.<LevelMethod>() without knowing the concrete logger
	// type nor the level.
```

testutil/testpolylog/event.go
```diff
@@ -0,0 +1,166 @@
...
+	NewEventWithLevelFn func(
+		*testing.T,
+		polylog.Logger,
+		polylog.Level,
+	) polylog.Event
+)
+
+// FnMethodSpy is a mock which implements a #Fn() method that is intended to be
```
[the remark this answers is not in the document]
Olshansk: What if we rename `Fn` to `SpyOnEvent`?

testutil/testpolylog/event.go
```diff
@@ -45,26 +45,50 @@ func (m *FnMethodSpy) Fn(event polylog.Event) {
...
+	// of MsgFmt is mutually exclusive with Msg. If neither are provided, then
+	// polylog.Event#Send() is called after the event method instead.
+	MsgFmt string
+
+	// MsgFmtArgs are the args to pass to polylog.Event#Msgf(). It is an error
+	// to provide MsgFmtArgs without also providing MsgFmt or while providing
+	// Msg.
+	MsgFmtArgs []any
+
+	// Key is the key to pass to the event method under test. If EventMethodName
```
Olshansk, line 68: Reminder to remove `If EventMethodName...` or not depending on the outcome of our other discussion.

We'd be able to avoid the comment here, below, in the implementation in exchange for verbosity in test case definition.

testutil/testpolylog/event.go
```diff
@@ -93,11 +117,14 @@ func RunEventMethodTests(
 		// Ensure that calls to #Msg(), #Msgf(), and #Send() are mutually exclusive.
 		switch {
 		case tt.Msg != "":
-			// Set up call args for polylog.Event#Msg() if tt.msg is not emtpy.
+			require.Emptyf(t, tt.MsgFmt, "Msg and MsgFmt are mutually exclusive but MsgFmt was not empty: %s", tt.MsgFmt)
```
Olshansk, line 120: Nice 🙌

Olshansk (changes requested):

testutil/testpolylog/event.go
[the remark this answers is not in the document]
bryanchriswhite: My rationale behind `FnMethodSpy` is that:
1. The object is a spy by definition of how we intend on using it.
2. Because the spy object must encapsulate the function we're spying on, it's becomes method which subsequently needs a name.
3. As we only need a single spy function, it seemed reasonable to name the method `Fn`.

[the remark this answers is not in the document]
Olshansk: > As we only need a single spy function

I agree, but this seems like an implicit requirement. If we rename it to `FnMethodSpy`, the code is self documenting without needing to understand it as deeply.

testutil/testpolylog/event.go
Olshansk: Fair enough. Don't feel too strongly.

I feel more strongly about using `desc` instead of `name`, which we started doing in various places.

testutil/testpolylog/event.go
Olshansk: +1 to levelMethodName!

testutil/testpolylog/event.go
Olshansk: Now that we also specify the method name explicitly, in addition to your new comments, the codepath is clear and approachable!

Olshansk (approved):
LGTM.

I left one last comment for consideration w.r.t renaming `Fn` to `FnMethodSpy` [1], but it's a personal preference and will leave it to your discretion. All that other changes are 💯 

[1] https://github.com/pokt-network/poktroll/pull/209#discussion_r1414868762

testutil/testpolylog/event.go
bryanchriswhite: :handshake: on `tests` and `desc`? :sweat_smile:

PR #220, opened 2023-11-23
[RingCache] Add `EventDelegateeChange` event emission and cache invalidation

h5law:
### Human Summary

This PR adds the `EventDelegateeChange` typed event and emits it on a successfull application delegation or undelegation to/from a gateway. This is listened for by the `RingCache` and used to invalidate cache entries for rings that need to be reloaded.

### AI Summary

reviewpad:summary

## Issue

- Fixes N/A

## Type of change

Select one or more:

- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

## Sanity Checklist

### DIFF
--- a/proto/pocket/application/event.proto
+++ b/proto/pocket/application/event.proto
@@ -1,0 +1,13 @@
+syntax = "proto3";
+package pocket.application;
+
+option go_package = "github.com/pokt-network/poktroll/x/application/types";
+
+import "cosmos_proto/cosmos.proto";
+
+// EventDelegateeChange is an event emitted whenever an application changes its
+// delegatee gateways on chain. This is in response to both a DelegateToGateway
+// and UndelegateFromGateway message.
+message EventDelegateeChange {
+  string app_address = 1 [(cosmos_proto.scalar) = "cosmos.AddressString"]; // The Bech32 address of the application using cosmos' ScalarDescriptor to ensure deterministic encoding
+}
--- a/x/application/keeper/msg_server_delegate_to_gateway.go
+++ b/x/application/keeper/msg_server_delegate_to_gateway.go
@@ -3,10 +3,10 @@ package keeper
 import (
 	"context"
 
-	"github.com/pokt-network/poktroll/x/application/types"
-
 	sdkerrors "cosmossdk.io/errors"
 	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	"github.com/pokt-network/poktroll/x/application/types"
 )
 
 func (k msgServer) DelegateToGateway(goCtx context.Context, msg *types.MsgDelegateToGateway) (*types.MsgDelegateToGatewayResponse, error) {
@@ -56,6 +56,9 @@ func (k msgServer) DelegateToGateway(goCtx context.Context, msg *types.MsgDelega
 	// Update the application store with the new delegation
 	k.SetApplication(ctx, app)
 	logger.Info("Successfully delegated application to gateway for app: %+v", app)
+
+	// Emit the application delegation change event
+	ctx.EventManager().EmitTypedEvent(msg.NewDelegateeChangeEvent())
 
 	return &types.MsgDelegateToGatewayResponse{}, nil
 }
--- a/x/application/keeper/msg_server_delegate_to_gateway_test.go
+++ b/x/application/keeper/msg_server_delegate_to_gateway_test.go
@@ -1,6 +1,7 @@
 package keeper_test
 
 import (
+	"fmt"
 	"testing"
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
@@ -56,6 +57,11 @@ func TestMsgServer_DelegateToGateway_SuccessfullyDelegate(t *testing.T) {
 	// Delegate the application to the gateway
 	_, err = srv.DelegateToGateway(wctx, delegateMsg)
 	require.NoError(t, err)
+	events := ctx.EventManager().Events()
+	require.Equal(t, 1, len(events))
+	require.Equal(t, "pocket.application.EventDelegateeChange", events[0].Type)
+	require.Equal(t, "app_address", events[0].Attributes[0].Key)
+	require.Equal(t, fmt.Sprintf("\"%s\"", appAddr), events[0].Attributes[0].Value)
 
 	// Verify that the application exists
 	foundApp, isAppFound := k.GetApplication(ctx, appAddr)
@@ -73,6 +79,11 @@ func TestMsgServer_DelegateToGateway_SuccessfullyDelegate(t *testing.T) {
 	// Delegate the application to the second gateway
 	_, err = srv.DelegateToGateway(wctx, delegateMsg2)
 	require.NoError(t, err)
+	events = ctx.EventManager().Events()
+	require.Equal(t, 2, len(events))
+	require.Equal(t, "pocket.application.EventDelegateeChange", events[1].Type)
+	require.Equal(t, "app_address", events[1].Attributes[0].Key)
+	require.Equal(t, fmt.Sprintf("\"%s\"", appAddr), events[1].Attributes[0].Value)
 	foundApp, isAppFound = k.GetApplication(ctx, appAddr)
 	require.True(t, isAppFound)
 	require.Equal(t, 2, len(foundApp.DelegateeGatewayAddresses))
@@ -120,6 +131,11 @@ func TestMsgServer_DelegateToGateway_FailDuplicate(t *testing.T) {
 	// Delegate the application to the gateway
 	_, err = srv.DelegateToGateway(wctx, delegateMsg)
 	require.NoError(t, err)
+	events := ctx.EventManager().Events()
+	require.Equal(t, 1, len(events))
+	require.Equal(t, "pocket.application.EventDelegateeChange", events[0].Type)
+	require.Equal(t, "app_address", events[0].Attributes[0].Key)
+	require.Equal(t, fmt.Sprintf("\"%s\"", appAddr), events[0].Attributes[0].Value)
 
 	// Verify that the application exists
 	foundApp, isAppFound := k.GetApplication(ctx, appAddr)
@@ -137,6 +153,8 @@ func TestMsgServer_DelegateToGateway_FailDuplicate(t *testing.T) {
 	// Attempt to delegate the application to the gateway again
 	_, err = srv.DelegateToGateway(wctx, delegateMsg2)
 	require.ErrorIs(t, err, types.ErrAppAlreadyDelegated)
+	events = ctx.EventManager().Events()
+	require.Equal(t, 1, len(events))
 	foundApp, isAppFound = k.GetApplication(ctx, appAddr)
 	require.True(t, isAppFound)
 	require.Equal(t, 1, len(foundApp.DelegateeGatewayAddresses))
@@ -242,10 +260,19 @@ func TestMsgServer_DelegateToGateway_FailMaxReached(t *testing.T) {
 		require.True(t, isAppFound)
 		require.Equal(t, int(i+1), len(foundApp.DelegateeGatewayAddresses))
 	}
+	events := ctx.EventManager().Events()
+	require.Equal(t, int(maxDelegatedParam), len(events))
+	for _, event := range events {
+		require.Equal(t, "pocket.application.EventDelegateeChange", event.Type)
+		require.Equal(t, "app_address", event.Attributes[0].Key)
+		require.Equal(t, fmt.Sprintf("\"%s\"", appAddr), event.Attributes[0].Value)
+	}
 
 	// Attempt to delegate the application when the max is already reached
 	_, err = srv.DelegateToGateway(wctx, delegateMsg)
 	require.ErrorIs(t, err, types.ErrAppMaxDelegatedGateways)
+	events = ctx.EventManager().Events()
+	require.Equal(t, int(maxDelegatedParam), len(events))
 	foundApp, isAppFound := k.GetApplication(ctx, appAddr)
 	require.True(t, isAppFound)
 	require.Equal(t, maxDelegatedParam, int64(len(foundApp.DelegateeGatewayAddresses)))
--- a/x/application/keeper/msg_server_delegate_to_gateway_test.go
+++ b/x/application/keeper/msg_server_delegate_to_gateway_test.go
@@ -24,12 +24,8 @@ func TestMsgServer_DelegateToGateway_SuccessfullyDelegate(t *testing.T) {
 	gatewayAddr1 := sample.AccAddress()
 	gatewayAddr2 := sample.AccAddress()
 	// Mock the gateway being staked via the staked gateway map
-	keepertest.StakedGatewayMap[gatewayAddr1] = struct{}{}
-	keepertest.StakedGatewayMap[gatewayAddr2] = struct{}{}
-	t.Cleanup(func() {
-		delete(keepertest.StakedGatewayMap, gatewayAddr1)
-		delete(keepertest.StakedGatewayMap, gatewayAddr2)
-	})
+	keepertest.AddGatewayToStakedGatewayMap(t, gatewayAddr1)
+	keepertest.AddGatewayToStakedGatewayMap(t, gatewayAddr2)
 
 	// Prepare the application
 	stakeMsg := &types.MsgStakeApplication{
@@ -100,10 +96,7 @@ func TestMsgServer_DelegateToGateway_FailDuplicate(t *testing.T) {
 	appAddr := sample.AccAddress()
 	gatewayAddr := sample.AccAddress()
 	// Mock the gateway being staked via the staked gateway map
-	keepertest.StakedGatewayMap[gatewayAddr] = struct{}{}
-	t.Cleanup(func() {
-		delete(keepertest.StakedGatewayMap, gatewayAddr)
-	})
+	keepertest.AddGatewayToStakedGatewayMap(t, gatewayAddr)
 
 	// Prepare the application
 	stakeMsg := &types.MsgStakeApplication{
@@ -210,10 +203,7 @@ func TestMsgServer_DelegateToGateway_FailMaxReached(t *testing.T) {
 	appAddr := sample.AccAddress()
 	gatewayAddr := sample.AccAddress()
 	// Mock the gateway being staked via the staked gateway map
-	keepertest.StakedGatewayMap[gatewayAddr] = struct{}{}
-	t.Cleanup(func() {
-		delete(keepertest.StakedGatewayMap, gatewayAddr)
-	})
+	keepertest.AddGatewayToStakedGatewayMap(t, gatewayAddr)
 
 	// Prepare the application
 	stakeMsg := &types.MsgStakeApplication{
@@ -244,10 +234,7 @@ func TestMsgServer_DelegateToGateway_FailMaxReached(t *testing.T) {
 		// Prepare the delegation message
 		gatewayAddr := sample.AccAddress()
 		// Mock the gateway being staked via the staked gateway map
-		keepertest.StakedGatewayMap[gatewayAddr] = struct{}{}
-		t.Cleanup(func() {
-			delete(keepertest.StakedGatewayMap, gatewayAddr)
-		})
+		keepertest.AddGatewayToStakedGatewayMap(t, gatewayAddr)
 		delegateMsg := &types.MsgDelegateToGateway{
 			AppAddress:     appAddr,
 			GatewayAddress: gatewayAddr,
--- a/x/application/types/message_delegate_to_gateway.go
+++ b/x/application/types/message_delegate_to_gateway.go
@@ -37,6 +37,12 @@ func (msg *MsgDelegateToGateway) GetSignBytes() []byte {
 	return sdk.MustSortJSON(bz)
 }
 
+func (msg *MsgDelegateToGateway) NewDelegateeChangeEvent() *EventDelegateeChange {
+	return &EventDelegateeChange{
+		AppAddress: msg.AppAddress,
+	}
+}
+
 func (msg *MsgDelegateToGateway) ValidateBasic() error {
 	// Validate the application address
 	if _, err := sdk.AccAddressFromBech32(msg.AppAddress); err != nil {
--- a/x/application/types/message_undelegate_from_gateway.go
+++ b/x/application/types/message_undelegate_from_gateway.go
@@ -37,6 +37,12 @@ func (msg *MsgUndelegateFromGateway) GetSignBytes() []byte {
 	return sdk.MustSortJSON(bz)
 }
 
+func (msg *MsgUndelegateFromGateway) NewDelegateeChangeEvent() *EventDelegateeChange {
+	return &EventDelegateeChange{
+		AppAddress: msg.AppAddress,
+	}
+}
+
 func (msg *MsgUndelegateFromGateway) ValidateBasic() error {
 	// Validate the application address
 	if _, err := sdk.AccAddressFromBech32(msg.AppAddress); err != nil {
--- a/go.mod
+++ b/go.mod
@@ -10,6 +10,7 @@ require (
 	github.com/athanorlabs/go-dleq v0.1.0
 	github.com/cometbft/cometbft v0.37.2
 	github.com/cometbft/cometbft-db v0.8.0
+	github.com/cosmos/cosmos-proto v1.0.0-beta.2
 	github.com/cosmos/cosmos-sdk v0.47.3
 	github.com/cosmos/gogoproto v1.4.10
 	github.com/cosmos/ibc-go/v7 v7.1.0
@@ -30,6 +31,7 @@ require (
 	go.uber.org/multierr v1.11.0
 	golang.org/x/crypto v0.12.0
 	golang.org/x/sync v0.3.0
+	google.golang.org/genproto v0.0.0-20230410155749-daa745c078e1
 	google.golang.org/grpc v1.56.1
 	gopkg.in/yaml.v2 v2.4.0
 )
@@ -73,7 +75,6 @@ require (
 	github.com/containerd/cgroups v1.1.0 // indirect
 	github.com/coreos/go-systemd/v22 v22.5.0 // indirect
 	github.com/cosmos/btcutil v1.0.5 // indirect
-	github.com/cosmos/cosmos-proto v1.0.0-beta.2 // indirect
 	github.com/cosmos/go-bip39 v1.0.0 // indirect
 	github.com/cosmos/gogogateway v1.2.0 // indirect
 	github.com/cosmos/iavl v0.20.0 // indirect
@@ -269,7 +270,6 @@ require (
 	gonum.org/v1/gonum v0.11.0 // indirect
 	google.golang.org/api v0.122.0 // indirect
 	google.golang.org/appengine v1.6.7 // indirect
-	google.golang.org/genproto v0.0.0-20230410155749-daa745c078e1 // indirect
 	google.golang.org/protobuf v1.31.0 // indirect
 	gopkg.in/ini.v1 v1.67.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
--- a/testutil/keeper/application.go
+++ b/testutil/keeper/application.go
@@ -21,12 +21,14 @@ import (
 	gatewaytypes "github.com/pokt-network/poktroll/x/gateway/types"
 )
 
-// StakedGatewayMap is used to mock whether a gateway is staked or not for use
+// stakedGatewayMap is used to mock whether a gateway is staked or not for use
 // in the application's mocked gateway keeper. This enables the tester to
 // control whether a gateway is "staked" or not and whether it can be delegated to
 // WARNING: Using this map may cause issues if running multiple tests in parallel
-var StakedGatewayMap = make(map[string]struct{})
+var stakedGatewayMap = make(map[string]struct{})
 
+// ApplicationKeeper returns a mocked application keeper and context for testing
+// it mocks the chain having staked gateways via the use of the stakedGatewayMap
 func ApplicationKeeper(t testing.TB) (*keeper.Keeper, sdk.Context) {
 	storeKey := sdk.NewKVStoreKey(types.StoreKey)
 	memStoreKey := storetypes.NewMemoryStoreKey(types.MemStoreKey)
@@ -51,7 +53,7 @@ func ApplicationKeeper(t testing.TB) (*keeper.Keeper, sdk.Context) {
 	mockGatewayKeeper := mocks.NewMockGatewayKeeper(ctrl)
 	mockGatewayKeeper.EXPECT().GetGateway(gomock.Any(), gomock.Any()).DoAndReturn(
 		func(_ sdk.Context, addr string) (gatewaytypes.Gateway, bool) {
-			if _, ok := StakedGatewayMap[addr]; !ok {
+			if _, ok := stakedGatewayMap[addr]; !ok {
 				return gatewaytypes.Gateway{}, false
 			}
 			stake := sdk.NewCoin("upokt", sdk.NewInt(10000))
@@ -84,4 +86,22 @@ func ApplicationKeeper(t testing.TB) (*keeper.Keeper, sdk.Context) {
 	k.SetParams(ctx, types.DefaultParams())
 
 	return k, ctx
+}
+
+// AddGatewayToStakedGatewayMap adds the given gateway address to the staked
+// gateway map for use in the application's mocked gateway keeper and ensures
+// that it is removed from the map when the test is complete
+func AddGatewayToStakedGatewayMap(t *testing.T, gatewayAddr string) {
+	t.Helper()
+	stakedGatewayMap[gatewayAddr] = struct{}{}
+	t.Cleanup(func() {
+		delete(stakedGatewayMap, gatewayAddr)
+	})
+}
+
+// RemoveGatewayFromStakedGatewayMap removes the given gateway address from the
+// staked gateway map for use in the application's mocked gateway keeper
+func RemoveGatewayFromStakedGatewayMap(t *testing.T, gatewayAddr string) {
+	t.Helper()
+	delete(stakedGatewayMap, gatewayAddr)
 }
--- a/x/application/keeper/msg_server_undelegate_from_gateway.go
+++ b/x/application/keeper/msg_server_undelegate_from_gateway.go
@@ -9,7 +9,10 @@ import (
 	"github.com/pokt-network/poktroll/x/application/types"
 )
 
-func (k msgServer) UndelegateFromGateway(goCtx context.Context, msg *types.MsgUndelegateFromGateway) (*types.MsgUndelegateFromGatewayResponse, error) {
+func (k msgServer) UndelegateFromGateway(
+	goCtx context.Context,
+	msg *types.MsgUndelegateFromGateway,
+) (*types.MsgUndelegateFromGatewayResponse, error) {
 	ctx := sdk.UnwrapSDKContext(goCtx)
 
 	logger := k.Logger(ctx).With("method", "UndelegateFromGateway")
@@ -46,6 +49,9 @@ func (k msgServer) UndelegateFromGateway(goCtx context.Context, msg *types.MsgUn
 	// Update the application store with the new delegation
 	k.SetApplication(ctx, app)
 	logger.Info("Successfully undelegated application from gateway for app: %+v", app)
+
+	// Emit the application delegation change event
+	ctx.EventManager().EmitTypedEvent(msg.NewDelegateeChangeEvent())
 
 	return &types.MsgUndelegateFromGatewayResponse{}, nil
 }
--- a/x/application/keeper/msg_server_undelegate_from_gateway_test.go
+++ b/x/application/keeper/msg_server_undelegate_from_gateway_test.go
@@ -179,7 +179,7 @@ func TestMsgServer_UndelegateFromGateway_SuccessfullyUndelegateFromUnstakedGatew
 	appAddr := sample.AccAddress()
 	gatewayAddr := sample.AccAddress()
 	// Mock the gateway being staked via the staked gateway map
-	keepertest.RemoveGatewayFromStakedGatewayMap(t, gatewayAddr)
+	keepertest.AddGatewayToStakedGatewayMap(t, gatewayAddr)
 
 	// Prepare the application
 	stakeMsg := &types.MsgStakeApplication{
@@ -220,7 +220,7 @@ func TestMsgServer_UndelegateFromGateway_SuccessfullyUndelegateFromUnstakedGatew
 	require.Equal(t, gatewayAddr, foundApp.DelegateeGatewayAddresses[0])
 
 	// Mock unstaking the gateway
-	keepertest.AddGatewayToStakedGatewayMap(t, gatewayAddr)
+	keepertest.RemoveGatewayFromStakedGatewayMap(t, gatewayAddr)
 
 	// Prepare an undelegation message
 	undelegateMsg := &types.MsgUndelegateFromGateway{
--- a/x/application/keeper/msg_server_undelegate_from_gateway_test.go
+++ b/x/application/keeper/msg_server_undelegate_from_gateway_test.go
@@ -26,14 +26,9 @@ func TestMsgServer_UndelegateFromGateway_SuccessfullyUndelegate(t *testing.T) {
 	for i := 0; i < len(gatewayAddresses); i++ {
 		gatewayAddr := sample.AccAddress()
 		// Mock the gateway being staked via the staked gateway map
-		keepertest.StakedGatewayMap[gatewayAddr] = struct{}{}
+		keepertest.AddGatewayToStakedGatewayMap(t, gatewayAddr)
 		gatewayAddresses[i] = gatewayAddr
 	}
-	t.Cleanup(func() {
-		for _, gatewayAddr := range gatewayAddresses {
-			delete(keepertest.StakedGatewayMap, gatewayAddr)
-		}
-	})
 
 	// Prepare the application
 	stakeMsg := &types.MsgStakeApplication{
@@ -113,12 +108,8 @@ func TestMsgServer_UndelegateFromGateway_FailNotDelegated(t *testing.T) {
 	gatewayAddr1 := sample.AccAddress()
 	gatewayAddr2 := sample.AccAddress()
 	// Mock the gateway being staked via the staked gateway map
-	keepertest.StakedGatewayMap[gatewayAddr1] = struct{}{}
-	keepertest.StakedGatewayMap[gatewayAddr2] = struct{}{}
-	t.Cleanup(func() {
-		delete(keepertest.StakedGatewayMap, gatewayAddr1)
-		delete(keepertest.StakedGatewayMap, gatewayAddr2)
-	})
+	keepertest.AddGatewayToStakedGatewayMap(t, gatewayAddr1)
+	keepertest.AddGatewayToStakedGatewayMap(t, gatewayAddr2)
 
 	// Prepare the application
 	stakeMsg := &types.MsgStakeApplication{
@@ -188,7 +179,7 @@ func TestMsgServer_UndelegateFromGateway_SuccessfullyUndelegateFromUnstakedGatew
 	appAddr := sample.AccAddress()
 	gatewayAddr := sample.AccAddress()
 	// Mock the gateway being staked via the staked gateway map
-	keepertest.StakedGatewayMap[gatewayAddr] = struct{}{}
+	keepertest.RemoveGatewayFromStakedGatewayMap(t, gatewayAddr)
 
 	// Prepare the application
 	stakeMsg := &types.MsgStakeApplication{
@@ -229,7 +220,7 @@ func TestMsgServer_UndelegateFromGateway_SuccessfullyUndelegateFromUnstakedGatew
 	require.Equal(t, gatewayAddr, foundApp.DelegateeGatewayAddresses[0])
 
 	// Mock unstaking the gateway
-	delete(keepertest.StakedGatewayMap, gatewayAddr)
+	keepertest.AddGatewayToStakedGatewayMap(t, gatewayAddr)
 
 	// Prepare an undelegation message
 	undelegateMsg := &types.MsgUndelegateFromGateway{
--- a/x/application/keeper/msg_server_undelegate_from_gateway_test.go
+++ b/x/application/keeper/msg_server_undelegate_from_gateway_test.go
@@ -1,6 +1,7 @@
 package keeper_test
 
 import (
+	"fmt"
 	"testing"
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
@@ -20,7 +21,8 @@ func TestMsgServer_UndelegateFromGateway_SuccessfullyUndelegate(t *testing.T) {
 
 	// Generate an address for the application and gateways
 	appAddr := sample.AccAddress()
-	gatewayAddresses := make([]string, int(k.GetParams(ctx).MaxDelegatedGateways))
+	maxDelegatedGateways := k.GetParams(ctx).MaxDelegatedGateways
+	gatewayAddresses := make([]string, int(maxDelegatedGateways))
 	for i := 0; i < len(gatewayAddresses); i++ {
 		gatewayAddr := sample.AccAddress()
 		// Mock the gateway being staked via the staked gateway map
@@ -60,9 +62,15 @@ func TestMsgServer_UndelegateFromGateway_SuccessfullyUndelegate(t *testing.T) {
 		_, err = srv.DelegateToGateway(wctx, delegateMsg)
 		require.NoError(t, err)
 	}
+	events := ctx.EventManager().Events()
+	require.Equal(t, int(maxDelegatedGateways), len(events))
+	for _, event := range events {
+		require.Equal(t, "pocket.application.EventDelegateeChange", event.Type)
+		require.Equal(t, "app_address", event.Attributes[0].Key)
+		require.Equal(t, fmt.Sprintf("\"%s\"", appAddr), event.Attributes[0].Value)
+	}
 
 	// Verify that the application exists
-	maxDelegatedGateways := k.GetParams(ctx).MaxDelegatedGateways
 	foundApp, isAppFound := k.GetApplication(ctx, appAddr)
 	require.True(t, isAppFound)
 	require.Equal(t, appAddr, foundApp.Address)
@@ -80,6 +88,11 @@ func TestMsgServer_UndelegateFromGateway_SuccessfullyUndelegate(t *testing.T) {
 	// Undelegate the application from the gateway
 	_, err = srv.UndelegateFromGateway(wctx, undelegateMsg)
 	require.NoError(t, err)
+	events = ctx.EventManager().Events()
+	require.Equal(t, int(maxDelegatedGateways)+1, len(events))
+	require.Equal(t, "pocket.application.EventDelegateeChange", events[7].Type)
+	require.Equal(t, "app_address", events[7].Attributes[0].Key)
+	require.Equal(t, fmt.Sprintf("\"%s\"", appAddr), events[7].Attributes[0].Value)
 	foundApp, isAppFound = k.GetApplication(ctx, appAddr)
 	require.True(t, isAppFound)
 	require.Equal(t, appAddr, foundApp.Address)
@@ -137,6 +150,8 @@ func TestMsgServer_UndelegateFromGateway_FailNotDelegated(t *testing.T) {
 	require.True(t, isAppFound)
 	require.Equal(t, appAddr, foundApp.Address)
 	require.Equal(t, 0, len(foundApp.DelegateeGatewayAddresses))
+	events := ctx.EventManager().Events()
+	require.Equal(t, 0, len(events))
 
 	// Prepare a delegation message
 	delegateMsg := &types.MsgDelegateToGateway{
@@ -147,10 +162,17 @@ func TestMsgServer_UndelegateFromGateway_FailNotDelegated(t *testing.T) {
 	// Delegate the application to the gateway
 	_, err = srv.DelegateToGateway(wctx, delegateMsg)
 	require.NoError(t, err)
+	events = ctx.EventManager().Events()
+	require.Equal(t, 1, len(events))
+	require.Equal(t, "pocket.application.EventDelegateeChange", events[0].Type)
+	require.Equal(t, "app_address", events[0].Attributes[0].Key)
+	require.Equal(t, fmt.Sprintf("\"%s\"", appAddr), events[0].Attributes[0].Value)
 
 	// Ensure the failed undelegation did not affect the application
 	_, err = srv.UndelegateFromGateway(wctx, undelegateMsg)
 	require.ErrorIs(t, err, types.ErrAppNotDelegated)
+	events = ctx.EventManager().Events()
+	require.Equal(t, 1, len(events))
 	foundApp, isAppFound = k.GetApplication(ctx, appAddr)
 	require.True(t, isAppFound)
 	require.Equal(t, 1, len(foundApp.DelegateeGatewayAddresses))
@@ -193,6 +215,11 @@ func TestMsgServer_UndelegateFromGateway_SuccessfullyUndelegateFromUnstakedGatew
 	// Delegate the application to the gateway
 	_, err = srv.DelegateToGateway(wctx, delegateMsg)
 	require.NoError(t, err)
+	events := ctx.EventManager().Events()
+	require.Equal(t, 1, len(events))
+	require.Equal(t, "pocket.application.EventDelegateeChange", events[0].Type)
+	require.Equal(t, "app_address", events[0].Attributes[0].Key)
+	require.Equal(t, fmt.Sprintf("\"%s\"", appAddr), events[0].Attributes[0].Value)
 
 	// Verify that the application exists
 	foundApp, isAppFound := k.GetApplication(ctx, appAddr)
@@ -213,6 +240,11 @@ func TestMsgServer_UndelegateFromGateway_SuccessfullyUndelegateFromUnstakedGatew
 	// Undelegate the application from the gateway
 	_, err = srv.UndelegateFromGateway(wctx, undelegateMsg)
 	require.NoError(t, err)
+	events = ctx.EventManager().Events()
+	require.Equal(t, 2, len(events))
+	require.Equal(t, "pocket.application.EventDelegateeChange", events[1].Type)
+	require.Equal(t, "app_address", events[1].Attributes[0].Key)
+	require.Equal(t, fmt.Sprintf("\"%s\"", appAddr), events[1].Attributes[0].Value)
 	foundApp, isAppFound = k.GetApplication(ctx, appAddr)
 	require.True(t, isAppFound)
 	require.Equal(t, appAddr, foundApp.Address)


pkg/client/interface.go
```diff
@@ -94,33 +99,34 @@ type TxContext interface {
...
+// events as well as direct access to the latest event via some blockchain API.
+type MappedClient[T any] interface {
+	// EventsSequence returns an observable which emits new events.
+	EventsSequence(context.Context) EventsObservable[T]
+	// LatestEvent returns the latest event that has been received.
+	LatestEvent(context.Context) T
+	// Close unsubscribes all observers of the events sequence observable
+	// and closes the events query client.
+	Close()
+}
```
h5law, line 128: These is the generic MappedClient interface

Olshansk:
@h5law Skipping this over for now but please request a review when we should take look.

pkg/client/interface.go
bryanchriswhite: @h5law After thinking about it for a bit, I think this should be called `EventsReplayClient`.

bryanchriswhite: I still plan on looking more closely at the package structure but haven't had a chance to do so yet.

h5law: I actually called this the ReplayClient first then the TransformingClient 🤖 then MappedClient lmao

bryanchriswhite: I also think that the block and delegate clients should lightly wrap (as opposed to bring or embedding) this generic type so that they have control over their own method names and, more importantly, become open for extension (S**O**LID).

pkg/client/block/interface.go
```diff
@@ -0,0 +1,28 @@
...
+import (
+	"context"
+
+	"github.com/pokt-network/poktroll/pkg/client"
+	"github.com/pokt-network/poktroll/pkg/observable"
+)
+
+type (
+	// Observable wraps the generic observable.ReplayObservable[Block] type
+	Observable observable.ReplayObservable[client.Block]
```
bryanchriswhite, line 13: I think this should be called `BlocksObservable` to improve clarity. I get that it's defined in the block package but when reading code that uses is in the block package, it can be a bit confusing.

pkg/client/delegation/interface.go
```diff
@@ -0,0 +1,28 @@
...
+import (
+	"context"
+
+	"github.com/pokt-network/poktroll/pkg/client"
+	"github.com/pokt-network/poktroll/pkg/observable"
+)
+
+type (
+	// Observable wraps the generic observable.ReplayObservable[DelegateeChange] type
+	Observable observable.ReplayObservable[client.DelegateeChange]
```
bryanchriswhite, line 13: Similar feedback here, I feel like `DelegateeChangeReplayObservable` (or something similar) would improve readability.

pkg/client/block/godoc.go
```diff
@@ -0,0 +1,5 @@
+// Package block contains the client implementation of the MappedClient
+// generic which listens for committed block events on chain and emits them
```
bryanchriswhite, line 2: Don't forget to update comments referencing anything that's getting renamed.

pkg/client/events_replay_client/client.go
bryanchriswhite, line 1: This package should be `eventsreplay`. No need to append ”client".

bryanchriswhite: On second thought, perhaps this should be `pkg/client/event/` instead. That would be more consistent with what you've done with block and delegate. Wdyt?

pkg/client/delegation/interface.go
```diff
@@ -0,0 +1,29 @@
...
+)
+
+type (
+	// DelegateeChangeReplayObservable wraps the generic
+	// observable.ReplayObservable[DelegateeChange] type
+	DelegateeChangeReplayObservable observable.ReplayObservable[client.DelegateeChange]
+
+	// Client is an interface that wraps the EventsReplayClient interface
+	// specific for the EventsReplayClient[DelegateeChange] implementation
+	Client interface {
```
bryanchriswhite, line 18: (same as https://github.com/pokt-network/poktroll/pull/220#discussion_r1405852929)

pkg/client/event/godoc.go
```diff
@@ -0,0 +1,12 @@
+// Package event provides a generic client for subscribing to on-chain events
```
bryanchriswhite, line 1: I feel like it would make more sense for this and `pkg/client/events_query` **to be the same**. However, upon reflection, I don't see any reason not to simplify the name to `pkg/client/events`, like you have here.

pkg/client/block/client.go
```diff
@@ -2,222 +2,75 @@ package block
...
-
-// Close unsubscribes all observers of the committed blocks sequence observable
-// and closes the events query client.
-func (bClient *blockClient) Close() {
-	// Closing eventsClient will cascade unsubscribe and close downstream observers.
-	bClient.eventsClient.Close()
+// blockClient is a wrapper around an EventsReplayClient that implements the
+// interface for use in network.
+type blockClient struct {
+	mappedClient client.EventsReplayClient[client.Block, client.EventsObservable[client.Block]]
```
bryanchriswhite, line 52:
```suggestion
	eventsReplayClient client.EventsReplayClient[client.Block, client.EventsObservable[client.Block]]
```

pkg/client/block/client.go
```diff
@@ -2,222 +2,75 @@ package block
...
-// CommittedBlocksSequence returns a ReplayObservable, with a replay buffer size
-// of 1, which is notified when block commit events are received by the events
-// query subscription.
-func (bClient *blockClient) CommittedBlocksSequence(ctx context.Context) client.BlocksObservable {
-	// Get the latest block observable from the replay observable. We only ever
-	// want the last 1 as any prior latest block observable values are closed.
-	// Directly accessing the zeroth index here is safe because the call to Last
-	// is guaranteed to return a slice with at least 1 element.
-	return bClient.latestBlockObsvbls.Last(ctx, 1)[0]
+	return &blockClient{mappedClient: client}, nil
```
bryanchriswhite, line 46:
```suggestion
	return &eventsReplayClient{mappedClient: client}, nil
```

pkg/client/block/client.go
```diff
@@ -2,222 +2,75 @@ package block
...
-	// If we get here, the retry limit was reached and the retry loop exited.
-	// Since this function runs in a goroutine, we can't return the error to the
-	// caller. Instead, we panic.
-	if publishErr != nil {
-		panic(fmt.Errorf("BlockClient.goPublishBlocks should never reach this spot: %w", publishErr))
-	}
+// CommittedBlocksSequence returns a replay observable of observables for Block events
+// from the BlockClient.
+func (b *blockClient) CommittedBlocksSequence(ctx context.Context) BlockReplayObservable {
+	return b.mappedClient.EventsSequence(ctx).(BlockReplayObservable)
```
bryanchriswhite, line 58:
```suggestion
	return b.eventsReplayClient.EventsSequence(ctx).(BlockReplayObservable)
```

pkg/client/block/client.go
```diff
@@ -2,222 +2,75 @@ package block
...
-			newEventsBytesToBlockMapFn(errCh),
-		)
-
-		// Initially set latestBlockObsvbls and update if after retrying on error.
-		bClient.latestBlockObsvblsReplayPublishCh <- blocksObsvbl
-
-		return errCh
+// LatestsNEvents returns the latest n blocks from the BockClient.
+func (b *blockClient) LastNBlocks(ctx context.Context, n int) []client.Block {
+	events := b.mappedClient.LastNEvents(ctx, n)
```
bryanchriswhite, line 63:
```suggestion
	events := b.eventsReplayClient.LastNEvents(ctx, n)
```

pkg/client/block/client.go
```diff
@@ -2,222 +2,75 @@ package block
...
-				"unexpected error deserializing block event: %s; eventBz: %s",
-				err, string(eventBz),
-			))
-		}
-		return block, false
-	}
+// Close closes the underlying websocket connection for the EventsQueryClient
+// and closes all subsequent connections.
+func (b *blockClient) Close() {
+	b.mappedClient.Close()
```
bryanchriswhite, line 75:
```suggestion
	b.eventsReplayClient.Close()
```

pkg/client/block/client.go
```diff
@@ -2,222 +2,75 @@ package block
...
-
-		return errCh
+// LatestsNEvents returns the latest n blocks from the BockClient.
+func (b *blockClient) LastNBlocks(ctx context.Context, n int) []client.Block {
+	events := b.mappedClient.LastNEvents(ctx, n)
+	for _, event := range events {
+		// Casting here is safe as this is the generic type of
+		// the EventsReplayClient
+		event = event.(client.Block)
 	}
```
bryanchriswhite, line 68: I don't think this should be necessary any more.

docs/pkg/client/events.md
```diff
@@ -0,0 +1,376 @@
...
+  mechanism.
+- **Dialer Interface**: Offers a `Dialer` interface for constructing connections,
+  which can be easily mocked for tests.
+- **Observable Pattern**: Integrates the observable pattern, making it easier to
+  react to chain events.
+- **Generic Replay Client**: Offers a generic typed replay client to listen for
+  specifc events on chain, and handles reconnection and subscription on error,
+  if the `EventsQueryClient` returns an error/is closed.
+
+## Usage (`EventsQueryClient`)
```
bryanchriswhite, line 203: I like how you've added structure to make room here. :sunglasses: 

Wdyt about adding links to the godoc? They will eventually be updated after this is merged. In the meantime you can use the `make go_docs` target to double-check the URL path.

docs/pkg/client/events.md
```diff
@@ -0,0 +1,376 @@
...
+- **WithDialer**: Configure the client to use a custom dialer for connections.
+
+## Usage (`EventsReplayClient`)
+
+### Basic Usage
+
+```go
+// Define a query string to provide to the EventsQueryClient
+// See: https://docs.cosmos.network/main/learn/advanced/events#subscribing-to-events
+// And: https://docs.cosmos.network/main/learn/advanced/events#default-events
```
bryanchriswhite, line 265: I had linked to the v0.47 page specifically as some things around events have changed in v0.50 IIRC. I find myself wanting to also suggest adding a TODO to update this (and the analogous comment in `EventsQueryClient`) but I don't think there's a good place for that in the docs. I've started #240 as a way to capture things like this (and I've already added this one).

```suggestion
// See: https://docs.cosmos.network/v0.47/learn/advanced/events#subscribing-to-events
// And: https://docs.cosmos.network/v0.47/learn/advanced/events#default-events
```

pkg/client/tx/client.go
bryanchriswhite, line 1: Let's add a  TODO around the `txClient` definition to refactor it in terms of `EventsReplayClient`. I think it can probably be simplified by its usage as well but this PR is already big enough. :sweat_smile:

pkg/client/delegation/client_test.go
```diff
@@ -0,0 +1,100 @@
...
+	eventsQueryClient := testeventsquery.NewAnyTimesEventsBytesEventsQueryClient(
+		ctx, t,
+		delegationEventQuery,
+		expectedEventBz,
+	)
+
+	deps := depinject.Supply(eventsQueryClient)
+
+	// Set up delegation client.
+	delegationClient, err := delegation.NewDelegationClient(ctx, deps, testclient.CometLocalWebsocketURL)
```
bryanchriswhite, line 48: NIT:
```suggestion
	// NB: the URL passed to  `NewDelegationClient` is irrelevant here because `eventsQueryClient` is a mock.
	delegationClient, err := delegation.NewDelegationClient(ctx, deps, testclient.CometLocalWebsocketURL)
```

pkg/client/delegation/client_integration_test.go
```diff
@@ -0,0 +1,214 @@
...
+// 	cfg.GenesisState[gatewaytypes.ModuleName] = buf
+//
+// 	return net
+// }
+
+// TestDelegationClient_DelegateeChangesObservable tests that the DelegationClient
+// can subscribe to the DelegateeChange events and that the events contain
+// the correct AppAddress, it does so by simulating the delegation
+// and undelegation of two applications to a gateway.
+// func TestDelegationClient_DelegateeChangesObservables(t *testing.T) {
```
bryanchriswhite, line 67: You should be able to uncomment most of this and add a `t.SkipNow()` call to prevent it from running:
```suggestion
// TODO_TEST/TODO_TECHDEBT: ...
func TestDelegationClient_DelegateeChangesObservables(t *testing.T) {
	t.SkipNow()
```

docs/pkg/client/events.md
```diff
@@ -0,0 +1,376 @@
...
+flowchart
+
+  a[Component A]
+  b[Component B]
+  c[Component C]
+  d[Component D]
+
+  a --"A uses B via B#MethodName()"--> b
+  a =="A returns C from A#MethodName()"==> c
+  b -."A uses D via network IO".-> d
```
bryanchriswhite, line 63:
```suggestion
  b -."B uses D via network IO".-> d
```

docs/pkg/client/events.md
```diff
@@ -0,0 +1,376 @@
...
+  q1_dial =="#DialContext()"==> q1_conn
+  q1_eb --"#Receive()"--> q1_conn
+
+  q1_conn -.-> sub
+```
+
+```mermaid
+---
+title: EventsReplayClient Components
+---
```
bryanchriswhite, line 102: I'm still thinking about this flowchart but the first thing I notice is that it's upside down; take a look at the legend one more time.

bryanchriswhite: I'm noticing that `EventsQueryClient` is described here in terms of inert sub-components. Whereas, in `EventsQueryClient Components`, it's (already) described using sub-components which actually associate with others.

bryanchriswhite: `Publish Events Factory` seems like it should be named `retryPublishEventsFactory`. Additionally, it's a a function represented as a rectangle, but rectangles represent components in the legend (and rest of the diagram). 

In other places I've used a rounded rectangle to represent functions or methods. Regardless of how that "kind" of thing is represented, it should be:

1. Distinct from other "kinds" which are already defined (for other purposes; i.e. component, usage associations, etc.).
2. Used consistently for that "kind", currently `retry.OnError` is depicted as a usage association. It also needs to change and should be represented the same way as `retryPublishEventsFactory` (assuming  both are still represented in the next iteration).
3. Sufficiently described in the legend to be unambiguous (i.e. hard to misinterpret).

bryanchriswhite: FWIW the legend also doesn't currently define rules around encapsulation. That could be taken as an opportunity either to add some and express them in here, or to reconsider representing encapsulation so strictly (e.g. `EventsQueryClient`/`EventsReplayClient`). I don't have a strong opinion here but wanted to share some additional perspective.

docs/pkg/client/events.md
```diff
@@ -0,0 +1,376 @@
...
+### Basic Usage
+
+```go
+// Define a query string to provide to the EventsQueryClient
+// See: https://docs.cosmos.network/main/learn/advanced/events#subscribing-to-events
+// And: https://docs.cosmos.network/main/learn/advanced/events#default-events
+const eventQueryString = "tm.event='Tx' AND message.action='eventType'"
+
+// Define the websocket URL the EventsQueryClient will subscribe to
+const cometWebsocketURL = "ws://example.com:36657"
```
bryanchriswhite, line 269:
```suggestion
const cometWebsocketURL = "ws://example.com:36657/websocket"
```

docs/pkg/client/events.md
```diff
@@ -0,0 +1,376 @@
...
+
+// Define an interface to represent the onchain event
+type EventType interface {
+  GetName() string
```
bryanchriswhite, line 273:
```suggestion
  GetName() string // illustrative only; arbitrary interfaces are supported
```

docs/pkg/client/events.md
```diff
@@ -0,0 +1,376 @@
...
+  }
+
+  // Confirm the event is correct by checking its fields
+  if eventMsg.Name == "" {
+    return nil, events.ErrEventsUnmarshalEvent.
+      Wrapf("unable to unmarshal eventType: %s", string(eventBz))
+  }
+
+  return eventMsg, nil
+}
```
bryanchriswhite, line 299: Wdyt about making this example use a factory function as that's a pattern that may prove useful in many consumer contexts? We can use a `context.Context` in this example but perhaps there are other and/or more idiomatic factory function parameters.

```suggestion
func eventTypeFactory(ctx context.Context) func ([]byte) EventType {
  return function(eventBz []byte) EventType {  eventMsg := new(eventType)
    logger := polylog.Ctx(ctx)
  
    if err := json.Unmarshal(eventBz, eventMsg); err != nil {
      return nil, err
    }

    // Confirm the event is correct by checking its fields
    if eventMsg.Name == "" {
      return nil, events.ErrEventsUnmarshalEvent.
        Wrapf("unable to unmarshal eventType: %s", string(eventBz))
    }

    return eventMsg, nil
  }
}
```

bryanchriswhite: Perhaps it would also be worth defining a `NewEventFn = func([]byte) EventType` generic (see suggestion above for usage).

docs/pkg/client/events.md
bryanchriswhite: I'm wondering if it makes sense for either `retry.OnError` or `retryPublishEventsFactory`  to be in this diagram. This diagram should be about the relationship betwen the components that are necessary for this unit to function. 

I think it makes more sense for `retry.OnError` and `retryPublishEventsFactory` to be central in some other diagram about the retry loop; perhaps a flowchart or state diagram showing the ways in which the client can get disconnected, how it recovers, and ultimately aborts and errors. There should be an analogous diagram for the `EventsQueryClient` as it does a similar thing IIRC. Perhaps they can be represented in the same diagram, but it's okay if not.

pkg/client/events/errors.go
```diff
@@ -0,0 +1,14 @@
...
+	sdkerrors "cosmossdk.io/errors"
+)
+
+var (
+	codespace = "events"
+
+	ErrDial                 = sdkerrors.Register(codespace, 1, "dialing for connection failed")
+	ErrConnClosed           = sdkerrors.Register(codespace, 2, "connection closed")
+	ErrSubscribe            = sdkerrors.Register(codespace, 3, "failed to subscribe to events")
+	ErrEventsUnmarshalEvent = sdkerrors.Register(codespace, 4, "failed to unmarshal event bytes")
```
bryanchriswhite, line 13: Again this is my fault, but wdyt about renaming these to be more consistent with the naming structure our on-chain errors?

docs/pkg/client/events.md
```diff
@@ -0,0 +1,376 @@
...
+func newEventType(eventBz []byte) EventType {
+  eventMsg := new(eventType)
+  if err := json.Unmarshal(eventBz, eventMsg); err != nil {
+    return nil, err
+  }
+
+  // Confirm the event is correct by checking its fields
+  if eventMsg.Name == "" {
+    return nil, events.ErrEventsUnmarshalEvent.
+      Wrapf("unable to unmarshal eventType: %s", string(eventBz))
```
bryanchriswhite, line 295: Wdyt about something like this?

```suggestion
      Wrapf("with eventType data: %s", string(eventBz))
```

docs/pkg/client/events.md
```diff
@@ -0,0 +1,376 @@
...
+)
+if err != nil {
+  return nil, fmt.Errorf("unable to create EventsReplayClient %w", err)
+}
+
+// Retrieve the latest emitted event
+lastEventType := client.LastNEvents(ctx, 1)[0]
+
+// Get the latest replay observable
+latestEventsObsvbl := client.EventsSequence(ctx)
```
bryanchriswhite, line 328: NIT: 

```suggestion
latestEventsObs := client.EventsSequence(ctx)
```

docs/pkg/client/events.md
```diff
@@ -0,0 +1,376 @@
...
+// Get the latest events from the sequence
+lastEventType := latestEventsObsvbl.Last(ctx, 1)[0]
```
bryanchriswhite, line 330:
```suggestion
latestEventsObs := client.EventsSequence(ctx)
// Get the latest events from the sequence
lastEventType := latestEventsObs.Last(ctx, 1)[0]
```

pkg/client/block/block.go
```diff
@@ -37,7 +38,8 @@ func newCometBlockEvent(blockMsgBz []byte) (client.Block, error) {
 
 	// If msg does not match the expected format then the block's height has a zero value.
 	if blockMsg.Block.Header.Height == 0 {
-		return nil, ErrUnmarshalBlockEvent.Wrap(string(blockMsgBz))
+		return nil, events.ErrEventsUnmarshalEvent.
+			Wrapf("unable to unmarshal block: %s", string(blockMsgBz))
```
bryanchriswhite, line 42: The error message is a bit redundant as `ErrEventsUnmarshalEvent` contains `"failed to unmarshal event bytes"` (similar to https://github.com/pokt-network/poktroll/pull/220#discussion_r1414581768).

```suggestion
		return nil, events.ErrEventsUnmarshalEvent.
			Wrapf("with block data: %s", string(blockMsgBz))
```

docs/pkg/client/events.md
bryanchriswhite: I don't think the `#EventBytes` association is accurately represented as network I/O.

bryanchriswhite: > I'm wondering if it makes sense for either `retry.OnError` or `retryPublishEventsFactory` to be in this diagram. This diagram should be about the relationship betwen the components that are necessary for this unit to function.

Similar feedback on `DecoderFunction`. It probably make sense to have a diagram that focuses on data transformation through `EventsReplayClient`, which `DecoderFunction` would be a critical part of. Also, is there a name that's more consistent with the code or have I just not read that part yet?

pkg/client/events/replay_client.go
```diff
@@ -0,0 +1,244 @@
...
+// bytes into the type defined by the EventsReplayClient's generic type parameter.
+//
+// Required dependencies:
+//   - client.EventsQueryClient
+func NewEventsReplayClient[T any, U observable.ReplayObservable[T]](
+	ctx context.Context,
+	deps depinject.Config,
+	cometWebsocketURL string,
+	queryString string,
+	eventBytesToTypeDecoder func([]byte) (T, error),
```
bryanchriswhite, line 88: Wdyt about `newEventFn` (also see https://github.com/pokt-network/poktroll/pull/220#discussion_r1414568543)

pkg/client/events/replay_client.go
```diff
@@ -0,0 +1,244 @@
...
+	// message bytes into the type defined by the EventsReplayClient's generic type
+	// parameter.
+	eventBytesToTypeDecoder func([]byte) (T, error)
+	// latestObsvbls is a replay observable with replay buffer size 1,
+	// which holds the "active latest observable" which is notified when
+	// new events are received by the events query client subscription
+	// created in goPublishEvents. This observable (and the one it emits) closes
+	// when the events bytes observable returns an error and is updated with a
+	// new "active" observable after a new events query subscription is created.
+	latestObsvbls observable.ReplayObservable[U]
```
bryanchriswhite, line 69: Wdyt about renaming this to `latestObs`. Open to better suggestions but I think the consistency would be an improvement.

pkg/client/events/replay_client.go
```diff
@@ -0,0 +1,244 @@
...
+	// latestObsvblsReplayPublishCh is the publish channel for latestObsvbls.
+	// It's used to set latestObsvbls initially and subsequently update it, for
+	// example, when the connection is re-established after erroring.
+	latestObsvblsReplayPublishCh chan<- U
```
bryanchriswhite, line 73: Update this as well if we rename `latestObsvbls`.

pkg/client/block/godoc.go
```diff
@@ -0,0 +1,5 @@
+// Package block contains the client implementation of the EventsReplayClient
```
bryanchriswhite, line 1: This statement is no longer true as the `EventsReplayClient` is encapsulated now.

pkg/client/delegation/client.go
```diff
@@ -0,0 +1,71 @@
...
+
+// NewDelegationClient creates a new delegation client from the given
+// dependencies and cometWebsocketURL. It uses the defined delegationEventQuery
+// to subscribe to new delegation events and maps them to DelegateeChange
+// objects, using the newDelegateeChangeEvent function as the mapping function.
+//
+// This is an implementation of the EventsReplayClient[DelegateeChange] generic
+//
+// Required dependencies:
+//   - client.EventsQueryClient
```
bryanchriswhite, line 26: :100:

pkg/client/delegation/client.go
```diff
@@ -0,0 +1,71 @@
...
+	)
+	if err != nil {
+		return nil, err
+	}
+	return &delegationClient{eventsReplayClient: client}, nil
+}
+
+// delegationClient is a wrapper around a mapped client that implements the same
+// interface for use in network. This is due to the lack of support from
+// gomock for generic types.
```
bryanchriswhite, line 50: #PUC. I don't think we need to mention gomock at this level.

pkg/client/delegation/delegate.go
```diff
@@ -0,0 +1,39 @@
...
+
+	"github.com/pokt-network/poktroll/pkg/client"
+	"github.com/pokt-network/poktroll/pkg/client/events"
+)
+
+var _ client.DelegateeChange = (*delegate)(nil)
+
+// delegate wraps the EventDelegateeChange event emitted by the application
+// module, for use in the observable
+type delegate struct {
```
bryanchriswhite, line 14: Should this be `delegatee`?

pkg/client/events/replay_client.go
```diff
@@ -0,0 +1,244 @@
...
+		eventsBzObsvbl, err := rClient.eventsClient.EventsBytes(ctx, rClient.queryString)
+		if err != nil {
+			errCh <- err
+			return errCh
+		}
+
+		// NB: must cast back to generic observable type to use with Map.
+		// client.BlocksObservable cannot be an alias due to gomock's lack of
+		// support for generic types.
+		eventsBz := observable.Observable[either.Either[[]byte]](eventsBzObsvbl)
```
bryanchriswhite, line 181:
```suggestion
		eventsBzObs, err := rClient.eventsClient.EventsBytes(ctx, rClient.queryString)
		if err != nil {
			errCh <- err
			return errCh
		}

		// NB: must cast back to generic observable type to use with Map.
		// client.BlocksObservable cannot be an alias due to gomock's lack of
		// support for generic types.
		eventsBz := observable.Observable[either.Either[[]byte]](eventsBzObs)
```

pkg/client/events/replay_client.go
```diff
@@ -0,0 +1,244 @@
...
+		typedObsrvbl := channel.MapReplay(
+			ctx,
+			latestReplayBufferSize,
+			eventsBz,
+			rClient.newEventsBytesToTypeMapFn(errCh),
+		)
+
+		// Initially set latestObsvbls and update if after retrying on error.
+		rClient.latestObsvblsReplayPublishCh <- typedObsrvbl.(R)
```
bryanchriswhite, line 190:
```suggestion
		typedObs := channel.MapReplay(
			ctx,
			latestReplayBufferSize,
			eventsBz,
			rClient.newEventsBytesToTypeMapFn(errCh),
		)

		// Initially set latestObs and update if after retrying on error.
		rClient.latestObsvblsReplayPublishCh <- typedObs.(R)
```

Makefile
```diff
@@ -516,7 +516,7 @@ trigger_ci: ## Trigger the CI pipeline by submitting an empty commit; See https:
 
 .PHONY: go_docs
 go_docs: check_godoc ## Generate documentation for the project
-	echo "Visit http://localhost:6060/pkg/pocket/"
+	echo "Visit http://localhost:6060/pkg/github.com/pokt-network/poktroll/"
```
bryanchriswhite, line 519: :sunglasses:

pkg/client/events/replay_client.go
```diff
@@ -0,0 +1,244 @@
...
+//
+// The map function itself attempts to deserialize the given byte slice as a
+// the EventsReplayClient's generic typed event, using the decoder function provided.
+// If the events bytes observable contained an error, this value is not emitted
+// (skipped) on the destination observable of the map operation.
+//
+// If deserialisation failed because the event bytes were for a different event
+// type, this value is also skipped. If deserialisation failed for some other
+// reason, this function panics.
+func (rClient *replayClient[T, R]) newEventsBytesToTypeMapFn(errCh chan<- error) func(
```
bryanchriswhite, line 210: Thus far, I've been naming functions that are intended to be used with `channel.Map()` as `mapXXX`. Wdyt about something like this?:

```suggestion
// newMapEventsBytesToTFn is a factory for a function which is intended
// to be used as a transformFn in a channel.Map() call. Since the map function
// is called asynchronously, this factory creates a closure around an error
// channel which can be used for asynchronous error signaling from within the
// map function, and handling from the Map call context.
//
// The map function itself attempts to deserialize the given byte slice as a
// the EventsReplayClient's generic typed event, using the decoder function provided.
// If the events bytes observable contained an error, this value is not emitted
// (skipped) on the destination observable of the map operation.
//
// If deserialisation failed because the event bytes were for a different event
// type, this value is also skipped. If deserialisation failed for some other
// reason, this function panics.
func (rClient *replayClient[T, R]) newMapEventsBytesToTFn(errCh chan<- error) func(
```

pkg/client/events/websocket/connection.go
```diff
@@ -18,7 +18,7 @@ type websocketConn struct {
 func (wsConn *websocketConn) Receive() ([]byte, error) {
 	_, msg, err := wsConn.conn.ReadMessage()
 	if err != nil {
-		return nil, ErrReceive.Wrapf("%s", err)
+		return nil, ErrEventsWebsocketReceive.Wrapf("%s", err)
```
bryanchriswhite, line 21: :100:

proto/pocket/application/event.proto
```diff
@@ -0,0 +1,13 @@
...
+
+option go_package = "github.com/pokt-network/poktroll/x/application/types";
+
+import "cosmos_proto/cosmos.proto";
+
+// EventDelegateeChange is an event emitted whenever an application changes its
+// delegatee gateways on chain. This is in response to both a DelegateToGateway
+// and UndelegateFromGateway message.
+message EventDelegateeChange {
+  string app_address = 1 [(cosmos_proto.scalar) = "cosmos.AddressString"]; // The Bech32 address of the application using cosmos' ScalarDescriptor to ensure deterministic encoding
```
Olshansk, line 12: Let's add `gateway_address` or `delegatee_address` here too.

pkg/client/events/websocket/errors.go
```diff
@@ -0,0 +1,11 @@
+package websocket
+
+import (
+	sdkerrors "cosmossdk.io/errors"
+)
+
+var (
+	codespace = "events_query_client_websocket_connection"
+
+	ErrEventsWebsocketReceive = sdkerrors.Register(codespace, 1, "failed to receive event")
```
Olshansk, line 10:
```suggestion
	ErrEventsWebsocketReceive = sdkerrors.Register(codespace, 1, "failed to receive event over websocket connection to pocket node")
```

pkg/client/delegation/client_integration_test.go
```diff
@@ -0,0 +1,190 @@
...
+	"context"
+	"sync"
+	"testing"
+	"time"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/stretchr/testify/require"
+
+	"github.com/pokt-network/poktroll/pkg/client"
+	keepertest "github.com/pokt-network/poktroll/testutil/keeper"
```
Olshansk, line 24: One of the reasons I believe you're struggling with this test is because we're trying to use `keeper` outside of on-chain code. Seems to be the first instance of it:

Screenshot 2023-12-04 at 7 06 48 PM

If you look in `x/supplier/client/cli/query_claim_test.go`, what we do is:
1. Create an in-memory network
2. Use the test CLI to create claim objects

@bryanchriswhite is reworking that code in #236 but the idea is the same and what I think you want instead of using the keepers directly.

testutil/network/network.go
```diff
@@ -199,7 +200,48 @@ func SupplierModuleGenesisStateWithAccounts(t *testing.T, addresses []string) *s
 	return state
 }
 
-// Initialize an Account by sending it some funds from the validator in the network to the address provided
+// ApplicationModuleGenesisStateWithAccounts generates a GenesisState object with
+// an application list full of applications with the given addresses.
+func ApplicationModuleGenesisStateWithAccounts(t *testing.T, addresses []string) *apptypes.GenesisState {
```
Olshansk, line 205: We already have `DefaultApplicationModuleGenesisState`, so I don't think copy-pasting that is correct.

Why not consolidate the two?

testutil/network/network.go
```diff
@@ -199,7 +200,48 @@ func SupplierModuleGenesisStateWithAccounts(t *testing.T, addresses []string) *s
...
+			},
+		}
+		state.ApplicationList = append(state.ApplicationList, application)
+	}
+	return state
+}
+
+// GatewayModuleGenesisStateWithAccounts generates a GenesisState object with
+// a gateway list full of gateways with the given addresses.
+func GatewayModuleGenesisStateWithAccounts(t *testing.T, addresses []string) *gatewaytypes.GenesisState {
```
Olshansk, line 229: Ditto for DefaultGatewayModuleGenesisState

Olshansk (dismissed):
Not a complete review, but leaving some comments in the interim.

1. Really good PR overall with lots of good comments, good code, great functionality, etc...

2. Left some non-actionable comments throughout. Feel free to treat them as FYI and resolve

3. I think there is a big opportunity for smaller PRs when we hit something like this in the future. For example, this was titled `off-chain` but you ended up introducing `on-chain` components. Even though the ordering of implementation may have led to this, we could've split out on-chain vs off-chain, the `EventsReplayClient` into its own thing, the actual business logic, etc... I think it could be 2-4 modular PRs that are easier to review in isolation.

4. The majority of the comments in this document are great but sometimes the comment just describes the actual code line underneath which isn't useful.

As an extreme example: `const n = 5 is a constant with the name n equaling to 5`
	 
In the particularly hard to understand parts (e.g. observable of observables observing observables), explaining `why` would go a long way!



Screenshot 2023-12-04 at 7 14 08 PM

pkg/client/block/block.go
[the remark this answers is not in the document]
bryanchriswhite: :+1: I would even suggest omitting the reference to "unmarshaling" because it's already expressed in the error description.

pkg/client/events/websocket/godoc.go
```diff
@@ -0,0 +1,4 @@
+// Package websocket provides a websocket client used to connect to a node
```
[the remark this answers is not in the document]
bryanchriswhite: I think we can broaden the constraint, at this low level, to be cosmos-sdk based chains/networks/nodes. Wdyt?

pkg/client/events/replay_client.go
bryanchriswhite: (cc @Olshansk)

pkg/client/block/client.go
```diff
@@ -83,141 +30,41 @@ func NewBlockClient(
...
+	eventsReplayClient client.EventsReplayClient[client.Block, client.EventsObservable[client.Block]]
 }
 
-// Close unsubscribes all observers of the committed blocks sequence observable
-// and closes the events query client.
-func (bClient *blockClient) Close() {
-	// Closing eventsClient will cascade unsubscribe and close downstream observers.
-	bClient.eventsClient.Close()
+// CommittedBlocksSequence returns a replay observable of observables for Block events
+// from the BlockClient.
```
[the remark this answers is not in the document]
bryanchriswhite:
```
// CommittedBlocksSequence returns a replay observable of Block new block events.
```

pkg/client/delegation/client.go
```diff
@@ -0,0 +1,71 @@
...
+type delegationClient struct {
+	eventsReplayClient client.EventsReplayClient[client.DelegateeChange, client.EventsObservable[client.DelegateeChange]]
```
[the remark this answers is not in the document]
bryanchriswhite: Tbh that seems totally fair, this specific example involves quite high-level abstractions/objects. A future rename may go a long way but in the meantime, this indirection is the trade-off for:

1. Smaller step sizes between components (e.g. `EventsQueryClient` --> `EventsReplayClient` --> `BlockClient`).
2. Delaying migration away from gomock.

pkg/client/interface.go
```diff
@@ -98,33 +99,70 @@ type TxContext interface {
...
 // Block is an interface which abstracts the details of a block to its minimal
 // necessary components.
 type Block interface {
 	Height() int64
 	Hash() []byte
 }
 
+// DelegateeChange is an interface which wraps the EventDelegateeChange event
+// emitted by the application module.
+type DelegateeChange interface {
```
bryanchriswhite, line 111: Wdyt about renaming this (see https://github.com/pokt-network/poktroll/pull/239/files#r1415897609)?

```suggestion
type Redelegation interface {
```

pkg/client/interface.go
```diff
@@ -98,33 +99,70 @@ type TxContext interface {
...
+	EventsSequence(context.Context) R
+	// LastNEvents returns the latest N events that has been received.
+	LastNEvents(ctx context.Context, n int) []T
+	// Close unsubscribes all observers of the events sequence observable
+	// and closes the events query client.
+	Close()
+}
+
+// BlockReplayObservable wraps the generic
+// observable.ReplayObservable[Block] type
```
bryanchriswhite, line 131:
```suggestion
// BlockReplayObservable is a defined type which is a replay observable of type Block.
```

bryanchriswhite: Rataionale behind "defined type": https://chat.openai.com/share/8f5daa07-dc01-4434-a978-c728923f6e78

pkg/client/events/replay_client.go
```diff
@@ -0,0 +1,244 @@
...
+	// latestObsvblsReplayBufferSize is the replay buffer size of the
+	// latestObsvbls replay observable which is used to cache the latest observable.
+	// It is updated with a new "active" observable when a new
+	// events query subscription is created, for example, after a non-persistent
+	// connection error.
+	latestObsvblsReplayBufferSize = 1
+	// latestReplayBufferSize is the replay buffer size of the latest replay
+	// observable which is notified when block commit events are received
+	// by the events query client subscription created in goPublishEvents.
+	latestReplayBufferSize = 1
```
bryanchriswhite, line 35:
```suggestion
	latestObsvblsReplayBufferSize = 1
	// replayObsBufferSize is the replay buffer size of the latest replay
	// observable which is notified when events are received by the 
	// underlying `EventsQueryClient`.
	replayObsBufferSize = 1
```

bryanchriswhite, line 35: Wdyt about something like this?

```suggestion
	// replayObsCacheBufferSize is the replay buffer size of the
	// replayObsCache replay observable which is used to cache the replay
	// observable that is notified of new events.
	// It, replayObsCache, is updated with a new "active" observable when a new
	// events query subscription is created, for example, after a non-persistent
	// connection error.
	replayObsCacheBufferSize = 1
	// replayObsBufferSize is the replay buffer size of the replay observable
	// which is notified of new events.
	replayObsBufferSize = 1
```

pkg/client/events/replay_client.go
bryanchriswhite: See https://github.com/pokt-network/poktroll/pull/220#discussion_r1415968048

bryanchriswhite, line 69:
```suggestion
	// replayObsCache is a replay observable with replay buffer size 1,
	// which holds the "active replay observable" which is notified of new
	// events. The "active replay observable" closes in the event of an upstream
	// error, and a new "active" observable is published to replayObsCache after
	//  a new events query subscription is created.
	replayObs observable.ReplayObservable[U]
```

pkg/client/events/replay_client.go
```diff
@@ -0,0 +1,244 @@
...
+		queryString:             queryString,
+		eventBytesToTypeDecoder: eventBytesToTypeDecoder,
+	}
+	latestObsvbls,
+		latestObsvblsReplayPublishCh := channel.NewReplayObservable[U](
+		ctx,
+		latestReplayBufferSize,
+	)
+	rClient.latestObsvbls = observable.ReplayObservable[U](latestObsvbls)
+	rClient.latestObsvblsReplayPublishCh = latestObsvblsReplayPublishCh
```
bryanchriswhite, line 102:
```suggestion
	replayObsCache,
		replayObsCachePublishCh := channel.NewReplayObservable[U](
		ctx,
		latestReplayBufferSize,
	)
	rClient.replayObsCache = observable.ReplayObservable[U](replayObsCache)
	rClient.replayObsPublishCh = replayObsPublishCh
```

pkg/client/events/replay_client.go
```diff
@@ -0,0 +1,244 @@
...
+	// Initialise the mapped client
+	rClient := &replayClient[T, U]{
+		endpointURL:             cometWebsocketURL,
+		queryString:             queryString,
+		eventBytesToTypeDecoder: eventBytesToTypeDecoder,
+	}
+	latestObsvbls,
+		latestObsvblsReplayPublishCh := channel.NewReplayObservable[U](
+		ctx,
+		latestReplayBufferSize,
```
bryanchriswhite, line 99: I think this should be received as a parameter and should not have a default (i.e. I think we should remove this const).

pkg/client/events/replay_client.go
```diff
@@ -0,0 +1,244 @@
...
+}
+
+// EventsSequence returns a ReplayObservable, with a replay buffer size of 1,
+// which is notified when new events are received by the events query subscription.
+func (rClient *replayClient[T, R]) EventsSequence(ctx context.Context) R {
+	// Get the latest events observable from the replay observable. We only ever
+	// want the last 1 as any prior latest events observable values are closed.
+	// Directly accessing the zeroth index here is safe because the call to Last
+	// is guaranteed to return a slice with at least 1 element.
+	replayObs := observable.ReplayObservable[R](rClient.latestObsvbls)
```
bryanchriswhite, line 122:
```suggestion
	replayObs := observable.ReplayObservable[R](rClient.replayObsCache)
```

pkg/client/events/replay_client.go
```diff
@@ -0,0 +1,244 @@
...
+		return nil, err
+	}
+
+	// Concurrently publish blocks to the observable emitted by latestObsvbls.
+	go rClient.goPublishEvents(ctx)
+
+	return rClient, nil
+}
+
+// EventsSequence returns a ReplayObservable, with a replay buffer size of 1,
```
bryanchriswhite, line 115: #PUC if we make `replayObsBufferSize` a parameter of the constructor.

pkg/client/events/replay_client.go
```diff
@@ -0,0 +1,244 @@
...
+// which is notified when new events are received by the events query subscription.
```
bryanchriswhite, line 116:
```suggestion
// which is notified when new events are received by the encapsulated EventsQueryClient.
```

pkg/client/events/replay_client.go
```diff
@@ -0,0 +1,244 @@
...
+func (rClient *replayClient[T, R]) EventsSequence(ctx context.Context) R {
+	// Get the latest events observable from the replay observable. We only ever
+	// want the last 1 as any prior latest events observable values are closed.
+	// Directly accessing the zeroth index here is safe because the call to Last
+	// is guaranteed to return a slice with at least 1 element.
```
bryanchriswhite, line 121:
```suggestion
	// Get the active event replay observable from replayObsCache. Only the last element
	// is useful as any prior elements are closed replay observables.
	// Directly accessing the zeroth index here is safe because the call to Last
	// is guaranteed to return a slice with at least 1 element.
```

pkg/client/delegation/client.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+// DelegateeChangesSequence returns a replay observable of observables for
+// delegation events from the DelegationClient.
+func (b *delegationClient) DelegateeChangesSequence(ctx context.Context) client.DelegateeChangeReplayObservable {
```
bryanchriswhite, line 57: I'm imagining something like this:

```suggestion
// RedelegationsSequence returns a replay observable of Redelgation events
// observed by the  DelegationClient.
func (b *delegationClient) RedelegationssSequence(ctx context.Context) client.RedelegationsReplayObservable {
```

pkg/client/block/godoc.go
h5law: WDYT of the update? @Olshansk @bryanchriswhite

pkg/client/delegation/client_integration_test.go
h5law: @Olshansk https://arc.net/l/quote/npxdzlgu this approach was in progress and I will go back to it after your comment

Olshansk:
@h5law Leaving a comment just to avoid miscommunication. Re-request a review once this is ready for another look.

proto/pocket/application/event.proto
h5law: No we only need the app address there is no need for any other fields, adds bloat we dont use

testutil/network/network.go
h5law: Remnants from my old test strategy
removing

x/application/types/message_delegate_to_gateway.go
```diff
@@ -37,6 +37,12 @@ func (msg *MsgDelegateToGateway) GetSignBytes() []byte {
 	return sdk.MustSortJSON(bz)
 }
 
+func (msg *MsgDelegateToGateway) NewDelegateeChangeEvent() *EventDelegateeChange {
+	return &EventDelegateeChange{
+		AppAddress: msg.AppAddress,
```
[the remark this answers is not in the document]
h5law: Am against this as it is irrelevent to its usage

x/application/types/message_undelegate_from_gateway.go
```diff
@@ -37,6 +37,12 @@ func (msg *MsgUndelegateFromGateway) GetSignBytes() []byte {
 	return sdk.MustSortJSON(bz)
 }
 
+func (msg *MsgUndelegateFromGateway) NewDelegateeChangeEvent() *EventDelegateeChange {
+	return &EventDelegateeChange{
+		AppAddress: msg.AppAddress,
```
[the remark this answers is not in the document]
h5law: See other comments on this

pkg/client/delegation/client_integration_test.go
h5law: Tried and failed left comment, think its something to do with the networks subscribiton to events unsure atm.

docs/pkg/client/events.md
```diff
@@ -0,0 +1,391 @@
...
+  r1_ob --"#EventsSequence"--> c1_ob
+  r1_ob --"#LastNEvents"--> c1_ev
+```
+
+### Subscriptions
+
+```mermaid
+---
+title: Event Subscription Data Flow
+---
```
bryanchriswhite, line 145: This diagram needs one of:

A. Its own legend, in which the usage via network I/O association direction matches the diagram
B. To be rewritten in terms of the legend provided in the previous section (i.e. turn it upside-down)
C. A note or some other indicator that the legend in the section above only applies to that section and/or does not apply to this section.

I think A is ideal but C would be sufficient for now. I think this diagram reads better as written (as opposed to upside-down). I also think that the component diagrams above read best  they way they are (i.e. shouldn't be turned upside-down either).

docs/pkg/client/events.md
```diff
@@ -0,0 +1,391 @@
...
+  // See: https://docs.cosmos.network/v0.47/learn/advanced/events#subscribing-to-events
+  // And: https://docs.cosmos.network/v0.47/learn/advanced/events#default-events
+  eventQueryString = "message.action='eventName'"
+  // Define the websocket URL the EventsQueryClient will subscribe to
+  cometWebsocketURL = "ws://example.com:36657/websocket"
+	// the amount of events we want before they are emitted
+  replayObsBufferSize = 1
+)
+
+// Define an interface to represent the onchain event
```
bryanchriswhite, line 278:
```suggestion
// Define an interface to represent an arbitrary onchain event
```

pkg/client/block/block.go
```diff
@@ -26,19 +28,23 @@ func (blockEvent *cometBlockEvent) Hash() []byte {
...
-// newCometBlockEvent attempts to deserialize the given bytes into a comet block.
+// newCometBlockEventFactoryFn is a factory function that returns a functon
+// that attempts to deserialize the given bytes into a comet block.
 // if the resulting block has a height of zero, assume the event was not a block
 // event and return an ErrUnmarshalBlockEvent error.
-func newCometBlockEvent(blockMsgBz []byte) (client.Block, error) {
-	blockMsg := new(cometBlockEvent)
-	if err := json.Unmarshal(blockMsgBz, blockMsg); err != nil {
-		return nil, err
+func newCometBlockEventFactoryFn(ctx context.Context) events.NewEventsFn[client.Block] {
```
bryanchriswhite, line 35: Why does this function need to close over the `ctx`? It doesn't look like the returned function uses it at all. :thinking:

pkg/client/block/client.go
```diff
@@ -2,79 +2,28 @@ package block
...
-	// latestBlockObsvbls replay observable which is used to cache the latest block observable.
-	// It is updated with a new "active" observable when a new
-	// events query subscription is created, for example, after a non-persistent
-	// connection error.
-	latestBlockObsvblsReplayBufferSize = 1
-	// latestBlockReplayBufferSize is the replay buffer size of the latest block
-	// replay observable which is notified when block commit events are received
-	// by the events query client subscription created in goPublishBlocks.
-	latestBlockReplayBufferSize = 1
+	replayObsBufferSize  = 1 // the amount of events we want before they are emitted
```
bryanchriswhite, line 18: Now that the block client exposes `#LastNBlocks()`  instead of `#LatestBlock()`, I think it's prudent that we increase this. I think we should also rename it to `defaultReplayObsBufferSize` and add a TODO: 

```suggestion
	// TODO_TECHDEBT/TODO_FUTURE: add a `blocksLimit` field to the block client struct that defaults to this but can be overridden via an option in future work.
	defaultBlocksLimit  = 1 // the number of blocks that the replay observable returned by LastNBlocks() will be able to replay.
```

pkg/client/block/client.go
```diff
@@ -83,141 +32,46 @@ func NewBlockClient(
...
-			eventsBz,
-			newEventsBytesToBlockMapFn(errCh),
-		)
-
-		// Initially set latestBlockObsvbls and update if after retrying on error.
-		bClient.latestBlockObsvblsReplayPublishCh <- blocksObsvbl
-
-		return errCh
-	}
+// LatestsNEvents returns the latest n blocks from the BockClient.
```
bryanchriswhite, line 68:
```suggestion
// LatestsNBlocks returns the last n blocks observed by the BockClient.
```

pkg/client/block/godoc.go
```diff
@@ -0,0 +1,5 @@
+// Package block contains a light wrapper of the EventsReplayClient[Block]
+// generic which listens for committed block events on chain and emits them
+// through a ReplayObservable. This enables consumers to listen for on-chain
+// application delegation changes and react to them asynchronously.
```
bryanchriswhite, line 4:
```suggestion
// through a ReplayObservable. This enables consumers to listen for newly
// committed blocks and react to them asynchronously.
```

pkg/client/delegation/client.go
```diff
@@ -0,0 +1,79 @@
...
+	"github.com/pokt-network/poktroll/pkg/client/events"
+)
+
+const (
+	// delegationEventQuery is the query used by the EventsQueryClient to subscribe
+	// to new delegation events from the the application module on chain.
+	// See: https://docs.cosmos.network/v0.47/learn/advanced/events#subscribing-to-events
+	// And: https://docs.cosmos.network/v0.47/learn/advanced/events#default-events
+	delegationEventQuery = "message.action='pocket.application.EventRedelegation'"
+	replayObsBufferSize  = 1 // the amount of events we want before they are emitted
```
bryanchriswhite, line 18: Similar feedback to https://github.com/pokt-network/poktroll/pull/220#discussion_r1417030309.

pkg/client/events/replay_client.go
```diff
@@ -0,0 +1,250 @@
...
+//   - client.EventsQueryClient
+func NewEventsReplayClient[T any, U observable.ReplayObservable[T]](
+	ctx context.Context,
+	deps depinject.Config,
+	cometWebsocketURL string,
+	queryString string,
+	newEventFn NewEventsFn[T],
+	replayObsBufferSize int,
+) (client.EventsReplayClient[T, U], error) {
+	// Initialise the mapped client
```
bryanchriswhite, line 95: NIT:
```suggestion
	// Initialize the mapped client
```

pkg/client/events/replay_client.go
```diff
@@ -0,0 +1,250 @@
...
+		queryString:  queryString,
+		eventDecoder: newEventFn,
+	}
+	latestObsvbls,
+		latestObsvblsReplayPublishCh := channel.NewReplayObservable[U](
+		ctx,
+		replayObsBufferSize,
+	)
+	rClient.replayObsCache = observable.ReplayObservable[U](latestObsvbls)
+	rClient.replayObsCachePublishCh = latestObsvblsReplayPublishCh
```
bryanchriswhite, line 107:
```suggestion
	replayObsCache, replayObsCachePublishCh := channel.NewReplayObservable[U](
		ctx,
		replayObsBufferSize,
	)
	rClient.replayObsCache = observable.ReplayObservable[U](replayObsCache)
	rClient.replayObsCachePublishCh = replayObsCachePublishCh
```

pkg/client/events/websocket/errors.go
```diff
@@ -0,0 +1,11 @@
+package websocket
+
+import (
+	sdkerrors "cosmossdk.io/errors"
+)
+
+var (
+	codespace = "events_query_client_websocket_connection"
+
+	ErrEventsWebsocketReceive = sdkerrors.Register(codespace, 1, "failed to receive event over websocket connection to pocket node")
```
bryanchriswhite, line 10: :sunglasses::pray::fire:

pkg/client/interface.go
```diff
@@ -98,33 +99,69 @@ type TxContext interface {
...
-}
-
 // Block is an interface which abstracts the details of a block to its minimal
 // necessary components.
 type Block interface {
 	Height() int64
 	Hash() []byte
 }
 
+// Redelegation is an interface which wraps the EventRedelegation event
```
bryanchriswhite, line 109: Is  `EventRedeletation` a type somethwhere? It's unclear to what this is referring.

pkg/client/interface.go
```diff
@@ -98,33 +99,69 @@ type TxContext interface {
...
+// emitted by the application module.
+type Redelegation interface {
+	AppAddress() string
+}
+
+// EventsObservable is a replay observable for events of some type T.
```
bryanchriswhite, line 115: It's probably worth adding TODO about how this can't be an alias due to the gomock constraint. I wouldn't go into much detail, just enough to articulate why this is not an alias, which seems like it would be more appriate otherwise.

pkg/client/delegation/delegatee.go
```diff
@@ -0,0 +1,43 @@
...
+func (d redelegation) AppAddress() string {
+	return d.Address
+}
+
+// newRedelegationEventFactoryFn is a factory function that returns a
+// function that attempts to deserialise the given bytes into a redelegation
+// struct. If the delegate struct has an empty app address then an
+// ErrUnmarshalRedelegation error is returned. Otherwise if deserialisation
+// fails then the error is returned.
+func newRedelegationEventFactoryFn(ctx context.Context) events.NewEventsFn[client.Redelegation] {
```
bryanchriswhite, line 29: Same question as https://github.com/pokt-network/poktroll/pull/220#discussion_r1417018681.

testutil/network/network.go
```diff
@@ -199,7 +201,46 @@ func SupplierModuleGenesisStateWithAccounts(t *testing.T, addresses []string) *s
...
+		gateway := gatewaytypes.Gateway{
+			Address: addr,
+			Stake:   &sdk.Coin{Denom: "upokt", Amount: sdk.NewInt(10000)},
+		}
+		state.GatewayList = append(state.GatewayList, gateway)
+	}
+	return state
+}
+
+// InitAccount initialises an Account by sending it some funds from the validator
```
bryanchriswhite, line 242:
```suggestion
// InitAccount initializes an Account by sending it some funds from the validator
```

x/application/keeper/msg_server_delegate_to_gateway.go
```diff
@@ -57,5 +57,11 @@ func (k msgServer) DelegateToGateway(goCtx context.Context, msg *types.MsgDelega
 	k.SetApplication(ctx, app)
 	logger.Info("Successfully delegated application to gateway for app: %+v", app)
 
+	// Emit the application redelegation change event
+	if err := ctx.EventManager().EmitTypedEvent(msg.NewRedelegationEvent()); err != nil {
+		logger.Error("Failed to emit application redelegation event: %v", err)
+		return nil, err
+	}
```
bryanchriswhite, line 64: Nice!

x/application/keeper/msg_server_delegate_to_gateway_test.go
```diff
@@ -242,10 +247,19 @@ func TestMsgServer_DelegateToGateway_FailMaxReached(t *testing.T) {
 		require.True(t, isAppFound)
 		require.Equal(t, int(i+1), len(foundApp.DelegateeGatewayAddresses))
 	}
+	events := ctx.EventManager().Events()
+	require.Equal(t, int(maxDelegatedParam), len(events))
+	for _, event := range events {
+		require.Equal(t, "pocket.application.EventRedelegation", event.Type)
+		require.Equal(t, "app_address", event.Attributes[0].Key)
+		require.Equal(t, fmt.Sprintf("\"%s\"", appAddr), event.Attributes[0].Value)
+	}
```
bryanchriswhite, line 256: :popcorn::sunglasses:

x/application/types/message_delegate_to_gateway.go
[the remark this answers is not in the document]
bryanchriswhite: Perhaps this is true of the current usage. However, imagine building an indexer for this, for example. It would be extremely convenient for the event to include the extra detail of which gateway the given application is (re)delegating to:

1. It disambiguates the different types of the redelegation event (i.e. delegation, redelegation, undelegation).
2. It removes the need to make a query for the related effected on-chain state.

proto/pocket/application/event.proto
bryanchriswhite: I chimed in on the other thread: https://github.com/pokt-network/poktroll/pull/220#discussion_r1417126855

bryanchriswhite (changes requested):
I finally got through it! :sweat_smile::tada: 

This PR is in great shape, especially for its size. I have one more round of comments and then I think it'll be good to go.

I also wanted to call our https://github.com/pokt-network/poktroll/pull/220#discussion_r1414790100 and https://github.com/pokt-network/poktroll/pull/220#discussion_r1414810089 to make sure that we don't loose track of them.

testutil/network/network.go
```diff
@@ -219,7 +260,8 @@ func InitAccount(t *testing.T, net *Network, addr sdk.AccAddress) {
 	require.Equal(t, float64(0), responseJson["code"], "code is not 0 in the response: %v", responseJson)
 }
 
-// Initialize an Account by sending it some funds from the validator in the network to the address provided
+// InitAccountWithSequence initialises an Account by sending it some funds from
```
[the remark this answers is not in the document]
h5law: dont try americanise me

docs/pkg/client/events.md
h5law: Added todos to add legends in the appropriate places

h5law:
@Olshansk @bryanchriswhite Should be good to go now <3

docs/pkg/client/events.md
bryanchriswhite, line 1: It looks like your latest diagram changes (https://github.com/pokt-network/poktroll/pull/220/commits/18f953c6aaed3840043999903ab9e3a8e536eafc?short_path=877a552#diff-877a552200311af4985616701be119cf6adc302755c4421136aabd6b7b847003) got reverted by the merge with main (e53078bf723a231c5e45b81a3d4cc7973857cde2).

bryanchriswhite: I've added a [section in a "Documentation in progress" page](https://www.notion.so/buildwithgrove/Documentation-in-progress-a2c3bd75c7f14606b2c8e9603d4f907f?pvs=4#b8e57f6bf84b44159cb8bce6ab39469a) that I started on notion.

1. We can use this as a (more) collaborative environment than mermaid.live alone. At the very least, it's better than copy/pasting through discord. :sweat_smile: 
2. I've made some changes that I would've added as a suggestion comment in GH if  the changes hadn't been reverted.

bryanchriswhite: I'm also noticing that the todo's don't render in the docs whereas I was expecting them to. Part of the reason I suggested it was to signal to readers that the legends are missing and to avoid the confusion of applying the wrong legend. Two questions:

1. Is this intentional?
2. Do you think we should prefer this (i.e. rationale, either way)?

h5law: I intentionally made them comments but can change to be visible, also unsure what you mean by changes reverted the branch has the latest copy?

docs/pkg/client/events.md
```diff
@@ -0,0 +1,421 @@
...
+    participant Q as EventsQueryClient
+    participant D as EventDecoder
+    Q ->>+ D: EventBytes
+    D ->> D: DecoderFn()
+    D ->>- Q: resubscribeOnError()
+    participant P as Replay Observable Cache Publish Channel
+    participant O as Replay Observable Cache
+    D ->> P: Typed Events
+    P ->> O: Typed Events
+```
```
bryanchriswhite, line 210: Let's shelve this diagram for now (i.e. remove and iterate in [notion](https://www.notion.so/Clients-a24e75fe6f70450890c8e12092d5f8e1?d=653b16f968c34e09bed1fe06cb0e0439&pvs=4#9c938bc617374de0a95d911e1d6d2641))

bryanchriswhite, line 210: Let's shelve this diagram for now (i.e. remove and iterate in [notion](https://www.notion.so/Clients-a24e75fe6f70450890c8e12092d5f8e1?d=653b16f968c34e09bed1fe06cb0e0439&pvs=4#9c938bc617374de0a95d911e1d6d2641)), it's not ready and isn't worth blocking this PR over.

bryanchriswhite (approved):
Preemptively approving but I have one last change request regarding the sequence diagram. 

Thanks for all your patience! :see_no_evil:

docs/pkg/client/events.md
bryanchriswhite: :pray:

pkg/client/delegation/client_integration_test.go
Olshansk: @h5law Wanted to ask to leave these sort of threads open in the future or let the author of the thread resolve it. Even if we don't necessarily tackle the work in this PR, it helps keep track of followup work.

PR #222, opened 2023-11-24
chore: correct links to blog posts in README.md

studna:
Updating links to blog posts as they are resulting with 404 now.


bryanchriswhite (approved):
Thanks @studna! :raised_hands: 

@Olshansk these both check out for me. The first link was indeed broken, the second one redirects to the updated URL. Unless there's a reason we should prefer the keeping the original second URL the same, this LGTM.

PR #206, opened 2023-11-21
[RelayMiner] feat: Add relay miner config file handling

red-0ne:
### Human Summary

Source the RelayMiner configuration from a yaml config file instead of flags

### Summary generated by Reviewpad on 22 Nov 23 08:52 UTC
This pull request introduces several changes across multiple files. 

In the file `proxy.go`, the type `servicesEndpointsMap` has been changed from `map[serviceId]url.URL` to `map[serviceId]*url.URL`. This means that the values in the map are now pointers to `url.URL` instead of the actual `url.URL` objects.

A new file named "errors.go" has been added to the "config" package. It defines error variables related to the RelayMiner configuration, covering cases such as unmarshaling YAML content, invalid URLs, and invalid configuration parameters.

The file `relayminer_config.yaml` has been added, containing various changes:
- `query_node_url` parameter has been added to specify the URL of a full pocket node for reading data and listening for on-chain events.
- `network_node_url` parameter has been added to specify the URL of a pocket node that gossips transactions throughout the network.
- `signing_key_name` parameter has been added to specify the name of the key in the keyring used for signing transactions.
- `proxied_service_endpoints` parameter has been added to specify the endpoints of the services that the relay miner proxies.
- `smt_store_path` parameter has been added to specify the path to the data backing SMT KV store on disk.

A new file called `yaml.go` has been added to the `testutil/yaml` directory. It defines a function `NormalizeYAMLIndentation` that normalizes the indentation of YAML content before parsing it, ensuring consistent indentation regardless of the original style.

The `jsonrpc.go` file in the `pkg/relayer/proxy` directory has been modified. It introduces a type `jsonRPCServer`, a function `NewJSONRPCServer`, and related modifications for implementing the `relayer.RelayServer` interface.

A new unit test file named `relayminer_configs_reader_test.go` has been added to the `pkg/relayer/config` package. It contains a comprehensive test suite for the `ParseRelayMinerConfigs` function, covering various scenarios and ensuring correct parsing and validation of RelayMiner configurations.

Lastly, the implementation for parsing and unmarshaling a RelayMiner configuration file has been added. It introduces structures `YAMLRelayMinerConfig` and `RelayMinerConfig` to represent the configuration in different formats and provides the `ParseRelayMinerConfigs` function for parsing and validating the configuration.

These changes collectively enhance the functionality and maintainability of the codebase for the RelayMiner project.

## Issue

- #205 

## Type of change

Select one or more:

- [x] New feature, functionality or library
- [ ] Bug fix
[...]

Issue #205, opened 2023-11-21
[RelayMiner] Add config file for RelayMiner


red-0ne:
## Objective

Productionize the RelayMiner configuration directive and allow them to specify proxied services endpoints

## Goals

- Enable a better Ops experience by allowing operators to specify different aspects/configurations of the RelayMiner in a yaml file

## Deliverables

- [ ] A PR with the feature and tests that enables configuring the RelayMiner from a yaml config file

## General deliverables

- [ ] **Comments**: Add/update TODOs and comments alongside the source code so it is easier to follow.
- [ ] **Testing**: Add new tests (unit and/or E2E) to the test suite.
- [ ] **Documentation**: Update architectural or development READMEs; use [mermaid](https://mermaid-js.github.io/mermaid/) diagrams where appropriate.

---

**Creator**: @red-0ne

Olshansk:
@red-0ne I'm beginning to lose picture of all the different configs and why we have them so I think a README with examples is becoming a higher priority. Here's what I see so far:

Staking configs:
- `App` config -> for `staking` applications
- `Gateway` config -> for `staking` gateways
- `Supplier` config -> for `staking` suppliers

Operations configs:
- `AppGate` config -> for `operating` a gateway or an application
- `RelayMiner` config -> for `operating` a supplier

Olshansk:
> Exactly. Haven't worked on anything regarding Gateway config though.

- #224

PR #223, opened 2023-11-24
wip: use viper as SoT for overridable config values & add select bound flags

bryanchriswhite:
### Human Summary

- Integrates viper into the relayminer subcommand for use with the relayminer config.
- Adds viper-bound flags for overriding specified config fields.

### AI Summary

reviewpad:summary

## Issue

- #205 
- #206 

image

## Type of change

Select one or more:

- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

## Sanity Checklist

### DIFF
--- a/pkg/relayer/cmd/cmd.go
+++ b/pkg/relayer/cmd/cmd.go
@@ -11,7 +11,9 @@ import (
 	cosmosflags "github.com/cosmos/cosmos-sdk/client/flags"
 	cosmostx "github.com/cosmos/cosmos-sdk/client/tx"
 	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
 
+	"github.com/pokt-network/poktroll/cmd/flags"
 	"github.com/pokt-network/poktroll/cmd/signals"
 	"github.com/pokt-network/poktroll/pkg/client/supplier"
 	"github.com/pokt-network/poktroll/pkg/client/tx"
@@ -25,11 +27,6 @@ import (
 
 // We're `explicitly omitting default` so the relayer crashes if these aren't specified.
 const omittedDefaultFlagValue = "explicitly omitting default"
-
-// TODO_CONSIDERATION: Consider moving all flags defined in `/pkg` to a `flags.go` file.
-var (
-	flagRelayMinerConfig string
-)
 
 func RelayerCmd() *cobra.Command {
 	cmd := &cobra.Command{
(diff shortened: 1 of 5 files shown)


pkg/relayer/cmd/flags.go
```diff
@@ -0,0 +1,9 @@
+package cmd
+
+const (
+	FlagRelayMinerConfigPath = "config"
```
Olshansk, line 4:
```suggestion
	FlagRelayMinerConfigPath = "relay-miner-config"
```

Since we'll have an `app-gate-config` as well

pkg/relayer/cmd/cmd.go
```diff
@@ -50,8 +47,47 @@ via an SMT KV store. The miner will monitor the current block height and periodi
...
+			FlagName:    FlagSmtStorePath,
+			ConfigKey:   relayerconfig.KeySmtStorePath,
+			Description: "Path to where the data backing SMT KV store exists on disk",
+		},
+	}
+	if err := flags.BindFlags(cmd, relayMinerFlags...); err != nil {
+		panic(err)
 	}
 
+	// TODO_IN_THIS_COMMIT: default config path?
```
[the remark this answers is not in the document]
Olshansk: We've got two (non mutually exclusive) options if set correctly:

1. `./localnet/poktrolld/config/relayminer_config.yaml` (for development & testing)
2. The value of the `--home` flag appendeded with `relayminer_config.yaml`.
Similar to how we do `poktrolld --home ./localnet/poktrolld/`, I believe `--home` will like default to `$HOME`

pkg/relayer/cmd/cmd.go
```diff
@@ -66,7 +102,8 @@ func runRelayer(cmd *cobra.Command, _ []string) error {
 	// Handle interrupt and kill signals asynchronously.
 	signals.GoOnExitSignal(cancelCtx)
 
-	configContent, err := os.ReadFile(flagRelayMinerConfig)
+	// TODO: come back to this... needs to be overridden by flags...
```
Olshansk, line 105: 1. TODO_IN_THIS_COMMIT?
2. If we set up the defaults correctly above, shouldn't it be what we expect?

pkg/relayer/config/config.go
```diff
@@ -0,0 +1,34 @@
...
+	KeyNetworkNodeUrl       = "network_node_url"
+	KeySigningKeyName       = "signing_key_name"
+	KeySmtStorePath         = "smt_store_path"
+)
+
+func ReadConfig(configPath string) {
+	if configPath != "" {
+		viper.SetConfigFile(configPath)
+	} else {
+		//// TODO_IN_THIS_COMMIT: what exactly does this do?
```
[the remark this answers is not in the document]
Olshansk: Off the top of my head (didn't look into it), I recall some stuff related to requiring prefixes in env variables. E.g. `POKTROLL_ENV_X` would be treated at `ENV_X` if we set a config correctly, but I don't remember the details related to the path.

pkg/relayer/config/config.go
```diff
@@ -0,0 +1,34 @@
...
+		//viper.SetConfigName("config")
+		//viper.AddConfigPath(".")
+
+		// TODO_IN_THIS_COMMIT: establish a default...
```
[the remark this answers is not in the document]
Olshansk: I'm still missing full context but does my comment in the previous file help?

pkg/relayer/config/relayminer_configs_reader.go
```diff
@@ -37,37 +38,66 @@ func ParseRelayMinerConfigs(configContent []byte) (*RelayMinerConfig, error) {
 		return nil, ErrRelayMinerConfigUnmarshalYAML.Wrapf("%s", err)
 	}
 
+	// TODO_IN_THIS_COMMIT: fix comment...
+	// Check if the --query-node flag was set and use it if so.
 	// Check that the query node URL is provided
-	if yamlRelayMinerConfig.QueryNodeUrl == "" {
+	queryNodeUrlString := viper.GetString(KeyQueryNodeUrl)
+	switch {
+	case queryNodeUrlString != "":
```
Olshansk, line 46: NIT: I think we shold leave the yaml reading function untouched and make it more functional.

```go
yamlRelayMinerConfigs = reader.ParseRelayMinerYamlConfigs(configContents)
relayMinerConfigs = reader.ViperConfigOverride(yamlRelayMinerConfigs)
```

That way things we explicitly mention via cli always take precedence.

Olshansk (changes requested):

pkg/relayer/cmd/flags.go
bryanchriswhite: I think the flag should stay as `--config`, the corresponding key will be unique and unambiguous.

pkg/relayer/cmd/cmd.go
bryanchriswhite: This is being done in the function that parses the config at the moment. It's not automatic because viper's API won't provide a deserialized data structure from the config (which is what we want). Instead it will parse the config and make everything available via the `viper.Get...` API, which we can use to override the sensitized config struct.

pkg/relayer/config/relayminer_configs_reader.go
bryanchriswhite: 💯

pkg/relayer/cmd/cmd.go
```diff
@@ -50,8 +47,47 @@ via an SMT KV store. The miner will monitor the current block height and periodi
 submit claim and proof messages according to the protocol as sessions become eligible
 for such operations.`,
 		RunE: runRelayer,
+		PreRun: func(_ *cobra.Command, _ []string) {
```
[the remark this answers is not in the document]
bryanchriswhite: If we intend to support flag and/or env var overrides for config keys/values, yes.

bryanchriswhite:
Tracked by #331. Closing.

Issue #140, opened 2023-11-03
[Protocol] MVP implementation off the on-chain `CreateClaim` message handling


Olshansk:
## Objective

MVP Implementation of `CreateClaim` to enable Shannon TestNet.

## Origin Document

- #43

## Goals

- Enable E2E validation of the entire protocol
- Unblock iterative development of the claim & proof lifecycle
- Unblock iterative development of on-chain tokenomics

## Deliverables

- [x] A single PR that introduces the business logic for the on-chain `CreateClaim` message (verification, etc...)
- [x] A single PR that updates the CLI to submit CreateClaim messages
- [x] Introduce make targets to enable development (visibility, triggering, debugging , etc...) of on-chain claims
- [ ] Add extensive unit tests 

## Non-goals / Non-deliverables

- Account Management (i.e. Burn & Mint) business logic related to the Claim & Proof lifecycle (leave TODOs)
- Governance Params necessary for the Claim & Proof lifecycle (leave TODOs)

## General deliverables

- [x] **Comments**: Add/update TODOs and comments alongside the source code so it is easier to follow.
- [x] **Testing**: Add new tests (unit and/or E2E) to the test suite.
- [ ] **Makefile**: Add new targets to the Makefile to make the new functionality easier to use.
- [ ] **Documentation**: Update architectural or development READMEs; use [mermaid](https://mermaid-js.github.io/mermaid/) diagrams where appropriate.

---

**Creator**: @Olshansk 
**Co-Owners**: @bryanchriswhite

Olshansk:
@bryanchriswhite Picking this up

Olshansk:
@bryanchriswhite I'm unassigning myself from this in the meantime and will let you pick it up.

I think the comments you left (copy-pastedbelow) in `func (k msgServer) CreateClaim(goCtx context.Context, msg *types.MsgCreateClaim) ` is a great starting point!

```go
	/*
		TODO_INCOMPLETE: Handling the message

		## Validation

		### Session validation
		1. [ ] claimed session ID matches on-chain session ID
		2. [ ] this supplier is in the session's suppliers list

		### Msg distribution validation (depends on session validation)
		1. [ ] governance-based earliest block offset
		2. [ ] pseudo-randomize earliest block offset

		### Claim validation
		1. [ ] session validation
		2. [ ] msg distribution validation

		## Persistence
		1. [ ] create claim message
			- supplier address
			- session header
			- claim
		2. [ ] last block height commitment; derives:
[...]

PR #237, opened 2023-11-30
[Testing] chore: add testkeyring pkg

bryanchriswhite:
### Human Summary

Adds a set of pre-generated accounts (i.e. addresses and mnemonics) and APIs which can be used to consume them. This is useful for populating the keyring and/or module genesis states.

It also includes the package used to generate the accounts and adds a make target for its usage (`go_testgen_accounts`).

### Summary generated by Reviewpad on 04 Dec 23 11:51 UTC

1. In the file `miner_test.go`, the comment regarding regenerating fixtures has been updated. The previous comment mentioned using "`make go_fixturegen`", while the updated comment suggests using "`make go_testgen_fixtures`" to regenerate fixtures.

2. In the file `template.go` in the `pkg/relayer/miner/gen/` directory, there are changes to comments regarding regenerating fixtures. The comment now recommends using the command `make go_testgen_fixtures` to regenerate all fixtures and `go generate ./pkg/relayer/miner/miner_test.go` to regenerate only this test's fixtures.

3. In the file `accounts.go` in the `testutil/testkeyring` package, a new file has been introduced. It defines several structs and functions related to pre-generated accounts used in tests. The file provides a way to iterate over pre-generated accounts, retrieve accounts at specific indexes, and encode/decode account information using base64 and JSON.

4. The diff in the file `relay_fixtures_test.go` in the package `pkg/relayer/miner` includes changes to generated test fixtures. It is advised not to manually edit this file as any changes will be overwritten upon regeneration. If you need to regenerate all fixtures, use the command `make go_testgen_fixture`. If you only want to regenerate fixtures for this specific test, run `go generate ./pkg/relayer/miner/miner_test.go`.

5. There are changes made to the file `testkeyring/keyring.go`. It includes the addition of a new file with changes related to the creation and management of pre-generated accounts in a keyring for testing purposes.

6. A new file `testutil/testkeyring/gen_accounts/gen.go` has been added. It is a Go program that generates pre-generated accounts. It reads command-line flags, creates a new in-memory keyring, generates accounts, and stores their information in a file.

Please review and verify these changes.

## Issue

- #140 

## Type of change

Select one or more:

- [ ] Bug fix
- [x] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

## Sanity Checklist


testutil/testkeyring/accounts_table.go
```diff
@@ -0,0 +1,111 @@
+// DO NOT EDIT. This Code is generated by gen_accounts/gen.go,
+// changes will be overwritten upon regeneration.
+//
+// To regenerate this file, use make go_test_accountgen or go generate ./testutil/network/keyring.go.
```
bryanchriswhite, line 4: Forgot to add the make target.

testutil/testkeyring/gen_accounts/gen.go
```diff
@@ -0,0 +1,91 @@
...
+	flag.StringVar(&flagOut, "out", defaultOutPath, "the path to the generated file.")
+	flag.IntVar(&flagAccountsLimit, "limit", 100, "the number of accounts to generate.")
+}
+
+func main() {
+	flag.Parse()
+
+	kr := keyring.NewInMemory(app.MakeEncodingConfig().Marshaler)
+
+	preGeneratedAccountsGobHexLines := make([]string, flagAccountsLimit)
```
Olshansk, line 40: Two questions about this naming:

1. The serialized value returned `hex`, so I think we can replace `HexLines` with `Hex`
2. We do `JSON` marshalling, so I believe we can remove `Gob`

testutil/testkeyring/gen_accounts/gen.go
```diff
@@ -0,0 +1,91 @@
...
+)
+
+var (
+	flagOut           string
+	flagAccountsLimit int
+	defaultOutPath    = "accounts_table.go"
+)
+
+func init() {
+	flag.StringVar(&flagOut, "out", defaultOutPath, "the path to the generated file.")
```
Olshansk, line 31:
```suggestion
	flag.StringVar(&flagOut, "accounts.gen.go", defaultOutPath, "the path to the generated go file of keyring test accounts.")
```

testutil/testkeyring/gen_accounts/gen.go
```diff
@@ -0,0 +1,91 @@
...
+	flag.IntVar(&flagAccountsLimit, "limit", 100, "the number of accounts to generate.")
```
Olshansk, line 32: Thoughts on s/limit/count or s/limit/n?

I always think of `limit` as `<=`, not `==`

testutil/testkeyring/accounts.go
```diff
@@ -0,0 +1,128 @@
...
+func MustPreGeneratedAccountAtIndex(index uint32) *PreGeneratedAccount {
+	account, err := PreGeneratedAccountAtIndex(index)
+	if err != nil {
+		panic(err)
+	}
+	return account
+}
+
+// PreGeneratedAccounts returns a new PreGeneratedAccountIterator with the
+// accounts which were pre-generated by the `gen_accounts` tool.
```
Olshansk, line 49:
```suggestion
// accounts which were pre-generated by the `gen_accounts` packag.
```

testutil/testkeyring/accounts.go
```diff
@@ -0,0 +1,128 @@
...
+func PreGeneratedAccounts() *PreGeneratedAccountIterator {
+	return preGeneratedAccounts.Clone()
+}
+
+// NewPreGeneratedAccountIterator returns a new PreGeneratedAccountIterator with
+// the given accounts. It is primarily used by the generated code to intially
```
Olshansk, line 55:
```suggestion
// the given accounts. It is primarily used by the generated code to initially
```

Olshansk (changes requested):

testutil/testkeyring/accounts.go
```diff
@@ -0,0 +1,128 @@
...
+	next, err := iter.Next()
+	if err != nil {
+		panic(err)
+	}
+	return next
+}
+
+// Next returns the next account in the iterator. It is safe to call
+// concurrently and is guaranteed to return a unique account on each call.
+func (iter *PreGeneratedAccountIterator) Next() (*PreGeneratedAccount, error) {
```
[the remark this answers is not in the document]
bryanchriswhite: :star_struck: Nice find - thanks for sharing!

I definitely want to iterate on he current design but I think an iterator pattern will support much simpler and  easier to reason about tests/helpers. My goal was to still support access by index, as in some specific cases that makes more sense; however, I imagine that the majority of the time it will just be more convenient to call a `#Next()` method to get a unique account instead of having to track an index and/or a container yourself. I also plan to add some examples to illustrate what I'm imagining.

testutil/testkeyring/gen_accounts/gen.go
bryanchriswhite: I could imagine both `count` or `n` describing a boundary or a counter. I would agree with the argument  that it wouldn't really make much sense for it to be the latter in this case; nonetheless, I thought the word `limit` would sufficiently disambiguate the intended usage.

testutil/testkeyring/accounts.go
```diff
@@ -0,0 +1,128 @@
...
+	ErrPreGeneratedAccountIndexOutOfRange = errorsmod.New(codespace, 1, "index out of range of pre-generated accounts list")
+	codespace                             = "testutil/network"
+)
+
+type PreGeneratedAccount struct {
+	Address  sdktypes.AccAddress
+	Mnemonic string
+}
+
+type PreGeneratedAccountIterator struct {
```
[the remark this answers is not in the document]
bryanchriswhite: (see https://github.com/pokt-network/poktroll/pull/237#discussion_r1412793351)

testutil/testkeyring/accounts.go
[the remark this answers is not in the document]
Olshansk: > I definitely want to iterate on he current design but I think an iterator pattern will support much simpler and easier to reason about tests/helpers. My goal was to still support access by index, as in some specific cases that makes more sense; however, I imagine that the majority of the time it will just be more convenient to call a #Next() method to get a unique account instead of having to track an index and/or a container yourself. I also plan to add some examples to illustrate what I'm imagining.

I understand where this is coming from but still feel like we're overengineering something that can be a basic list. I'm not 100% convinced we'll need this at the moment, so will push back unless unless using `#Next()` makes the first application of this pattern easier to reason about.

testutil/testkeyring/gen_accounts/gen.go
Olshansk: Seems like you did think this through so not strong opinion either way. I think keeping as `limit` is fine.

testutil/testkeyring/accounts.go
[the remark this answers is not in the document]
bryanchriswhite: I believe that it already does. The fact that the keyring can be populated in one line completely separately from where the keys are designated is indicative of the kind of coordination avoidance this enables.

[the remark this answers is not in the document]
bryanchriswhite: The requirements will continue to get more complex until we finish with proof validation. Wdyt about waiting until then before rendering final judgement?

testutil/testkeyring/gen_accounts/gen.go
```diff
@@ -0,0 +1,91 @@
...
+	); err != nil {
+		log.Fatal(err)
+	}
+
+	if err := os.WriteFile(flagOut, outputBuffer.Bytes(), 0644); err != nil {
+		log.Fatal(err)
+	}
+}
+
+func serializePreGeneratedAccount(account *testkeyring.PreGeneratedAccount) (string, error) {
```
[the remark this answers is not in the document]
bryanchriswhite: I went with `Marshal()` and `Unmarshal()` since both can retrun an error and it is more consistent with the JSON interface used underneath.

testutil/testkeyring/accounts.go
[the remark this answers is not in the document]
Olshansk: > Wdyt about waiting until then before rendering final judgement?

I personally think that if we merge it in, we'll never go back to remove it.

However, you seem pretty adamant that this is he right approach so I trust you on that 🤝

testutil/testkeyring/accounts.go
```diff
@@ -94,35 +86,33 @@ func (iter *PreGeneratedAccountIterator) Clone() *PreGeneratedAccountIterator {
 	return NewPreGeneratedAccountIterator(iter.accounts...)
 }
 
-func (pga *PreGeneratedAccount) ToTestAccount(name ...string) testutil.TestAccount {
-	return testutil.TestAccount{
-		Name:    name[0],
-		Address: pga.Address,
+// Marshal returns the base64 and JSON encoded account string.
+func (pga *PreGeneratedAccount) Marshal() (string, error) {
```
Olshansk, line 90: Nice, ty for doing this!

Olshansk (approved):
Approved!!

PR #219, opened 2023-11-23
[Off-chain, Polylog] refactor: replace std logger usage in pkg/... with polylog

bryanchriswhite:
### Human Summary

Refactors usage of go std `log` pkg with analogous usage of polylog in the off-chain (i.e. pkg/...) packages.

## Issue

- #181
- #210 
- #209 
- #211 
- #218

Now that polylog is available, use it instead instead of `log`.

## Type of change

Select one or more:

- [x] New feature, functionality or library
- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

- [ ] **Run all unit tests**: `make go_develop_and_test`
- [ ] **Verify Localnet manually**: See the instructions [here](TODO: add link to instructions)

## Sanity Checklist

- [ ] I have tested my changes using the available tooling
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, updated documentation and left TODOs throughout the codebase

PR #244, opened 2023-12-05
[Logging] Enforce polylog usage after #219 for new logs added

Olshansk:
## Summary

Add a reviewdog workflow to make sure we do not use `log.*` for logging purposes anymore.

### AI Summary

reviewpad:summary

## Issue
- Follow up to PR #219

## Type of change

Select one or more:

- [ ] Bug fix
- [x] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

## Sanity Checklist

### DIFF
--- a/.github/workflows/reviewdog.yml
+++ b/.github/workflows/reviewdog.yml
@@ -7,10 +7,10 @@ on:
 concurrency:
   group: ${{ github.workflow }}-${{ github.head_ref || github.ref_name }}
   cancel-in-progress: true
-  
+
 jobs:
-  # Makes sure that comments like TODO_IN_THIS_PR or TODO_IN_THIS_COMMIT block merging to main
-  # More info: https://github.com/pokt-network/action-fail-on-found
+  # Makes sure that comments like TODO_IN_THIS_PR or TODO_IN_THIS_COMMIT block
+  # merging to main. See https://github.com/pokt-network/action-fail-on-found
   check_todo_in_this:
     name: Check TODO_IN_THIS_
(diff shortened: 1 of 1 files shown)


.github/workflows/reviewdog.yml
```diff
@@ -24,6 +24,23 @@ jobs:
           fail_on_error: true
           pattern: TODO_IN_THIS_
 
+  # Ensure that we are using polylog (via logger.) instead of the golang's default
+  # log package.
+  check_polylog_in_off_chain_package:
+    name: Check polylog in off-chain source code
```
bryanchriswhite, line 30: Is there any reason to limit this to off-chain code?

---

NIT:

```suggestion
  # Ensure that we are using polylog (via logger.) instead of the golang's stdlib
  # log package.
  check_stdlog_in_off_chain_package:
    name: Check stdlog in off-chain source code
```

bryanchriswhite (approved):
:rocket: Thanks for doing this! :raised_hands: 

Approving but with a small question/suggestion. :pinching_hand:

.github/workflows/reviewdog.yml
Olshansk: @bryanchriswhite I committed your suggestion, but the reason it's limited to `off-chain` is because my goal (with this change) was to prevent someone accidentally using the std logger in our off-chain packages. Some of the default scaffolded work still uses `log.`. I haven't gone through it's uses and simply considered it out of scope.

Do you want to add a TODO or something else?

PR #267, opened 2023-12-12
[EventsReplayClient] Fix Replay Client Bugs

h5law:
### Human Summary

This PR implements remapping of events to allow the usage of `EventsSequence` (`CommittedBlocksSequence` and `RedelegationsSequence`) to be maintained between the `EventsQueryClient` closing.

It also fixes numerous bugs where the `EventsReplayClient` was not behaving as expected:

- `retry.OnError` logic has been fixed
- The `EventsQueryClient` closing causing the above to occur has been added
- Logic for remapping events from one observable to another has been improved

These are all covered in the regression integration test added which now acts as a backstop protecting these features from breaking.

### Summary generated by Reviewpad on 22 Dec 23 16:23 UTC
- Added a new error variable `ErrEventsConsClosed` in `pkg/client/events/errors.go`.
- Changes to the file `client.go` indicate the removal of `cometWebsocketURL` from the `NewDelegationClient()` function and its related calls.
- Changes in `query_client.go` include modifications to code blocks and locking/unlocking calls.
- Modifications in `pkg/client/block/client_integration_test.go` involve the removal of build tags and the addition of test functions and import statements.
- Changes in `connection.go` involve imports, addition of new functions, and modifications related to managing mock objects for testing.
- Modifications in the "openapi.yml" file include updates to titles and options for the RPC type.
- Changes in `cmd.go` indicate the removal of arguments from function calls in `NewSupplyBlockClientFn()` and `NewSupplyPOKTRollSDKFn()`.
- Modifications in a file indicate the removal of a method and changes to other methods in terms of their functionality and comments.
- Changes in the file `client.go` involve variable declaration and function call modifications.
- The diff in the file `client.go` shows changes related to import statements and function arguments.
- Modifications in the file `client.go` involve the removal of import statements and changes in variable initialization.
- Changes in the file `client.go` involve the removal of a function call argument.
- A new integration test file `replay_client_integration_test.go` has been added with a test function.
- Changes in `replay.go` involve synchronization, error messages, and comment updates related to context cancellation and unsubscribe functionality.
- Changes in a file include import statements, variable assignments, and function call modifications.
- The file `client_integration_test.go` has experienced syntactical modifications, changes in comments, and modifications in test functions and assertions.
- The diff in the file `deps_builder.go` shows the removal of a parameter in a function call.
- Changes in the file `godoc.go` involve updating package documentation comments with a focus on redelegation events.
- Changes in the file `replay_client.go` involve retrying the events query, modification of the replay observable buffer size, and changes to function signatures.
- The diff in the file `client.go` shows the removal of a function call argument.
- Changes in the file involve logging error messages during retries with improved logging output.
- The file `client_test.go` underwent changes related to import statements and function call modifications.
- The diff in the file includes the removal of an import statement and modification of a function call argument.
- The summary of the file diff includes the removal of an import statement and modification of a function call argument.
- Changes in a file involve modifying a function call by removing an argument.

Please let me know if you need more information or if you have any other questions regarding this pull request.

## Issue

- N/A

## Type of change

Select one or more:

- [x] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

- [ ] **Run E2E tests on DevNet`: Add the `devnet-test-e2e` label to the PR. This is VERY expensive, o only do it after all the reviews are complete.

## Sanity Checklist


pkg/client/block/client_integration_test.go
```diff
@@ -31,11 +31,17 @@ func TestBlockClient_LastNBlocks(t *testing.T) {
...
-	blockClient := testblock.NewLocalnetClient(ctx, t)
+	blockClient, queryClient := testblock.NewLocalnetClient(ctx, t)
 	require.NotNil(t, blockClient)
 
 	blockSub := blockClient.CommittedBlocksSequence(ctx).Subscribe(ctx)
 
+	// Simulate the connection closing
+	go func() {
+		time.Sleep(1 * time.Second)
+		queryClient.Close()
```
bryanchriswhite, line 42: If the intended effect is to simulate a network failure and/or connection retry, I don't think this is the right way to go because while `eventsQueryClient#Close()` does close the underlying connection, it also calls `#UnsubscribeAll()` on the events bytes observable, which will effectively close the block client.

bryanchriswhite: Additionally, if this is the motivation behind the `testblock.NewLocalClient()` refactor, I would suggest reverting it if we decide not to do this here.

pkg/client/block/client_test.go
```diff
@@ -44,7 +44,7 @@ func TestBlockClient(t *testing.T) {
 	expectedEventBz, err := json.Marshal(expectedBlockEvent)
 	require.NoError(t, err)
 
-	eventsQueryClient := testeventsquery.NewAnyTimesEventsBytesEventsQueryClient(
+	eventsQueryClient := testeventsquery.NewAnyTimesClosableEventsQueryClient(
```
bryanchriswhite, line 47: What's the rationale for this? The pattern behind the naming thus far has been `New<num_times><main_mocked_method><interface_name>()`.

testutil/testclient/testeventsquery/client.go
```diff
@@ -117,3 +117,39 @@ func NewAnyTimesEventsBytesEventsQueryClient(
 
 	return eventsQueryClient
 }
+
+// NewAnyTimesClosableEventsQueryClient returns a new events query client which
+// is configured to return the expected event bytes when queried with the expected
+// query, any number of times. The returned client also expects to be closed any
+// number of times.
+func NewAnyTimesClosableEventsQueryClient(
```
bryanchriswhite, line 125: I'm not convinced that we need a new helper for this. Is there a reason that we can't add the same "any times" expectation to the existing helper?

pkg/client/delegation/client_test.go
```diff
@@ -72,6 +72,22 @@ func TestDelegationClient(t *testing.T) {
...
 				require.Equal(t, expectedAddress, lastRedelegation.GetAppAddress())
 
+				return lastRedelegation
+			},
+		},
+		{
+			name: "RedelegationsSequence successfully remaps closure",
+			fn: func() client.Redelegation {
+				// Close the events query client to force a reconnection.
+				eventsQueryClient.Close()
```
bryanchriswhite, line 82: (same as https://github.com/pokt-network/poktroll/pull/267#discussion_r1425089102)

pkg/client/block/client_integration_test.go
bryanchriswhite: I think this behavior should be tested in an unit test, not an integration test.

bryanchriswhite: This behavior might be more easily tested in a unit test where the `Connection` could be mocked and closed independently of the other components involved.

bryanchriswhite: I don't think we need to be testing for this case in an integration test.

pkg/client/block/client_test.go
```diff
@@ -79,6 +79,23 @@ func TestBlockClient(t *testing.T) {
...
 				require.Equal(t, expectedHash, lastBlock.Hash())
 
+				return lastBlock
+			},
+		},
+		{
+			name: "CommittedBlocksSequence successfully remaps closure",
+			fn: func() client.Block {
+				// Close the events query client to force a reconnection.
+				eventsQueryClient.Close()
```
bryanchriswhite, line 89: I think we should be closing a mocked `Connection` instead:
(see: https://github.com/pokt-network/poktroll/pull/267#discussion_r1425089102)

bryanchriswhite (changes requested):
Nice, thanks for the assist! :sunglasses: 

I left a few comments regarding how/what we should be testing but otherwise, this is a solid improvement and bug fix. :mechanical_arm:

pkg/client/block/client_integration_test.go
h5law: Agreed removing this and keeping it to unit tests

pkg/client/events/query_client.go
```diff
@@ -244,8 +245,8 @@ func (eqc *eventsQueryClient) goUnsubscribeOnDone(
 	// Wait for the context to be done.
 	<-ctx.Done()
 	// Only close the eventsBytes for the given query.
-	eqc.eventsBytesAndConnsMu.RLock()
-	defer eqc.eventsBytesAndConnsMu.RUnlock()
+	eqc.eventsBytesAndConnsMu.Lock()
+	defer eqc.eventsBytesAndConnsMu.Unlock()
```
bryanchriswhite, line 249: Nice catch! :fire:

pkg/client/events/replay_client.go
```diff
@@ -61,12 +58,17 @@ type replayClient[T any, U observable.ReplayObservable[T]] struct {
 	// message bytes into the type defined by the EventsReplayClient's generic type
 	// parameter.
 	eventDecoder NewEventsFn[T]
+	// replayObsBufferSize is the buffer size for the replay observable returned
+	// by EventsSequence
```
bryanchriswhite, line 62: NIT:
```suggestion
	// by EventsSequence.
```

pkg/client/events/replay_client.go
```diff
@@ -61,12 +58,17 @@ type replayClient[T any, U observable.ReplayObservable[T]] struct {
...
+	replayObsBufferSize int
 	// replayObsCache is a replay observable with replay buffer size 1,
 	// which holds the "active latest observable" which is notified when
 	// new events are received by the events query client subscription
 	// created in goPublishEvents. This observable (and the one it emits) closes
 	// when the events bytes observable returns an error and is updated with a
 	// new "active" observable after a new events query subscription is created.
+	// TODO(@h5law): consider retyping this to a Observable of ReplayObservables
+	// as it does not need to be replayable.
```
bryanchriswhite, line 71: :pray:

pkg/client/events/replay_client.go
```diff
@@ -87,20 +89,19 @@ type replayClient[T any, U observable.ReplayObservable[T]] struct {
...
-		queryString:  queryString,
-		eventDecoder: newEventFn,
+		queryString:         queryString,
+		eventDecoder:        newEventFn,
+		replayObsBufferSize: replayObsBufferSize,
 	}
 	replayObsCache, replayObsCachePublishCh := channel.NewReplayObservable[U](
 		ctx,
-		replayObsBufferSize,
+		1, // replay buffer size 1
```
bryanchriswhite, line 104: NIT: maybe we should add a comment similar to that above as a note that this will go away.

bryanchriswhite: I feel like otherwise, I would like the comment to speak to the "why" a bit more (the cache only ever needs to return the last cached value).

pkg/client/events/replay_client.go
```diff
@@ -110,22 +111,51 @@ func NewEventsReplayClient[T any, U observable.ReplayObservable[T]](
...
+			return
+		default:
+		}
+		// Publish events from the new "active" replay observable to the given
+		// publish channel.
+		eventObserver := eventObs.Subscribe(ctx)
+		for event := range eventObserver.Ch() {
+			publishCh <- event
+		}
+	}
```
bryanchriswhite, line 158: As discussed, I think this can be simplified with a `channel.Map()`.

pkg/client/events/replay_client.go
```diff
@@ -191,7 +219,20 @@ func (rClient *replayClient[T, R]) retryPublishEventsFactory(ctx context.Context
...
-		// Initially set latestObsvbls and update if after retrying on error.
+		// Subscribe to the eventBzObs and block until the channel closes.
+		// Then pass this as an error to force the retry.OnError to resubscribe.
+		go func() {
+			eventsBzSub := eventsBz.Subscribe(ctx)
+			for range eventsBzSub.Ch() {
+				// Wait for the channel to close.
+				continue
+			}
+			typedObs.UnsubscribeAll()
```
bryanchriswhite, line 230: Would you mind adding a comment to remind us why we need to do this?

pkg/client/events/replay_client.go
```diff
@@ -191,7 +219,20 @@ func (rClient *replayClient[T, R]) retryPublishEventsFactory(ctx context.Context
 			rClient.newMapEventsBytesToTFn(errCh),
 		)
 
-		// Initially set latestObsvbls and update if after retrying on error.
+		// Subscribe to the eventBzObs and block until the channel closes.
+		// Then pass this as an error to force the retry.OnError to resubscribe.
+		go func() {
+			eventsBzSub := eventsBz.Subscribe(ctx)
+			for range eventsBzSub.Ch() {
```
bryanchriswhite, line 226: Did you forget to push some changes, I thought we did this already?

```suggestion
			eventsBzObserver := eventsBz.Subscribe(ctx)
			for range eventsBzObserver.Ch() {
```

pkg/client/events/replay_client.go
```diff
@@ -181,8 +211,6 @@ func (rClient *replayClient[T, R]) retryPublishEventsFactory(ctx context.Context
 		}
 
 		// NB: must cast back to generic observable type to use with Map.
-		// client.BlocksObservable cannot be an alias due to gomock's lack of
-		// support for generic types.
 		eventsBz := observable.Observable[either.Either[[]byte]](eventsBzObs)
```
bryanchriswhite, line 214: Wdyt about renaming this to `eventsBzObs`, as it's used more than the variable that currently uses that name, and renaming that one to something like `eventsBytesObs`?

pkg/client/events/replay_client.go
bryanchriswhite: IIRC it's because `typedObs` is a derivative of `eventsBzObs` + the fact that calling `UnsubscribeAll()` on an observable has no effect observables derived from it (i.e. their observers would still be subscribed).

pkg/client/events/replay_client_integration_test.go
```diff
@@ -0,0 +1,174 @@
...
+			received := eventsReceived.Add(1)
+			if received >= eventsToRecv {
+				errCh <- nil
+				return
+			}
+		}
+	}()
+
+	// Wait for ~2 events to be received before closing the connection
+	time.Sleep(51675 * time.Microsecond)
```
bryanchriswhite, line 160: This isn't going to provide consistent timing across different CPU architectures and scales. Could you instead subscribe to an observable and cancel a context (or send on a "done" channel, etc.) after 2 notifications?

bryanchriswhite (changes requested):
I think this will be good to go after one last pass. :sunglasses:

pkg/client/events/replay_client_integration_test.go
bryanchriswhite: I think it would be better to couple the timing via an asynchronous primitive rather than aligning `time.Sleep()` durations. At the very least thought I would prefer the durations have a common unit and/or share some constant.

testutil/testclient/testeventsquery/connection.go
```diff
@@ -47,3 +48,66 @@ func NewOneTimeMockDialer(
...
+) *mockclient.MockConnection {
+	ctrl := gomock.NewController(t)
+	connMock := mockclient.NewMockConnection(ctrl)
+	// Expect the connection to be closed and the dialer to be re-established
+	connMock.EXPECT().Close().DoAndReturn(func() error {
+		connClosed.CompareAndSwap(false, true)
+		return nil
+	}).
+		// Close once by hand and the EventQueryClient will close it again
+		Times(2)
```
Olshansk, line 84: NIT: Fix spacing

testutil/testclient/testeventsquery/connection.go
```diff
@@ -47,3 +48,66 @@ func NewOneTimeMockDialer(
...
+	connMock.EXPECT().
+		Send(gomock.Any()).
+		DoAndReturn(func(eventBz []byte) error {
+			if connClosed.Load() {
+				connClosed.CompareAndSwap(true, false)
+			}
+			delayEvent.CompareAndSwap(true, false)
+			return nil
+		}).
+		// Once to estabslish the connection and once to re-establish
```
Olshansk, line 95:
```suggestion
		// Once to establish the connection and once to re-establish
```

testutil/testclient/testeventsquery/connection.go
```diff
@@ -47,3 +48,66 @@ func NewOneTimeMockDialer(
...
+	t *testing.T,
+	connClosed *atomic.Bool,
+	delayEvent *atomic.Bool,
+) (*mockclient.MockConnection, *mockclient.MockDialer) {
+	connMock := NewTwiceCloseSendConnectionMock(t, connClosed, delayEvent)
+	dialerMock := NewAnyTimesMockDailer(t, connMock)
+	return connMock, dialerMock
+}
+
+// NewTwiceCloseSendConnectionMock returns a mock connection that will expect
```
Olshansk, line 66: What's so special about the number 2? I'm trying to understand why we don't parameterize it.

pkg/client/events/replay_client_integration_test.go
```diff
@@ -0,0 +1,150 @@
...
+
+			eventNum := readEventCounter.Add(1) - 1
+			event := newMessageEventBz(eventNum)
+			if eventNum == 2 {
+				// Simulate the connection closing
+				connMock.Close()
+			}
+
+			// Simulate IO delay between sequential events.
+			time.Sleep(50 * time.Microsecond)
```
Olshansk, line 93: I'm worried about different machines having different performance characteristics, so can you explain where/how 50ms was validated?

Olshansk (changes requested):
Still need to take a closer look at `pkg/client/events/replay_client_integration_test.go` and @bryanchriswhite is def the best person to approve this, but left a few minor nits throughout.

pkg/client/events/replay_client.go
```diff
@@ -61,18 +58,22 @@ type replayClient[T any, U observable.ReplayObservable[T]] struct {
...
+	// replayObsBufferSize is the buffer size for the replay observable returned
+	// by EventsSequence.
+	replayObsBufferSize int
+	// replayObsCache is a replay observable with a buffer size of 1, which
+	// holds the "active latest replay observable" which is notified when new
+	// events are received by the events query client subscription created in
+	// goPublishEvents. This observable (and the one it emits) closes when the
+	// events bytes observable returns an error and is updated with a new
+	// "active" observable after a new events query subscription is created.
+	// TODO_REFACTOR(@h5law): Look into making this a regular observable
```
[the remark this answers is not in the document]
bryanchriswhite: It would be a simplification.

pkg/client/events/replay_client.go
```diff
@@ -84,48 +85,74 @@ type replayClient[T any, U observable.ReplayObservable[T]] struct {
...
-		endpointURL:  cometWebsocketURL,
-		queryString:  queryString,
-		eventDecoder: newEventFn,
+	rClient := &replayClient[T, R]{
+		queryString:         queryString,
+		eventDecoder:        newEventFn,
+		replayObsBufferSize: replayObsBufferSize,
 	}
-	replayObsCache, replayObsCachePublishCh := channel.NewReplayObservable[U](
+	// TODO_REFACTOR(@h5law): Look into making this a regular observable
```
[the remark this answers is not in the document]
bryanchriswhite: It would be a simplification.

pkg/client/events/replay_client_integration_test.go
bryanchriswhite: It **needs to be non-zero** because `#Recieve()` is called in tightest loop possible (i.e. max speed limited by go runtime). If this delay is too small, in this environment where there is no real delay between events being available for publishing, calls to `#Receive()` will bog down the runtime, leading to other unexpected errors.

We should **prefer to keep it as small as possible** (individual orders of magnitude matter here) in order to minimize the feedback loop in the event that we need to debug flaky behavior in the future; I typically consider a test likely to be flake-free if I can get it to pass 10K times contiguously.

bryanchriswhite: IIRC, there are a few other places in tests where we have to slow down some loops for similar reasons.

pkg/client/events/replay_client.go
```diff
@@ -84,48 +85,74 @@ type replayClient[T any, U observable.ReplayObservable[T]] struct {
...
+	// Concurrently publish events to the observable emitted by replayObsCache.
 	go rClient.goPublishEvents(ctx)
 
 	return rClient, nil
 }
 
-// EventsSequence returns a ReplayObservable, with a replay buffer size of 1,
-// which is notified when new events are received by the encapsulated
-// EventsQueryClient.
+// EventsSequence returns a ReplayObservable, with the buffer size provided
```
bryanchriswhite, line 122:
```suggestion
// EventsSequence returns a new ReplayObservable, with the buffer size provided
```

pkg/client/events/replay_client.go
```diff
@@ -61,18 +58,22 @@ type replayClient[T any, U observable.ReplayObservable[T]] struct {
...
-	// which holds the "active latest observable" which is notified when
-	// new events are received by the events query client subscription
-	// created in goPublishEvents. This observable (and the one it emits) closes
-	// when the events bytes observable returns an error and is updated with a
-	// new "active" observable after a new events query subscription is created.
-	replayObsCache observable.ReplayObservable[U]
+	// replayObsBufferSize is the buffer size for the replay observable returned
+	// by EventsSequence.
+	replayObsBufferSize int
+	// replayObsCache is a replay observable with a buffer size of 1, which
```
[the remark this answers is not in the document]
h5law: Yes so the replayObsCache is a replay observable of replay observables it hold the "active"/open observable so it needs a buffer size of 1 - the `replayObsBufferSize` refers to the buffer size of the replay observables emitted by the replayObsCache and these are customisable

pkg/client/events/replay_client_integration_test.go
h5law: Yeah there needs to be a delay but its not super important on the timing of the delay the only delay that has some significance is the delayed event (line 81) as this gives the events query client time to subscribe

pkg/client/events/replay_client.go
```diff
@@ -84,48 +85,74 @@ type replayClient[T any, U observable.ReplayObservable[T]] struct {
...
+	// replay observable.
+	cachedEventTypeObserver := rClient.replayObsCache.Subscribe(ctx)
+	for eventObs := range cachedEventTypeObserver.Ch() {
+		// Publish events from the new "active" replay observable to the given
+		// publish channel.
+		eventObserver := eventObs.Subscribe(ctx)
+		for event := range eventObserver.Ch() {
+			publishCh <- event
+		}
+	}
```
bryanchriswhite, line 155:
```suggestion
	channel.ForEach[R](
		ctx,
		rClient.replayObsCache,
		func(ctx context.Context, eventTypeObs R) {
			// TODO_CONSIDERATION: hold a reference to the previous event type
			// observable and call `#UnsubscribeAll()` here in the event that it
			// somehow starts receiving notifications again.
			eventObserver := eventTypeObs.Subscribe(ctx)
			for event := range eventObserver.Ch() {
				publishCh <- event
			}
		},
	)
```

testutil/testclient/testeventsquery/connection.go
h5law: I can make it a parameter 👍🏼

pkg/client/events/replay_client.go
```diff
@@ -84,48 +85,74 @@ type replayClient[T any, U observable.ReplayObservable[T]] struct {
...
+		queryString:         queryString,
+		eventDecoder:        newEventFn,
+		replayObsBufferSize: replayObsBufferSize,
 	}
-	replayObsCache, replayObsCachePublishCh := channel.NewReplayObservable[U](
+	// TODO_REFACTOR(@h5law): Look into making this a regular observable
+	replayObsCache, replayObsCachePublishCh := channel.NewReplayObservable[R](
 		ctx,
-		replayObsBufferSize,
+		// buffer size of 1 as the cache only needs to hold the latest
```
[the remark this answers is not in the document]
bryanchriswhite: Yes, it was incorrect previously. In the current construction (i.e. the cache is a replay observable), we only ever need to replay the most recent value from the cache because older values are references observables which have been abandoned (i.e. will never be notified again) due to network interruption.

Since this PR refactors `#EventSequence()` to no longer depend on `replayObsCache.Last()`, it need not be a replay observable.

testutil/testclient/testeventsquery/connection.go
h5law: This is the correct spacing from gofumpt?

testutil/testclient/testeventsquery/connection.go
bryanchriswhite: @Olshansk in this particular case, I would agree so long as the function were renamed to something like `NewNTimesReconnectConnectionMock`. @h5law wdyt?

bryanchriswhite: @Olshansk in general though, I still want to resist this tendency to avoid arbitrarily coupling expectations and/or exploding the signature sizes of these testutils.

testutil/testclient/testeventsquery/connection.go
bryanchriswhite, line 84:
```suggestion
	// Expect the connection to be closed and the dialer to be re-established;
	// closed once manually in the test and again when the EventQueryClient closes.
	connMock.EXPECT().
		Close().
		DoAndReturn(
			func() error {
				connClosed.CompareAndSwap(false, true)
				return nil
			},
		).Times(2)

```

pkg/client/events/replay_client_integration_test.go
```diff
@@ -0,0 +1,150 @@
...
+			// Simulate ErrConnClosed if connection is isClosed.
+			if connClosed.Load() {
+				return nil, events.ErrEventsConnClosed
+			}
+
+			// Delay the event if needed
+			if !delayEvent.Load() {
+				time.Sleep(50 * time.Millisecond)
+				delayEvent.CompareAndSwap(false, true)
+			}
```
bryanchriswhite, line 83:
```suggestion
			if delayEvent.CompareAndSwap(false, true) {
				time.Sleep(50 * time.Millisecond)
			}
```

pkg/client/events/replay_client_integration_test.go
```diff
@@ -0,0 +1,150 @@
...
+// newMessageEventBz returns a new message event in JSON format
+func newMessageEventBz(eventNum int32) []byte {
+	return []byte(fmt.Sprintf(`{"message":"message_%d"}`, eventNum))
+}
+
+func TestReplayClient_Remapping(t *testing.T) {
+	var (
+		ctx              = context.Background()
+		connClosed       atomic.Bool
+		delayEvent       atomic.Bool
```
bryanchriswhite, line 61: I think the name is the opposite of what it actually is.
```suggestion
		firstEventDelayed       atomic.Bool
```

pkg/client/events/replay_client_integration_test.go
```diff
@@ -0,0 +1,150 @@
...
+		100, // replay buffer size
+	)
+	require.NoError(t, err)
+
+	go func() {
+		// Subscribe to the replay clients events
+		replayObs := replayClient.EventsSequence(ctx)
+		replaySub := replayObs.Subscribe(ctx)
+		var previousMessage messageEvent
+		for msgEvent := range replaySub.Ch() {
```
bryanchriswhite, line 119: NIT: naming consistency:

```suggestion
		eventTypeObs := replayClient.EventsSequence(ctx)
		eventObserver := eventTypeObs.Subscribe(ctx)
		var previousMessage messageEvent
		for msgEvent := range replaySub.Ch() {
```

pkg/client/events/replay_client_integration_test.go
```diff
@@ -0,0 +1,147 @@
...
+			previousMessage = msgEvent
+
+			require.NotEmpty(t, msgEvent)
+			received := eventsReceived.Add(1)
+			if received >= eventsToRecv {
+				errCh <- nil
+				return
+			}
+		}
+	}()
```
bryanchriswhite, line 134: I think this can be expressed more and concisely as a `ForEach()`. Additionally, It looks like the event was being deserialized but for no real reason.

```suggestion
	channel.ForEach(
		ctx,
		observable.Observable[messageEvent](replayClient.EventsSequence(ctx)),
		func(ctx context.Context, msgEvent messageEvent) {
			require.NotEmpty(t, msgEvent)
			received := eventsReceived.Add(1)
			if received >= eventsToRecv {
				errCh <- nil
				return
			}
		},
	)
```

bryanchriswhite (approved):
Last last comment from me! :vulcan_salute: :ship:

pkg/client/events/replay_client.go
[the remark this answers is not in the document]
Olshansk: If you can think of a way to update the existing comment to make it clearer by integrating this explanation, #PUC

pkg/client/events/replay_client.go
```diff
@@ -142,17 +144,19 @@ func (rClient *replayClient[T, R]) EventsSequence(ctx context.Context) R {
...
-			publishCh <- event
-		}
-	}
+	channel.ForEach[R](
+		ctx,
+		rClient.replayObsCache,
+		func(ctx context.Context, eventTypeObs R) {
+			// TODO_CONSIDERATION: hold a reference to the previous event type
+			// observable and call `#UnsubscribeAll()` here in the event that it
+			// somehow starts receiving notifications again.
```
Olshansk, line 153: `somehow starts receiving notifications again.` 

Is this a bug you encountered?

pkg/client/events/replay_client_integration_test.go
Olshansk: #PUC with bryan's explanation

pkg/client/events/replay_client.go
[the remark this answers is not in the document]
Olshansk: So when we have a replay observable of observables, is there a use case of it being greater than one?

(no changes necessary, just trying to understand)

Olshansk (approved):
@h5law Please see my last NITs w.r.t comments + Bryan's comment, but otherwise let's get it in!

pkg/client/events/replay_client.go
bryanchriswhite: So long as we're using the gorilla websocket implementation and their comment about **all** receive errors being persistent, this should not be possible. However, if that assumption fails, due to the fact that obervables don't have a "closed" state (either has observers or not), a non-persistent error could cause the retry loop to trigger which would result in this scenario.

bryanchriswhite: Maybe #PUC with some deets from :point_up_2:

pkg/client/events/replay_client.go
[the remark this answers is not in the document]
bryanchriswhite: Correct, the idea was that we could just always have the most up-to-date reference to the "active" event data stream by calling `#Last()`. However, what we discovered is that there's a race between the caller and when the observable is notified with the new reference. This is the reason that `EventsReplayClient#EventsSequence()` now composes and returns a new observable instead, it perfectly synchronizes where the race was before.

As we're no longer using the replay-specific API (i.e. `#Last()`), It doesn't matter anymore and we should be able to simplify this again in a future iteration.

pkg/client/events/replay_client.go
h5law: Implemented instead

pkg/client/events/replay_client_integration_test.go
```diff
@@ -69,13 +69,20 @@ func TestReplayClient_Remapping(t *testing.T) {
 	connMock, dialerMock := testeventsquery.NewNTimesReconnectMockConnAndDialer(t, 2, &connClosed, &firstEventDelayed)
 	// Mock the connection receiving events
 	connMock.EXPECT().Receive().
+		// Receive is called in the tightest loop possible (max speed limited
+		// by a goroutine) and as such the sleep's within are used to slow down
+		// the time between events to prevent unexpected behavior. As in this
+		// test environment, there are no "rea" delays between "#Receive" calls
```
Olshansk, line 75:
```suggestion
		// test environment, there are no "real" delays between "#Receive" calls
```

Olshansk (approved):
@h5law Can you merge with main & rerun e2e tests?

h5law:
@Olshansk @okdas merged with main and still no luck on devnet e2e tests - locally fine tho so....

okdas:
> merged with main and still no luck on devnet e2e tests - locally fine tho so....

The issue has been resolved, update from Discord:
Screenshot 2023-12-22 at 12 27 06 PM

pkg/client/block/client_integration_test.go
```diff
@@ -19,6 +17,7 @@ import (
 const blockIntegrationSubTimeout = 5 * time.Second
 
 func TestBlockClient_LastNBlocks(t *testing.T) {
+	t.Skip("TODO(@h5law): Figure out how to subscribe to events on the simulated localnet")
```
Olshansk, line 20: Nice

Olshansk (approved):
🚢

PR #272, opened 2023-12-14
[Cleanup] Centralzie websocket url -> endpoint changes

Olshansk:
### Summary generated by Reviewpad on 14 Dec 23 04:16 UTC

- Removed unused variables and imports in `pkg/deps/config/suppliers.go`
- Deleted `pkg/deps/config/types.go`
- In `pkg/relayer/config/relayminer_configs_reader.go`, replaced the `fmt.Sprintf` call with `sdk.HostToWebsocketURL`
- In `pkg/relayer/proxy/relay_verifier.go`, replaced the import for `github.com/noot/ring-go` with `ring "github.com/noot/ring-go"`
- Removed unused imports and the function `queryNodeToWebsocketURL` in `pkg/sdk/deps_builder.go`
- Created a new file `pkg/sdk/urls.go` with a function `HostToWebsocketURL` which converts the provided host into a websocket URL that can be used for various purpose

## Issue

- #{ISSUE_NUMBER}

## Type of change

Select one or more:

- [ ] Bug fix
- [x] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

## Sanity Checklist


pkg/sdk/urls.go
```diff
@@ -0,0 +1,10 @@
+package sdk
+
+import "fmt"
+
+// HostToWebsocketURL converts the provided host into a websocket URL that can
+// be used to subscribe to onchain events and query the chain via a client
+// context or send transactions via a tx client context.
+func HostToWebsocketURL(host string) string {
```
red-0ne, line 8: I believe, it would be more robust to pass `*url.URL` down to this function which will ensure using the `Host` component to build the websocket URL. Its name would probably change too.

WDYT?

Olshansk: I would disagree for multiple reasons:

1. We have some usages where we use a string directly. For example: `sdk.HostToWebsocketURL(queryHost)`
2. I don't want introduce an added dependency in this little package.
3. What if the source of the URL varies (url.URL, string, etc...)

Issue #274, opened 2023-12-14
[Configs] Document & Productionize Staking & Operational Configurations


Olshansk:
## Objective

Productionize & document our configurations

## Origin Document

<img width="851" alt="Screenshot 2023-12-14 at 3 37 02 PM" src="https://github.com/pokt-network/poktroll/assets/1892194/ff897b5f-b1d5-4919-8fc6-f5d88f3849b0">

https://www.notion.so/buildwithgrove/Actors-configs-dependencies-etc-ab9a93322b8744a2aaa089d5b4758b75?pvs=4

## Goals

- Settle on the MVP configs for staking configs:
	- Application
	- Gateway
	- Supplier
- Settle on the MVP configs for operational configs
	- AppGate
	- RelayMiner

## Deliverables

- [x] Finalize the configurations we are using & update the source code appropriately
- [x] Document & provide examples of all our configuations

## Non-goals / Non-deliverables

- Adding any new features or functionality

## General deliverables

- [ ] **Comments**: Add/update TODOs and comments alongside the source code so it is easier to follow.
- [ ] **Testing**: Add new tests (unit and/or E2E) to the test suite.
- [ ] **Makefile**: Add new targets to the Makefile to make the new functionality easier to use.
- [ ] **Documentation**: Update architectural or development READMEs; use [mermaid](https://mermaid-js.github.io/mermaid/) diagrams where appropriate.

---

**Creator**: @Olshansk 
**Co-Owners**: @red-0ne @bryanchriswhite

Issue #271, opened 2023-12-13
[RelayMiner] Integrate new RelayMiner configuration


red-0ne:
## Objective

The current  `RelayMiner` naive configuration file structure and parser is lacking capabilities to handle real life use cases like reverse proxy usage, server containerization and elaborate backend setups.

## Origin Document

The new config discussion happened in this page:
https://www.notion.so/buildwithgrove/Actors-configs-dependencies-etc-ab9a93322b8744a2aaa089d5b4758b75?pvs=4#8eaa5c5a3f4743319cc1e8360d058be3

Final config structure proposed:
https://github.com/pokt-network/poktroll/pull/246

```yaml
pocket:
  query_node_rpc_url: tcp://localhost:36657
  query_node_grpc_url: tcp://127.0.0.1:36658
  tx_node_grpc_url: tcp://127.0.0.1:36658
signing_key_name: servicer1
proxies:
  - name: http-example
    host: 127.0.0.1
    port: 8080
    type: http 
suppliers:
  - name: ethereum
    type: http
    service_config:
      url: http://anvil.servicer:8545
      authentication:
        username: user
        password: pwd
      headers: {}
    hosts:
      - ethereum.devnet1.poktroll.com 
      - ethereum # <- this part can be added automatically.
    proxy_names:
      - http-example # when the RelayMiner server builder runs.
  - name: 7b-llm-model
    type: http
    service_config:
      url: http://llama-endpoint
    hosts:
      - 7b-llm-model.devnet1.poktroll.com 
      - 7b-llm-model # <- this part can be added automatically.
    proxy_names:
      - http-example # when the RelayMiner server builder runs.
```

## Goals

- Be able to parse the new config structure
- Adapt the `RelayerProxy` to handle the new configuration and start the adequate servers

## Deliverables

- [x] PR for parsing the config into the adequate structure
- [x] PR for adapting the `RelayerProxy` to use the new config

## Non-goals / Non-deliverables

- Implement additional types other than `http`
- Implement authentication mechanisms beyond basic auth

## General deliverables

- [x] **Comments**: Add/update TODOs and comments alongside the source code so it is easier to follow.
- [x] **Testing**: Add new tests (unit and/or E2E) to the test suite.

---

**Creator**: @red-0ne
**Co-Owners**: @okdas

red-0ne:
Closed by #284

PR #257, opened 2023-12-11
Use gRPC client connection

red-0ne:
### Human Summary

* Add gRPC connection `FlagGRPC` and `FlagGRPCInsecure` to `AppGateServer` and `RelayMiner` `cmd.go` to make `cosmosClient.Context` build a gRPC client.
* Change `AppGateServer` config files to accept
  * `query_node_grpc_url` to be used by `QueryClient`s
  * `query_node_rpc_url` to be used by `BlockClient` and `EventsQueryClient`
  * Adapt its config reader and tests
* Change `RelayMiner` config files to accept
  * `query_node_grpc_url` for `QueryClient`s usage
  * `tx_node_grpc_url` for transactions usage
  * `query_node_rpc_url` for `BlockClient` and `EventsQueryClient`
  * Adapt its config reader and tests
* Adapt `deps/config/supplier.go` to handle these changes

_Note: e2e tests for this change need changes from `helm-charts` repo to enable gRPC connections_

### Summary generated by Reviewpad on 21 Dec 23 01:11 UTC
This pull request modifies multiple files in the codebase. Here is a summary of the changes:

1. `values-relayminer.yaml`:
   - Replaced `query_node_url` and `network_node_url` properties with `query_node_rpc_url`, `query_node_grpc_url`, and `tx_node_grpc_url`.

2. `relayminer_config.yaml`:
   - Added `query_node_rpc_url`, `query_node_grpc_url`, and `tx_node_grpc_url` properties for Pocket node URLs.
   - Removed `network_node_url` property.
   - Updated the value of `query_node_url` property to `tcp://poktroll-sequencer:36657`.
   - Removed duplicate `smt_store_path` property.
   - Removed `proxied_service_endpoints` section.
   - Removed `svc1` endpoint from the `proxied_service_endpoints` section.

3. `RelayMinerConfig` structure and its related functionality:
   - Renamed `QueryNodeUrl` field to `QueryNodeRPCUrl`.
   - Added new fields `QueryNodeGRPCUrl` and `TxNodeGRPCUrl`.
   - Changed `ProxiedServiceEndpoints` field type from `map[string]string` to `map[string]*url.URL`.
   - Added error handling logic for empty or invalid field values.

4. `appgate_configs_reader.go`:
   - Added new fields `QueryNodeRPCUrl` and `QueryNodeGRPCUrl` of type `string` to the `YAMLAppGateServerConfig` struct.
   - Changed the order of fields in the `YAMLAppGateServerConfig` struct.
[...]

PR #246, opened 2023-12-06
[META] Add a new relayminer config

okdas:
### Human Summary

A proposed configuration for RelayMiner

### Summary generated by Reviewpad on 06 Dec 23 01:58 UTC
This pull request adds a new configuration file for the relay miner beta. It includes considerations for future soft config changes, proxy settings, and supplier configurations. The file also defines proxies and their properties, as well as suppliers and their connections to proxies.

## Type of change

Select one or more:

- [ ] New feature, functionality or library
- [ ] Bug fix
- [x] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Sanity Checklist

- [ ] I have tested my changes using the available tooling
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, updated documentation and left TODOs throughout the codebase

PR #284, opened 2023-12-19
[Configs] Add foundation for RelayMiner operation configs.

red-0ne:
### Human Summary

* Implement new `RlayMiner` configuration (#246) parser
* Adapt `RelayMiner` server initialization to
  * Source from the config
  * Check if the config matches the on-chain `Supplier`'s data
  * Support multiple proxy server instances 
* Adapt the `RelayMiner` config parser tests
* Adapt the `RelayMiner` unit tests

This PR is based on #257 and deprecates the `RelayMiner` part of it.

## Issue

- #274
- #271

## Type of change

Select one or more:

- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

## Sanity Checklist


pkg/relayer/config/relayminer_configs_reader.go
```diff
@@ -4,98 +4,264 @@ import (
...
+	if err != nil {
+		return nil, ErrRelayMinerConfigInvalidNodeUrl.Wrapf(
+			"invalid query node rpc url %s",
+			err.Error(),
+		)
 	}
 
-	if yamlRelayMinerConfig.ProxiedServiceEndpoints == nil {
-		return nil, ErrRelayMinerConfigInvalidServiceEndpoint.Wrapf("proxied service endpoints are required")
+	// Proxies section
```
Olshansk, line 81: This might require a refactor, so want to hear your thoughts first. 

Rather than having comments for sections, what if you create small helper functions and have `RelayMinerConfig,` be the receiver? Then the code might look like:

```go
err := relayMinerConfig.HydrateProxies(yamlConfig)
if err != nil {...}
relayMinerConfig.HydrateSuppliers(yamlConfig)
if err != nil {...}
```

pkg/relayer/config/relayminer_configs_reader.go
```diff
@@ -4,98 +4,264 @@ import (
...
+	// At least one proxy is required
+	if len(yamlRelayMinerConfig.Proxies) == 0 {
+		return nil, ErrRelayMinerConfigInvalidProxy.Wrap("no proxies provided")
+	}
+
+	proxies := yamlRelayMinerConfig.Proxies
```
Olshansk, line 87: is the copy necessary or can we just iterate over the yaml data directly?

pkg/relayer/config/types.go
```diff
@@ -0,0 +1,132 @@
+package config
+
+import "net/url"
+
+// YAMLRelayMinerConfig is the structure used to unmarshal the RelayMiner config file
+// TODO_DOCUMENT(@red-0ne): Add proper README documentation for yaml config files.
```
Olshansk, line 6:
```suggestion
// TODO_DOCUMENT(@red-0ne): Add proper README documentation for yaml config files and update inline comments as well.
```

pkg/relayer/config/types.go
```diff
@@ -0,0 +1,132 @@
...
+type YAMLRelayMinerConfig struct {
+	SigningKeyName string                         `yaml:"signing_key_name"`
+	SmtStorePath   string                         `yaml:"smt_store_path"`
+	Pocket         YAMLRelayMinerPocketConfig     `yaml:"pocket"`
```
Olshansk, line 10: Not a fan of the word `Pocket` since it's too generic/overloaded.

How about `Node` or PocketNode`?

pkg/relayer/config/types.go
```diff
@@ -0,0 +1,132 @@
...
+	Proxies        []YAMLRelayMinerProxyConfig    `yaml:"proxies"`
+	Suppliers      []YAMLRelayMinerSupplierConfig `yaml:"suppliers"`
+}
+
+// YAMLRelayMinerPocketConfig is the structure used to unmarshal the pocket
+// node URLs section of the RelayMiner config file
+type YAMLRelayMinerPocketConfig struct {
+	QueryNodeGRPCUrl string `yaml:"query_node_grpc_url"`
```
Olshansk, line 18: We'll need to add inline comments here (not now) when we add the proper documentation as well.

pkg/relayer/config/types.go
```diff
@@ -0,0 +1,132 @@
...
+	// back to the Host header.
+	XForwardedHostLookup bool
+}
+
+// RelayMinerSupplierConfig is the structure resulting from parsing the supplier
+// section of the RelayMiner config file.
+type RelayMinerSupplierConfig struct {
+	// Name is the serviceId corresponding to the current configuration.
+	Name string
+	// Type is the transport protocol used by the supplier, it must match the
```
Olshansk, line 100: Why don't we use a `golang iota` for this or a `protobuf enum`?

pkg/relayer/config/types.go
```diff
@@ -0,0 +1,132 @@
...
+	// Hosts is a list of hosts advertised on-chain by the supplier, the corresponding
+	// proxy server will accept relay requests for these hosts.
+	Hosts []string
+	// ServiceConfig is the config of the service that relays will be proxied to.
+	ServiceConfig *RelayMinerSupplierServiceConfig
+}
+
+// RelayMinerSupplierServiceConfig is the structure resulting from parsing the supplier
+// service sub-section of the RelayMiner config file.
+// If the supplier is of type "http", it may embed a basic auth structure and
```
Olshansk, line 112: I think this is TMI (Too much Information). The reason we have attribute specific comments is to capture this sort of detail.

pkg/relayer/proxy/errors.go
```diff
@@ -15,4 +15,6 @@ var (
 	ErrRelayerProxyInvalidRelayRequest               = sdkerrors.Register(codespace, 7, "invalid relay request")
 	ErrRelayerProxyInvalidRelayResponse              = sdkerrors.Register(codespace, 8, "invalid relay response")
 	ErrRelayerProxyEmptyRelayRequestSignature        = sdkerrors.Register(codespace, 9, "empty relay response signature")
+	ErrRelayerProxyServiceEndpointNotHandled         = sdkerrors.Register(codespace, 10, "service endpoint not handled by relayer proxy")
+	ErrRelayerProxyUnsupportedTransportType          = sdkerrors.Register(codespace, 11, "unsupported proxy transport type")
```
Olshansk, line 19: perf my comment above, I really do think we should be using an enum (golang iota) for this.

pkg/relayer/proxy/server_builder.go
```diff
@@ -28,58 +31,68 @@ func (rp *relayerProxy) BuildProvidedServices(ctx context.Context) error {
...
+					endpoint.Url,
 				)
-			default:
-				return ErrRelayerProxyUnsupportedRPCType
 			}
+		}
+	}
+
+	// Build the advertised relay servers map. For each service's endpoint, create the appropriate RelayServer.
+	providedServices := make(map[string]relayer.RelayServer)
```
Olshansk, line 70: Another opportunity to use smaller helper functions

pkg/relayer/proxy/synchronous.go
```diff
@@ -82,18 +85,44 @@ func (sync *synchronousRPCServer) Stop(ctx context.Context) error {
...
-
 // ServeHTTP listens for incoming relay requests. It implements the respective
 // method of the http.Handler interface. It is called by http.ListenAndServe()
 // when synchronousRPCServer is used as an http.Handler with an http.Server.
 // (see https://pkg.go.dev/net/http#Handler)
 func (sync *synchronousRPCServer) ServeHTTP(writer http.ResponseWriter, request *http.Request) {
 	ctx := request.Context()
 
+	var originHost string
+	// Get the original host from X-Forwarded-Host header if specified in the proxy config.
```
Olshansk, line 96: Why do we ned to do this?

pkg/relayer/proxy/synchronous.go
```diff
@@ -82,18 +85,44 @@ func (sync *synchronousRPCServer) Stop(ctx context.Context) error {
...
+	}
+
+	if originHost == "" {
+		originHost = request.Host
+	}
+
+	var supplierService *sharedtypes.Service
+	var serviceUrl *url.URL
+
+outer:
```
Olshansk, line 109: I'm not a fan of using labels in production code. IMO it means we can either:
1. Use a helper that returns a couple variables.
2. Check in the outer loop if there's a value set and break then

Olshansk (changes requested):
@red-0ne Posting a partial review and will finish it later.

Overall, looks good w/ no real blockers but just some cleanup. I think there's an opportunity to use smaller compartmentalized helper functions to help with code maintainability.

Also, I'd rename the PR title to `[Configs] Add foundation for RelayMiner operation configs.`

Screenshot 2023-12-19 at 5 30 05 PM

pkg/relayer/cmd/cmd.go
```diff
@@ -63,9 +64,11 @@ for such operations.`,
 	cmd.Flags().StringVar(&flagRelayMinerConfig, "config", "", "The path to the relayminer config file")
 
 	// Cosmos flags
+	// TODO_TECHDEBT(#256): Remove unneeded cosmos flags.
 	cmd.Flags().String(cosmosflags.FlagKeyringBackend, "", "Select keyring's backend (os|file|kwallet|pass|test)")
-	cmd.Flags().
-		StringVar(&flagCosmosNodeURL, cosmosflags.FlagNode, omittedDefaultFlagValue, "Register the default Cosmos node flag, which is needed to initialize the Cosmos query and tx contexts correctly. It can be used to override the `QueryNodeUrl` and `NetworkNodeUrl` fields in the config file if specified.")
+	cmd.Flags().StringVar(&flagNodeRPCUrl, cosmosflags.FlagNode, omittedDefaultFlagValue, "Register the default Cosmos node flag, which is needed to initialize the Cosmos query and tx contexts correctly. It can be used to override the `QueryNodeUrl` and `NetworkNodeUrl` fields in the config file if specified.")
```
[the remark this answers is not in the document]
red-0ne: Renamed to `txNodeGRPCUrl`

pkg/relayer/config/proxy_http_config_parser.go
```diff
@@ -0,0 +1,67 @@
+package config
+
+import (
+	"fmt"
+	"net/url"
+)
+
+// parseRelayMinerConfigs populates the proxy fields of the target structure that
```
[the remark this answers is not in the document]
red-0ne: Comments updated.

This function is actually handling the proxy config when its type is `http`, where `Host` needs to be a valid `http` host (eg. not a UNIX socket`.

In the future, when we support `https` for example, a similar `https` dedicated function would look for `tls` configs and knows how to parse them.

As to the `tcp://` scheme, it's only to have a valid url to parse, and make sure its host is valid. Added a comment and prepended a `http` scheme instead.

pkg/relayer/config/relayminer_configs_reader.go
```diff
@@ -4,98 +4,264 @@ import (
...
+	// SigningKeyName is required
+	if yamlRelayMinerConfig.SigningKeyName == "" {
+		return nil, ErrRelayMinerConfigInvalidSigningKeyName
 	}
 
-	// Parse the query node URL
-	queryNodeUrl, err := url.Parse(yamlRelayMinerConfig.QueryNodeUrl)
-	if err != nil {
-		return nil, ErrRelayMinerConfigInvalidQueryNodeUrl.Wrapf("%s", err)
+	// SmtStorePath is required
```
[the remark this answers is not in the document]
red-0ne: Not sure this is where we should indicate if it's relative or absolute path, unless we check the directory existence which I'm not against.

pkg/relayer/config/relayminer_configs_reader.go
```diff
@@ -4,98 +4,264 @@ import (
...
+			err.Error(),
+		)
 	}
 
-	// Parse the websocket URL of the Pocket Node to connect to for subscribing to on-chain events.
-	pocketNodeWebsocketUrl := sdk.HostToWebsocketURL(queryNodeUrl.Host)
+	// If the query node grpc url is empty, use the tx node grpc url
+	if pocket.QueryNodeGRPCUrl == "" {
+		relayMinerPocketConfig.QueryNodeGRPCUrl = relayMinerPocketConfig.TxNodeGRPCUrl
+	} else {
```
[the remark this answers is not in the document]
red-0ne: Not sure what you want to achieve here.

We want to use `TxNodeGRPCUrl` which is already validated above., if `QueryNodeGRPCUrl` is not set.
Otherwise (`QueryNodeGRPCUrl` has its own value) we need to make sure it's a valid URL.

pkg/relayer/proxy/errors.go
red-0ne: Do you mean, getting rid of `sdkerrors.Register`? How to not lose the info captured by the `Register` error description?

pkg/relayer/proxy/synchronous.go
red-0ne: Updated the comments to explain why we have this.

testutil/testproxy/relayerproxy.go
```diff
@@ -209,14 +216,25 @@ func WithDefaultSessionSupplier(
 // MarshalAndSend marshals the request and sends it to the provided service
```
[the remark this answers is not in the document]
Olshansk: #PUC with a TODO and/or explanation

pkg/relayer/config/relayminer_configs_reader.go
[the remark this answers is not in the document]
Olshansk: I thought it was just an easier way to write the code (more readable) but feel free to ignore if that's not the case from your POV.

pkg/relayer/proxy/errors.go
Olshansk: Not for the errors, for the transport type.

red-0ne: Done

pkg/relayer/config/proxies_config_hydrator.go
```diff
@@ -0,0 +1,63 @@
+package config
```
Olshansk, line 1: All of these hydrators are SO easy to read and understand!

pkg/relayer/config/relayminer_configs_reader_test.go
```diff
@@ -16,7 +16,7 @@ func Test_ParseRelayMinerConfigs(t *testing.T) {
 	tests := []struct {
```
Olshansk, line 16: I'm going to skip reviewing this file 😅

pkg/relayer/config/relayminer_configs_reader.go
```diff
@@ -1,102 +1,68 @@
...
 
-	relayMinerCMDConfig := &RelayMinerConfig{
-		QueryNodeUrl:            queryNodeUrl,
-		NetworkNodeUrl:          networkNodeUrl,
-		PocketNodeWebsocketUrl:  pocketNodeWebsocketUrl,
-		SigningKeyName:          yamlRelayMinerConfig.SigningKeyName,
-		ProxiedServiceEndpoints: proxiedServiceEndpoints,
-		SmtStorePath:            yamlRelayMinerConfig.SmtStorePath,
+	// Check if a proxy is referencing a host more than once
+	for _, proxyConfig := range relayMinerConfig.Proxies {
```
Olshansk, line 51: Can you add a `Validate` function (with the appropriate receiver as the config) so this is just a one liner that returns an error?

pkg/relayer/config/types.go
```diff
@@ -0,0 +1,141 @@
+package config
+
+import "net/url"
+
+type ProxyType int
+
+const (
+	ProxyTypeHTTP ProxyType = iota
```
Olshansk, line 8: Please add a TODO on what we want to support in the future

pkg/relayer/config/types.go
```diff
@@ -0,0 +1,141 @@
...
+}
+
+// RelayMinerSupplierServiceConfig is the structure resulting from parsing the supplier
+// service sub-section of the RelayMiner config file.
+type RelayMinerSupplierServiceConfig struct {
+	// Url is the URL of the service that relays will be proxied to.
+	Url *url.URL
+	// Authentication is the basic auth structure used to authenticate to the
+	// request being proxied from the current proxy server.
+	// If `RelayMinerSupplierConfig.Type` is `ProxyTypeHTTP` then `Authentication`
```
Olshansk, line 126: Whe we go type specific, can you add a `For example`. Ditto below

localnet/poktrolld/config/relayminer_config_full_example.yaml
```diff
@@ -0,0 +1,108 @@
...
+  tx_node_grpc_url: tcp://sequencer-poktroll-sequencer:36658
+
+# Proxies are endpoints that expose different suppliers to the internet.
+proxies:
+    # Name of the proxy. It will be used to reference in a supplier. Must be unique.
+    # Required.
+    # TODO_CONSIDERATION: if we enforce DNS compliant names, it can potentially
+    # become handy in the future.
+    # More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names
+  - name: http-example
```
Olshansk, line 29: Can you add a `TODO_CONSIDERATION: Rename to supplier_name`

localnet/poktrolld/config/relayminer_config_full_example.yaml
```diff
@@ -0,0 +1,108 @@
...
+  #     key: /path/to/key
+
+# Suppliers are different services that can be offered through RelayMiner.
+# When a supplier is configured to use a proxy and staked appropriately,
+# the relays will start flowing through RelayMiner.
+suppliers:
+    # Name of the supplier offered to the network.
+    # Must be unique.
+    # Required.
+  - name: ethereum
```
Olshansk, line 57: Does this nee dto match the on-chain `ServiceId`? If so, #PUC

localnet/poktrolld/config/relayminer_config_full_example.yaml
```diff
@@ -0,0 +1,108 @@
...
+    # Type of how the supplier offers service through the network.
+    # Must match the type of the proxy the supplier is connected to.
+    # Required.
+    type: http
+    # Configuration of the service offered through RelayMiner.
+    service_config:
+      # URL RelayMiner proxies the requests to.
```
Olshansk, line 64:
```suggestion
      # URL RelayMiner proxies the requests to.
      # Also known as the data node, or service node in some cases.
```

localnet/poktrolld/config/relayminer_config_full_example.yaml
```diff
@@ -0,0 +1,108 @@
...
+    # Required.
+    proxy_names:
+      - http-example # when the RelayMiner server builder runs.
+  - name: 7b-llm-model
+    type: http
+    service_config:
+      url: http://llama-endpoint
+    hosts:
+      - 7b-llm-model.devnet1.poktroll.com
+      # - 7b-llm-model # <- this part can be added automatically.
```
Olshansk, line 106: can or will?

I'm still a bit confused why hosts is a list so #PUC

localnet/poktrolld/config/relayminer_config_full_example.yaml
```diff
@@ -0,0 +1,108 @@
...
+    # request to the correct supplier.
+    # That hostname is what the user should stake the supplier for.
+    # Must be unique within a proxy/proxies it is set up on;
+    # in other words, one proxy can't offer the same hostname more than once.
+    # The `name` of the supplier is automatically added to the hosts section
+    # for potential troubleshooting/debugging purposes
+    # Required.
+    hosts:
+      - ethereum.devnet1.poktroll.com
+      # - ethereum # <- this part is be added automatically.
```
Olshansk, line 92: Can you explain WHY? for what purposes / use case?
```suggestion
      # - ethereum # <- this part is added automatically.
```

localnet/poktrolld/config/relayminer_config_full_example.yaml
```diff
@@ -0,0 +1,108 @@
...
+      headers: {}
+
+    # A list of hosts the HTTP service is offering.
+    # When linked to the proxy, that hostname is going to be used to route the
+    # request to the correct supplier.
+    # That hostname is what the user should stake the supplier for.
+    # Must be unique within a proxy/proxies it is set up on;
+    # in other words, one proxy can't offer the same hostname more than once.
+    # The `name` of the supplier is automatically added to the hosts section
+    # for potential troubleshooting/debugging purposes
```
Olshansk, line 88: This is a bit wordy and confusing.

When we stake "stake", try to add details on "has the on-chain equivalent when staking .... "

localnet/poktrolld/config/relayminer_config_full_example.yaml
```diff
@@ -0,0 +1,108 @@
...
+      # Optional.
+      headers: {}
+
+    # A list of hosts the HTTP service is offering.
+    # When linked to the proxy, that hostname is going to be used to route the
+    # request to the correct supplier.
+    # That hostname is what the user should stake the supplier for.
+    # Must be unique within a proxy/proxies it is set up on;
+    # in other words, one proxy can't offer the same hostname more than once.
+    # The `name` of the supplier is automatically added to the hosts section
```
Olshansk, line 87: Can you move this comment next to the name section and provide more details on why & how it can be used for troubleshooting? E.g. updating /etc/hosts or a custom k8s cluster, etc...

Olshansk (changes requested):
I really like this after the last iteration. 

Can't say I reviewed everything line by line, but it's probably the best codebase for configurations that I've ever seen!

localnet/poktrolld/config/appgate_server_config_example.yaml
```diff
@@ -0,0 +1,12 @@
+# Pocket node URL that exposes CometBFT JSON-RPC API.
+# This can be used by the Cosmos client SDK, event subscriptions, etc...
+query_node_rpc_url: tcp://sequencer-poktroll-sequencer:36657
+# Pocket node URL that exposes the Cosmos gRPC service, dedicated to querying purposes.
+# If unspecified, defaults to `tx_node_grpc_url`.
```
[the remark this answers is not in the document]
red-0ne: Yes, `AppGateServer` don't need a tx node.

pkg/relayer/interface.go
```diff
@@ -64,7 +64,7 @@ type RelayerProxy interface {
 	VerifyRelayRequest(
 		ctx context.Context,
 		relayRequest *servicetypes.RelayRequest,
-		service *sharedtypes.Service,
+		serviceId *sharedtypes.Service,
```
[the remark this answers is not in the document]
red-0ne: Good catch!

pkg/relayer/config/types.go
red-0ne: Good catch! I confused it with proxy type related fields. This is independent of the proxy type indeed.

localnet/poktrolld/config/relayminer_config_full_example.yaml
red-0ne: It would rather be `proxy_name` since it identifies a `proxy`

localnet/poktrolld/config/relayminer_config_full_example.yaml
red-0ne: This **is** added automatically.

@okdas  may provide more arguments.

The main reasons I see are
* That the operator may want to route the requests coming from these domains to a single proxy
* Migrating to a new domain, where requests may still arrive from the old one 
* Having internal requests coming from non FQDNs because of complex routing and debugging purposes.

localnet/poktrolld/config/relayminer_config_full_example.yaml
red-0ne: Moved to the right section but will let @okdas explain how this is useful. I will then update the comment accordingly.

localnet/poktrolld/config/relayminer_config_full_example.yaml
okdas: @red-0ne is correct. If we can support multiple, I'd say this is the way. It allows for easier migrations across domains, it gives more flexibility. Kubernetes Ingresses work the same way, heres what ChatGPT has to say about why that is: https://chat.openai.com/share/286502d2-021b-4caa-8c90-eb6138a57c01 - not all points apply to our case (yet)

localnet/poktrolld/config/relayminer_config_full_example.yaml
```diff
@@ -21,9 +21,11 @@ pocket_node:
 
 # Proxies are endpoints that expose different suppliers to the internet.
 proxies:
-    # Name of the proxy. It will be used to reference in a supplier. Must be unique.
+    # Name of the proxy. It will be used to reference in a supplier.
+    # Must be unique.
     # Required.
     # TODO_CONSIDERATION: if we enforce DNS compliant names, it can potentially
+    # TODO_CONSIDERATION: Rename to `proxy_name`
```
Olshansk, line 28: I personally think we should do it. Wdyt @okdas @red-0ne?

localnet/poktrolld/config/relayminer_config_full_example.yaml
```diff
@@ -78,18 +83,25 @@ suppliers:
...
-    # When linked to the proxy, that hostname is going to be used to route the
-    # request to the correct supplier.
-    # That hostname is what the user should stake the supplier for.
-    # Must be unique within a proxy/proxies it is set up on;
-    # in other words, one proxy can't offer the same hostname more than once.
-    # The `name` of the supplier is automatically added to the hosts section
-    # for potential troubleshooting/debugging purposes
+    # A list of hosts the proxy is accepting requests from.
+    # When linked to the proxy, each host is going to be used to lookup the
+    # the Supplier.Service in the Pocket Network.
```
Olshansk, line 88:
```suggestion
    # the Supplier.Service in Pocket Network.
```

localnet/poktrolld/config/relayminer_config_full_example.yaml
```diff
@@ -78,18 +83,25 @@ suppliers:
...
+    # Each host in the list must match a Supplier.Service.Endpoint that the Supplier
+    # has advertised on-chain when staking for that Service.
+    # The reasons to have multiple hosts for the same supplier service are:
```
Olshansk, line 91: Same comment as in #288

Screenshot 2023-12-22 at 4 21 03 PM

localnet/poktrolld/config/relayminer_config_full_example.yaml
Olshansk: @red-0ne I know you already documneted a subset of these, but I think there's opportunity to add even a few more of the points here. Make sure to add the appropriate qualifiers (see my comment)

Olshansk (changes requested):
Will this require additional changes if we go with `proxy_name` and `service_id` instead of `name` or would we want to do this in a followup PR?

localnet/poktrolld/config/relayminer_config_full_example.yaml
red-0ne: Updated on the `ethereum` example entry

Olshansk (approved):
Thanks for the multiple iterations on this!

localnet/kubernetes/values-relayminer.yaml
```diff
@@ -1,3 +1,20 @@
...
+      host: 0.0.0.0:8545
+  suppliers:
+    - service_id: anvil
+      type: http
+      service_config:
+        url: http://anvil:8547/
+      proxy_names:
+        - http-proxy
+      hosts:
+        - tcp://relayminers:8545
```
okdas, line 20: Should this be just `relayminers`?

red-0ne: Since we don't have any proxy/load banalncer yet, the port needs to be added here. Also the `tcp://` scheme is here for url validation, it is then stripped to only take the host part.

Olshansk (approved):
Let's frikkin 🚢 this thing!

Issue #301, opened 2024-01-05
[Infra] Rollkit image used in LocalNet fails DNS name resolution of kubernetes services 


adshmh:
## Objective

Celestia secret is created by the `secret-update-sidecar` container.

## Origin Document

The `secret-update-sidecar` container fails to create the Celestia secret, causing failure in downstream containers like Sequencer. This seems to be due to DNS name resolution failing for k8s services in the Rollkit image used in LocalNet.

Pod used for testing:

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: dnsutils
  namespace: default
spec:
  containers:
  - name: dnsutils
    image: ghcr.io/rollkit/local-celestia-devnet:v0.12.1
    command:
      - sleep
      - "infinity"
    imagePullPolicy: IfNotPresent
  restartPolicy: Always
```

Name resolution failure:

```bash
kubectl exec -it pod/dnsutils -- /bin/bash -c 'nslookup kubernetes.default'
Server:         10.96.0.10
Address:        10.96.0.10:53

** server can't find kubernetes.default: NXDOMAIN
```

The same pod template, with a different image, succeeds (image from https://kubernetes.io/docs/tasks/administer-cluster/dns-debugging-resolution/)

```image:  registry.k8s.io/e2e-test-images/jessie-dnsutils:1.3```

Name resolution success:

```
kubectl exec -it pod/dnsutils -- /bin/bash -c 'nslookup kubernetes.default'
Server:         10.96.0.10
Address:        10.96.0.10#53

Name:   kubernetes.default.svc.cluster.local
Address: 10.96.0.1
```

## Goals

- The `update-secret-sidecar` container should successfully create the secret.

## Deliverables

- [ ] The Celestia secret is successfully created, allowing the subsequent containers to operate normally.

**Creator**: @adshmh 
**Co-Owners**: @okdas @Olshansk

**Comment by okdas** [comment]
Hey @adshmh, seems like tweaking the DNS endpoint works:

```
dnsutils:/$ nslookup kubernetes.default.svc
Server:		10.96.0.10
Address:	10.96.0.10:53

** server can't find kubernetes.default.svc: NXDOMAIN

** server can't find kubernetes.default.svc: NXDOMAIN

dnsutils:/$ nslookup kubernetes.default.svc.cluster.local
Server:		10.96.0.10
Address:	10.96.0.10:53

Name:	kubernetes.default.svc.cluster.local
Address: 10.96.0.1
```

[...]

PR #314, opened 2024-01-10
[LocalNet] Update Kubernetes DNS Endpoint for secret-update-sidecar

okdas:
### Human Summary

There is an issue with older versions of busybox not being able to resolve `kubernetes.default.svc` domain under some circumstances. `kubernetes.default.svc.cluster.local` works, though!

### Summary generated by Reviewpad on 10 Jan 24 23:10 UTC
This pull request changes the Kubernetes DNS endpoint for the celestia secret in the celestia-rollkit.yaml file. The patch modifies a line of code that sets up a loop that waits for a specific node key to exist, signs a JWT, and pushes it to a Kubernetes secret. The endpoint URL is updated to use the "kubernetes.default.svc.cluster.local" instead of "kubernetes.default.svc".

## Issue

- #301

## Type of change

Select one or more:

- [x] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Sanity Checklist


Olshansk (changes requested):
@okdas Before approving, two things:

1. We should enable E2E tests to make sure it works on DevNet
2. Can you explain why this is working for some people but not others?
3. Should we add (2) as a comment inside the yaml file so others don't accidently revert it?

okdas:
>     1. We should enable E2E tests to make sure it works on DevNet
No reason to; this YAML manifest is only used for LocalNet.

>     2. Can you explain why this is working for some people but not others?
I am surprised as well. With @adshmh's example failing for me locally, the script still works in my case.
I did go through the K8S documentation when debugging the problem, and I've seen that this exact issue can happen on WSL. I can't find that particular page right now, though.
Also, there is a selection of DNS servers for K8s (CoreDNS, kube-dns, etc.), and different cluster setups with different DNS servers might potentially have different behavior. I haven't seen such issues personally, but that is a variable that can change.

>     3. Should we add (2) as a comment inside the yaml file so others don't accidently revert it?
I don't think we will touch that part in the near future. That ugly one-liner should ideally go away! There must be a better way to push a K8s secret, but that hack worked then and stuck around.

Olshansk (approved):

PR #319, opened 2024-01-12
[Linting] Enforce 120 max lines, and import order with custom linting

h5law:
### Human Summary

- Enforce max lines as 120 in all files except errors.go files (using lll)
- Enforce import order (std, external, internal) (using gci)
  - Excludes generated files
- Enforce gofumpt formatting
- Disable enforcement for certain files
- Create workflow for this for CI
- Fixes all linter errors for the entire repo so we can enable them for new PRs only

`make go_gci` will order all eligible `.go` files imports to be std,external,internal
	- here non eligible files are ones with ignite scaffold comments you need to do them yourself as `gci` will remove the comment
`make go_gofumpt` will format (scricter than gofmt) all eligible '.go' files
	- non-eligible files mostly include generated files like mocks and protobuf generated code
`make go_lint` will show any linter errors

These three makefile targets rely on you first running `make install_ci_deps` which will install `golangci-lint`, `gci` and `gofumpt` if you don't already have them. These are configured in the `.golangci.yml` file and the workflow file for running tests.

From now on PRs will (🤞🏼) show `checks` not comments where you have linter errors.

## Issue

- Fixes N/A

## Type of change

Select one or more:

- [ ] Bug fix
- [x] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

## Sanity Checklist


.golangci.yml
```diff
@@ -1,17 +1,30 @@
...
+    check-shadowing: true
+  lll:
+    line-length: 120
+    tab-width: 4
+  gci:
+    sections:
+      - standard # std lib
+      - default # external
+      - prefix(github.com/pokt-network/poktroll) # local imports
+    skip-generated: true
```
bryanchriswhite, line 26: Can we remove this makefile target and script now?
- https://github.com/pokt-network/poktroll/blob/main/Makefile#L155
- https://github.com/pokt-network/poktroll/blob/main/tools/scripts/goimports/main.go

bryanchriswhite: Or perhaps it still makes sense to have the make target to be able to attempt to fix fixable issues. :thinking:

bryanchriswhite: > I believe so will have to make sure this works first but running `golangci-lint` should do the job of these two

The reason why it's not a simple as running `golangci-lint` is because there is some additional generated code that we need to ignore. The silliest example is the multi-line import statement which ignite seems to rely on but the linter would rewrite as a single line. :roll_eyes: This is the reason for the script package's existend.

h5law: running `golangci-lint run` should do show errors and idk if it fixes for u this PR probs gunna have to include me fixing everything that doesn't fit the rules anyway

h5law: > The silliest example is the files that contain a multi-line import statement that only has one package, which ignite seems to rely on, but the linter would rewrite as a single line

An example of one of these files would help me a lot @bryanchriswhite

h5law: I have replaced the tool you mentioned and added another and same with makefiles that are relevant to the new rules

.github/workflows/run-tests.yml
```diff
@@ -44,8 +44,11 @@ jobs:
       - name: Generate mocks
         run: make go_mockgen
 
-      - name: Run golangci-lint
-        run: make go_lint
+      - name: Fix imports
+        run: make gci
+
+      - name: Format the repo
+        run: make gofumpt
```
bryanchriswhite, line 51: Does this modify the code in CI (i.e. outside of version control)?

Makefile
```diff
@@ -71,6 +72,26 @@ check_godoc:
...
+		exit 1; \
+	fi; \
+	}
+
+.PHONY: check_gofumpt
+# Internal helper target - check if gofumpt is installed
+check_gofumpt:
+	{ \
+	if ( ! ( command -v gci >/dev/null )); then \
+		echo "Seems like you don't have gci installed. Make sure you install it via 'go install mvdan.cc/gofumpt@latest' before continuing"; \
```
bryanchriswhite, line 90:
```suggestion
		echo "Seems like you don't have gofumpt installed. Make sure you install it via 'go install mvdan.cc/gofumpt@latest' before continuing"; \
```

.github/workflows/linter.yml
bryanchriswhite, line 1: Is this workflow intended to replace the "Run golangci-lint" step in the `go-test` job of the run-test.yaml workflow? If so, does it make sense to either move this into a new job in the same workflow in place of that one. My understanding is that its purpose is to validate linting before we spend any CI time building. It seems redundant to lint twice though.

Makefile
```diff
@@ -150,10 +171,15 @@ localnet_regenesis: ## Regenerate the localnet genesis file
 
 .PHONY: go_lint
 go_lint: ## Run all go linters
-	golangci-lint run --timeout 5m --build-tags test
+	golangci-lint run
```
bryanchriswhite, line 174: Why no more `--timeout` or `--build-tags`?

e2e/tests/session_steps_test.go
```diff
@@ -11,19 +11,20 @@ import (
...
 	"github.com/pokt-network/poktroll/testutil/testclient"
 	suppliertypes "github.com/pokt-network/poktroll/x/supplier/types"
-	"github.com/stretchr/testify/require"
 )
 
 const (
 	createClaimTimeoutDuration   = 10 * time.Second
 	eitherEventsReplayBufferSize = 100
-	msgClaimSenderQueryFmt       = "tm.event='Tx' AND message.sender='%s' AND message.action='/pocket.supplier.MsgCreateClaim'"
+	msgClaimSenderQueryFmt       = "tm.event='Tx' AND message.sender='%s' AND message.action='/pocket.supplier.MsgCreateClaim'" //nolint:lll
```
bryanchriswhite, line 27: #PUC anywhere we make exceptions to the rules.

pkg/relayer/miner/gen/gen_fixtures.go
```diff
@@ -108,7 +108,7 @@ func main() {
 	}
 
 	// Write the output buffer to the file at flagOut path.
-	if err := os.WriteFile(flagOut, outputBuffer.Bytes(), 0644); err != nil {
+	if err := os.WriteFile(flagOut, outputBuffer.Bytes(), 0o644); err != nil {
```
bryanchriswhite, line 111: I find the explicit octal representation less intuitive and idiomatic. Which rule enforces this?

testutil/keeper/pocket.go
```diff
@@ -18,6 +18,7 @@ import (
 	"github.com/pokt-network/poktroll/x/pocket/types"
 )
 
+// PocketKeeper returns a mocked pocket keeper and context for testing.
```
bryanchriswhite, line 21: Is it really returning a mock? It looks like a real keeper.

testutil/keeper/service.go
```diff
@@ -18,6 +18,7 @@ import (
 	"github.com/pokt-network/poktroll/x/service/types"
 )
 
+// ServiceKeeper returns a mocked service keeper and context for testing.
```
bryanchriswhite, line 21: (same as https://github.com/pokt-network/poktroll/pull/319#discussion_r1453313810)

testutil/keeper/supplier.go
```diff
@@ -22,7 +22,11 @@ import (
 	"github.com/pokt-network/poktroll/x/supplier/types"
 )
 
-func SupplierKeeper(t testing.TB, sessionByAppAddr supplier.SessionsByAppAddress) (*keeper.Keeper, sdk.Context) {
+// SupplierKeeper returns a mocked supplier keeper and context for testing.
```
bryanchriswhite, line 25: (same as https://github.com/pokt-network/poktroll/pull/319#discussion_r1453313810). It's more like it returns a keeper with mocked dependencies.

testutil/keeper/tokenomics.go
```diff
@@ -18,6 +18,7 @@ import (
 	"github.com/pokt-network/poktroll/x/tokenomics/types"
 )
 
+// TokenomicsKeeper returns a mocked tokenomics keeper and context for testing.
```
bryanchriswhite, line 21: (same as https://github.com/pokt-network/poktroll/pull/319#discussion_r1453313810)

.github/workflows/linter.yml
okdas: I agree - I think we should have linters set up in the main job.

h5law: I left a lot of comments explaining this in the file

Makefile
h5law: my bad

pkg/relayer/miner/gen/gen_fixtures.go
h5law: That was my linter again will revert

h5law: I think go prefers octals here

.github/workflows/linter.yml
```diff
@@ -0,0 +1,84 @@
...
+
+      - name: Install CI dependencies
+        run: make install_ci_deps
+
+      - name: Generate protobufs
+        run: make proto_regen
+
+      # Fetch a list of all modified files in this PR.
+      - name: Get Modified Files
+        run: git diff --name-only ${{ github.event.before }} ${{ github.sha }} > modified_files.txt
```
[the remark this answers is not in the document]
h5law: Here I need to filter the issues to not do gci check on certain files with scaffold comments then do a second lint on all files disabling gci linint

.github/workflows/run-tests.yml
h5law: Yes according to filters but yes

pkg/relayer/miner/gen/gen_fixtures.go
h5law: gofumpt does this - it is actually a go standard it seems.

.github/workflows/run-tests.yml
h5law: It did previously with the `make goimports` step so I added these if this isnt intended then Ill remove them altogether

.github/workflows/linter.yml
h5law: I can revert this but there needs to be 2 linting steps AFAIK due to the ignite scaffold comment issure gci removes this so if we don't double lint with it on file where its not present and lint without it on all files then its not all files are covered

e2e/tests/session_steps_test.go
```diff
@@ -11,22 +11,25 @@ import (
...
 
 const (
 	createClaimTimeoutDuration   = 10 * time.Second
 	eitherEventsReplayBufferSize = 100
-	msgClaimSenderQueryFmt       = "tm.event='Tx' AND message.sender='%s' AND message.action='/pocket.supplier.MsgCreateClaim'"
-	testServiceId                = "anvil"
-	eitherEventsBzReplayObsKey   = "eitherEventsBzReplayObsKey"
-	preExistingClaimsKey         = "preExistingClaimsKey"
+	// Not liniting as this is a long query string.
+	//nolint:lll
```
Olshansk, line 28:
```suggestion
	//nolint:lll	// Not liniting as this is a long query string.
```

Olshansk: Whenever I would use `nolint` in the past, I usually had the reasoning on the same line.

Is there a best practice we're following?

pkg/client/supplier/client_integration_test.go
```diff
@@ -13,7 +13,8 @@ import (
 )
 
 func TestNewSupplierClient_Localnet(t *testing.T) {
-	t.Skip("TODO_TECHDEBT: this test depends on some setup which is currently not implemented in this test: staked application and servicer with matching services")
+	t.Skip("TODO_TECHDEBT: this test depends on some setup which is currently" +
```
Olshansk, line 16: Why do we go with `+` for multiline strings instead of backticks? `

https://stackoverflow.com/questions/7933460/how-do-you-write-multiline-strings-in-go

.github/workflows/run-tests.yml
```diff
@@ -44,8 +49,44 @@ jobs:
...
+
+      # Fetch a list of all modified files in this PR.
+      - name: Get Modified Files
+        run: git diff --name-only ${{ github.event.before }} ${{ github.sha }} > modified_files.txt
+
+      # Filter the list of files modified in this PR to exclude those that have
+      # ignite scaffold comments in their import block, this is because gci
```
Olshansk, line 63:
```suggestion
      # ignite scaffold comments in their import block. This is because gci
```

.github/workflows/run-tests.yml
```diff
@@ -44,8 +49,44 @@ jobs:
...
+      - name: Run golangci-lint (filtered for scaffold comments)
+        uses: golangci/golangci-lint-action@v3
+        with:
+          version: latest
+          args: --timeout=15m --config=.golangci.yml $(cat files_to_lint.txt)
+          only-new-issues: true
+          skip-cache: true
+
+      # Do a second lint pass without import checking (after filtered lint) to
+      # catcy any linter errors in files that were excluded in the first pass.
```
Olshansk, line 86:
```suggestion
      # catch any linter errors in files that were excluded in the first pass.
```

.golangci.yml
```diff
@@ -1,17 +1,39 @@
...
 # Don't forget to delete the `disable-all: true` line as well.
 linters:
   disable-all: true
   enable:
-#    - govet
-#    - revive
-#    - errcheck
-#    - unused
-    - goimports
+    #    - govet
```
Olshansk, line 13: Can we remove these comments and remove TECHDEBT above?

Olshansk (changes requested):
Leaving a few preliminary comments on the core logic. 

Not sure if it's worth going through all the `.go` files, but do think it'll be easier to merge all at once (assuming the proper docs and tooling is in place) and adjust if/when need be.

.github/workflows/run-tests.yml
```diff
@@ -1,4 +1,4 @@
-name: Run tests
+Name: Run tests
```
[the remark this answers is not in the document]
h5law: Updated the description this should help. LMK if u need anything more before we make a sytle guide or smt but thats outside scope of this PR.

.github/workflows/run-tests.yml
bryanchriswhite: It doesn't make sense to me for linting fixes to be applied in CI. The CI linter should fail if there are any issues, as opposed to fixing and then abandoning the changes.

Makefile
```diff
@@ -71,6 +72,26 @@ check_godoc:
...
+		echo "Seems like you don't have gci installed. Make sure you install it via 'go install github.com/daixiang0/gci@latest' before continuing"; \
+		exit 1; \
+	fi; \
+	}
+
+.PHONY: check_gofumpt
+# Internal helper target - check if gofumpt is installed
+check_gofumpt:
+	{ \
+	if ( ! ( command -v gci >/dev/null )); then \
```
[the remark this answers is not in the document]
bryanchriswhite: Unresolving, as the change hasn't been pushed yet.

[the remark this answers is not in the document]
bryanchriswhite: (I'm noticing that there are other resolved comments with no accompanying change as well.)

Makefile
```diff
@@ -71,6 +73,36 @@ check_godoc:
 	fi; \
 	}
 
+.PHONY: check_golanci_lint
+# Internal helper target - check if golangci-lint is installed
+check_golanci_lint:
+	{ \
+	if ( ! ( command -v gci >/dev/null )); then \
```
Olshansk, line 80: Should this be s/gci/golangci-lint?

e2e/tests/init_test.go
```diff
@@ -100,13 +106,13 @@ func (s *suite) TheUserRunsTheCommand(cmd string) {
 	res, err := s.pocketd.RunCommand(cmds...)
 	s.pocketd.result = res
 	if err != nil {
-		s.Fatalf("error running command %s: %s", cmd, err)
+		s.Fatalf("error running command %s: %s", cmd, err) //nolint:typecheck // This method does exist
```
Olshansk, line 109: Do you know what the issue with `s.Fatalf` is?

go.mod
```diff
@@ -14,7 +14,9 @@
 // module pocket
 module github.com/pokt-network/poktroll
 
-go 1.20
+go 1.21
```
Olshansk, line 17: Can we revert this? This can have downstream effects, so I'd want to make it a separate PR.

Olshansk (changes requested):

tools/scripts/gci/filters/filters.go
```diff
@@ -0,0 +1,132 @@
+// The filters package contains functions that can be used to filter file paths.
```
Olshansk, line 1:
```suggestion
// The filters package contains functions that can be used to filter file paths
// in limiting the scope of which files the gci linter touches. 
```

tools/scripts/gci/filters/filters.go
```diff
@@ -0,0 +1,132 @@
...
+
+import (
+	"bufio"
+	"bytes"
+	"os"
+	"path/filepath"
+	"strings"
+)
+
+const igniteScaffoldComment = "// this line is used by starport scaffolding"
```
Olshansk, line 13:
```suggestion
// This is a comment inserted by the ignite CLI that we want to maintain in the
// source code and ignore for linting purposes.
const igniteScaffoldComment = "// this line is used by starport scaffolding"
```

Olshansk (changes requested):
Halfway through

e2e/tests/init_test.go
h5law: Its embedded in the struct `s` as a gocuke.T type and the linter doesn't detect this - and `s` inherits this method from the embedded field

x/supplier/keeper/query_proof.go
```diff
@@ -2,7 +2,6 @@ package keeper
 
 import (
 	"context"
-	"fmt"
```
red-0ne, line 5: This seems to be required

red-0ne (commented):
Thanks for going through all the codebase :pray:

Olshansk:
@h5law Should we move this one to a draft state if it's not under active review/development?

h5law:
This is probably a good community task as picking up from what I started shouldn't be too hard with the tooling added. Moved into draft and put into iteration 14 «Developer Experience»

Olshansk:
> This is probably a good community task as picking up from what I started shouldn't be too hard with the tooling added.
> Moved into draft and put into iteration 14 «Developer Experience»

@h5law I'd rather have someone on the team do this. When you introduce a (*trivial*) change that touches a large surface area of the codebase, it's better to be done by the core team.

Pick it up in iteration 14 and we should be 🏅

Olshansk:
Outdated.

Issue #275, opened 2023-12-15
[Protocol] Implement Session Rollovers


Olshansk:
## Objective

Resolve the infamous session rollover issue once and for all

## Origin Document

<img width="827" alt="Screenshot 2023-12-14 at 3 54 21 PM" src="https://github.com/pokt-network/poktroll/assets/1892194/411aec2c-4091-4079-b306-8e7773af3ed7">

While working on #252, we had to add a temporary hack below due to:
- Very fast block times
- Small number of blocks per session

Overall, this is us hitting the session rollover issue in Shannon

```go
	// TODO_IN_THIS_PR: Either slow down blocks, or increase numBlocksPerSession,
	// or create a ticket related to session rollovers and link to it here.
	// currentBlock := rp.blockClient.LastNBlocks(ctx, 1)[0]
	// session, err := rp.sessionQuerier.GetSession(ctx, appAddress, service.Id, currentBlock.Height())
	// session, err := rp.sessionQuerier.GetSession(ctx, appAddress, service.Id, relayRequest.Meta.SessionHeader.SessionStartBlockHeight)
	currentBlock := rp.blockClient.LastNBlocks(ctx, 1)[0]
	session, err := rp.sessionQuerier.GetSession(ctx, appAddress, service.Id, currentBlock.Height())
```

## Goals

- Design & Implement an on-chain solution for session rollovers in Shannon

## Deliverables

- [ ] A PR that implements and solves session rollovers
- [ ] A governance parameter that modulates how much leeway session rollovers have
- [ ] Document how session rollovers are solved in Shannon

## Non-goals / Non-deliverables

- Using the same solution for session rollovers as in Morse (i.e. client side)

## General deliverables

- [ ] **Comments**: Add/update TODOs and comments alongside the source code so it is easier to follow.
- [ ] **Testing**: Add new tests (unit and/or E2E) to the test suite.
- [ ] **Makefile**: Add new targets to the Makefile to make the new functionality easier to use.
- [ ] **Documentation**: Update architectural or development READMEs; use [mermaid](https://mermaid-js.github.io/mermaid/) diagrams where appropriate.

---

**Creator**: @Olshansk 
**Co-Owners**: @red-0ne

Olshansk:
tl;dr In `pkg/relayer/proxy/relay_verifier.go`:

```diff
[...]

PR #326, opened 2024-01-17
feat: Implement session rollover logic

red-0ne:
### Human Summary
* Implement the session rollover mechanism that delays `Session.Flush`ing and `Claim` creation saving late `RelayRequests` from being rejected.
* Factor `Session`'s start/end block heights calculation
* Fix an off-by-one bug on the `SessionEndBlockHeight` calculation

### Summary generated by Reviewpad on 26 Jan 24 03:44 UTC

1. In the file "claim.go":

- Added an import statement for the package "sessionkeeper" from the module "github.com/pokt-network/poktroll/x/session/keeper".
- Modified the function "waitForEarliestCreateClaimHeight" to calculate the value of "createClaimWindowStartHeight" with an additional offset.

2. In the file "testutil/testproxy/relayerproxy.go":

- Added fields `SessionStartBlockHeight` and `SessionEndBlockHeight` to the `SessionHeader` struct.

3. In the file "session.feature":

- Added scenarios for a late relay inside and outside the SessionGracePeriod, including steps to send a relay request, wait for blocks, get session information, and verify the relay response.

4. In the file "session_hydrator_test.go":

- Modified the expected values of `sessionHeader.SessionEndBlockHeight` and `expectedSessionEndBlock` in several test cases.

5. In a file with no specified name:

- Added an import for "github.com/pokt-network/poktroll/x/session/keeper".
- Updated the logic in the function "mapBlockToSessionsToClaim" to check if sessions need to enter the claim/proof phase based on their end block height.
- Updated the logic in the function "mapBlockToSessionsToClaim" to iterate over session trees that have a grace period ending at the current block height and add them to the list of session trees to be published.
- Added a new function "IsWithinGracePeriod" to check if the grace period for a session has ended and determine if it's time to create a claim for it.

6. In a file with no specified name:

- Updated various sections of the file, including constants, helper functions, and struct fields.

7. In the file "errors.go":

- Added a new error `ErrSessionTreeAlreadyMarkedAsClaimed` with the code `6`.

8. In the file "sessiontree.go":

- Added a new field `isClaiming` to the `sessionTree` struct.
- Added a new method `StartClaiming()` to mark the session tree as being picked up for claiming.

9. In the file "pkg/relayer/session/session_test.go":

- Added an import statement for `sessionkeeper` from `github.com/pokt-network/poktroll/x/session/keeper`.
- Modified the implementation of the `TestRelayerSessionsManager_Start` function, including calculating the session grace period end block height and triggering the creation of a claim for the session.

10. In the file "interface.go":

- Added a new method `StartClaiming()` to the `SessionTree` interface.

11. In the file "sessionquerier.go":

- Modified the `Header` struct to include additional fields.
- Added the `NumBlocksPerSession` and `SessionNumber` fields to the struct.

12. In the file "relay_verifier.go":

- Added an import statement for the sessiontypes package.
- Added a new function called `getTargetSessionBlockHeight`.
- Modified the `VerifyRelayRequest` function to use the `getTargetSessionBlockHeight` function and handle session block height validation.

These changes encompass various improvements and additions related to session management, claiming, and error handling.
Let me know if you require further assistance.

## Issue

- #275 

## Type of change

Select one or more:

- [x] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

## Sanity Checklist


pkg/relayer/session/session.go
```diff
@@ -165,18 +166,21 @@ func (rs *relayerSessionsManager) mapBlockToSessionsToClaim(
 	defer rs.sessionsTreesMu.Unlock()
 
 	// Check if there are sessions that need to enter the claim/proof phase
```
Olshansk, line 168: I think you should change this comment back. The PREVIOUS comment is the business logic. Inside of it, we add an explanation regarding the grace period.

It's similar to a database `GET`. You explain what the `GET` does, but not the "caching" that may or may not happen inside.

pkg/relayer/session/session.go
```diff
@@ -165,18 +166,21 @@ func (rs *relayerSessionsManager) mapBlockToSessionsToClaim(
...
-	// as their end block height was the one before the last committed block.
-	// Iterate over the sessionsTrees map to get the ones that end at a block height
-	// lower than the current block height.
+	// as grace period block height was the one before the last committed block.
+	// Iterate over the sessionsTrees map to get the ones that have a grace period
+	// ending at the current block height.
 	for endBlockHeight, sessionsTreesEndingAtBlockHeight := range rs.sessionsTrees {
 		// TODO_BLOCKER(@red-0ne): We need this to be == instead of <= because we don't want to keep sending
```
Olshansk, line 173: Is this comment still relevant?

pkg/relayer/session/session.go
```diff
@@ -165,18 +166,21 @@ func (rs *relayerSessionsManager) mapBlockToSessionsToClaim(
...
 		// the same session while waiting the next step. This does not address the case
 		// where the block client misses the target block which should be handled by the
 		// retry mechanism. See the discussion in the following GitHub thread for next
 		// steps: https://github.com/pokt-network/poktroll/pull/177/files?show-viewed-files=true&file-filters%5B%5D=#r1391957041
-		if endBlockHeight == block.Height() {
-			// Iterate over the sessionsTrees that end at this block height (or
-			// less) and add them to the list of sessionTrees to be published.
+		// TODO_DISCUSS: Add a `isClosing()` method to SessionTree to check if the session is closing
```
Olshansk, line 178: What do you want to dicuss? WHy not do it in this PR?

pkg/relayer/session/session.go
```diff
@@ -165,18 +166,21 @@ func (rs *relayerSessionsManager) mapBlockToSessionsToClaim(
...
+		// so we can use the `<=` operator here while still avoiding sending the same session
+		// multiple times.
+		if endBlockHeight == block.Height()-sessionkeeper.SessionGracePeriod {
```
Olshansk, line 181: I think this check should be moved into a helper function with an appropriate name + comment. It'll be more lines/character of code but less confusing.

pkg/relayer/proxy/relay_verifier.go
```diff
@@ -6,7 +6,10 @@ import (
 	sdkerrors "cosmossdk.io/errors"
```
Olshansk, line 6: Leaving a comment here but it is appropriate elsewhere.

Can you write out an E2E test (in englih/comments) on how we can test & verify session rollover?

Just write out the steps (that do not compile) and add a TODO to add it in a followup PR.

pkg/relayer/session/session_test.go
```diff
@@ -62,18 +63,19 @@ func TestRelayerSessionsManager_Start(t *testing.T) {
 	noopBlock := testblock.NewAnyTimesBlock(t, zeroByteSlice, sessionStartHeight)
 	blockPublishCh <- noopBlock
 
+	sessionGracePeriodEndBlockHeight := int64(sessionEndHeight + sessionkeeper.SessionGracePeriod)
```
Olshansk, line 66:
```suggestion
	sessionGracePeriodEndBlockHeight := int64(sessionEndHeight + sessionkeeper.SessionGracePeriod)

```

pkg/relayer/session/session_test.go
```diff
@@ -62,18 +63,19 @@ func TestRelayerSessionsManager_Start(t *testing.T) {
 	noopBlock := testblock.NewAnyTimesBlock(t, zeroByteSlice, sessionStartHeight)
 	blockPublishCh <- noopBlock
```
Olshansk, line 65: Can you add a comment to what/why we needing `sessionGracePeriodEngBlockHeight`

Olshansk (changes requested):

pkg/relayer/session/session.go
red-0ne: Yes, until we implement what's on the `TODO_DISCUSS` below

pkg/relayer/session/session.go
red-0ne: @Olshansk , I wanted to discuss whether it's worth it to introduce more state into the `SessionTree` struct. IMO it's OK since it'll be only a derived state.

I'll add it to the current PR if you're OK with the change involved.

x/session/keeper/session_hydrator.go
```diff
@@ -21,6 +21,7 @@ var SHA3HashLen = crypto.SHA3_256.Size()
 // TODO(#21): Make these configurable governance param
 const (
 	NumBlocksPerSession         = 4
+	SessionGracePeriod          = NumBlocksPerSession
```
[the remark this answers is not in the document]
red-0ne: We will anyway end up converting it to blocks since we'll be using it's `blockHeight` representation when it's involved in `currentBlockHeight` comparisons or passing blockHeights to `waitForBlock` to create `Claim`s.

We'll be replacing `SessionGracePeriod` by `SessionGracePeriod * sessionkeeper.NumBlocksPerSession` in all its usage occurrences.

It also makes it non-trivial to switch to a block granular representation if we want to do so in the future IMO

x/session/keeper/session_hydrator.go
```diff
@@ -243,8 +243,24 @@ func sha3Hash(bz []byte) []byte {
 	return hasher.Sum(nil)
 }
 
+// GetSessionStartBlockHeight returns the block height at which the session starts
+func GetSessionStartBlockHeight(blockHeight int64) int64 {
```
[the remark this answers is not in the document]
red-0ne: Interesting... I added them as helper functions to factor these calculations since they are needed in contexts other than this keeper (mainly `testutils`) I wonder if we're not going to end-up passing the `NumBlocksPerSession` to these functions :thinking:

e2e/tests/session.feature
```diff
@@ -8,3 +8,8 @@ Feature: Session Namespace
 #  TODO_IMPROVE: ...
 #    And an event should be emitted...
 #  TODO_INCOMPLETE: add step(s) for proof validation.
+
+#  Scenario: Late RelayRequests are handled within the session grace period
+#    When a RelayRequest is signed for an Application within session number N
+#    And the RelayMiner gets the RelayRequest within session number N+1
+#    Then the RelayMiner should process the RelayRequest and include it in the claim N
```
Olshansk, line 15: Wdyt of something like this instead
```suggestion
# TODO_BLOCKER(@red-0ne): Make sure to implement and validate this test
Scenario: A late Relay inside the SessionGracePeriod is handled
    Given the user has the pocketd binary installed
    And the parameter "NumBlockPerSession" is "4"
    And the parameter "SessionGracePeriod" is "1"
    When the supplier "supplier1" has serviced a session with service "svc1" for application "app1" with session number "1"
    And we have waited "5" block
    And the application "app1" sends a relay request to supplier "supplier1" for service "svc" with session number "1"
    Then the application "app1" receives a successful relay response signed by "supplier1" for session number "1"
    And after the supplier "supplier1" updates a claim for session number "1" for service "svc1" for application "app1"
    Then the claim created by supplier "supplier1" for service "svc1" for application "app1" should be persisted on-chain

# TODO_BLOCKER(@red-0ne): Make sure to implement and validate this test
Scenario: A late Relay outside the SessionGracePeriod is rejected
    Given the user has the pocketd binary installed
    And the parameter "NumBlockPerSession" is "4"
    And the parameter "SessionGracePeriod" is "1"
    When the supplier "supplier1" has serviced a session with service "svc1" for application "app1" with session number "1"
    And we have waited "10" block
    And the application "app1" sends a relay request to supplier "supplier1" for service "svc" with session number "1"
    Then the application "app1" receives a failed relay response
    And the supplier "supplier1" fails to update a claim for session number "1" for service "svc1" for application "app1"
```

pkg/relayer/proxy/relay_verifier.go
```diff
@@ -128,3 +136,33 @@ func (rp *relayerProxy) VerifyRelayRequest(
 
 	return ErrRelayerProxyInvalidSupplier
 }
+
+// getTargetSessionBlockHeight returns the block height at which the session
+// for the given relayRequest should be processed. If the session is in the
```
Olshansk, line 141:
```suggestion
// for the given relayRequest should be processed. If the session is within the
```

pkg/relayer/proxy/relay_verifier.go
```diff
@@ -128,3 +136,33 @@ func (rp *relayerProxy) VerifyRelayRequest(
...
+	ctx context.Context,
+	relayRequest *types.RelayRequest,
+) (sessionBlockHeight int64, err error) {
+	currentBlockHeight := rp.blockClient.LastNBlocks(ctx, 1)[0].Height()
+	sessionEndblockHeight := relayRequest.Meta.SessionHeader.GetSessionEndBlockHeight()
+
+	// Check if the `RelayRequest`'s session has expired.
+	if sessionEndblockHeight < currentBlockHeight {
+		// Do not process the `RelayRequest` if the session has expired and the current
+		// block height is past the session's grace period.
```
Olshansk, line 155:
```suggestion
		// block height is outside the session's grace period.
```

x/session/keeper/session_hydrator.go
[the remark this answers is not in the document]
Olshansk: I want to push back on this.

The governance parameter (what we will document, update through governance votes, and what people read) will be in sessions. That is `SessionGracePeriod`.

Internally, we do math as a function of `SessionGracePeriod * NumBlocksPerSession`, so I think the separation will make maintenance simpler (especially if we leave the appropriate comments).

Lmk if jumping on a call is easier.

pkg/relayer/session/session.go
```diff
@@ -261,3 +265,9 @@ func (rs *relayerSessionsManager) mapAddMinedRelayToSessionTree(
 	// Skip because this map function only outputs errors.
 	return nil, true
 }
+
+// checkGracePeriodEnd checks if the grace period for the session has ended
+// and signals whether it is time to create a claim for it.
+func checkGracePeriodEnd(endBlockHeight, currentBlockHeight int64) bool {
```
Olshansk, line 271:
```suggestion
func checkGracePeriodEnd(sessionEndBlockHeight, currentBlockHeight int64) bool {
```

pkg/relayer/session/session.go
```diff
@@ -261,3 +265,9 @@ func (rs *relayerSessionsManager) mapAddMinedRelayToSessionTree(
...
+	return endBlockHeight == currentBlockHeight-sessionkeeper.SessionGracePeriod
```
Olshansk, line 272: Why is this not a `<=`?

pkg/relayer/session/session.go
```diff
@@ -261,3 +265,9 @@ func (rs *relayerSessionsManager) mapAddMinedRelayToSessionTree(
 	// Skip because this map function only outputs errors.
 	return nil, true
 }
+
+// checkGracePeriodEnd checks if the grace period for the session has ended
```
Olshansk, line 269: Since it just returns a bool, I think something like `isWithinGracePeriod` or `isOutsideGracePeriod` or `didGracePeriodEnd`, etc is more appropriate based on how you're thinking about it.

pkg/relayer/session/session.go
Olshansk: Consider making it a `TODO_UPNEXT(@red-0ne)`. Let's get this done.

pkg/relayer/proxy/relay_verifier.go
```diff
@@ -128,3 +136,33 @@ func (rp *relayerProxy) VerifyRelayRequest(
...
+		if sessionEndblockHeight < currentBlockHeight-sessionkeeper.SessionGracePeriod {
```
Olshansk, line 156: Why don't we use the `checkGracePeriodEnd` helper you introduce in the other file?

Olshansk (changes requested):

e2e/tests/session.feature
red-0ne: @Olshansk, Merged these as they are more explicit but I don't think they will exercise the session rollover.

First it has to start with the application sending a relay in a block height that resolves to session number `1`
Then blocks has to pass (wait n blocks) so when the supplier receives the relay from the application, the supplier is already in a block that resolves to session number `2`.

The nuance is that the blocks has to pass between the application sending the relay request and the supplier receiving that relay request.

pkg/relayer/session/session.go
red-0ne: Will do in this PR

red-0ne: Implemented as `ClaimOnce`

e2e/tests/session.feature
Olshansk: > The nuance is that the blocks has to pass between the application sending the relay request and the supplier receiving that relay request.

Can you update the steps to reflect that? Can we not do something like:

1. The supplier services a relay for session number 1
2. The supplier waits 5 blocks
3. The supplier calls GetSession and gets session number 2
4. The supplier services a relay session number 1

I think that if we get really explicit with the waiting and the numbers, we can make it deterministic, which is the goal of the test. What do you think we're missing here?

e2e/tests/session.feature
```diff
@@ -8,3 +8,26 @@ Feature: Session Namespace
 #  TODO_IMPROVE: ...
 #    And an event should be emitted...
 #  TODO_INCOMPLETE: add step(s) for proof validation.
+
+# # TODO_BLOCKER(@red-0ne): Make sure to implement and validate this test
```
Olshansk, line 12: Double comment

Olshansk (changes requested):
@red-0ne I want to make sure we really resolve [1] before emerging this in.

[1] https://github.com/pokt-network/poktroll/pull/326#discussion_r1463661493

e2e/tests/session.feature
red-0ne: @Olshansk , I updated the test, WDYT of of it's stated now?

e2e/tests/session.feature
```diff
@@ -8,3 +8,30 @@ Feature: Session Namespace
...
```
Olshansk, line 12: s/# #/#

pkg/relayer/session/session.go
```diff
@@ -261,3 +258,9 @@ func (rs *relayerSessionsManager) mapAddMinedRelayToSessionTree(
 	// Skip because this map function only outputs errors.
 	return nil, true
 }
+
+// IsWithinGracePeriod checks if the grace period for the session has ended
+// and signals whether it is time to create a claim for it.
+func IsWithinGracePeriod(sessionEndBlockHeight, currentBlockHeight int64) bool {
+	return sessionEndBlockHeight+sessionkeeper.GetSessionGracePeriodBlockCount() >= currentBlockHeight
```
Olshansk, line 265: Shouldn't this be `<=`?

x/session/keeper/session_hydrator.go
```diff
@@ -18,9 +18,13 @@ import (
...
 var SHA3HashLen = crypto.SHA3_256.Size()
 
-// TODO(#21): Make these configurable governance param
+// TODO_BLOCKER(#21): Make these configurable governance param
 const (
-	NumBlocksPerSession         = 4
+	// TODO_BLOCKER: Remove direct usage of these constants in helper functions
+	// when they will be replaced by governance params
+	NumBlocksPerSession = 4
+	// Duration of the grace period in sessions
```
Olshansk, line 26:
```suggestion
	// Duration of the grace period in number of sessions
```

x/session/keeper/session_hydrator.go
```diff
@@ -255,11 +274,30 @@ func GetSessionId(
...
+		serviceIdBz,
+		appPubKeyBz,
+		blockHeightBz,
+	)
 	sessionId = hex.EncodeToString(sha3Hash(sessionIdBz))
 
 	return sessionId, sessionIdBz
 }
+
+// GetSessionGracePeriodBlockCount returns the number of blocks in the session
```
Olshansk, line 290: 🔥

pkg/relayer/session/sessiontree.go
```diff
@@ -206,3 +208,18 @@ func (st *sessionTree) Delete() error {
 	// Delete the KVStore from disk
 	return os.RemoveAll(filepath.Dir(st.storePath))
 }
+
+// ClaimOnce adds the sessionTree to the list of sessionTrees to be claimed.
+// It returns the updated list of sessionTrees or the original list if the sessionTree
+// is already included in a list for sessions to be claimed.
+func (st *sessionTree) ClaimOnce(sessionTreesToClaim []relayer.SessionTree) []relayer.SessionTree {
```
Olshansk, line 215: This is a very confusing function for me. 

An object receives a list of objects and returns that list while potentially (or not) mutating its own state. 

Doesn't feel like it follows SOLID principles and I just struggle to follow along.

Olshansk (changes requested):

pkg/relayer/session/session.go
red-0ne: It was
```go
if sessionEndblockHeight < currentBlockHeight-sessionkeeper.SessionGracePeriod` {
  return error
}
```
Which was indicating "isNOTWithingGracePeriod"

It's more readable now IMO

pkg/relayer/session/sessiontree.go
red-0ne: I see what you mean.

The constraints here are:
* `sessionMu` is private to the `SessionTree`
* `isClaiming` has to be altered only once
* Append the the slice only when `isClaiming` has changed

I'll try another approach but open to any suggestion.

red-0ne: Updated, PTAL

pkg/relayer/session/sessiontree.go
```diff
@@ -206,3 +208,17 @@ func (st *sessionTree) Delete() error {
 	// Delete the KVStore from disk
 	return os.RemoveAll(filepath.Dir(st.storePath))
 }
+
+// MarkAsClaimed marks the session tree as claimed
+// It returns an error if it has already been marked as claimed.
+func (st *sessionTree) MarkAsClaimed() error {
```
Olshansk, line 214: I think this should be called `Claim` or `StartClaiming`. Wdyt?

pkg/relayer/session/session.go
```diff
@@ -261,3 +266,9 @@ func (rs *relayerSessionsManager) mapAddMinedRelayToSessionTree(
...
```
Olshansk, line 273:
```suggestion
	return currentBlockHeight <= sessionEndBlockHeight+sessionkeeper.GetSessionGracePeriodBlockCount()
```

pkg/relayer/session/sessiontree.go
Olshansk: See new comment in latest review.

pkg/relayer/session/session.go
Olshansk: See suggested change in latest review.

pkg/relayer/session/sessiontree.go
red-0ne: `StartClaiming` is perfect

Olshansk (approved):
@red-0ne I updated the branched and kicked off E2E tests. Once everything is 🟢 , let's merge this in!

I just realized this rhymes: "Once everything is green, merge it in!"

Issue #141, opened 2023-11-03
[Protocol] MVP implementation off the on-chain `SubmitProof` message handling


Olshansk:
## Objective

MVP Implementation of `SubmitProof` to enable Shannon TestNet.

## Origin Document

- #44

## Goals

- Enable E2E validation of the entire protocol
- Unblock iterative development of the claim & proof lifecycle
- Unblock iterative development of on-chain tokenomics

## Deliverables

- [ ] A single PR that introduces the business logic for the on-chain `SubmitProof` message (verification, etc...)
- [x] A single PR that updates the CLI to submit SubmitProof messages
- [ ] Introduce make targets to enable development (visibility, triggering, debugging , etc...) of on-chain proofs
- [ ] Add extensive unit tests 

## Non-goals / Non-deliverables

- Account Management (i.e. Burn & Mint) business logic related to the Claim & Proof lifecycle (leave TODOs)
- Governance Params necessary for the Claim & Proof lifecycle (leave TODOs)

## General deliverables

- [ ] **Comments**: Add/update TODOs and comments alongside the source code so it is easier to follow.
- [ ] **Testing**: Add new tests (unit and/or E2E) to the test suite.
- [ ] **Makefile**: Add new targets to the Makefile to make the new functionality easier to use.
- [ ] **Documentation**: Update architectural or development READMEs; use [mermaid](https://mermaid-js.github.io/mermaid/) diagrams where appropriate.

---

**Creator**: @Olshansk 
**Co-Owners**: @bryanchriswhite

bryanchriswhite:
@Olshansk I'm starting on; got a happy path unit test already.
> A single PR that introduces the business logic for the on-chain SubmitProof message (verification, etc...)

Olshansk:
@bryanchriswhite Have you (planning to, considering, open-to) implemented the equivalent of #148 and #151 for proofs?

I created #243 for the work we discussed today and will drive that myself but wanted to know if the above is on your radar.

bryanchriswhite:
> @bryanchriswhite Have you (planning to, considering, open-to) implemented the equivalent of #148 and #151 for proofs?

Yessir! The starting point I was referring to is analogous to #148 + #236 - testutils (which I will endeavor to isolate as reasonable). I think this is the best place to start as it's the most "outside", allowing me to work outside-in. Parts of the #151 analogue may be driven out along the way but otherwise I think it will be quite natural to lay out more isolated PRs for the various pieces.

[...]

PR #266, opened 2023-12-12
[Supplier] refactor: proof store indices

bryanchriswhite:
### Human Summary

Refactors on-chain proof store to support multiple indices/stores in the multistore (similar to claims): "primary key", "supplier address", and "session end block height".

## Issue

- #141 
- #263
- #264
- #265 
- #289
- #290 
- #285
- :point_right: THIS PR :point_left: 
- #250

## Type of change

Select one or more:

- [ ] New feature, functionality or library
- [ ] Bug fix
- [x] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

- [ ] **Run all unit tests**: `make go_develop_and_test`
- [ ] **Run E2E tests locally**: `make test_e2e`
- [ ] **Run E2E tests on DevNet**: Add the `devnet-test-e2e` label to the PR. This is VERY expensive, o only do it after all the reviews are complete.

## Sanity Checklist

- [x] I have tested my changes using the available tooling
- [x] I have performed a self-review of my own code
- [x] I have commented my code, updated documentation and left TODOs throughout the codebase

PR #250, opened 2023-12-07
[Supplier] feat: persistence proofs in submit proof msg handler

bryanchriswhite:
### Human Summary

WIP

## Issue

- #141 
- #263
- #264
- #265 
- #289
- #290 
- #285
- #266 
- :point_right: THIS PR :point_left: 

The `SubmitProof` msg handler function is receiving proof messages but is not persisting the respective proofs yet.

## Type of change

Select one or more:

- [x] New feature, functionality or library
- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

- [ ] **Run all unit tests**: `make go_develop_and_test`
- [ ] **Verify Localnet manually**: See the instructions [here](TODO: add link to instructions)

## Sanity Checklist

- [ ] I have tested my changes using the available tooling
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, updated documentation and left TODOs throughout the codebase

PR #328, opened 2024-01-17
[Supplier] feat: persistence proofs in submit proof msg handler (2)

bryanchriswhite:
### Human Summary

- Validate and persist proofs received in `MsgSubmitProof` message by the supplier module.
- ~~Simplify `ReplayEventsClient` to reduce number of generic parameters~~
   *(merged in #330)*

## Issue

- #141

## Type of change

Select one or more:

- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

## Sanity Checklist


bryanchriswhite:
Supersedes #250; shelves unit & integration coverage due to excessive refactor effort required.

e2e/tests/session_steps_test.go
```diff
@@ -145,6 +119,41 @@ func (s *suite) TheSupplierHasServicedASessionWithRelaysForServiceForApplication
...
+	allProofsRes, err := s.supplierQueryClient.AllProofs(ctx, &suppliertypes.QueryAllProofsRequest{
+		Filter: &suppliertypes.QueryAllProofsRequest_SupplierAddress{
+			SupplierAddress: accNameToAddrMap[supplierName],
+		},
+	})
+	require.NoError(s, err)
+	require.NotNil(s, allProofsRes)
+
+	// Assert that the number of proofs has increased by one.
+	preExistingProofs := s.scenarioState[preExistingProofsKey].([]suppliertypes.Proof)
```
Olshansk, line 138: I believe when you do this sort of type casting, you can get back an error that we can assert on.

Olshansk (changes requested):
Partial review in case I don't finish it today

Screenshot 2024-01-17 at 6 11 41 PM

pkg/client/interface.go
```diff
@@ -123,9 +123,9 @@ type EventsObservable[T any] observable.ReplayObservable[T]
 
 // EventsReplayClient is an interface which provides notifications about newly received
 // events as well as direct access to the latest event via some blockchain API.
-type EventsReplayClient[T any, R observable.ReplayObservable[T]] interface {
+type EventsReplayClient[T any] interface {
```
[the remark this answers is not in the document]
bryanchriswhite: Yes. I considered isolating it but IIRC it was (much) less than 100 LoC diff.

x/supplier/keeper/msg_server_create_claim.go
```diff
@@ -13,7 +16,9 @@ func (k msgServer) CreateClaim(goCtx context.Context, msg *suppliertypes.MsgCrea
 	// TODO_BLOCKER: Validate the signature on the Claim message corresponds to the supplier before Upserting.
 
 	ctx := sdk.UnwrapSDKContext(goCtx)
-	logger := k.Logger(ctx).With("method", "CreateClaim")
+	// TODO_DISCUSS_IN_THIS_COMMIT: log lines from the cosmos-sdk logger aren't printed in tilt for some reason.
```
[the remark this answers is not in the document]
bryanchriswhite: After more careful examination, what I was observing was the fact that our localnet [config.toml](https://github.com/pokt-network/poktroll/blob/main/localnet/poktrolld/config/config.toml#L53) is currently set to the `"info"`, by default. I also find it worthwhile to point out that is is also set to `"plain"` format, whereas we could set it to JSON, which may simplify downstream log aggregation, parsing, searching, etc. (in addition to being more consistent with the default output format of our off-chain logs):

image

image

Yields:

image

When setting the log level to `"debug"` the log output becomes much more verbose but with cosmos-sdk internal output (which is not typically useful, especially for debugging business logic). For this reason, I **would not** suggest that we change the localnet config to the `"debug"` level. However, this puts us in a position where we have to either A) manually change the config and wade through the excess output in order to effectively use debug logs that we can leave in the code, or to B) use an alternative logger which provides the debugging degree of freedom without including all the cosmos-sdk specific debug log lines. The latter option puts us in a position of inconsistency with respect to which logging library to use so long as we want to keep the debug lines in the code, which we should definitely have the option to do.

---

I'm also noticing that we don't seem to be able to take advantage of tilt's log level filtering (it hasn't a clue):

image

I'm not sure if this has to do with whether logs are printed to stderr vs. stdout or if it's something we can configure tilt to be sensitive to but it would be extremely convenient to be able to use this feature.

pkg/client/interface.go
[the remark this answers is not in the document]
Olshansk: Can you make sure the PR description & title captures this?

x/supplier/keeper/msg_server_submit_proof.go
```diff
@@ -4,55 +4,128 @@ import (
...
+	}
+
+	if err := k.queryAndValidateClaimForProof(ctx, &proof); err != nil {
+		return nil, status.Error(codes.FailedPrecondition, err.Error())
+	}
+
+	// TODO_CONSIDERATION: check if this proof already exists and return an appropriate error
+	// in any case where the supplier should no longer be able to update the given proof.
+	k.Keeper.UpsertProof(ctx, proof)
+
```
Olshansk, line 70:
```suggestion

		// TODO_BLOCKER(@bryanchriswhite, @Olshansk): Call `tokenomics.SettleSessionAccounting()` here

```

x/supplier/types/message_submit_proof.go
```diff
@@ -40,10 +41,42 @@ func (msg *MsgSubmitProof) GetSignBytes() []byte {
...
+	}
+
+	if msg.GetSessionHeader().GetService().GetId() == "" {
+		return ErrSupplierInvalidService.Wrap("proof service ID %q cannot be empty")
+	}
+
+	if len(msg.GetProof()) == 0 {
+		return sdkerrors.ErrInvalidRequest.Wrap("proof cannot be empty")
+	}
+	if errMsg != "" {
```
Olshansk, line 77: `errMsg` isn't really being used anywhere so can we just remove it?

x/supplier/types/message_submit_proof.go
```diff
@@ -40,10 +41,42 @@ func (msg *MsgSubmitProof) GetSignBytes() []byte {
...
+			msg.GetSessionHeader().GetApplicationAddress(),
+			err,
+		)
+	}
+
+	if msg.GetSessionHeader().GetService().GetId() == "" {
+		return ErrSupplierInvalidService.Wrap("proof service ID %q cannot be empty")
+	}
+
+	if len(msg.GetProof()) == 0 {
```
Olshansk, line 74: Should we try to parse it (using the proto) as well? Similar to how we do address validation?

Feel free to add a TODO.

x/supplier/keeper/msg_server_submit_proof.go
```diff
@@ -4,55 +4,128 @@ import (
...
+// queryAndValidateClaimForProof ensures that  a claim corresponding to the given proof's
+// session exists & has a matching supplier address and session header.
+func (k msgServer) queryAndValidateClaimForProof(sdkCtx sdk.Context, proof *suppliertypes.Proof) error {
+	sessionId := proof.GetSessionHeader().GetSessionId()
+	// NB: no need to assert the testSessionId or supplier address as it is retrieved
+	// by respective values of the give proof. I.e., if the claim exists, then these
+	// values are guaranteed to match.
+	claim, found := k.GetClaim(sdkCtx, sessionId, proof.GetSupplierAddress())
+	if !found {
+		return suppliertypes.ErrSupplierClaimNotFound.Wrapf("no claim found for session ID %q", sessionId)
```
Olshansk, line 91:
```suggestion
		return suppliertypes.ErrSupplierClaimNotFound.Wrapf("no claim found for session ID %q and supplier %q", sessionId, proof.GetSupplierAddress())
```

Olshansk (changes requested):

x/supplier/keeper/msg_server_submit_proof.go
```diff
@@ -4,55 +4,130 @@ import (
...
+		3. relay.Req.signer == session.appAddr
+		4. relay.Res.signer == msg.supplier
+
+		## Msg distribution validation (governance based params)
+		1. Validate Proof submission is not too early; governance-based param + pseudo-random variation
+		2. Validate Proof submission is not too late; governance-based param + pseudo-random variation
+
+		## Relay Mining validation
+		1. verify(proof.path) is the expected path; pseudo-random variation using on-chain data
+		2. verify(proof.ValueHash, expectedDiffictul); governance based
```
red-0ne, line 40:
```suggestion
		2. verify(proof.ValueHash, expectedDiffictulty); governance based
```

x/supplier/keeper/msg_server_submit_proof.go
```diff
@@ -4,55 +4,130 @@ import (
...
+
+	return &suppliertypes.MsgSubmitProofResponse{}, nil
+}
+
+// queryAndValidateClaimForProof ensures that  a claim corresponding to the given proof's
+// session exists & has a matching supplier address and session header.
+func (k msgServer) queryAndValidateClaimForProof(sdkCtx sdk.Context, proof *suppliertypes.Proof) error {
+	sessionId := proof.GetSessionHeader().GetSessionId()
+	// NB: no need to assert the testSessionId or supplier address as it is retrieved
+	// by respective values of the give proof. I.e., if the claim exists, then these
```
red-0ne, line 89:
```suggestion
	// by respective values of the given proof. I.e., if the claim exists, then these
```

x/supplier/types/message_create_claim.go
```diff
@@ -47,27 +47,27 @@ func (msg *MsgCreateClaim) GetSignBytes() []byte {
...
-	_, err := sdk.AccAddressFromBech32(msg.SupplierAddress)
+	_, err := sdk.AccAddressFromBech32(msg.GetSupplierAddress())
 	if err != nil {
-		return sdkerrors.Wrapf(ErrSupplierInvalidAddress, "invalid supplierAddress address (%s)", err)
+		return sdkerrors.Wrapf(ErrSupplierInvalidAddress, "%s", msg.GetSupplierAddress())
 	}
 
 	// Validate the session header
 	sessionHeader := msg.SessionHeader
 	if sessionHeader.SessionStartBlockHeight < 1 {
```
red-0ne, line 57: First session ever has a `SessionStartBlockHeight` of `0`, see https://github.com/pokt-network/poktroll/blob/main/x/session/keeper/session_hydrator_test.go#L71

x/supplier/keeper/msg_server_submit_proof.go
```diff
@@ -4,55 +4,130 @@ import (
...
+
+		## Actions (error if anything fails)
+		1. Retrieve a fully hydrated `session` from on-chain store using `msg` metadata
+		2. Retrieve a fully hydrated `claim` from on-chain store using `msg` metadata
+		3. Retrieve `relay.Req` and `relay.Res` from deserializing `proof.ClosestValueHash`
+
+		## Basic Validations (metadata only)
+		1. proof.testSessionId == claim.testSessionId
+		2. msg.supplier in session.suppliers
+		3. relay.Req.signer == session.appAddr
```
red-0ne, line 31: Is extracting the `relay` process captured somewhere? Can't find any `TODO` for it.

I liked the checkbox comments.

e2e/tests/session.feature
```diff
@@ -1,10 +1,15 @@
...
 
+  # TODO_TECHDEBT(@Olshansk, #180): This test requires you to run `make supplier1_stake && make app1_stake` first
+  # As a shorter workaround, we can also add steps that stake the application and supplier as part of the scenario.
   Scenario: Supplier completes claim/proof lifecycle for a valid session
     Given the user has the pocketd binary installed
     When the supplier "supplier1" has serviced a session with "5" relays for service "svc1" for application "app1"
     And after the supplier creates a claim for the session for service "svc1" for application "app1"
     Then the claim created by supplier "supplier1" for service "svc1" for application "app1" should be persisted on-chain
 #  TODO_IMPROVE: ...
 #    And an event should be emitted...
```
h5law, line 11: Remove - duplicated below
```suggestion
```

e2e/tests/session_steps_test.go
```diff
@@ -10,76 +10,41 @@ import (
...
-	testServiceId                = "anvil"
-	eitherEventsBzReplayObsKey   = "eitherEventsBzReplayObsKey"
-	preExistingClaimsKey         = "preExistingClaimsKey"
+	// txEventTimeout is the duration of time to wait after sending a valid tx
+	// before the test should time out (fail).
+	txEventTimeout = 10 * time.Second
+	// txSenderEventSubscriptionQueryFmt is the format string which yields the
+	// cosmos-sdk event subscription "query" string for a given sender address.
+	// This is used by an events replay client to subscribe to tx events from the supplier.
+	// See: https://docs.cosmos.network/v0.46/core/events.html#subscribing-to-events.
```
h5law, line 31:
```suggestion
	// See: https://docs.cosmos.network/v0.47/learn/advanced/events#subscribing-to-events.
```

e2e/tests/session_steps_test.go
```diff
@@ -145,6 +131,42 @@ func (s *suite) TheSupplierHasServicedASessionWithRelaysForServiceForApplication
...
+	// Assert that the number of proofs has increased by one.
+	preExistingProofs, ok := s.scenarioState[preExistingProofsKey].([]suppliertypes.Proof)
+	require.True(s, ok, "preExistingProofsKey not found in scenarioState")
+	// NB: We are avoiding the use of require.Len here because it provides unreadable output
+	// TODO_TECHDEBT: Due to the speed of the blocks of the LocalNet sequencer, along with the small number
+	// of blocks per session, multiple proofs may be created throughout the duration of the test. Until
+	// these values are appropriately adjusted, we assert on an increase in proofs rather than +1.
+	require.Greater(s, len(allProofsRes.Proof), len(preExistingProofs), "number of proofs must have increased")
+
+	// TODO_IMPROVE: assert that the root hash of the proof contains the correct
```
h5law, line 159: I think this can be added now - if not already merged with main. `MerkleRoot.Sum()` exists now

e2e/tests/session_steps_test.go
```diff
@@ -163,3 +185,42 @@ func (s *suite) sendRelaysForSession(
...
+// waitForMessageAction waits for an event to be observed which has the given message action.
+func (s *suite) waitForMessageAction(action string) {
+	ctx, done := context.WithCancel(context.Background())
+
+	eventsReplayClient, ok := s.scenarioState[eventsReplayClientKey].(client.EventsReplayClient[*abci.TxResult])
+	require.True(s, ok, "eventsReplayClientKey not found in scenarioState")
+	require.NotNil(s, eventsReplayClient)
+
+	// For each observed event, **asynchronously** check if it contains the given action.
+	channel.ForEach[*abci.TxResult](
```
h5law, line 198: 🔥

x/supplier/keeper/msg_server_create_claim.go
```diff
@@ -25,7 +28,7 @@ func (k msgServer) CreateClaim(goCtx context.Context, msg *suppliertypes.MsgCrea
 		msg.GetSupplierAddress(),
 	)
 	if err != nil {
-		return nil, err
+		return nil, status.Error(codes.InvalidArgument, err.Error())
```
h5law, line 31: Is this something we should be doing in message handlers, over returning an error from `types/errors.go`? Or is it specific for this use case that we are not doing so?

x/supplier/keeper/msg_server_create_claim_test.go
```diff
@@ -96,21 +107,21 @@ func TestMsgServer_CreateClaim_Error(t *testing.T) {
...
 			createClaimRes, err := srv.CreateClaim(ctx, tt.claimMsgFn(t))
-			require.ErrorIs(t, err, tt.expectedErr)
+			require.ErrorContains(t, err, tt.expectedErr.Error())
 			require.Nil(t, createClaimRes)
 		})
 	}
 }
 
-func newTestClaimMsg(t *testing.T) *suppliertypes.MsgCreateClaim {
+func newTestClaimMsg(t *testing.T, sessionId string) *suppliertypes.MsgCreateClaim {
```
h5law, line 116: NIT: 
```suggestion
func newTestClaimMsg(t *testing.T, sessionID string) *suppliertypes.MsgCreateClaim {
```

x/supplier/keeper/msg_server_submit_proof.go
```diff
@@ -4,55 +4,130 @@ import (
...
+		4. relay.Res.signer == msg.supplier
+
+		## Msg distribution validation (governance based params)
+		1. Validate Proof submission is not too early; governance-based param + pseudo-random variation
+		2. Validate Proof submission is not too late; governance-based param + pseudo-random variation
+
+		## Relay Mining validation
+		1. verify(proof.path) is the expected path; pseudo-random variation using on-chain data
+		2. verify(proof.ValueHash, expectedDiffictulty); governance based
```
h5law, line 40: Does this step include validation of the relay `proof.ValueHash` itself? Signatures etc? Ensuring the relay is an actual relay should be here, if I have missed that ignore this.

x/supplier/keeper/msg_server_submit_proof.go
```diff
@@ -4,55 +4,130 @@ import (
...
+		3. verify(claim.Root, proof.ClosestProof); verify the closest proof is correct
+	*/
 
 	if err := msg.ValidateBasic(); err != nil {
-		return nil, err
+		return nil, status.Error(codes.InvalidArgument, err.Error())
```
h5law, line 45: This seems like a pattern you have found, is it worth updating the codebase **post migration** to do this for all message handlers? Should we create an issue for this?

It looks like it stops us from using `require.ErrorIs` / `require.ErrorAs` so some tests will need to be updated too.

x/supplier/types/message_create_claim.go
```diff
@@ -47,27 +47,27 @@ func (msg *MsgCreateClaim) GetSignBytes() []byte {
 
 func (msg *MsgCreateClaim) ValidateBasic() error {
 	// Validate the supplier address
-	_, err := sdk.AccAddressFromBech32(msg.SupplierAddress)
+	_, err := sdk.AccAddressFromBech32(msg.GetSupplierAddress())
 	if err != nil {
-		return sdkerrors.Wrapf(ErrSupplierInvalidAddress, "invalid supplierAddress address (%s)", err)
+		return sdkerrors.Wrapf(ErrSupplierInvalidAddress, "%s", msg.GetSupplierAddress())
```
h5law, line 52:
```suggestion
		return ErrSupplierInvalidAddress.Wrapf("%s", msg.GetSupplierAddress())
```
Same for all the others

x/supplier/types/message_submit_proof.go
```diff
@@ -40,10 +40,38 @@ func (msg *MsgSubmitProof) GetSignBytes() []byte {
...
+// ValidateBasic ensures that the bech32 address strings for the supplier and
+// application addresses are valid and that the proof and service ID are not empty.
+//
+// TODO_TECHDEBT: Call `msg.GetSessionHeader().ValidateBasic()` once its implemented
 func (msg *MsgSubmitProof) ValidateBasic() error {
-	_, err := sdk.AccAddressFromBech32(msg.SupplierAddress)
+	_, err := sdk.AccAddressFromBech32(msg.GetSupplierAddress())
 	if err != nil {
-		return sdkerrors.Wrapf(sdkerrors.ErrInvalidAddress, "invalid supplierAddress address (%s)", err)
+		return sdkerrors.ErrInvalidAddress.Wrapf(
```
h5law, line 50:
```suggestion
		return ErrSupplierInvalidAddress.Wrapf(
```
I think you should define this type something like `invalid supplier address`

x/supplier/types/message_submit_proof.go
```diff
@@ -40,10 +40,38 @@ func (msg *MsgSubmitProof) GetSignBytes() []byte {
...
+			err,
+		)
+	}
+
+	if msg.GetSessionHeader().GetService().GetId() == "" {
+		return ErrSupplierInvalidService.Wrap("proof service ID %q cannot be empty")
+	}
+
+	if len(msg.GetProof()) == 0 {
+		return sdkerrors.ErrInvalidRequest.Wrap("proof cannot be empty")
```
h5law, line 71: Same as above - this type should be defined `ErrSupplierEmptyProof`

x/supplier/types/message_submit_proof_test.go
```diff
@@ -7,34 +7,93 @@ import (
...
+				SessionHeader: &sessiontypes.SessionHeader{
+					ApplicationAddress:      "not_a_bech32_address",
+					Service:                 testService,
+					SessionStartBlockHeight: 0,
+					SessionId:               "mock_session_id",
+					SessionEndBlockHeight:   5,
+				},
+				Proof: testClosestMerkleProof,
+			},
+			expectedErr: sdkerrors.ErrInvalidAddress.Wrapf(
```
h5law, line 36: See the messages above about defining a type for this specific to the supplier module

h5law (changes requested):
Nice work, looks great - saw a pattern for error handling in messages emerge left some comments on that but overall this is coming along great!

e2e/tests/session_steps_test.go
Olshansk: Or a `TODO_UPNEXT(@bryanchriswhite)` in a small PR.

x/supplier/keeper/msg_server_submit_proof.go
Olshansk: Step #3 from the first section?

x/supplier/keeper/msg_server_submit_proof.go
```diff
@@ -4,55 +4,130 @@ import (
...
+		## Basic Validations (metadata only)
+		1. proof.sessionId == claim.sessionId
+		2. msg.supplier in session.suppliers
+		3. relay.Req.signer == session.appAddr
+		4. relay.Res.signer == msg.supplier
+
+		## Msg distribution validation (governance based params)
+		1. Validate Proof submission is not too early; governance-based param + pseudo-random variation
+		2. Validate Proof submission is not too late; governance-based param + pseudo-random variation
+
```
Olshansk, line 37: Per @h5law's comments:

```suggestion

## Relay Signature validation
1. verify(relay.Req.Signature, app.pubKey)
2. verify(relay.Res.Signature, supplier.pubKey)

```

Olshansk (changes requested):
Last minor set of comments in addition to those left by others

x/supplier/keeper/msg_server_submit_proof.go
h5law: To be clear here we would need to grab the ring for the app not the pubkey to validate the request's signature

bryanchriswhite: Great clarification, thank you!

e2e/tests/session.feature
bryanchriswhite: My thinking was that there should be an event signaling claim creation and another signaling proof submission.

x/supplier/keeper/msg_server_submit_proof.go
bryanchriswhite: > Is extracting the relay process captured somewhere? Can't find any TODO for it.
As per @Olshansk's comment, I think that's implied by "Actions" #3.

e2e/tests/session_steps_test.go
bryanchriswhite:
```suggestion
	// TODO_UPNEXT(@bryanchriswhite): assert that the root hash of the proof contains the correct
```

x/supplier/keeper/msg_server_submit_proof.go
bryanchriswhite: The CLI should ALWAYS return a gRPC status error (i.e. `status.Error(...)`), in addition to consistency, this is an important aspect of separating client-facing errors from internal errors. Details of internal errors should only appear in logs that the operator sees and NEVER be sent to the client to avoid leaking potentially sensitive information.

This pattern fell out of the refactoring that was necessary in #266/#250. The tests in those branches take this into consideration (and will be re-applied in subsequent work).

IIRC the reason I pushed usage of `status.Error` all the way down into the keeper was because there was a lot of redundancy appearing in the CLI code involving wrapping keeper errors in the same status code error. Given that the only consumer of the keepers should be the CLI, or one another, it didn't seem inappropriate to make that simplification. I don't have a strong opinion about it though.

x/supplier/types/message_create_claim.go
bryanchriswhite: Thanks for calling this out! :raised_hands: 

I thought I got all of these but I either missed some or there were some reversions between the cherry-picking, etc.

bryanchriswhite: After looking into this for second:

image

It seems like we have lots of usages of `sdkerrors.Wrapf()` across the modules (none of which are in tests). This leads me to believe that this is baked in to the ignite generator template (in whatever ignite version was used). 

I would prefer to batch this out in a separate PR **after v0.50 migration**, in case we do decide to transplant regenerate module code.

bryanchriswhite: Captured in #347.

x/supplier/types/message_submit_proof.go
bryanchriswhite: I created #348 to capture the higher-level concerns that I see as in play here.

In this particular case, my rationale for using `sdkerrors.ErrInvalidAddress` over any of our own error variables is that the usage here is identical to that of the what I interpreted as its intended usage, as indicated by its actual usage in cosmos-sdk code (i.e. in cases where `sdk.AccAddressFromBech32()` returns an error in `#ValidateBasic()` methods).

bryanchriswhite: I could imagine a middle-of-the-road alternative as us defining an alias to that error, such that we reference the alias instead. However, this begs the question, "in which module should that error variable be defined?"

x/supplier/types/message_submit_proof_test.go
bryanchriswhite: This must match whatever error variable we decide to go with in the above. Let's continue the conversation there (https://github.com/pokt-network/poktroll/pull/328#discussion_r1466082964).

x/supplier/types/message_submit_proof.go
bryanchriswhite: This one I feel less strongly about and would tend to agree; however, I find myself looking for guidance related to the uncertainty outlined in #348. At this point, my intuition would be to create a new `supplier.ErrSupplierPoof` variable and use it instead (with the same message string). Wdyt?

x/supplier/keeper/msg_server_create_claim.go
bryanchriswhite: (see: https://github.com/pokt-network/poktroll/pull/328#discussion_r1465982213)

Olshansk (approved):
Looks good from my end!

I know @h5law has some open comments for current/future work, so definitely should wait for a 👍 from his end as well.

h5law (approved):

Issue #230, opened 2023-11-28
[Service] Require payment when staking for a new service


Olshansk:
## Objective

Prevent permissionless Sybil attacks by the introduction of new services.

## Goals

- Disincentivize Sybil attacks on the network
- Allow Pocket Network to scale to any number of actors/services permissionlessley
- Prevent state/chain bloat by allowing applications to introduce an infinite number of new services on the network

## Deliverables

- [ ] Introduce a new `AddService` command on the network that allows anyone (actor or not) to stake a new `serviceID` in exchange for some fee (a governance param)
- [ ] Block/fail/ignore any on-chain state transition that is performed for `serviceId` that has not been added.

## Non-goals / Non-deliverables

- Modelling out all the tokenomic 'repercussions' of this feature

## General deliverables

- [ ] **Comments**: Add/update TODOs and comments alongside the source code so it is easier to follow.
- [ ] **Testing**: Add new tests (unit and/or E2E) to the test suite.
- [ ] **Makefile**: Add new targets to the Makefile to make the new functionality easier to use.
- [ ] **Documentation**: Update architectural or development READMEs; use [mermaid](https://mermaid-js.github.io/mermaid/) diagrams where appropriate.

---

**Creator**: @Olshansk 
**Co-Owners**: @h5law @RawthiL

RossiNYC:
@Olshansk - who collects the fee here? PNF?

Olshansk:
> @Olshansk - who collects the fee here? PNF?

It will either be burnt or paid to the DAO.

PR #337, opened 2024-01-19
[Service] Add `AddServiceFee` enforce and deduct it during the `AddService` command

h5law:
## Summary

- Adds the `AddServiceFee` parameter
- Enforces the `AddServiceFee` parameter by deducting the balance from the
  user if they don't have enough funds
- Adds extensive tests and mocks

### Summary generated by Reviewpad on 24 Jan 24 11:45 UTC

- `genesis_test.go` in the `x/service/types` package:
  - Added a new service `svc3` with ID and name.
  - Modified the test cases to include a `Params` field in `GenesisState` and update the `ServiceList` field accordingly.
  - Added new test cases to check for duplicate service ID and name, and invalid add service fee parameter.

- `msg_server_add_service.go`:
  - Added check for signer's funds to cover the `AddServiceFee` parameter.
  - If sufficient funds, deduct the fee from signer's account and debit to service module's account, then add the service on-chain.
  - Other changes include validation of incoming message, checking for existing service, validating actor's address, checking for actor's sufficient funds, and checking if token balance is enough.

- `msg_server_add_service_test.go`:
  - Added new test cases with various scenarios (valid service addition, invalid service supplier address, missing/mismatched service ID and name, existing service, and insufficient/upokt balance).

- `keeper.go` in `x/service/keeper`:
  - Added new `Keeper` type and related functions for the service module.
  - Updated `NewKeeper` function with additional parameters.
  - Added `Logger` function to provide module-specific logger.

- `params.proto`:
  - Added new field `add_service_fee` of type uint64 to represent the amount of uPOKT required to add a new service.

- `openapi.yml`:
  - Added `add_service_fee` property of type `string` and format `uint64` to `pocket.service.Params` and `pocket.service.QueryParamsResponse` definitions.

- `keeper.go` in `testutil/keeper/service`:
  - Added import statements and functions for testing purposes.

- `genesis.go` in `x/service/types`:
  - Added new maps and modified validation logic for `ServiceList` in `GenesisState` struct.

- `app.go`:
  - Updated function calls with additional parameters in various modules.

- `Makefile`:
  - Added new `go generate` command to generate mocks in `x/service/types/` directory.

- Error handling changes, configuration changes, and addition of `go:generate` comment.

Please review the changes for accuracy and logic. Let me know if you need further assistance with this diff.

## Issue

- Fixes (in part) [#230](https://github.com/pokt-network/poktroll/issues/230)

## Type of change

Select one or more:

- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

## Sanity Checklist


x/service/keeper/msg_server_add_service.go
```diff
@@ -23,14 +24,57 @@ func (k msgServer) AddService(
...
 	if _, found := k.GetService(ctx, msg.Service.Id); found {
 		logger.Error(fmt.Sprintf("Service already exists: %v", msg.Service))
-		return nil, types.ErrServiceAlreadyExists
+		return nil, types.ErrServiceAlreadyExists.Wrapf(
+			"duplicate ID: %s", msg.Service.Id,
+		)
+	}
+
+	// Retrieve the address of the actor adding the service.
+	accAddr, err := sdk.AccAddressFromBech32(msg.Address)
```
Olshansk, line 42: _Not actionable, just a question._

I looked at the docs in [1], and some of the underlying code. Do you know if we are guaranteed that the signature is validated by the time `msgServer` receives it?

[1] https://docs.cosmos.network/main/build/building-modules/protobuf-annotations#signer

x/service/keeper/msg_server_add_service.go
```diff
@@ -23,14 +24,57 @@ func (k msgServer) AddService(
...
+	if err != nil {
+		logger.Error(fmt.Sprintf("could not parse address %s", msg.Address))
+		return nil, types.ErrServiceInvalidAddress.Wrapf(
+			"%s is not in Bech32 format", msg.Address,
+		)
+	}
+
+	// Check the actor has sufficient funds to pay for the add service fee.
```
Olshansk, line 50: 🔥

x/service/keeper/msg_server_add_service.go
```diff
@@ -23,14 +24,57 @@ func (k msgServer) AddService(
...
+	if accCoins.Len() == 0 {
+		logger.Error(fmt.Sprintf("%s doesn't have any funds to add service: %v", msg.Address, err))
+		return nil, types.ErrServiceNotEnoughFunds.Wrapf(
+			"account has no spendable coins",
+		)
+	}
+
+	// Check the balance of upokt is enough to cover the AddServiceFee.
+	accBalance := accCoins.AmountOf("upokt")
+	if accBalance.LTE(sdk.NewIntFromUint64(k.GetParams(ctx).AddServiceFee)) {
```
Olshansk, line 61: NIT: Can you move `sdk.NewIntFromUint64(k.GetParams(ctx).AddServiceFee))` into a local var for readability?

x/service/keeper/msg_server_add_service.go
```diff
@@ -23,14 +24,57 @@ func (k msgServer) AddService(
...
+		logger.Error(fmt.Sprintf("%s doesn't have enough funds to add service: %v", msg.Address, err))
+		return nil, types.ErrServiceNotEnoughFunds.Wrapf(
+			"account has %d uPOKT, but the service fee is %d uPOKT",
+			accBalance.Uint64(), k.GetParams(ctx).AddServiceFee,
+		)
+	}
+
+	// Deduct the service fee from the actor's balance.
+	serviceFee := sdk.Coins{sdk.NewCoin("upokt", sdk.NewIntFromUint64(k.GetParams(ctx).AddServiceFee))}
+	err = k.bankKeeper.DelegateCoinsFromAccountToModule(ctx, accAddr, types.ModuleName, serviceFee)
```
Olshansk, line 71: Per my other comment, I think this should be send, not delegate.

x/service/keeper/msg_server_add_service.go
```diff
@@ -23,14 +24,57 @@ func (k msgServer) AddService(
...
+	if accBalance.LTE(sdk.NewIntFromUint64(k.GetParams(ctx).AddServiceFee)) {
+		logger.Error(fmt.Sprintf("%s doesn't have enough funds to add service: %v", msg.Address, err))
+		return nil, types.ErrServiceNotEnoughFunds.Wrapf(
+			"account has %d uPOKT, but the service fee is %d uPOKT",
+			accBalance.Uint64(), k.GetParams(ctx).AddServiceFee,
+		)
+	}
+
+	// Deduct the service fee from the actor's balance.
+	serviceFee := sdk.Coins{sdk.NewCoin("upokt", sdk.NewIntFromUint64(k.GetParams(ctx).AddServiceFee))}
```
Olshansk, line 70: You'd be able to reuse the local var from above here.

x/service/keeper/msg_server_add_service_test.go
```diff
@@ -13,18 +13,29 @@ import (
 	sharedtypes "github.com/pokt-network/poktroll/x/shared/types"
 )
 
+// oneUPOKTGreaterThanFee is 1 upokt more than the AddServiceFee of 10000000000
+const oneUPOKTGreaterThanFee = 10000000001
```
Olshansk, line 17: NIT: Use `DefalutXXX + 1`

x/service/keeper/msg_server_add_service_test.go
```diff
@@ -107,26 +126,67 @@ func TestMsgServer_AddService(t *testing.T) {
...
+			},
+			address:       validAddr2,
+			service:       svc1,
+			expectedError: types.ErrServiceNotEnoughFunds,
+		},
 	}
 
 	for _, tt := range tests {
 		t.Run(tt.desc, func(t *testing.T) {
+			tt.preFn(t)
```
Olshansk, line 182: This is a cool pattern, but I'm a bit worried about the implicit expectation of order without parallelizability or other bugs that could happen.

Why don't we need a cleanup function that would be called `t.Cleanup(tt.teardown)`? If we do this, I think renaming `preFn` to `setup` just do be a bit more idiomatic with testing lingo.

Olshansk (changes requested):

testutil/keeper/service.go
```diff
@@ -12,12 +13,23 @@ import (
...
+	"github.com/golang/mock/gomock"
 	"github.com/stretchr/testify/require"
 
+	"github.com/pokt-network/poktroll/testutil/service/mocks"
 	"github.com/pokt-network/poktroll/x/service/keeper"
 	"github.com/pokt-network/poktroll/x/service/types"
 )
 
+var (
+	// mapAccAddrCoins is used by the mock BankModule to determine who has what
```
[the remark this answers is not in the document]
h5law: I think this is simpler than the in-memory network personally but will be good to use it once we have the in-memory network available.

x/service/types/params.go
```diff
@@ -1,10 +1,16 @@
...
 import (
+	sdkerrors "cosmossdk.io/errors"
 	paramtypes "github.com/cosmos/cosmos-sdk/x/params/types"
 	"gopkg.in/yaml.v2"
 )
 
+// DefaultAddServiceFee is the default value for the add service fee
+// parameter in the genesis state of the service module.
+// TODO: Revisit default param values
+const DefaultAddServiceFee = 1000000000 // 1000 POKT
```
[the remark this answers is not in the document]
h5law: @Olshansk can you explain this name change to me? This would imply there is a variable fee to add a service - if there is a min there should also be a maximum no?

x/service/types/params.go
```diff
@@ -29,6 +35,14 @@ func (p *Params) ParamSetPairs() paramtypes.ParamSetPairs {
 
 // Validate validates the set of params
 func (p Params) Validate() error {
+	// TODO(@h5law): Look into better validation
+	if p.AddServiceFee < 1000000 { // 1 POKT
```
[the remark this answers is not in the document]
h5law: @Olshansk I like the idea of validating against the service fee default value - still unsure on adding the `Minimum` could you explain a bit more around your reasoning here?

x/service/keeper/msg_server_add_service_test.go
h5law: We dont need a cleanup bc the method called in the `setup` method has it's own cleanup as a helper which is called when the `t.Run` closes.

x/service/types/params.go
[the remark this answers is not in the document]
Olshansk: Yea, I was thinking that it would be like variable tx fees (at the time of writing this comment), but believe it overcomplicated things so let's ignore.

Olshansk (commented):
@h5law Changes LGTM, but not approving yet because tests seem broken. Can you merge with main and run E2E tests before we get this in?

Olshansk (approved):

Issue #366, opened 2024-02-06
[Sessions] Block hash store does not update correctly


red-0ne:
## Objective

Fix  `session mismatch` errors occurring when `RelayRequests` are processed at some specific block heights.

Have a flawless relaying process at any time in a session or a rolled-over session and at any `RelayRequest` volume.

## Origin Document

**Resulting `RelayResponse`**
![image](https://github.com/pokt-network/poktroll/assets/231488/28846ed0-15d6-4932-81c1-42248504e44a)

**Sequencer's logs**
![image](https://github.com/pokt-network/poktroll/assets/231488/f86bbf07-3a6b-4218-8fb0-9ca17601f2fc)

## Goals

- Ensure that no unwanted `session mismatch` error is occurring during `RelayRequest` verification that may be caused by `GetSession` being called at sensitive points in time.
- Ensure deterministic `GetSession` queries.

## Deliverables

- [ ] A single PR addressing the fact that `GetSession` may not have access to the  stored `blockHash` used in `sessionId` construction.

## Non-goals / Non-deliverables

- Implement a new solution for getting past block hashes.
- Change the Pocket protocol to address the issue.

## General deliverables

- [ ] **Comments**: Add/update TODOs and comments alongside the source code so it is easier to follow.
- [ ] **Testing**: Add new tests (unit and/or E2E) to the test suite.
---

**Creator**: @red-0ne 
**Co-Owners**: @Olshansk

red-0ne:
It seems that Tendermint is broadcasting new blocks to the off-chain actors (through websocket) and to Cosmos at the same time.

Making the off-chain actors aware of the latest block height `N` while Cosmos is currently processing block height `N`.

This results in having the session querier trying to hydrate `Sessions` for blocks that does not have their `hash` stored yet  (`blockHeight == ctx.BlockHeight()`). 

_`ctx.BlockHeight()` being the block that is currently being processed by Cosmos._

Olshansk:
> It seems that Tendermint is broadcasting new blocks to the off-chain actors (through websocket) and to Cosmos at the same time.

Is this not expected behaviour?

> Making the off-chain actors aware of the latest block height N while Cosmos is currently processing block height N.

I looked at the [list of events](https://pkg.go.dev/github.com/tendermint/tendermint/types#pkg-constants) and found this:

[...]

PR #323, opened 2024-01-16
[Tokenomics] SettleSessionAccounting first implementation (emissions=burn)

Olshansk:
## Summary

Implement `SettleSessionAccounting` where emissions=burn.

## Issue

- #243 

## Type of change

Select one or more:

- [x] New feature, functionality or library
- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

- [x] **Run all unit tests**: `make go_develop_and_test`
- [ ] **Run E2E tests locally**: `make test_e2e`
- [ ] **Run E2E tests on DevNet**: Add the `devnet-test-e2e` label to the PR. This is VERY expensive, only do it after all the reviews are complete.

## Sanity Checklist

- [x] I have tested my changes using the available tooling
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, updated documentation and left TODOs throughout the codebase

PR #368, opened 2024-02-07
[Session] fix: session mismatch due to off-by-one block height

red-0ne:
### Human Summary

Fix off-by-one block height causing `session mismatch` errors during relaying.

### AI Summary

reviewpad:summary

## Issue

- #366 

## Type of change

Select one or more:

- [x] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

## Sanity Checklist

### DIFF
--- a/e2e/tests/session.feature
+++ b/e2e/tests/session.feature
@@ -5,14 +5,16 @@ Feature: Session Namespace
   Scenario: Supplier completes claim/proof lifecycle for a valid session
     Given the user has the pocketd binary installed
     When the supplier "supplier1" has serviced a session with "5" relays for service "svc1" for application "app1"
+    # TODO_TECHDEBT: Once the session grace period is configurable, set it to 0 at the beginning of this test.
+    # The timeout for when a claim can be submitted on-chain depends on `createClaimWindowStartHeight`, which
+    # is a function of `SessionGracePeriod`. The higher this value, the higher this timeout needs to be. Since
+    # this test is not dependant on the grace period, setting it to 0 and having a lower grace period will simplify it.
     And the user should wait for "7" seconds
     Then the claim created by supplier "supplier1" for service "svc1" for application "app1" should be persisted on-chain
-#  TODO_IMPROVE: ...
-#    And an event should be emitted...
+    # TODO_IMPROVE: And an event should be emitted...
     And after the supplier submits a proof for the session for service "svc1" for application "app1"
     Then the proof submitted by supplier "supplier1" for service "svc1" for application "app1" should be persisted on-chain
-#  TODO_IMPROVE: ...
-#    And an event should be emitted...
+    # TODO_IMPROVE: And an event should be emitted...
 
 # TODO_BLOCKER(@red-0ne): Make sure to implement and validate this test
 # One way to exercise this behavior is to close the `RelayMiner` port to prevent
--- a/e2e/tests/session.feature
+++ b/e2e/tests/session.feature
@@ -5,7 +5,7 @@ Feature: Session Namespace
   Scenario: Supplier completes claim/proof lifecycle for a valid session
     Given the user has the pocketd binary installed
     When the supplier "supplier1" has serviced a session with "5" relays for service "svc1" for application "app1"
-    And the user should wait for "5" seconds
+    And the user should wait for "7" seconds
     Then the claim created by supplier "supplier1" for service "svc1" for application "app1" should be persisted on-chain
 #  TODO_IMPROVE: ...
 #    And an event should be emitted...
--- a/x/session/keeper/keeper.go
+++ b/x/session/keeper/keeper.go
@@ -62,9 +62,9 @@ func (k Keeper) BeginBlocker(ctx sdk.Context) {
 	hash := ctx.BlockHeader().LastBlockId.Hash
 	// ctx.BlockHeader().Height is the height of the last committed block.
 	height := ctx.BlockHeader().Height
-	// When height is 1, ctx.BlockHeader().LastBlockId.Hash will not be populated,
-	// and leads to a panic when trying to save the hash in the store.
-	// Explicitly setting hash to an empty byte slice when height is 1 prevents the panic.
+	// Block height 1 is the genesis block, there is no state root since there is
+	// no block preceding it.
+	// Fallback to an empty byte slice in this case.
 	if height == 1 {
 		hash = []byte{}
 	}
--- a/x/session/keeper/keeper.go
+++ b/x/session/keeper/keeper.go
@@ -62,8 +62,9 @@ func (k Keeper) BeginBlocker(ctx sdk.Context) {
 	hash := ctx.BlockHeader().LastBlockId.Hash
 	// ctx.BlockHeader().Height is the height of the last committed block.
 	height := ctx.BlockHeader().Height
-	// If height is 1, ctx.BlockHeader().LastBlockId.Hash will be nil,
-	// fall back to an empty byte slice.
+	// When height is 1, ctx.BlockHeader().LastBlockId.Hash will not be populated,
+	// and leads to a panic when trying to save the hash in the store.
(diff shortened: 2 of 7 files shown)


x/session/keeper/keeper.go
```diff
@@ -60,12 +60,11 @@ func (k Keeper) Logger(ctx sdk.Context) log.Logger {
...
-	// ctx.BlockHeader().Height is the height of the current block being processed,
-	// decrementing it by 1 gives us the height of the last block committed.
-	height := ctx.BlockHeader().Height - 1
-	// If height is 0, ctx.BlockHeader().LastBlockId.Hash will be nil,
+	// ctx.BlockHeader().Height is the height of the last committed block.
+	height := ctx.BlockHeader().Height
+	// If height is 1, ctx.BlockHeader().LastBlockId.Hash will be nil,
 	// fall back to an empty byte slice.
-	if height == 0 {
+	if height == 1 {
```
Olshansk, line 67: Why is this 1?

If the last commit block is 1, it means we have committed a real block. Genesis is 0.

x/session/keeper/keeper.go
```diff
@@ -60,12 +60,11 @@ func (k Keeper) Logger(ctx sdk.Context) log.Logger {
 func (k Keeper) BeginBlocker(ctx sdk.Context) {
 	// ctx.BlockHeader().LastBlockId.Hash is the hash of the last block committed
 	hash := ctx.BlockHeader().LastBlockId.Hash
-	// ctx.BlockHeader().Height is the height of the current block being processed,
-	// decrementing it by 1 gives us the height of the last block committed.
-	height := ctx.BlockHeader().Height - 1
-	// If height is 0, ctx.BlockHeader().LastBlockId.Hash will be nil,
+	// ctx.BlockHeader().Height is the height of the last committed block.
+	height := ctx.BlockHeader().Height
```
Olshansk, line 64: @red-0ne Can you explain what we missed or misunderstood in our last review cycle? I want to understand why it got through code review.

Screenshot 2024-02-06 at 9 26 08 PM

Olshansk (changes requested):

x/session/keeper/keeper.go
red-0ne: Yes, block `1` should behave just like the others, but for some reason it doesn't have its `ctx.BlockHeader().LastBlockId` populated and the `Hash` has its 0 value that panics when the CosmosSDK store checks for valid values.

The empty byte slice seems to fix it (but we can assign it any valid byte slice value)

image

Will have a better comment if you are OK with the solution.

e2e/tests/session.feature
```diff
@@ -5,7 +5,7 @@ Feature: Session Namespace
   Scenario: Supplier completes claim/proof lifecycle for a valid session
     Given the user has the pocketd binary installed
     When the supplier "supplier1" has serviced a session with "5" relays for service "svc1" for application "app1"
-    And the user should wait for "5" seconds
+    And the user should wait for "7" seconds
```
Olshansk, line 8:
```suggestion
    # TODO_TECHDEBT: Once the session grace period is configurable, set it to 0
    # at the beginning of the test to reduce this wait time as much as possible.
    And the user should wait for "7" seconds
```

x/session/keeper/keeper.go
Olshansk: Can you #PUC to a link that proves `ctx.BlockHeader().Height is the height of the last committed block.`

x/session/keeper/keeper.go
Olshansk: I appreciate the workaround and the detailed comment, but I still don't understand "why". maybe it's a bug in cosmos or rollkit? I want to be convinced that we're not bandainding this but actually understand why its the solution.

For example, see this comment where I reached out to cosmos when I didn't understand something in the past: https://github.com/pokt-network/poktroll/pull/323#discussion_r1463541516

Olshansk (changes requested):

x/session/keeper/keeper.go
red-0ne: This has been confirmed by comparing the block heights and hashes we got from the CometBFT websocket client and the values of `ctx.BlockHeader().Height` and `ctx.BlockHeader().LastBlockId.Hash`.

I couldn't find any statement of that in the CosmosSDK or code. The closest I got was this reference:

The only relevant info I found was this https://github.com/cosmos/cosmos-sdk/blob/main/baseapp/grpcserver.go#L60

e2e/tests/session.feature
red-0ne: I don't understand how this relates to session grace period configuration. To me it's the time we need to wait before querying for claims and see if they incremented. I believe it mostly serves as a timeout.

x/session/keeper/keeper.go
red-0ne: image

This should explain why there's no hash for block `1`. Will update the comments accordingly.

We are also assuming that genesis starts at block `0`. We should revisit our session start/end calculations and their corresponding tests. Will create a ticket for that.

x/session/keeper/query_get_session.go
```diff
@@ -21,16 +21,14 @@ func (k Keeper) GetSession(goCtx context.Context, req *types.QueryGetSessionRequ
 
 	ctx := sdk.UnwrapSDKContext(goCtx)
 
-	// If block height is not specified, use the current (context's latest) block height
+	// GetSession should be deterministic and always return the same session for
```
Olshansk, line 24: Can you please move this into a godoc comment above

x/session/keeper/query_get_session.go
```diff
@@ -21,16 +21,14 @@ func (k Keeper) GetSession(goCtx context.Context, req *types.QueryGetSessionRequ
...
+	// the same block height.
 	// Note that `GetSession` is called via the `Query` service rather than the `Msg` server.
 	// The former is stateful but does not lead to state transitions, while the latter one
 	// does. The request height depends on how much the node has synched and only acts as a read,
 	// while the `Msg` server handles the code flow of the validator/sequencer when a new block
 	// is being proposed.
 	blockHeight := req.BlockHeight
```
Olshansk, line 31: Consider adding a logigng statement "getting session for height blockHeight"

x/session/keeper/keeper.go
```diff
@@ -60,12 +60,12 @@ func (k Keeper) Logger(ctx sdk.Context) log.Logger {
...
 	hash := ctx.BlockHeader().LastBlockId.Hash
-	// ctx.BlockHeader().Height is the height of the current block being processed,
-	// decrementing it by 1 gives us the height of the last block committed.
-	height := ctx.BlockHeader().Height - 1
-	// If height is 0, ctx.BlockHeader().LastBlockId.Hash will be nil,
-	// fall back to an empty byte slice.
-	if height == 0 {
+	// ctx.BlockHeader().Height is the height of the last committed block.
+	height := ctx.BlockHeader().Height
+	// When height is 1, ctx.BlockHeader().LastBlockId.Hash will not be populated,
```
Olshansk, line 65: @red-0ne Please create a ticket for `We are also assuming that genesis starts at block 0. We should revisit our session start/end calculations and their corresponding tests. Will create a ticket for that.` and add a TODO_TECHDEBT in the code in the appropriate location.

x/session/keeper/keeper.go
```diff
@@ -60,12 +60,12 @@ func (k Keeper) Logger(ctx sdk.Context) log.Logger {
...
+	// and leads to a panic when trying to save the hash in the store.
```
Olshansk, line 66: @red-0ne This is a good comment but if we copy-paste (and reword) what marko told you as a comment, there will be no ambiguity fo anyone.

Olshansk (changes requested):

Olshansk (approved):

PR #397, opened 2024-02-23
[Makefile] Check if gnu grep and sed needs to be installed on MacOS

okdas:
Use GNU grep and sed on macos. Install if missing.


Olshansk (approved):
Will add some TODO_IMPROVE

Issue #401, opened 2024-02-29
[Proof] Relay signature & Merkle proof validation


bryanchriswhite:
## Objective

The proof module MUST validate Merkle proofs (i.e. [`MsgSubmitProof#closest_merkle_proof`](https://github.com/pokt-network/poktroll/blob/main/proto/poktroll/proof/proof.proto#L15)) in its [`msgServer#SubmitProof()` method](https://github.com/pokt-network/poktroll/blob/main/x/proof/keeper/msg_server_submit_proof.go#L12).

- May depend on #377.
- **Integration test implementation depends on #404.**

## Origin Document

![image](https://github.com/pokt-network/poktroll/assets/600733/298e9595-5837-4edd-b4bf-85936ff17773)

## Goals

- Validate that the merkle proof's root, path, value hash match on-chain expectations.
- Validate that the relay (proof value) request and response signatures match the application and supplier (respectively) for the given session.

## Deliverables

- [ ] A PR which implements merkle proof validation.
- [ ] A PR which implements relay unpacking and validation.
- [ ] A PR which adds `relay_min_difficulty` to the proof module.
- [ ] A PR which implements relay minimum difficulty validation.

## Non-goals / Non-deliverables

- Refactoring outside the scope of the goals outlined in this issue.
- Additional validation (e.g. distribution).

## General deliverables

- [ ] **Comments**: Add/update TODOs and comments alongside the source code so it is easier to follow.
- [ ] **Testing**: Add new tests (unit and/or E2E) to the test suite.
- [ ] **Makefile**: Add new targets to the Makefile to make the new functionality easier to use.
- [ ] **Documentation**: Update architectural or development READMEs; use [mermaid](https://mermaid-js.github.io/mermaid/) diagrams where appropriate.

---

**Creator**: @bryanchriswhite 
**Co-Owners**:

red-0ne:
Closed as per #406

PR #406, opened 2024-03-01
[Proof] Relay signature & Merkle proof validation

red-0ne:
### Human Summary

Validate relay signatures and and merkle proof:
1. Ensure that both relay and request session headers match the (already validated) proof session header
2. Unmarshal the `SparseMerkleClosestProof` and extract the `Relay` that is present in the `ClosestValueHash` property
3. Verify the `RelayResponse` signature
4. Verify the `RelayRequest` signature
5. Ensure that `proof.Path` matches the earliest proof submission block hash
6. Ensure that the mining difficulty meets the required difficulty
7. Ensure that the proof verifies against the expected root hash

### AI Summary

reviewpad:summary

## Issue

- #401 

## Type of change

Select one or more:

- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

## Sanity Checklist


x/service/types/relay.go
```diff
@@ -15,7 +15,12 @@ func (req RelayRequest) getSignableBytes() ([]byte, error) {
 // Hashing the marshaled request message guarantees that the signable bytes are
 // always of a constant and expected length.
 func (req *RelayRequest) GetSignableBytesHash() ([32]byte, error) {
+	signature := req.Meta.Signature
 	requestBz, err := req.getSignableBytes()
+
+	// Set the signature back to its original value
+	req.Meta.Signature = signature
```
bryanchriswhite, line 22: Why do we need to do this?

x/proof/keeper/msg_server_submit_proof.go
```diff
@@ -127,12 +209,354 @@ func (k msgServer) queryAndValidateClaimForProof(ctx context.Context, proof *typ
...
+		)
+	}
+
+	return nil
+}
+
+// verifyRelayRequestSignature verifies the signature on the relay request.
+// TODO_TECHDEBT: Factor out the relay request signature verification into a shared
+// function that can be used by both the proof and relayer packages.
+func (k msgServer) verifyRelayRequestSignature(
```
bryanchriswhite, line 330: As discussed, after #411, we should be able to move this to a method on the relay request type for use both on- and off-chain.

x/proof/keeper/msg_server_submit_proof.go
```diff
@@ -127,12 +209,354 @@ func (k msgServer) queryAndValidateClaimForProof(ctx context.Context, proof *typ
...
+		// Insert the point into the slice of points
+		points[i] = point
+	}
+	return points, nil
+}
+
+// verifyRelayResponseSignature verifies the signature on the relay response.
+// TODO_TECHDEBT: Factor out the relay response signature verification into a shared
+// function that can be used by both the proof and the SDK packages.
+func (k msgServer) verifyRelayResponseSignature(
```
bryanchriswhite, line 459: Same as above for the relay response type.

bryanchriswhite (commented):
Nice! :sunglasses: 

I had a couple questions, but otherwise this is looks great. :ok_hand:

x/proof/keeper/msg_server_submit_proof.go
```diff
@@ -85,22 +161,28 @@ func (k msgServer) SubmitProof(ctx context.Context, msg *types.MsgSubmitProof) (
 
 // queryAndValidateClaimForProof ensures that  a claim corresponding to the given proof's
 // session exists & has a matching supplier address and session header.
-func (k msgServer) queryAndValidateClaimForProof(ctx context.Context, proof *types.Proof) error {
-	sessionId := proof.GetSessionHeader().GetSessionId()
+func (k msgServer) queryAndValidateClaimForProof(
+	ctx context.Context,
+	msg *types.MsgSubmitProof,
```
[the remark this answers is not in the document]
bryanchriswhite: Maybe #PUC?

x/service/types/relay.go
bryanchriswhite: #PUC

pkg/crypto/pubkey_client/client.go
```diff
@@ -0,0 +1,58 @@
...
+	"github.com/pokt-network/poktroll/pkg/crypto"
+)
+
+var _ crypto.PubKeyClient = (*pubKeyClient)(nil)
+
+// pubKeyClient is an implementation of the PubKeyClient that uses an account
+// querier to get the public key of an address.
+type pubKeyClient struct {
+	// accountQuerier is the querier for the account module, it is used to get
+	// the account of an address.
```
bryanchriswhite, line 19:
```suggestion
	// the public key corresponding to an address.
```

pkg/crypto/rings/cache.go
```diff
@@ -23,23 +20,16 @@ type ringCache struct {
...
 	delegationClient client.DelegationClient
 
-	// applicationQuerier is the querier for the application module, and is
-	// used to get the addresses of the gateways an application is delegated to.
-	applicationQuerier client.ApplicationQueryClient
-
-	// accountQuerier is the querier for the account module, and is used to get
-	// the public keys of the application and its delegated gateways.
-	accountQuerier client.AccountQueryClient
+	ringClient crypto.RingClient
```
bryanchriswhite, line 32:
```suggestion
	// ringClient is used to retrieve the rings that are cached.
	ringClient crypto.RingClient
```

pkg/crypto/rings/client.go
```diff
@@ -0,0 +1,207 @@
...
+// converting them to their secp256k1 curve points.
+func (rc *ringClient) GetRingForAddress(
+	ctx context.Context,
+	appAddress string,
+) (*ring.Ring, error) {
+	pubKeys, err := rc.getDelegatedPubKeysForAddress(ctx, appAddress)
+	if err != nil {
+		return nil, err
+	}
+	// Cache the ring's points for future use
```
bryanchriswhite, line 69:
```suggestion

```

pkg/crypto/rings/client.go
```diff
@@ -0,0 +1,207 @@
...
+			"session_id":          relayRequest.Meta.SessionHeader.SessionId,
+			"application_address": relayRequest.Meta.SessionHeader.ApplicationAddress,
+			"service_id":          relayRequest.Meta.SessionHeader.Service.Id,
+		}).
+		Msg("verifying relay request signature")
+
+	// Extract the relay request's ring signature
+	if relayRequest.Meta == nil {
+		return ErrRingClientEmptyRelayRequestSignature.Wrapf(
+			"request payload: %s", relayRequest.Payload,
```
bryanchriswhite, line 101: `relayRequest.Payload` is a byte slice here. If you're sure about logging it, I would suggest formatting it as hex:

```suggestion
			"request payload: %x", relayRequest.Payload,
```

pkg/crypto/rings/client.go
```diff
@@ -0,0 +1,207 @@
...
+		)
+	}
+
+	signature := relayRequest.Meta.Signature
+	if signature == nil {
+		return ErrRingClientInvalidRelayRequest.Wrapf(
+			"missing signature from relay request: %v", relayRequest,
```
bryanchriswhite, line 108: Similar feedback as the above. I think including the `relayRequest` in the error message could easily make it difficult to read. Is there something more specific that we can include that would still be useful?

bryanchriswhite (commented):
I have to run away for a bit. Will resume later today.

image

x/proof/keeper/msg_server_submit_proof.go
```diff
@@ -127,12 +224,127 @@ func (k msgServer) queryAndValidateClaimForProof(ctx context.Context, proof *typ
...
+		return types.ErrProofInvalidRelay.Wrapf(
+			"sessionHeaders application addresses mismatch expect: %s, got: %s",
+			expectedSessionHeader.GetApplicationAddress(),
+			sessionHeader.GetApplicationAddress(),
+		)
+	}
+
+	if sessionHeader.GetService().GetId() != expectedSessionHeader.GetService().GetId() {
+		return types.ErrProofInvalidRelay.Wrapf(
+			"sessionHeaders service IDs mismatch expect: %s, got: %s",
```
bryanchriswhite, line 252:
```suggestion
			"sessionHeaders service IDs mismatch expect: %q, got: %q",
```

x/proof/keeper/msg_server_submit_proof.go
```diff
@@ -127,12 +224,127 @@ func (k msgServer) queryAndValidateClaimForProof(ctx context.Context, proof *typ
...
+		return types.ErrProofInvalidRelay.Wrapf(
+			"sessionHeaders session end heights mismatch expect: %d, got: %d",
+			expectedSessionHeader.GetSessionEndBlockHeight(),
+			sessionHeader.GetSessionEndBlockHeight(),
+		)
+	}
+
+	if sessionHeader.GetSessionId() != expectedSessionHeader.GetSessionId() {
+		return types.ErrProofInvalidRelay.Wrapf(
+			"sessionHeaders session IDs mismatch expect: %s, got: %s",
```
bryanchriswhite, line 276:
```suggestion
			"sessionHeaders session IDs mismatch expect: %q, got: %q",
```

x/proof/keeper/msg_server_submit_proof.go
```diff
@@ -127,12 +224,127 @@ func (k msgServer) queryAndValidateClaimForProof(ctx context.Context, proof *typ
...
+			"error counting difficulty bits: %s",
+			err,
+		)
+	}
+
+	if difficultyBits < relayDifficultyBits {
+		return types.ErrProofInvalidRelay.Wrapf(
+			"relay difficulty %d is less than the required difficulty %d",
+			difficultyBits,
+			relayDifficultyBits,
```
bryanchriswhite, line 325:
```suggestion
	if difficultyBits < relayMinDifficultyBits {
		return types.ErrProofInvalidRelay.Wrapf(
			"relay difficulty %d is less than the required difficulty %d",
			difficultyBits,
			relayMinDifficultyBits,
```

x/proof/keeper/option.go
bryanchriswhite, line 1: I think we can remove this file since you updated the `NewKeeper()` signature.

x/proof/types/account_query_client.go
```diff
@@ -0,0 +1,32 @@
...
+	"github.com/pokt-network/poktroll/pkg/client"
+)
+
+var _ client.AccountQueryClient = (*AccountKeeperQueryClient)(nil)
+
+type AccountKeeperQueryClient struct {
+	keeper AccountKeeper
+}
+
+func NewAccountKeeperQueryClient(accountKeeper AccountKeeper) client.AccountQueryClient {
```
bryanchriswhite, line 18:
```suggestion
// NewAccountKeeperQueryClient returns a new AccountQueryClient that is backed by an AccountKeeper instance.
// It is intended for use on-chain.
func NewAccountKeeperQueryClient(accountKeeper AccountKeeper) client.AccountQueryClient {
```

x/proof/types/application_query_client.go
```diff
@@ -0,0 +1,32 @@
...
+
+var (
+	_ client.ApplicationQueryClient = (*AppKeeperQueryClient)(nil)
+)
+
+type AppKeeperQueryClient struct {
+	keeper ApplicationKeeper
+}
+
+func NewAppKeeperQueryClient(appKeeper ApplicationKeeper) client.ApplicationQueryClient {
```
bryanchriswhite, line 18:
```suggestion
// NewAppKeeperQueryClient returns a new ApplicationQueryClient that is backed by an ApplicationKeeper instance.
// It is intended for use on-chain.
func NewAppKeeperQueryClient(appKeeper ApplicationKeeper) client.ApplicationQueryClient {
```

x/proof/types/application_query_client.go
```diff
@@ -0,0 +1,32 @@
...
+import (
+	"context"
+
+	"github.com/pokt-network/poktroll/pkg/client"
+	apptypes "github.com/pokt-network/poktroll/x/application/types"
+)
+
+var (
+	_ client.ApplicationQueryClient = (*AppKeeperQueryClient)(nil)
+)
```
bryanchriswhite, line 12:
```suggestion
var _ client.ApplicationQueryClient = (*AppKeeperQueryClient)(nil)
```

bryanchriswhite (changes requested):
Left a few more comments but otherwise looks good to go. :rocket:

pkg/crypto/rings/client.go
red-0ne: The error used here is wrong. I'll change this to:
```go
if relayRequest.GetMeta() == nil {
  return ErrRingClientInvalidRelayRequest.Wrap("missing meta from relay request")
}
```

As there's not much useful information to provide about a nil field.

pkg/crypto/rings/client.go
```diff
@@ -0,0 +1,209 @@
...
+// addresses in the ring.
+type ringClient struct {
+	// logger is the logger for the ring cache.
+	logger polylog.Logger
+
+	// applicationQuerier is the querier for the application module, and is
+	// used to get the addresses of the gateways an application is delegated to.
+	applicationQuerier client.ApplicationQueryClient
+
+	// pubKeyClient
```
Olshansk, line 34: What's this for? #PUC

testutil/keeper/proof.go
```diff
@@ -80,6 +80,8 @@ func ProofKeeper(
 		log.NewNopLogger(),
 		authority.String(),
 		mockSessionKeeper,
+		nil,
```
Olshansk, line 83: Thought on adding mocks for this?

x/proof/keeper/keeper.go
```diff
@@ -21,7 +28,11 @@ type (
 		// should be the x/gov module account.
 		authority string
 
-		sessionKeeper types.SessionKeeper
+		sessionKeeper     types.SessionKeeper
+		applicationKeeper types.ApplicationKeeper
+		accountKeeper     types.AccountKeeper
+		ringClient        crypto.RingClient
```
Olshansk, line 34: Let's separate the keepers and the clients w/ a new line

x/proof/keeper/keeper.go
```diff
@@ -32,18 +43,45 @@ func NewKeeper(
...
 	if _, err := sdk.AccAddressFromBech32(authority); err != nil {
 		panic(fmt.Sprintf("invalid authority address: %s", authority))
 	}
 
+	applicationQuerier := types.NewAppKeeperQueryClient(applicationKeeper)
+	accountQuerier := types.NewAccountKeeperQueryClient(accountKeeper)
+	// TODO_TECHDEBT: Use cosmos-sdk based polylog implementation once available. Also remove polyzero import.
+	polylogger := polylog.Ctx(context.TODO())
+
+	ringClientDeps := depinject.Supply(
```
Olshansk, line 58: Can you add a few comments on what's happening here as we set up the clients and why?

x/proof/keeper/msg_server_submit_proof.go
```diff
@@ -1,17 +1,35 @@
...
 
+	"github.com/pokt-network/poktroll/pkg/relayer/protocol"
 	"github.com/pokt-network/poktroll/x/proof/types"
+	servicetypes "github.com/pokt-network/poktroll/x/service/types"
+	sessionkeeper "github.com/pokt-network/poktroll/x/session/keeper"
+	sessiontypes "github.com/pokt-network/poktroll/x/session/types"
+)
+
+const (
+	// TODO_TECHDEBT: relayMinDifficultyBits should be a governance-based parameter
```
Olshansk, line 20:
```suggestion
	// TODO_BLOCKER: relayMinDifficultyBits should be a governance-based parameter
```

x/proof/keeper/msg_server_submit_proof.go
```diff
@@ -1,17 +1,35 @@
...
+	relayMinDifficultyBits = 0
+
+	// sumSize is the size of the sum of the relay request and response
```
Olshansk, line 23: `is the size of the sum of the relay request and response`

What do you mean by this?

I don't think it has anything to do with the relay request and response and is actually a function of the SMT leaf. I may be misunderstanding.

Olshansk (changes requested):
Posting a partial review
Screenshot 2024-03-08 at 7 57 27 PM

x/proof/types/application_query_client.go
```diff
@@ -0,0 +1,39 @@
...
+func NewAppKeeperQueryClient(appKeeper ApplicationKeeper) client.ApplicationQueryClient {
+	return &AppKeeperQueryClient{keeper: appKeeper}
+}
+
+// GetApplication returns the application corresponding to the given address.
+func (appQueryClient *AppKeeperQueryClient) GetApplication(
+	ctx context.Context,
+	appAddr string,
+) (apptypes.Application, error) {
+	app, _ := appQueryClient.keeper.GetApplication(ctx, appAddr)
```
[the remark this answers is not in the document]
bryanchriswhite: It's a `bool`, not an `error` (see: [`apptypes.Keeper#GetApplication()`](https://github.com/pokt-network/poktroll/blob/main/x/application/keeper/application.go#L25)). Do you think we should return an error if the app wasn't found?

testutil/keeper/proof.go
bryanchriswhite: This will be populated in #412.

x/service/types/relay.go
```diff
@@ -53,5 +89,31 @@ func (res *RelayResponse) ValidateBasic() error {
...
+	}
+
+	if len(res.GetMeta().GetSupplierSignature()) == 0 {
+		return ErrServiceInvalidRelayResponse.Wrap("missing supplier signature")
+	}
+
+	return nil
+}
+
+func (res *RelayResponse) VerifySignature(supplierPubKey cryptotypes.PubKey) error {
```
h5law, line 107:
```suggestion
// VerifySupplierSignature ensure the signature provided by the
// supplier is valid according to their relay response.
func (res *RelayResponse) VerifySupplierSignature(supplierPubKey cryptotypes.PubKey) error {
```

pkg/crypto/rings/client.go
```diff
@@ -0,0 +1,209 @@
...
+	pubkeyclient "github.com/pokt-network/poktroll/pkg/crypto/pubkey_client"
+	"github.com/pokt-network/poktroll/pkg/polylog"
+	"github.com/pokt-network/poktroll/x/service/types"
+)
+
+var _ crypto.RingClient = (*ringClient)(nil)
+
+// ringClient is an implementation of the RingClient interface that uses the
+// client.ApplicationQueryClient to get applications delegation information and
+// needed to construct the ring for signing relay requests.
```
Olshansk, line 23:
```suggestion
// client.ApplicationQueryClient to get application's delegation information
// needed to construct the ring for signing relay requests.
```

pkg/crypto/rings/client.go
```diff
@@ -0,0 +1,209 @@
...
+// It also uses the pubkeyclient.PubKeyClient to get the public keys for the
+// addresses in the ring.
```
Olshansk, line 25: Aren't these two sentences redundant? We need the public keys to construct the ring so I feel like we can just delete it.

pkg/crypto/rings/client.go
```diff
@@ -0,0 +1,209 @@
...
+	if valid := ringSig.Verify(requestSignableBz); !valid {
+		return ErrRingClientInvalidRelayRequestSignature.Wrapf("invalid ring signature")
+	}
+	return nil
+}
+
+// getDelegatedPubKeysForAddress returns the ring used to sign a message for
+// the given application address, by querying the application module for it's
+// delegated pubkeys and converting them to points on the secp256k1 curve in
+// order to create the ring.
```
Olshansk, line 159: Is this comment still accurate? I feel like it just retreves the list of public keys the app delegated to.

pkg/crypto/rings/ring.go
```diff
@@ -0,0 +1,40 @@
...
+
+import (
+	ring_secp256k1 "github.com/athanorlabs/go-dleq/secp256k1"
+	ringtypes "github.com/athanorlabs/go-dleq/types"
+	"github.com/cosmos/cosmos-sdk/crypto/keys/secp256k1"
+	cryptotypes "github.com/cosmos/cosmos-sdk/crypto/types"
+	"github.com/noot/ring-go"
+)
+
+// newRingFromPoints creates a new ring from points on the secp256k1 curve
```
Olshansk, line 11:
```suggestion
// newRingFromPoints creates a new ring from points (i.e. public keys) on the secp256k1 curve
```

pkg/relayer/proxy/relay_verifier.go
```diff
@@ -18,81 +14,21 @@ func (rp *relayerProxy) VerifyRelayRequest(
...
+	// been verified.
+	appAddress := relayRequest.GetMeta().GetSessionHeader().GetApplicationAddress()
 
 	// Query for the current session to check if relayRequest sessionId matches the current session.
 	rp.logger.Debug().
 		Fields(map[string]any{
-			"session_id":          relayRequest.Meta.SessionHeader.SessionId,
-			"application_address": relayRequest.Meta.SessionHeader.ApplicationAddress,
-			"service_id":          relayRequest.Meta.SessionHeader.Service.Id,
+			"session_id":          relayRequest.GetMeta().GetSessionHeader().GetSessionId(),
```
Olshansk, line 29: Let's assign sessionHeader to a local variable and simplify this and the other usage below.

pkg/sdk/sdk.go
```diff
@@ -113,3 +113,23 @@ func NewPOKTRollSDK(ctx context.Context, config *POKTRollSDKConfig) (POKTRollSDK
...
+	if pubKey, ok := sdk.supplierPubKeyCache[address]; ok {
+		return pubKey, nil
+	}
+
+	pubKey, err := sdk.pubKeyClient.GetPubKeyFromAddress(ctx, address)
+	if err != nil {
+		return nil, err
+	}
+
+	sdk.supplierPubKeyCache[address] = pubKey
```
Olshansk, line 133: The function name and comment is pretty general purpose (i.e. no mention of supplier) but the cache explicitly calls out `supplier`. 

Should something be renamed based on this discrepancy?

pkg/sdk/send_relay.go
```diff
@@ -101,13 +101,30 @@ func (sdk *poktrollSDK) SendRelay(
...
+	}
+
+	// Get the supplier's public key.
+	supplierPubKey, err := sdk.getPubKeyFromAddress(ctx, supplierEndpoint.SupplierAddress)
+	if err != nil {
+		return nil, ErrSDKHandleRelay.Wrapf("error getting supplier public key: %s", err)
+	}
+
+	sdk.logger.Debug().
+		Str("supplier", supplierEndpoint.SupplierAddress).
```
Olshansk, line 115: Let's assign sessionHeader to a local var

x/proof/keeper/msg_server_submit_proof.go
```diff
@@ -55,15 +77,84 @@ func (k msgServer) SubmitProof(ctx context.Context, msg *types.MsgSubmitProof) (
...
+
+	// Verify that the relay request session header matches the proof session header.
+	if err := compareSessionHeaders(msg.GetSessionHeader(), relay.GetReq().Meta.GetSessionHeader()); err != nil {
+		return nil, status.Error(codes.FailedPrecondition, err.Error())
+	}
+
+	// Verify that the relay response session header matches the proof session header.
+	if err := compareSessionHeaders(msg.GetSessionHeader(), relay.GetRes().Meta.GetSessionHeader()); err != nil {
+		return nil, status.Error(codes.FailedPrecondition, err.Error())
+	}
```
Olshansk, line 116: `compareSessionHeaders` duplicate? Are we missing another check that we needed?

x/proof/keeper/msg_server_submit_proof.go
```diff
@@ -55,15 +77,84 @@ func (k msgServer) SubmitProof(ctx context.Context, msg *types.MsgSubmitProof) (
...
+
+	supplierPubKey, err := k.pubKeyClient.GetPubKeyFromAddress(ctx, msg.GetSupplierAddress())
```
Olshansk, line 118: @bryanchriswhite Wanted to check who is responsible for documenting this series of checks in the docs? Could be a good opportunity for a mermaid diagram "waterfall".

x/proof/keeper/keeper.go
```diff
@@ -32,18 +43,45 @@ func NewKeeper(
...
+		polylogger,
+		applicationQuerier,
+		accountQuerier,
+	)
+
+	ringClient, err := rings.NewRingClient(ringClientDeps)
```
Olshansk, line 64: Can we rename these to `ringKeeperClient` and the one below to `pubKeyKeeperClient`. When I was reading the business logic, I got worried that we were making networking requests from the `x/` repo which I believe happened due to the naming.

x/proof/keeper/keeper.go
Olshansk: @red-0ne Given both mine & @h5law's confusion, this definitely needs a #PUC on what/why is happening.

x/proof/keeper/msg_server_submit_proof.go
Olshansk, line 118: Let's move the "simple checks" closer to the top so the E2E validation "reads easier"

x/proof/keeper/msg_server_submit_proof.go
```diff
@@ -127,12 +224,141 @@ func (k msgServer) queryAndValidateClaimForProof(ctx context.Context, proof *typ
...
 	}
 
+	return &foundClaim, nil
+}
+
+// compareSessionHeaders compares a session header against an expected session header.
+func compareSessionHeaders(
+	expectedSessionHeader *sessiontypes.SessionHeader,
+	sessionHeader *sessiontypes.SessionHeader,
+) error {
```
Olshansk, line 241:
```suggestion
func compareSessionHeaders(expectedSessionHeader, sessionHeader *sessiontypes.SessionHeader) error {
```

x/proof/keeper/msg_server_submit_proof.go
```diff
@@ -127,12 +224,141 @@ func (k msgServer) queryAndValidateClaimForProof(ctx context.Context, proof *typ
...
+
+	return nil
+}
+
+// verifyClosestProof verifies the closest merkle proof against the expected root hash.
+func verifyClosestProof(
+	proof *smt.SparseMerkleClosestProof,
+	expectedRootHash []byte,
+) error {
+	spec := smt.NoPrehashSpec(sha256.New(), true)
```
Olshansk, line 290: Can you move this into a local variable that's initiated on `init` and add a comment explaining it?

Seems like really important business logic that's embedded deep in the code.

Olshansk (changes requested):
Almost done.
Screenshot 2024-03-11 at 5 30 01 PM

pkg/relayer/proxy/errors.go
```diff
@@ -5,14 +5,12 @@ import sdkerrors "cosmossdk.io/errors"
...
-	ErrRelayerProxyUndefinedSigningKeyName           = sdkerrors.Register(codespace, 5, "undefined relayer proxy signing key name")
-	ErrRelayerProxyUndefinedProxiedServicesEndpoints = sdkerrors.Register(codespace, 6, "undefined proxied services endpoints for relayer proxy")
-	ErrRelayerProxyInvalidRelayRequest               = sdkerrors.Register(codespace, 7, "invalid relay request")
-	ErrRelayerProxyInvalidRelayResponse              = sdkerrors.Register(codespace, 8, "invalid relay response")
-	ErrRelayerProxyEmptyRelayRequestSignature        = sdkerrors.Register(codespace, 9, "empty relay response signature")
-	ErrRelayerProxyServiceEndpointNotHandled         = sdkerrors.Register(codespace, 10, "service endpoint not handled by relayer proxy")
-	ErrRelayerProxyUnsupportedTransportType          = sdkerrors.Register(codespace, 11, "unsupported proxy transport type")
+	ErrRelayerProxyInvalidSession                    = sdkerrors.Register(codespace, 2, "invalid session in relayer request")
+	ErrRelayerProxyInvalidSupplier                   = sdkerrors.Register(codespace, 3, "invalid relayer proxy supplier")
```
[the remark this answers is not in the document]
Olshansk: +1

I think leaving the var names as is is "good enough" but we can definitely add some detail in the strings so it's clear and descrptive without the context of where the error is used.

x/proof/keeper/msg_server_submit_proof.go
```diff
@@ -55,15 +77,84 @@ func (k msgServer) SubmitProof(ctx context.Context, msg *types.MsgSubmitProof) (
...
+	relayBz := closestValueHash[:len(closestValueHash)-sumSize]
+	relay := &servicetypes.Relay{}
+	if err := k.cdc.Unmarshal(relayBz, relay); err != nil {
+		return nil, types.ErrProofInvalidRelay.Wrapf(
+			"failed to unmarshal relay: %s",
+			err,
+		)
+	}
+
+	if err := relay.GetReq().ValidateBasic(); err != nil {
```
Olshansk, line 100: @red-0ne WHat's the thinking/plan around testing each of these?

I realize we'll have an E2E test that covers one happy path but there might be an opportunity for a few simple unit tests. 

1. Open to push back. Lmk
2. If there's pushback. I'd personally copy-paste this file into ChatGPT and have it do the first ~30% of the work for me (if you prompt it correctly)
3. TODO_TEST is recommended if we decide to do this.

x/proof/keeper/msg_server_submit_proof.go
```diff
@@ -127,12 +224,141 @@ func (k msgServer) queryAndValidateClaimForProof(ctx context.Context, proof *typ
...
+	// see: relayerSessionsManager.waitForEarliest{CreateClaim,SubmitProof}Height().
+	// The RelayMiner hast to wait this long to ensure that late relays are accepted
+	// and included in the SessionNumber=N tree.
+	// (i.e. relays initiated by Applications/Gateways in SessionNumber=N but
+	// arriving to the RelayMiner in SessionNumber=N + 1)
+	// Otherwise, the RelayMiner would not be able to include the late relays in
+	// the SessionNumber N claim and the proof.
+	// Since smt.ProveClosest is in terms of submitProofWindowStartHeight, this
+	// block's hash needs to be used for validation too.
+	// TODO_TECHDEBT(#409): Reference the session rollover documentation here.
```
Olshansk, line 351: Nice

Olshansk, line 351:
```suggestion
	// The RelayMiner has to wait until the createClaimWindowStartHeight and the
	// submitProofWindowStartHeight are open to create the claim & submit the proof respectively.
	//
	// These windows are calculated as (SessionEndBlockHeight + GracePeriodBlockCount).
	// For reference, see relayerSessionsManager.waitForEarliest{CreateClaim,SubmitProof}Height().
	//
	// The RelayMiner has to wait this long to ensure that late relays (i.e. submitted during
	// SessionNumber=(N+1) but created during SessionNumber=N) are still included as part
	// of SessionNumber=N. 
	//
	// Since smt.ProveClosest is defined in terms of submitProofWindowStartHeight, this
	// block's hash needs to be used for validation too.
	//
	// TODO_TECHDEBT(#409): Reference the session rollover documentation here.
```

x/proof/keeper/msg_server_submit_proof.go
```diff
@@ -127,12 +224,141 @@ func (k msgServer) queryAndValidateClaimForProof(ctx context.Context, proof *typ
...
+	blockHeight := sessionHeader.GetSessionEndBlockHeight() + sessionkeeper.GetSessionGracePeriodBlockCount()
+	blockHash := k.sessionKeeper.GetBlockHash(ctx, blockHeight)
+
+	if !bytes.Equal(proof.Path, blockHash) {
```
Olshansk, line 355: ⚠️IMPORTANT BLOCKER ⚠️

I don't understand this. Why are we comparing the proof path to the block hash?

This is the sort of thing we really need a unit test for or AT LEAST a TODO_TEST.

x/proof/keeper/msg_server_submit_proof.go
```diff
@@ -127,12 +224,141 @@ func (k msgServer) queryAndValidateClaimForProof(ctx context.Context, proof *typ
...
+	// The RelayMiner hast to wait this long to ensure that late relays are accepted
+	// and included in the SessionNumber=N tree.
+	// (i.e. relays initiated by Applications/Gateways in SessionNumber=N but
+	// arriving to the RelayMiner in SessionNumber=N + 1)
+	// Otherwise, the RelayMiner would not be able to include the late relays in
+	// the SessionNumber N claim and the proof.
+	// Since smt.ProveClosest is in terms of submitProofWindowStartHeight, this
+	// block's hash needs to be used for validation too.
+	// TODO_TECHDEBT(#409): Reference the session rollover documentation here.
+	blockHeight := sessionHeader.GetSessionEndBlockHeight() + sessionkeeper.GetSessionGracePeriodBlockCount()
```
Olshansk, line 352: Can you think of a more semantic name than blockHeight? For example, `sessionEndWithGracePeriodblockHeight`

Olshansk (changes requested):
Finally DONE. 

Planning to do a re-review after all comments are tended to.

x/proof/keeper/keeper.go
h5law: I would be more in favour of removing the pubkeyKeeperClient - these are already accessible through the accountkeeper which can just be a dependency of the proof keeper itself.

x/proof/keeper/msg_server_submit_proof.go
h5law: @Olshansk Here `proof.Path` is the path used in the `ClosestProof` method to obtain the pseudo-random leaf. So if these 2 are different someone is trying to do something dodgy or something has gone wrong.

bryanchriswhite:
I'm noticing an E2E failure locally:

image

e2e/tests/session.feature
```diff
@@ -9,7 +9,7 @@ Feature: Session Namespace
     # The timeout for when a claim can be submitted on-chain depends on `createClaimWindowStartHeight`, which
     # is a function of `SessionGracePeriod`. The higher this value, the higher this timeout needs to be. Since
     # this test is not dependant on the grace period, setting it to 0 and having a lower grace period will simplify it.
-    And the user should wait for "7" seconds
+    And the user should wait for "10" seconds
```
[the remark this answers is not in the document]
red-0ne: This should be implemented when this branch merges main.

x/service/types/relay.go
```diff
@@ -38,7 +66,15 @@ func (res RelayResponse) getSignableBytes() ([]byte, error) {
 // Hashing the marshaled response message guarantees that the signable bytes are
 // always of a constant and expected length.
 func (res *RelayResponse) GetSignableBytesHash() ([32]byte, error) {
+	// Save the signature and restore it after getting the signable bytes
```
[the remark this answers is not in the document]
red-0ne: Thie issue is definitely coming from the fact that `RelayRequest` is not a deep copy and `req/res.Meta` is a pointer to the original `Meta` field.

x/proof/module/module.go
```diff
@@ -201,6 +202,8 @@ func ProvideModule(in ModuleInputs) ModuleOutputs {
 		in.Logger,
 		authority.String(),
 		in.SessionKeeper,
+		in.ApplicationKeeper,
```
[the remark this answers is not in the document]
red-0ne: We can have it (BankKeeper) back if/when needed.

x/proof/keeper/msg_server_submit_proof.go
red-0ne: We currently have no SMT function that extracts the relay from the Proof's `ClosestValueHash` which contains both the sum and the relay bytes. We need this value to be able to truncate the `closestValueHash`. 

It is currently mimicking what's happening here https://github.com/pokt-network/smt/blob/main/proofs.go#L319.

While the `sumSize` is defined here https://github.com/pokt-network/smt/blob/main/types.go#L10.

We can get rid of this when the SMT package offer a method to extract the relay as stated in the `TODO_TECHDEBT` below.

pkg/sdk/sdk.go
```diff
@@ -65,17 +61,21 @@ type poktrollSDK struct {
 	// It is used to get the current block height to query for the current session.
 	blockClient client.BlockClient
 
-	// accountCache is a cache of the supplier accounts that has been queried
+	// pubKeyClient the client used to get the public key given an account address.
+	// TODO_TECHDEBT: Add a size limit to the cache.
```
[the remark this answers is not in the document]
red-0ne: IMO, an `LRU` cache would be the best option. As clearing every `N` session could create fetching spikes and have a bad time distribution of the fetching (expensive) operation.

pkg/sdk/sdk.go
```diff
@@ -65,17 +61,21 @@ type poktrollSDK struct {
...
+	supplierPubKeyCache map[string]cryptotypes.PubKey
 }
 
 // NewPOKTRollSDK creates a new POKTRollSDK instance with the given configuration.
 func NewPOKTRollSDK(ctx context.Context, config *POKTRollSDKConfig) (POKTRollSDK, error) {
 	sdk := &poktrollSDK{
 		config:                  config,
 		serviceSessionSuppliers: make(map[string]map[string]*SessionSuppliers),
-		supplierAccountCache:    make(map[string]cryptotypes.PubKey),
+		supplierPubKeyCache:     make(map[string]cryptotypes.PubKey),
```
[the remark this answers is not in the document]
red-0ne: We should consider using https://github.com/hashicorp/golang-lru for caching pubkeys.

x/proof/keeper/keeper.go
```diff
@@ -32,18 +43,45 @@ func NewKeeper(
...
+	accountKeeper types.AccountKeeper,
 ) Keeper {
 	if _, err := sdk.AccAddressFromBech32(authority); err != nil {
 		panic(fmt.Sprintf("invalid authority address: %s", authority))
 	}
 
+	applicationQuerier := types.NewAppKeeperQueryClient(applicationKeeper)
+	accountQuerier := types.NewAccountKeeperQueryClient(accountKeeper)
+	// TODO_TECHDEBT: Use cosmos-sdk based polylog implementation once available. Also remove polyzero import.
+	polylogger := polylog.Ctx(context.TODO())
```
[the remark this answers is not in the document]
red-0ne: cc @bryanchriswhite

x/proof/keeper/msg_server_submit_proof.go
```diff
@@ -1,17 +1,35 @@
...
+)
+
+const (
+	// TODO_TECHDEBT: relayMinDifficultyBits should be a governance-based parameter
+	relayMinDifficultyBits = 0
+
+	// sumSize is the size of the sum of the relay request and response
+	// in bytes. This is used to extract the relay request and response
+	// from the closest merkle proof.
+	// TODO_TECHDEBT: Have a method on the smst to extract the value hash or export
```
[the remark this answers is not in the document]
red-0ne: I'll use it as soon as we have it.

x/proof/keeper/msg_server_submit_proof.go
```diff
@@ -85,22 +176,28 @@ func (k msgServer) SubmitProof(ctx context.Context, msg *types.MsgSubmitProof) (
 
 // queryAndValidateClaimForProof ensures that  a claim corresponding to the given proof's
 // session exists & has a matching supplier address and session header.
```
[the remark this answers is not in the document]
red-0ne: It rather returns the found claim if the validation is successful.

x/proof/keeper/msg_server_submit_proof.go
```diff
@@ -127,12 +224,141 @@ func (k msgServer) queryAndValidateClaimForProof(ctx context.Context, proof *typ
...
+) error {
+	if sessionHeader.GetApplicationAddress() != expectedSessionHeader.GetApplicationAddress() {
+		return types.ErrProofInvalidRelay.Wrapf(
+			"sessionHeaders application addresses mismatch expect: %q, got: %q",
+			expectedSessionHeader.GetApplicationAddress(),
+			sessionHeader.GetApplicationAddress(),
+		)
+	}
+
+	if sessionHeader.GetService().GetId() != expectedSessionHeader.GetService().GetId() {
```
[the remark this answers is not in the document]
red-0ne: No strong opinion here, but since one side of the comparison is an off-chain header, we could ensure that it was properly constructed by comparing the service name too.

Will add it

x/proof/keeper/msg_server_submit_proof.go
```diff
@@ -127,12 +224,141 @@ func (k msgServer) queryAndValidateClaimForProof(ctx context.Context, proof *typ
...
+// difficulty.
+// TODO_TECHDEBT: Factor out the relay mining difficulty validation into a shared function
+// that can be used by both the proof and the miner packages.
+func validateMiningDifficulty(relayBz []byte) error {
+	hasher := sha256.New()
+	hasher.Write(relayBz)
+	realyHash := hasher.Sum(nil)
+
+	difficultyBits, err := protocol.CountDifficultyBits(realyHash)
+	if err != nil {
```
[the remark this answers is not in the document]
red-0ne: How would you handle an error returned by `CountDifficultyBits`?

x/proof/keeper/msg_server_submit_proof.go
```diff
@@ -127,12 +224,141 @@ func (k msgServer) queryAndValidateClaimForProof(ctx context.Context, proof *typ
...
+// validateClosestPath ensures that the proof's path matches the expected path.
+func (k msgServer) validateClosestPath(
+	ctx context.Context,
+	proof *smt.SparseMerkleClosestProof,
+	sessionHeader *sessiontypes.SessionHeader,
+) error {
+	// The RelayMiner has to wait until the createClaimWindowStartHeight and the
+	// submitProofWindowStartHeight are open to respectively create the claim and
+	// submit the proof.
+	// These windows are calculated as SessionEndBlockHeight + GracePeriodBlockCount.
```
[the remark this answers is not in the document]
red-0ne: It's currently a const in the [x/session/keeper/session_hydrator.go](https://github.com/pokt-network/poktroll/blob/main/x/session/keeper/session_hydrator.go#L27)

x/proof/types/account_query_client.go
```diff
@@ -0,0 +1,40 @@
+package types
+
+import (
+	"context"
+
+	"github.com/cosmos/cosmos-sdk/types"
```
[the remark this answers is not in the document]
red-0ne: `AccountI` from `"github.com/cosmos/cosmos-sdk/x/auth/types"` seems to be deprecated.

x/service/types/relay.go
```diff
@@ -15,7 +19,15 @@ func (req RelayRequest) getSignableBytes() ([]byte, error) {
 // Hashing the marshaled request message guarantees that the signable bytes are
 // always of a constant and expected length.
 func (req *RelayRequest) GetSignableBytesHash() ([32]byte, error) {
```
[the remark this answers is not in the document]
red-0ne: As discussed, this does not have any effect because of `Meta` being a pointer.

pkg/crypto/rings/client.go
```diff
@@ -0,0 +1,209 @@
...
+
+	ringSig := new(ring.RingSig)
+
+	if err := ringSig.Deserialize(ring_secp256k1.NewCurve(), signature); err != nil {
+		return ErrRingClientInvalidRelayRequestSignature.Wrapf(
+			"error deserializing ring signature: %s", err,
+		)
+	}
+
+	if sessionHeader.GetApplicationAddress() == "" {
```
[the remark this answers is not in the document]
red-0ne: This is a refactor artifact as we are now calling `sessionHeader.ValidateBasic` that is checking the `ApplicationAddress`.

Will remove this check.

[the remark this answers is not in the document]
red-0ne: We are already calling `sessionHeader.ValidateBasic` :+1:  above in line ~95

pkg/crypto/rings/client.go
red-0ne: Good catch!

x/proof/types/account_query_client.go
```diff
@@ -0,0 +1,40 @@
...
+// It should be injected into the PubKeyClient when initialized from within the a keeper.
+func NewAccountKeeperQueryClient(accountKeeper AccountKeeper) client.AccountQueryClient {
+	return &AccountKeeperQueryClient{keeper: accountKeeper}
+}
+
+// GetAccount returns the account associated with the provided address.
+func (accountQueryClient *AccountKeeperQueryClient) GetAccount(
+	ctx context.Context,
+	addr string,
+) (accounttypes.AccountI, error) {
```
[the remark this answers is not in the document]
bryanchriswhite: I think no reason. I noticed that it was deprecated but at the time and I wasn't sure (nor willing to discover) how deep the refactor was going to be. I looked into it the other day and it seems pretty minimal.

There should be no need to add a new type, only to update all the deprecated usages in our code.

x/proof/types/account_query_client.go
```diff
@@ -0,0 +1,40 @@
...
+	addrBz, err := types.AccAddressFromBech32(addr)
+	if err != nil {
+		return nil, err
+	}
+
+	return accountQueryClient.keeper.GetAccount(ctx, addrBz), nil
```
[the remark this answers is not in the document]
red-0ne: Weirdly `GetAccount` panics if the account is not found. Will catch it and return an error as we cannot allow it to panic at this stage.

image

x/proof/types/expected_keepers.go
```diff
@@ -7,21 +7,28 @@ import (
...
 type SessionKeeper interface {
 	GetSession(context.Context, *sessiontypes.QueryGetSessionRequest) (*sessiontypes.QueryGetSessionResponse, error)
+	GetBlockHash(ctx context.Context, height int64) []byte
 }
 
 // AccountKeeper defines the expected interface for the Account module.
 type AccountKeeper interface {
-	GetAccount(context.Context, sdk.AccAddress) sdk.AccountI // only used for simulation
-	// Methods imported from account should be defined here
+	GetAccount(context.Context, sdk.AccAddress) sdk.AccountI
```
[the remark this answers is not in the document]
red-0ne: Done,

It has to be implemented in the `AccountKeeperQueryClient` too.

x/proof/keeper/msg_server_submit_proof.go
red-0ne: We definitely need to have tests for the req/res `ValidateBasic` functions, and these function more love.

I'll opt for adding `TODO_TEST` above their declarations now and test them in another PR.

x/proof/keeper/msg_server_submit_proof.go
bryanchriswhite: > ⚠️IMPORTANT BLOCKER ⚠️
> 
> I don't understand this. Why are we comparing the proof path to the block hash?
> 
> This is the sort of thing we really need a unit test for or AT LEAST a TODO_TEST.

This case is covered in #417 by https://github.com/pokt-network/poktroll/pull/417/files#diff-9ebcaa7396c928d8fc0a011521652f5f7332d928e25291cf7d04805cf98ec4dcR725.

Olshansk (changes requested):

Olshansk:
@red-0ne Looks awesome!

Didn't have any major comments after the first round of reviews. As a next step:
1. Please see the changes I made (mostly comments): [`46c61dd` (#406)](https://github.com/pokt-network/poktroll/pull/406/commits/46c61dddcbb1f7cb19d2e5673713924085f500c0)
2. I left a couple TODO_IN_THIS_PR before we can merge
3. I think this is super close to getting in 🤝

x/proof/keeper/msg_server_submit_proof.go
bryanchriswhite: See: https://www.notion.so/buildwithgrove/RPC-Session-Life-cycle-4bcf9f97e46540a0962cbc41f29a485d

x/proof/types/expected_keepers.go
```diff
@@ -7,21 +7,22 @@ import (
...
 }
 
-// BankKeeper defines the expected interface for the Bank module.
-type BankKeeper interface {
-	SpendableCoins(context.Context, sdk.AccAddress) sdk.Coins
-	// Methods imported from bank should be defined here
```
bryanchriswhite, line 26: We still need the bank keeper to be available in this package for testing purposes.

bryanchriswhite: While it is no longer a dependency of the proof keeper, it's still a dpendency of gateway, application, supplier, and session keepers; all of which we need to construct in the integration tests added in #412 & #417.

bryanchriswhite: In fact, the expected keeper interface must also be updated for those PRs: https://github.com/pokt-network/poktroll/pull/412/files#diff-8395873d16465a1ae7a1601264eac2646e93594539f92d123270b5e9c98d7386

pkg/crypto/rings/client.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+
+	// Get and hash the signable bytes of the relay request.
+	requestSignableBz, err := relayRequest.GetSignableBytesHash()
+	if err != nil {
+		return ErrRingClientInvalidRelayRequest.Wrapf("error getting signable bytes: %v", err)
+	}
+
+	// Verify the relay request's signature.
+	if valid := ringSig.Verify(requestSignableBz); !valid {
```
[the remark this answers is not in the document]
red-0ne: Not sure if I understand. From where could we get the signature except from the `RelayRequest`?`rc.GetRingForAddress` returns the `Ring` but not `RingSig`.

x/service/types/relay.go
```diff
@@ -1,21 +1,29 @@
...
-
-	return req.Marshal()
-}
+	cryptotypes "github.com/cosmos/cosmos-sdk/crypto/types"
+)
 
 // GetSignableBytesHash returns the hash of the signable bytes of the relay request
 // Hashing the marshaled request message guarantees that the signable bytes are
 // always of a constant and expected length.
 func (req *RelayRequest) GetSignableBytesHash() ([32]byte, error) {
```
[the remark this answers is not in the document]
red-0ne: To get rid of the concurrency issue we would need to change `Meta` to a non pointer in the protobuf.

Let's do it!

x/proof/keeper/msg_server_submit_proof.go
h5law: It depends on the hashing function used. A cosmos block hash will be 32 bytes so it presumably uses some variation of `sha256`. As long as the trie is setup with a 256 bit path hasher then yes this is guarenteed to always be the same length (this is the case currently). However, there is no enforcement - a check could be done in the smt's `ClosestProof` logic that throws an error if the hash provided is not the same size as the path hasher of the trie I can add this to https://github.com/pokt-network/smt/pull/42 now

x/proof/keeper/msg_server_submit_proof.go
```diff
@@ -127,12 +259,162 @@ func (k msgServer) queryAndValidateClaimForProof(ctx context.Context, proof *typ
...
+	//
+	// Since smt.ProveClosest is defined in terms of submitProofWindowStartHeight,
+	// this block's hash needs to be used for validation too.
+	//
+	// TODO_TECHDEBT(#409): Reference the session rollover documentation here.
+	sessionEndBlockHeightWithGracePeriod := sessionHeader.GetSessionEndBlockHeight() +
+		sessionkeeper.GetSessionGracePeriodBlockCount()
+	blockHash := k.sessionKeeper.GetBlockHash(ctx, sessionEndBlockHeightWithGracePeriod)
+
+	// TODO_IN_THIS_PR: Finish off the conversation related to this check: https://github.com/pokt-network/poktroll/pull/406/files#r1520790083
```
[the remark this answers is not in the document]
Olshansk: @red-0ne FYI

x/proof/keeper/msg_server_submit_proof.go
Olshansk: > but instead ensure we are using the same sized path hashers (which AFAIK we are)

Exactly. How do we ensure this?

1. **Current**: Manual checking by someone (i.e. @h5law) across two github repositories with implicit institutional knowledge of how things work and relate.
2. Proposed (explicit, commented & clear):
```go
// BlockHashToClosestMerkleProofHashPath converts the block hash,
// a pseudo-random value that cannot be predicted, to a random path
// in the tree.
func BlockHashToClosestMerkleProofHashPath(blockHash string) string {
     smtPathSize = smt.GetPathSize() // this is a new function
     return blockHash[:smtPathSize]
}
```
3. I believe there are two `TODO_BLOCKER`s we need to add here:

```
TODO_BLOCKER: Update `blockHeight` to be the value when the `ProofWindow` opens once the variable is added
TODO_BLOCKER: The seed of the path should be `foo(blockHash, sessionId)` to prevent all proofs needing to use the same path.
```

Olshansk: @red-0ne Let's add TODO_BLOCKERs and pass-through functions where appropriate and get this PR in. It'll simplify and unblock a lot.

x/proof/keeper/msg_server_submit_proof.go
```diff
@@ -1,22 +1,56 @@
...
+const (
+	// relayMinDifficultyBits is the minimum difficulty that a relay must have
+	// to be reward (i.e. volume) applicable.
+	// TODO_BLOCKER: relayMinDifficultyBits should be a governance-based parameter
+	relayMinDifficultyBits = 0
+
+	// sumSize is the size of the sum of the relay request and response in bytes.
+	// This is needed to extract the relay request and response from the closest
+	// merkle proof.
+	// TODO_TECHDEBT(@h5law): Add a method on the smst to extract the value hash
```
h5law, line 28: Addressed in https://github.com/pokt-network/smt/pull/43 you can update to use this hash of the smt repo in go mod for all the functions e3dbbbd9f97d4f9007c67f5c7e512a93c8f3335c

x/proof/keeper/msg_server_submit_proof.go
```diff
@@ -1,22 +1,56 @@
...
+)
+
+// SMT specification used for the proof verification.
+var spec *smt.TrieSpec
+
+func init() {
+	// Use a spec that does not prehash values in the smst. This returns a nil
+	// value hasher for the proof verification in order to to avoid hashing the
+	// value twice.
+	spec = smt.NoPrehashSpec(sha256.New(), true)
```
h5law, line 40: Nothing actionable right now but this won't be needed anymore after https://github.com/pokt-network/smt/pull/42 merges you can just pass the spec normally.

x/proof/keeper/msg_server_submit_proof.go
```diff
@@ -43,64 +77,162 @@ func (k msgServer) SubmitProof(ctx context.Context, msg *types.MsgSubmitProof) (
...
+	if err := sparseMerkleClosestProof.Unmarshal(msg.GetProof()); err != nil {
+		return nil, types.ErrProofInvalidProof.Wrapf(
+			"failed to unmarshal closest merkle proof: %s",
+			err,
+		)
+	}
+
+	// Get the relay request and response from the proof.GetClosestMerkleProof.
+	closestValueHash := sparseMerkleClosestProof.ClosestValueHash
+	relayBz := closestValueHash[:len(closestValueHash)-sumSize]
```
h5law, line 112:
```suggestion
	relayBz := sparseMerkleClosestProof.GetValueHash(spec)
```

x/proof/keeper/msg_server_submit_proof.go
```diff
@@ -43,64 +77,162 @@ func (k msgServer) SubmitProof(ctx context.Context, msg *types.MsgSubmitProof) (
...
 	}
+	logger.Debug("successfully compared relay request session header")
+
+	// Verify that the relay response session header matches the proof session header.
+	if err := compareSessionHeaders(msg.GetSessionHeader(), relayRes.Meta.GetSessionHeader()); err != nil {
+		return nil, status.Error(codes.FailedPrecondition, err.Error())
+	}
+	logger.Debug("successfully compared relay response session header")
+
+	// Verify the relay request's signature.
```
h5law, line 154:
```suggestion
	// Verify the relay request's signature.
	// TODO_TECHDEBT(@h5law): Fetch the correct ring for the session
	// this relay is from.
```

x/proof/keeper/msg_server_submit_proof.go
h5law: @Olshansk @red-0ne https://github.com/pokt-network/smt/commit/e3dbbbd9f97d4f9007c67f5c7e512a93c8f3335c

Now there is enforcement in ClosestProof and its verification that the paths are the same size in bytes - both methods will error if they dont match which means this case is already covered by verifying the proof - and creating it.

x/proof/types/expected_keepers.go
```diff
@@ -7,21 +7,29 @@ import (
...
+	GetAccount(context.Context, sdk.AccAddress) sdk.AccountI
 }
 
 // BankKeeper defines the expected interface for the Bank module.
 type BankKeeper interface {
 	SpendableCoins(context.Context, sdk.AccAddress) sdk.Coins
-	// Methods imported from bank should be defined here
+	DelegateCoinsFromAccountToModule(ctx context.Context, senderAddr sdk.AccAddress, recipientModule string, amt sdk.Coins) error
+	UndelegateCoinsFromModuleToAccount(ctx context.Context, senderModule string, recipientAddr sdk.AccAddress, amt sdk.Coins) error
+}
```
h5law, line 29: Didn't you remove the `BankKeeper` as a dependency in `x/proof/module/module.go` in the `ModuleInputs`. Or am I missing something?

x/service/types/relay.go
```diff
@@ -1,57 +1,101 @@
...
+
+	return nil
 }
 
 // GetSignableBytesHash returns the hash of the signable bytes of the relay response
 // Hashing the marshaled response message guarantees that the signable bytes are
 // always of a constant and expected length.
-func (res *RelayResponse) GetSignableBytesHash() ([32]byte, error) {
-	responseBz, err := res.getSignableBytes()
+func (res RelayResponse) GetSignableBytesHash() ([32]byte, error) {
```
h5law, line 49: Glad we got this sorted with the protobuf annotations nice work 👍🏼

h5law (dismissed):
Gone thru it all 😅

Screenshot 2024-03-21 at 06 26 15

Overall great PR, left a few comments throughout - the SMT stuff you need is merged in main so updating `go.mod` to point to the latest commit will help with where you need them. Really great work!

x/proof/types/expected_keepers.go
red-0ne: Yes it's not needed by the Proof keeper itself but seems useful for integration tests:

https://github.com/pokt-network/poktroll/pull/406#discussion_r1530163406

pkg/crypto/pubkey_client/client.go
```diff
@@ -0,0 +1,59 @@
...
+	}
+
+	// If the account's public key is nil, then return an error.
+	pubKey := acc.GetPubKey()
+	if pubKey == nil {
+		return nil, ErrPubKeyClientEmptyPubKey
+	}
+
+	return pubKey, nil
+}
```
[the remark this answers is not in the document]
h5law: Nice one 👍🏼

pkg/client/query/accquerier.go
```diff
@@ -18,6 +20,10 @@ var _ client.AccountQueryClient = (*accQuerier)(nil)
 type accQuerier struct {
 	clientConn     grpc.ClientConn
 	accountQuerier accounttypes.QueryClient
+
+	// accountCache is a cache of accounts that have already been queried.
+	// TODO_TECHDEBT: Add a size limit to the cache and consider an LRU cache.
```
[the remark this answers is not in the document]
red-0ne: I imaging we could have a cache for all of the queriers.

x/proof/keeper/msg_server_submit_proof.go
red-0ne: Added the `TODO_BLOCKER`s. And now using https://github.com/pokt-network/smt/commit/e3dbbbd9f97d4f9007c67f5c7e512a93c8f3335c.

@Olshansk , Anything else preventing from removing the `TODO_IN_THIS_PR` ?

Olshansk (approved):
My update:
1. Reviewed the changes: looks good
2. Added a tiny comment update: linking to our discussion
3. Merged with main
4. Pushed
5. Enabled E2E tests

Let's get everything 🟢 and then :shipit: !!!

Olshansk (approved):
My update:
1. Reviewed the changes: looks good
2. Added a tiny comment update: linking to our discussion
3. Merged with main
4. Pushed
5. Enabled E2E tests

Let's get everything 🟢 and then :shipit: !!!

Issue #354, opened 2024-01-29
[bug] Goroutines leaks on RelayMiner/AppGate


okdas:
## Objective

Both AppGate and RelayMiner creating, but not cleaning up goroutines at approx 2 goroutines per relay.

## Origin Document

<img width="1467" alt="image12312" src="https://github.com/pokt-network/poktroll/assets/4950477/5306ab45-fcac-4493-af01-90b014fe3a48">

## Goals

- Make sure the amount of goroutines does not grow linearly with a number of relays. 

## How to reproduce

1. Start LocalNet: `make localnet_up`
2. Stake actors and run e2e tests to make sure LocalNet works: `make supplier1_stake && make app1_stake` && `make test_e2e`
3. Get the current numbers of goroutines:

AppGate:
```
curl localhost:9093 | grep go_goroutines
# HELP go_goroutines Number of goroutines that currently exist.
# TYPE go_goroutines gauge
go_goroutines 77
```

RelayMiner:
```
curl localhost:9094 | grep go_goroutines
# HELP go_goroutines Number of goroutines that currently exist.
# TYPE go_goroutines gauge
go_goroutines 846
```

(This number already seems kind of high after runnung e2e tests)

4. Now, run basic load test: `make load_test_simple`
5. Check the number of routines again:

AppGate:

```
curl localhost:9093 | grep go_goroutines
# HELP go_goroutines Number of goroutines that currently exist.
# TYPE go_goroutines gauge
go_goroutines 11743
```

RelayMiner:
```
curl localhost:9094 | grep go_goroutines
# HELP go_goroutines Number of goroutines that currently exist.
# TYPE go_goroutines gauge
go_goroutines 18219
```

## General deliverables

- [ ] **Comments**: Add/update TODOs and comments alongside the source code so it is easier to follow.
- [ ] **Testing**: Add new tests (unit and/or E2E) to the test suite.
- [ ] **Makefile**: Add new targets to the Makefile to make the new functionality easier to use.
- [ ] **Documentation**: Update architectural or development READMEs; use [mermaid](https://mermaid-js.github.io/mermaid/) diagrams where appropriate.

---

**Creator**: @okdas 
**Co-Owners**: @red-0ne @okdas

Olshansk:
CCing @ezeike @adshmh as well. This is the first (of many) tickets where protocol & backend work will start to overlap.

Olshansk:
@red-0ne 

[...]

PR #410, opened 2024-03-05
[Bug] Address go-routines leak

red-0ne:
### Human Summary

`AppGate Server` and `RelayMiner` leak go-routines when handling relays, resulting to an ever increasing number each time a relay is processed.

This PR addresses most of the leaks by cancelling the contexts passed to `ReplayObservables` and `ReplayClients` but does not eliminate them totally.

Means to test and observe these leaks are described in #354.

### AI Summary

reviewpad:summary

## Issue

- #354 

## Type of change

Select one or more:

- [x] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

## Sanity Checklist


pkg/relayer/session/session.go
```diff
@@ -225,8 +225,10 @@ func (rp *relayerSessionsManager) validateConfig() error {
 // waitForBlock blocks until the block at the given height (or greater) is
 // observed as having been committed.
 func (rs *relayerSessionsManager) waitForBlock(ctx context.Context, height int64) client.Block {
+	ctx, cancel := context.WithCancel(ctx)
```
Olshansk, line 228: #PUC why we need to cancel

pkg/relayer/session/session.go
```diff
@@ -225,8 +225,10 @@ func (rp *relayerSessionsManager) validateConfig() error {
...
+	defer cancel()
+
 	subscription := rs.blockClient.CommittedBlocksSequence(ctx).Subscribe(ctx)
```
Olshansk, line 231: Why don't we unsusbscripe anymore? #PUC

pkg/observable/channel/replay.go
```diff
@@ -111,7 +112,18 @@ func (ro *replayObservable[V]) Subscribe(ctx context.Context) observable.Observe
...
-	observer := NewObserver[V](ctx, ro.observerManager.remove)
+	if ctx == nil {
+		ctx = context.Background()
+	}
+
+	// caller can cancel context or close the publish channel to unsubscribe active observers
+	ctx, cancel := context.WithCancel(ctx)
+	removeAndCancel := func(toRemove observable.Observer[V]) {
+		ro.observerManager.remove(toRemove)
+		cancel()
```
Olshansk, line 123: If the caller never calles `removeAndCancel`, does this never get cancelled?

pkg/observable/channel/map.go
```diff
@@ -112,6 +121,7 @@ func goMapTransformNotification[S, D any](
 	srcObserver observable.Observer[S],
 	transformFn MapFn[S, D],
 	publishFn func(dstNotifications D),
+	doneFn func(),
```
Olshansk, line 124: #PUC on what's expected of this function.

Olshansk (changes requested):

pkg/client/events/replay_client.go
```diff
@@ -145,21 +145,10 @@ func (rClient *replayClient[T]) EventsSequence(ctx context.Context) observable.R
 // goRemapEventsSequence publishes events observed by the most recent cached
 // events type replay observable to the given publishCh
 func (rClient *replayClient[T]) goRemapEventsSequence(ctx context.Context, publishCh chan<- T) {
-	var prevEventTypeObs observable.ReplayObservable[T]
 	channel.ForEach[observable.ReplayObservable[T]](
 		ctx,
 		rClient.replayObsCache,
 		func(ctx context.Context, eventTypeObs observable.ReplayObservable[T]) {
-			if prevEventTypeObs != nil {
```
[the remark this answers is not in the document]
bryanchriswhite: This was originally added in reaction to a failing test at some point, IIRC. Looking at it now, I don't think it works like we thought at the time. I think in reality it was a band-aid that was covering up some of the other leaks that this PR fixes which are present deeper in dependency tree (i.e. `Observable` & `ReplayObservable`).

pkg/observable/channel/map.go
bryanchriswhite: Could this be a channel instead of an anonymous function? Then we would just call `close` on 134 instead of the func instead.

I think the comment should be something to the effect of: it's the sender's responsibility to close the channel, since the publish channel (`dstProducer`) and its associated observable are constructed in this scope, if they're not also closed here then the associated goroutine will leak.

bryanchriswhite: Could this be a channel instead of an anonymous function? Then we would just call `close` on 134 instead of the func instead.

I think the comment should be something to the effect of: it's the sender's responsibility to close the channel, since the publish channel (`dstProducer`) and its associated observable are constructed in this scope, if they're not also closed here then the associated goroutine will leak. Closing the channel ends the goroutine.

pkg/relayer/session/session.go
bryanchriswhite: I think the comment you're asking for already exists in `replayObservable#Subscribe()`:

```
// caller can cancel context or close the publish channel to unsubscribe active observers
```

pkg/observable/channel/replay.go
bryanchriswhite: `removeAndCancel` is only ever called in the `Observer` implementation (see: `#unsubscribe()`, where it's referenced as `observer#onUnsubscribe()`) via [`channelObserverManager#goUnsubscribeOnDone()`](https://github.com/pokt-network/poktroll/blob/ddfdd3ca0331991e98590a1366bcdffc7deee109/pkg/observable/channel/observer_manager.go#L135), below.

pkg/relayer/session/session.go
bryanchriswhite: Anywhere we `#Subscribe()` to an observable, we need to either cancel the associated context or close the associated publish channel. The former is preferable for consistency because we may not always have a reference to the publish channel. For example, an observable (and it's corresponding publish channel) could be constructed in another scope that only returns the observable itself. In this case the only way to "close" the observable is to cancel the context.

bryanchriswhite (approved):
Looks good to me - thanks for diving into this! :raised_hands:

pkg/client/events/replay_client.go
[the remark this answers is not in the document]
Olshansk: @bryanchriswhite To your knowledge, do you think we have any more similar bandaids in the code?

pkg/relayer/session/session.go
Olshansk: @bryanchriswhite Shold the observable be able to close itself?

pkg/client/tx/client.go
```diff
@@ -375,6 +375,10 @@ func (tClient *txClient) addPendingTransactions(
 // Parameters:
 // - ctx: Context for managing the function's lifecycle and child operations.
 func (tClient *txClient) goSubscribeToOwnTxs(ctx context.Context) {
+	// Create a cancellable child context for managing the EventsSequence lifecycle.
+	// Cancelling it at the end of the function, when listening for submitted
+	// transactions is no longer needed will ensure that the EventsSequence, its
```
Olshansk, line 380: `is no longer needed will`

Is this missing an `and`?

pkg/observable/channel/map.go
```diff
@@ -28,9 +28,7 @@ func Map[S, D any](
 		func(dstNotification D) {
 			dstProducer <- dstNotification
 		},
-		func() {
-			close(dstProducer)
-		},
+		dstProducer,
```
Olshansk, line 31: Nice

pkg/observable/channel/map.go
```diff
@@ -116,12 +110,16 @@ func ForEach[V any](
 
 // goMapTransformNotification transforms, optionally skips, and publishes
 // notifications via the given publishFn.
-func goMapTransformNotification[S, D any](
+func goMapTransformNotification[S, D any, P any](
```
Olshansk, line 113: Can we do this?

```suggestion
func goMapTransformNotification[S, D, P any](
```

Olshansk (approved):
A few small comments but approving regardless.

I can't say I have a 100% understanding of all the observable flows but trust that it's working as expected.

pkg/client/tx/client.go
red-0ne: It should not. What about:
```
// Cancelling it when listening for submitted transactions is no longer needed
// will ensure that the EventsSequence, its subscriptions and underlying channels
// are effectively closed.
```

Removed the `Cancelling it at the end of the function` since the `defer cancel()` is not far from the comment.

pkg/client/tx/client.go
```diff
@@ -376,9 +376,9 @@ func (tClient *txClient) addPendingTransactions(
 // - ctx: Context for managing the function's lifecycle and child operations.
 func (tClient *txClient) goSubscribeToOwnTxs(ctx context.Context) {
 	// Create a cancellable child context for managing the EventsSequence lifecycle.
-	// Cancelling it at the end of the function, when listening for submitted
-	// transactions is no longer needed will ensure that the EventsSequence, its
-	// subscriptions and underlying channels are effectively closed.
+	// Cancelling it when listening for submitted transactions is no longer needed
```
Olshansk, line 379: NIT: is/are but we can do it in a separate commit

pkg/client/tx/client.go
Olshansk: LGTM

PR #413, opened 2024-03-06
[PubKeyClient] Implement PubKeyClient for on/off-chain usage

red-0ne:
### Human Summary

Add a PubKeyClient interface and its implementation to handle pubkeys retrieval from `POKTRollSDK` and the `Proof` keeper.

It injects an `AccountQuerier` for its internal usage which will have different provisioning when on-chain or off-chain is initializing it.

Move `RelayResponse` signature to the `service` type's `relay.go` and add initial implementation of `ValidateBasic`.

### AI Summary

reviewpad:summary

## Type of change

Select one or more:

- [ ] Bug fix
- [x] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

## Sanity Checklist


x/proof/keeper/keeper.go
```diff
@@ -63,10 +63,19 @@ func NewKeeper(
 
 	ringClient, err := rings.NewRingClient(ringClientDeps)
 	if err != nil {
-		return Keeper{}, err
+		panic(err)
```
bryanchriswhite, line 66: #PUC explaining rationale. Is this because of an expectation that `NewKeeper()` should only be called on startup?

x/proof/keeper/keeper.go
```diff
@@ -63,10 +63,19 @@ func NewKeeper(
...
 	}
 
-	k := Keeper{
+	pubKeyClientDeps := depinject.Supply(
+		accountQuerier,
+	)
+
+	pubKeyClient, err := pubkeyclient.NewPubKeyClient(pubKeyClientDeps)
```
bryanchriswhite, line 73: `pubKeyClientDeps` could either be in-lined, or substituted for `replayClientDeps`.

x/proof/keeper/keeper.go
red-0ne: OK, I think I understand your use case for making `NewKeeper`s return an error. But then shouldn't we remove the `sdk.AccAddressFromBech32(authority)` panic from the other `Keeper`'s initializers?

x/proof/keeper/keeper.go
red-0ne: You mean substituded for `ringClientDeps`?

bryanchriswhite: Yes, my mistake. :facepalm:

bryanchriswhite (changes requested):
Nice :sunglasses::ok_hand: 

I had one more suggestion, I threw it in a commit and opened #416.

Issue #20, opened 2023-10-05
[Testing] Load Testing Relays Round #1


Olshansk:
## Objective

Complete load test #1.

## Origin Document

The result of the following ticket:
- #229 

Was the following PR:
- #286 

Which was merged here:
- https://dev.poktroll.com/infrastructure/testing/load_testing_plan

## Goals

- Complete load test #1
- Identify bugs & issues in the codebase and infrastructure
- Identify gaps in our tooling & visibility

## Deliverables

- [ ] Document the results (which will be uploaded to [dev.poktroll.com](https://dev.poktroll.com/)) capturing the results of the first load test found [here](https://dev.poktroll.com/infrastructure/testing/load_testing_plan)
- [ ] Document the tools that were added to perform the load test
- [x] Add Grafana visuals for Load Tests.
	* That would unblock drilling down the results ~~by k6 tags, as default functionality does not support that - tags are only available for time series data~~.

## Non-goals / Non-deliverables

- Solving issues or gaps we've identified in the load test
- Items outside the load testing plan

---

**Creator**: @Olshansk
**Co-Owners**: @okdas @red-0ne

okdas:
TODO: create a ticket to add Grafana visuals for Load Tests. That would unblock drilling down the results by k6 tags, as default functionality does not support that - tags are only available for time series data.

Olshansk:
@okdas Thank you for capturing that action item. I've started adding deliverables to avoid having too many "small tasks" for our one load testing operation. @red-0ne please do the same as things come to mind!

Olshansk:
@okdas @red-0ne When we go through [1], we should reference [2] put together by @RawthiL. It is a production version of the demand side scalability in practice on Morse.

[1] https://dev.poktroll.com/infrastructure/testing/load_testing_plan
[2] https://forum.pokt.network/t/block-sizes-claims-and-proofs-in-the-multi-gateway-era/5060

bryanchriswhite:
One of the outstanding deliverables is captured in #513, along with additional deliverables. I've moved the other outstanding deliverable ("Document the tools that were added to perform the load test") to #513 as well and am closing this issue for now.

PR #414, opened 2024-03-06
[LocalNet] Allow to scale up actors

okdas:
### Human Summary

Allows to scale actors on LocalNet. 

### AI Summary

reviewpad:summary

## Issue

We need to be able to have multiple actors of each kind staked and in working condition on LocalNet for load-testing development.

- #20 

## Testing

## Sanity Checklist

### DIFF
--- a/Tiltfile
+++ b/Tiltfile
@@ -132,7 +132,7 @@ for x in range(localnet_config["relayminers"]["count"]):
         resource_deps=["validator"],
         port_forwards=[
             str(8084+actor_number)+":8545", # relayminer1 - exposes 8545, relayminer2 exposes 8546, etc.
-            str(40044+actor_number)+":40005", # DLV port. relayminer1 - exposes 40045, relayminer2 exposes 40045, etc.
+            str(40044+actor_number)+":40005", # DLV port. relayminer1 - exposes 40045, relayminer2 exposes 40046, etc.
             # Run `curl localhost:PORT` to see the current snapshot of relayminer metrics.
             str(9069+actor_number)+":9090", # Relayminer metrics port. relayminer1 - exposes 9070, relayminer2 exposes 9071, etc.
         ],
--- a/Tiltfile
+++ b/Tiltfile
@@ -10,7 +10,6 @@ localnet_config_path = "localnet_config.yaml"
 localnet_config_defaults = {
     "validator": {"cleanupBeforeEachStart": True},
     "relayminers": {"count": 1},
-    "gateways": {"count": 1},
     "appgateservers": {"count": 1},
     # By default, we use the `helm_repo` function below to point to the remote repository
     # but can update it to the locally cloned repo for testing & development
@@ -98,7 +97,7 @@ k8s_yaml(
     ["localnet/kubernetes/anvil.yaml", "localnet/kubernetes/validator-volume.yaml"]
 )
 
-# Run pocket-specific nodes (validator, relayminers, etc...)
+# Provision validator
 helm_resource(
     "validator",
     chart_prefix + "poktroll-validator",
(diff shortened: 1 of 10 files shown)


localnet/poktrolld/config/supplier1_stake_config.yaml
```diff
@@ -4,7 +4,7 @@ stake_amount: 1000upokt
 services:
   - service_id: anvil
     endpoints:
-      - url: http://anvil.supplier1:8545
+      - url: http://relayminer1:8545
```
Olshansk, line 7: Why is the `relayminer` endpoint the one backing the `anvil` service?

I thought the `url` for `anvil` should be the "data node" or "service node". No?

Olshansk (changes requested):
Just one question for my personal understanding.

localnet/poktrolld/config/supplier1_stake_config.yaml
okdas: @Olshansk this is a stake config. That value should be pointing to the endpoint by which relay miner should be accessed to provide service. This staking config was actually never used because `supplier1` is staked in `genesis.json` from the first block (specified in `config.yml`). I just wanted this value to be the same as in `config.yml` to avoid confusion.

Olshansk: > which relay miner should be accessed to provide service

Can we reword this sentence to `which the supplier should be accessed to provide service`?

I understand that this works, but I'm already getting confused by why an on-chain advertised address (i.e. for supplying a service) uses a relayminer endpoint (an off-chain endpoint for handling business logic).

For example, `anvil.supplier:8454` was very clear. Seeing how these configs will be used as examples we need to make them clear.

Screenshot 2024-03-07 at 3 55 17 PM

okdas: @Olshansk, this endpoint is a DNS name that was created by Kubernetes/Helm chart, The name comes from the name of the Helm chart deployment [here](https://github.com/pokt-network/poktroll/blob/localnet-scale-up-actors/Tiltfile#L119). This DNS allows all resources within the same Kubernetes namespace (including AppGateServers) to access that RelayMiner.

> why an on-chain advertised address (i.e. for supplying a service) uses a relayminer endpoint (an off-chain endpoint for handling business logic).

I believe this is the purpose of on-chain address advertising: to point to suppliers (RelayMiner services).

In a real-world scenario, when resources are not just deployed in one single Kubernetes namespace, it would be a normal URL such as `supplier-1.poktroll.com`, or `relayminer-1.poktroll.com` or any other DNS.

What am I missing? I can hop on a call to discuss this.

Olshansk: Makes sense. In that case.

1. Can we add inline comments in the `xxx_stake_config.yaml` next/above relay miner explaining `the relayminer port is exposed since it proxies the request to the backing anvil data node on the bakend`
2. Can you update https://dev.poktroll.com/configs/supplier_staking_config to explain that it'll usually be the endpoint exposed by the relay miner?

For 2, try to keep it short (not long sentences) but clear w/ bullet points if necessary.

Tiltfile
```diff
@@ -111,56 +110,65 @@ helm_resource(
...
+        image_deps=["poktrolld"],
+        image_keys=[("image.repository", "image.tag")],
+    )
+    k8s_resource(
+        "relayminer"  + str(actor_number),
+        labels=["supplier_nodes"],
+        resource_deps=["validator"],
+        port_forwards=[
+            str(8084+actor_number)+":8545", # relayminer1 - exposes 8545, relayminer2 exposes 8546, etc.
+            str(40044+actor_number)+":40005", # DLV port. relayminer1 - exposes 40045, relayminer2 exposes 40045, etc.
```
red-0ne, line 135:
```suggestion
            str(40044+actor_number)+":40005", # DLV port. relayminer1 - exposes 40045, relayminer2 exposes 40046, etc.
```

red-0ne (commented):
Just a small change request on a comment but looks good to me.

Also, do these changes have a corresponding helm-chart configuration or should we edit the local one to reflect them?

I was thinking specifically, if I want to enable debugging on relayminer-2 for example, 
* Would `templates/deployment.yaml` affect all of them?
* How would the debug container port in `templates/deployment.yaml` behave?

Olshansk:
@okdas Let us know once the docs are updated!

okdas:
> Also, do these changes have a corresponding helm-chart configuration or should we edit the local one to reflect them?

Negative, this change didn't require any modifications to the helm chart.

> Would templates/deployment.yaml affect all of them?

Currently - yes! Would it interfere with your work? We can change that! I'm thinking we can expose the setting in `localnet_config.yml` so only one relayminer at a time would run dlv.

> How would the debug container port in templates/deployment.yaml behave?

The port would be open as it always did, but Tilt would map it to the different port. For example, for relayminer2, the port `40045` would point to the dlv (`40005` inside of the container).

okdas:
> @okdas Let us know once the docs are updated!

They are updated now! :)

Olshansk (approved):

Olshansk:
@okdas I approved this last week. Are we waiting on anything else?

okdas:
> @okdas I approved this last week. Are we waiting on anything else?

Nope - thanks for the reminder!

PR #417, opened 2024-03-07
[Proof] add proof submission integration tests

bryanchriswhite:
### Human Summary

- Adds integration tests exercising submit proof message handling.
- Moves `minRelayDifficultyBits` to a gov param.
- Bumps coverage of proof message handling to 90% (also bumps coverage of related files):
image

### AI Summary

reviewpad:summary

## Issue

- #403 

## Type of change

Select one or more:

- [ ] New feature, functionality or library
- [ ] Bug fix
- [x] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

- [x] **Run all unit tests**: `make go_develop_and_test`
- [x] **Run E2E tests locally**: `make test_e2e`
- [ ] **Run E2E tests on DevNet**: Add the `devnet-test-e2e` label to the PR. This is VERY expensive, only do it after all the reviews are complete.

## Sanity Checklist

- [x] I have tested my changes using the available tooling
- [x] I have performed a self-review of my own code
- [x] I have commented my code, updated documentation and left TODOs throughout the codebase

PR #423, opened 2024-03-12
chore: add todos for future iterations/improvements

h5law:
## Summary

Adds todos for work to be completed in the future

## Issue

- N/A

## Type of change

Select one or more:

- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [x] Other (comments)

## Testing

## Sanity Checklist


pkg/client/interface.go
```diff
@@ -46,6 +46,7 @@ type SupplierClient interface {
 	// smt.SparseMerkleClosestProof, corresponding to some previously created claim
 	// for the same session. The proof is validated on-chain as part of the pocket
 	// protocol.
+	// TODO_IMPROVE(@h5law): Use a SparseCompactClosestProof here instead.
```
Olshansk, line 49: #PUC on why.

Thoughts on just creating a ticket that says "migrate to compact proofs on-chain"?

That way:
1.  You can reference the same ticket in multiple places
2. Make a deliverable be "search for all #XXX in the codebase and resolve it"

pkg/relayer/session/sessiontree.go
```diff
@@ -122,6 +122,8 @@ func (st *sessionTree) Update(key, value []byte, weight uint64) error {
 // This function is intended to be called after a session has been claimed and needs to be proven.
 // If the proof has already been generated, it returns the cached proof.
 // It returns an error if the SMST has not been flushed yet (the claim has not been generated)
+// TODO_IMPROVE(h5law): Compress the proof into a SparseCompactClosestMerkleProof
+// prior to submitting to chain.
```
Olshansk, line 126:
```suggestion
// prior to submitting to chain to reduce on-chain storage requirements for proofs.
```

x/proof/keeper/msg_server_submit_proof.go
```diff
@@ -62,6 +62,7 @@ func (k msgServer) SubmitProof(ctx context.Context, msg *types.MsgSubmitProof) (
 		ClosestMerkleProof: msg.Proof,
 	}
 
+	// TODO_IMPROVE(@h5law): Utilise smt.VerifyCompactClosestProof here.
```
Olshansk, line 65: Ditto the comment above

Olshansk (changes requested):

pkg/client/interface.go
h5law: https://github.com/pokt-network/poktroll/issues/427

x/proof/keeper/msg_server_submit_proof.go
```diff
@@ -62,6 +62,8 @@ func (k msgServer) SubmitProof(ctx context.Context, msg *types.MsgSubmitProof) (
 		ClosestMerkleProof: msg.Proof,
 	}
 
+	// TODO_IMPROVE(@h5law #427): Utilise smt.VerifyCompactClosestProof here to
+	// reduce on-chain storage requirements for proofs.
```
bryanchriswhite, line 66: I don't think this comment is in quite the right place. Ideally, this PR would be based on #406 (or #417, which is downstream from there). Closest Merkle proof verification isn't done in `queryAndValidateClaimForProof`.

image

image

x/proof/keeper/proof.go
```diff
@@ -15,6 +15,8 @@ import (
 func (k Keeper) UpsertProof(ctx context.Context, proof types.Proof) {
 	logger := k.Logger().With("method", "UpsertProof")
 
+	// TODO_IMPROVE(@h5law #427): Use the marshal method on the SparseCompactClosestProof
+	// type here instead in order to reduce space stored on chain.
```
bryanchriswhite, line 19: `proof` here is referring to the protobuf type, not the closest Merkle proof, which is a field on that type. I think instead of this comment here, a comment should be added to proto/poktroll/proof/proof.proto to rename `closest_merkle_proof` to `compact_closest_merkle_proof`.

I think this comment belongs in pkg/relayer/session/proof.go#99, where `relayer.SessionTree#ProveClosest()` is called (the implementation of which you've already updated the godoc comment for in this PR).

pkg/relayer/protocol/difficulty.go
```diff
@@ -18,6 +18,13 @@ func MustCountDifficultyBits(bz []byte) int {
...
 // CountDifficultyBits returns the number of leading zero bits in the given byte
 // slice. It returns an error if the byte slice is all zero bits.
+// TODO_IMPROVE(@h5law): Remove the forloop logic and replace with a simplified
+// single method that accounts for the fact that paths are **always** 32 bytes.
+//
+//	if (lenBz) != 32 {
+//	    return 0, ErrDifficulty.Wrapf("invalid byte length; got %d, want 32")
+//	}
+//	return bits.LeadingZeros64(binary.LittleEndian.Uint64(bz)), nil
 func CountDifficultyBits(bz []byte) (int, error) {
```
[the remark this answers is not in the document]
bryanchriswhite: It looks like this was resolved but no change was made?

bryanchriswhite (changes requested):
Thanks for the quick comment additions! :raised_hands: 

I think it would be better if this PR were based off of either #406 or #417 the actual proof verification isn't present in main yet.

pkg/relayer/protocol/difficulty.go
```diff
@@ -18,6 +18,17 @@ func MustCountDifficultyBits(bz []byte) int {
 
 // CountDifficultyBits returns the number of leading zero bits in the given byte
 // slice. It returns an error if the byte slice is all zero bits.
+//
+// TODO_IMPROVE(@h5law): Remove the forloop logic and replace with a simplified
+// single method that accounts for the fact that block hashes/paths are always
```
Olshansk, line 23: I know I'm sounding like a broken record a bit but:

1. The assumption that block hahes will be 32 bytes is fair
2. The assumption that paths will remain 32 bytes I don't think should be stated. smt is a library so whether or not it will be configurable, we should tightly couple the two so much. 
3. We should change how we use hashes -> paths. The idea is that a `block hash` is a `seed` for the path and not the path itself. I was thinking `path = foo(block_hash, session_id)` and it does some sort of processing inside.

x/proof/keeper/msg_server_submit_proof.go
Olshansk: 👍

Olshansk (changes requested):
One small comment and also second the ones left by @bryanchriswhite

pkg/relayer/protocol/difficulty.go
h5law: > The assumption that paths will remain 32 bytes I don't think should be stated. smt is a library so whether or not it will be configurable, we should tightly couple the two so much.

I mean its because we are using `sha256` the `smt` library can have bigger or smaller paths but we have chosen a 32 byte hasher. I think this is something we should settle on and not change - maybe only change to https://pkg.go.dev/golang.org/x/crypto/sha3 `sha3.Sum256` I see no reason for this to be configurable as it would make things overly complex.

> We should change how we use hashes -> paths. The idea is that a block hash is a seed for the path and not the path itself. I was thinking path = foo(block_hash, session_id) and it does some sort of processing inside.

```go
seed := blockHash
seed = append(seed, []byte(serviceID))
path := sha256.Sum256(seed)
```

??? Does it have to be more complex than hashing some seed we can augment it with as much data as we want but hashing it needs to be done now as path sizes equalling one another is a requirement of the ClosestProof - unless this is not intended to be the same as the path provided to that?

x/proof/keeper/proof.go
h5law: My bad 😅 I will rebase onto #406 and add these comments in the appropriate places.

pkg/relayer/protocol/difficulty.go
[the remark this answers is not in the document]
Olshansk: I'd argue it's part of the TODO_IMPROVE but I did rename the function.

Olshansk (approved):

Olshansk:
@bryanchriswhite PTAL at this PR again when you have a chance.

bryanchriswhite (approved):
LGTM :+1:

PR #424, opened 2024-03-12
[LocalNet] Fix/localnet regenesis

red-0ne:
### Human Summary

`make localnet_regenesis` copies configuration files to the wrong directory, copying them to `localnet/poktrolld/config/config` instead of `localnet/poktrolld/config`.

This lead to the validator not able to locate `genesis.json` when deploying the actors from a clean repo (that does not have previous files in the right directory)

image

The PR also fixes the location of some off-chain actors interfaces that were badly referenced in `Makefile`'s `go_mockgen`

### AI Summary

reviewpad:summary

## Type of change

Select one or more:

- [x] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

## Sanity Checklist


Makefile
```diff
@@ -242,9 +242,8 @@ localnet_regenesis: check_yq acc_initialize_pubkeys_warn_message ## Regenerate t
 	@echo "Initializing chain..."
 	@set -e ;\
 	ignite chain init --skip-proto ;\
-	mkdir -p $(POKTROLLD_HOME)/config/ ;\
```
Olshansk, line 245: Why not keep `mkdir -p` but remove ` /config/` from the end?

Makefile
```diff
@@ -242,9 +242,8 @@ localnet_regenesis: check_yq acc_initialize_pubkeys_warn_message ## Regenerate t
...
 	cp -r ${HOME}/.poktroll/keyring-test $(POKTROLLD_HOME) ;\
-	cp -r ${HOME}/.poktroll/config/ $(POKTROLLD_HOME)/config/ ;\
+	cp -r ${HOME}/.poktroll/config $(POKTROLLD_HOME)/ ;\
```
Olshansk, line 246: I found a bunch of other references to `/config/`. 

1. Are you confident that this works?
2. Does any of the other code need to be modified?

Screenshot 2024-03-12 at 3 28 18 PM
Screenshot 2024-03-12 at 3 28 08 PM

Makefile
```diff
@@ -310,9 +309,9 @@ go_mockgen: ## Use `mockgen` to generate mocks used for testing purposes of all
 	go generate ./x/proof/types/
 	go generate ./x/tokenomics/types/
 	go generate ./pkg/client/interface.go
```
Olshansk, line 311: As an improvement, why not do something like: `find . -name "interface.go | xargs -I {} go generate {}"`

(I didn't test this so bash is from memory but I'm sure ChatGPT can help)

Olshansk (changes requested):

Olshansk (changes requested):
@red-0ne Left a couple of comments

Makefile
red-0ne: `cp -r ${HOME}/.poktroll/config $(POKTROLLD_HOME)/` creates the config directory. Having them both is redundant.

We can leave `mkidr -p` and have `cp -r ${HOME}/.poktroll/config/* $(POKTROLLD_HOME)/config/`

They achieve the same, except the later may skip dot prefixed files.

Makefile
red-0ne: On the first screenshot, the copy is commented but is fine too. Since each copy command copies a single file.

The second screenshot is using ` $(POKTROLLD_HOME)/config/` and not ` $(POKTROLLD_HOME)/config/config` which is fine too.

I'm confident we changed them only where needed.

okdas (approved):
LGTM! :) 

I've never had an issue with a nested `config` dir, but that could be macos/linux cli differences, as I think I've heard Bryan had a similar problem happen.

Thanks for fixing this!

Makefile
Olshansk: @red-0ne you 👍ed the comment, re-requested a review, but the code is still the same?

Olshansk (changes requested):

red-0ne:
@Olshansk , Fixed, I added `find` command to generate `pkg` interface mocks

Olshansk (approved):

PR #432, opened 2024-03-15
[Load Testing] gherkin/go implementation of test plan #1

bryanchriswhite:
## Summary

Implementing [load test plan #1](https://github.com/pokt-network/poktroll/issues/20) using gherkin and go tests.

### How to run:
* Make sure to have the appropriate `gateways` and `relayminers` in `localnet_config.yaml` (max number is 3 for each)
* Edit the `relays_stress.feature` to match the testing plan to execute
```cucumber
    And a rate of "1" relay requests per second is sent per application
    And the following initial actors are staked:
      | actor       | count |
      | gateway     | 1     |
      | application | 1     |
      | supplier    | 1     |
    And more actors are staked as follows:
      | actor       | actor inc rate | blocks per inc | max actors |
      | gateway     | 1              | 4              | 3          |
      | application | 2              | 4              | 5          |
      | supplier    | 1              | 4              | 3          |
```

Run the test with the following command:
```
go test -v -run TestLoadRelays ./load-testing/tests/...
```

## Issue

- #20 

## Type of change

Select one or more:

- [ ] New feature, functionality or library
- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

- [ ] **Unit Tests**: `make go_develop_and_test`
- [ ] **LocalNet E2E Tests**: `make test_e2e`
- [ ] **DevNet E2E Tests**: Add the `devnet-test-e2e` label to the PR. **THIS IS VERY EXPENSIVE**, so only do it after all the reviews are complete.
- [ ] **Documentation changes**: `make docusaurus_start`

## Sanity Checklist

- [ ] I have tested my changes using the available tooling
- [ ] I have commented my code
- [ ] I have performed a self-review of my own code; both comments & source code
- [ ] I create and referenced any new tickets, if applicable
- [ ] I have left TODOs throughout the codebase, if applicable

PR #433, opened 2024-03-18
[WIP][Test]: k6 for load testing plan 1

red-0ne:
## Summary

Use k6 to implement load testing plan 1 execution #20. This is to compare with #432 and make an informed decision on which tools to use for load testing.

This PR only handles load testing instrumentaiton and do not involve:
* Adding new metrics
* Enancing the `AppGate Server` supplier selection strategy (which is needed to distribute the load across `RelayMiners`

## Issue

- #20 

## Type of change

Select one or more:

- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

## Sanity Checklist

- [x] I create and referenced any new tickets, if applicable


red-0ne:
It appears that there are some challenges w.r.t. using k6 to implement our load testing plans executions, with the most impeding ones being:

### Shared mutable state
There seems to be no way to share mutable state across different scenarios or VUs. This is helpful to:
* Inform the `sendRelay` scenario of the available gateways and applications  to use since those are changing throughout the test progression.
* Inform the `application`s about the `gateway`s to delegate to

### Unreliable plugins
The plugins needed to execute the staking actions do not behave as desired:
* We need the `avitalique/xk6-file` plugin to write staking configuration files before execution the stake command. It seems that this function **does not** block until the file is effectively written, leading to the `poktrolld tx <actor> stake-<actor>` command to fail because of a lacking config file. A sleep command was needed to ensure that the file was created.
* We need the `grafana/xk6-exec` plugin to execute `poktroll` and `ignite`. This plugin offer a `command` function that does not provide a way to catch a command error or to get the `stderr` output. Making the scenario relying probing the command's output, loops and `sleep`s
  
### New block/event info
There is no way to  (globally/not per iteration) listen for blocks or transactions. The `poktrolld tx` commands rely on blocks duration to estimate when a transaction has been committed on-chain and proceed with the next step.


### Lack of precise iteration count needed to stake the actors.
Only one execution plan (`per-vu-iterations`) seems to fit that use case, as we need a precise number contrary to other scenarios (`sendRelay`) where losing a few iterations is not as critical.

red-0ne:
Closing it in favor of #432

PR #439, opened 2024-03-19
[RelayMiner] feat: Make the RelayMiner wait for its corresponding Supllier to stake

red-0ne:
## Summary

Make the `RelayMiner` to loop, query and wait for its associated supplier to be staked before continuing the start process.

## Issue

When a `RelayMiner` starts and its associated `Supplier` is not staked yet, the former will exit due to the non staked suppler.

This is problematic for pre-provisioned `RelayMiner`s used for load testing.

image

## Type of change

Select one or more:

- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

## Sanity Checklist


pkg/relayer/proxy/server_builder.go
```diff
@@ -26,10 +28,30 @@ func (rp *relayerProxy) BuildProvidedServices(ctx context.Context) error {
...
+
+	// Wait for the supplier to be staked before building the relay servers.
+	for {
+		// Get the supplier's advertised information from the blockchain
+		supplier, err = rp.supplierQuerier.GetSupplier(ctx, supplierAddress.String())
+
+		// If the supplier is not found, wait for the supplier to be staked.
+		if err != nil && suppliertypes.ErrSupplierNotFound.Is(err) {
+			rp.logger.Info().Msgf(
+				"Waiting for the supplier with address %s to stake",
```
bryanchriswhite, line 41: Is it worth mentioning that the next check will be in 5 seconds? At which point maybe that value should get its own variable.

pkg/relayer/proxy/server_builder.go
```diff
@@ -26,10 +28,30 @@ func (rp *relayerProxy) BuildProvidedServices(ctx context.Context) error {
 		return err
 	}
 
-	// Get the supplier's advertised information from the blockchain
-	supplier, err := rp.supplierQuerier.GetSupplier(ctx, supplierAddress.String())
-	if err != nil {
-		return err
+	var supplier sharedtypes.Supplier
+
+	// Wait for the supplier to be staked before building the relay servers.
```
Olshansk, line 33: OPTIONAL NIT: What do you think of moving this business logic into a helper and updating the comment that it's a while loop needed to get back a supplier?

Olshansk (changes requested):

pkg/relayer/proxy/server_builder.go
```diff
@@ -107,3 +111,36 @@ func (rp *relayerProxy) initializeProxyServers(
...
 }
+
+// waitForSupplierToStake waits in a loop until it gets the on-chain supplier's
+// information back.
+func (rp *relayerProxy) waitForSupplierToStake(
+	ctx context.Context,
+	supplierAddress string,
+) (supplier sharedtypes.Supplier, err error) {
+	for {
+		// Get the supplier's advertised information from the blockchain
```
Olshansk, line 122:
```suggestion
		// Get the supplier's on-chain record
```

pkg/relayer/proxy/server_builder.go
```diff
@@ -107,3 +111,36 @@ func (rp *relayerProxy) initializeProxyServers(
 
 	return proxyServers, nil
 }
+
+// waitForSupplierToStake waits in a loop until it gets the on-chain supplier's
```
Olshansk, line 115: I think we really need to qualify that this is for testing or debugging purposes.

The analogy I think of is a bank account. If we were building a web2 banking sass business, this function would be "waitForUserToCreateBankAccount" before opening up their app. It makes sense, but would almost never be necessary since it's a one-time thing.

Olshansk (changes requested):

Olshansk (approved):

PR #442, opened 2024-03-19
Zero retryLimit Support in ReplayClient

ezeike:
## Summary

Enables the replay client to continuously attempt to reconnect to the pocket node. 

## Issue

Current implementation of the replay client will panic after 10 failed connection attempts. 

## Type of change

Select one or more:

- [X] New feature, functionality or library
- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

- [x] **DevNet E2E Tests**: Add the `devnet-test-e2e` label to the PR. **THIS IS VERY EXPENSIVE**, so only do it after all the reviews are complete.

## Sanity Checklist


pkg/client/events/replay_client.go
```diff
@@ -21,7 +21,7 @@ const (
 	// eventsBytesRetryLimit is the maximum number of times to attempt to
 	// re-establish the events query bytes subscription when the events bytes
 	// observable returns an error or closes.
-	eventsBytesRetryLimit        = 10
+	eventsBytesRetryLimit        = 0
```
h5law, line 24: Can we add a TODO to make this a customisable parameter in the appgateserver and relayminer config files. It would reqiire some refactoring to make it variable not a fixed value but could provide some benefits.

h5law (approved):
LGTM Just want you to add a TODO then good to go!

h5law (dismissed):
After a second look at https://github.com/pokt-network/poktroll/blob/e0f7978cd4de8ef5bb77929a1ef9c61ee7edc80d/pkg/client/events/replay_client.go#L198 you need to change the logic in the retry.OnError method as well as this or you will panic if the connection ever closes and needs to be reestablihed

pkg/retry/retry.go
```diff
@@ -58,7 +58,7 @@ func OnError(
 				return nil
 			}
 
-			if retryCount >= retryLimit {
+			if retryLimit > 0 && retryCount >= retryLimit {
```
Olshansk, line 61: I'm a bit confused. So if `retryLimit == 0`, it'll always keep retrying?

I have a follow-up question if that's the case, but wanted to understand the reasoning & approach here first.

Olshansk (changes requested):
@bryanchriswhite Added you as as a reviewer instead of @red-0ne since you have more context into the reply client and wanted to get your 👀 on this.

pkg/retry/retry.go
ezeike: Yes retryLimit == 0 will keep retrying.

Olshansk:
@bryanchriswhite Can you PTAL at @ezeike questions before you go OOO?

h5law:
@ezeike

> Under what conditions is errCh (in OnError) closed? If that does happen, can we count that as an error and retry the same as if it was an error that came through errCh?

The `errCh` in `retry.OnError` closes after we have reached the max number of retry attempts. `retry.OnError` essentially takes a work function that returns an error channel and does this until it errors then retries a certain number of times. See [1] for the work function that is supplied by the replay client. The two instances where it returns an error on the channel (causing a retry in `retry.OnError` are:

1. We were unable to subscribe to the events query client for some reason.
2. The events query client closed for some reason.

These may be due to network issues or some unexpected error or whatever but these are naturally possibilities. If either of these reasons do happen we are already treating them as errors but in order for the retry logic to work using a channel makes sense.

I am a little unsure what you mean by retry in the same way but treat them as errors - it sounds like that would just be the same thing to me?

> Do we need a retryLimit at all? At the moment it does not appear we do, is there a future scenario where we might?

The `retryLimit` is essentially there to stop an infinite loop occurring if say a connection to the full-node providing the on-chain data a gateway is faulty or the node goes offline.

By setting the `eventsBytesRetryLimit` to `^uint64(0)` and removing the `if retryLimit > 0` check you will get the same result - infinitely retry (essentially). As soon as you have a successful connection to an events query client for 10 seconds your `retryCount` resets so you can again fail and retry to connect over and over.

@ezeike Are you finding that your connections are failing a lot and you are reaching the 10 retry limit often? If so I think investigating what is causing that could help us figure out a better solution than an infinite loop.

[1] https://github.com/pokt-network/poktroll/blob/e0f7978cd4de8ef5bb77929a1ef9c61ee7edc80d/pkg/client/events/replay_client.go#L207

pkg/retry/retry.go
bryanchriswhite: What about using -1 for "infinity" instead of 0, otherwise, there's no way to specify an actual retry limit of 0?

ezeike:
> @ezeike
> 
> > Under what conditions is errCh (in OnError) closed? If that does happen, can we count that as an error and retry the same as if it was an error that came through errCh?
> 
> The `errCh` in `retry.OnError` closes after we have reached the max number of retry attempts. `retry.OnError` essentially takes a work function that returns an error channel and does this until it errors then retries a certain number of times. See [1] for the work function that is supplied by the replay client. The two instances where it returns an error on the channel (causing a retry in `retry.OnError` are:
> 
> 1. We were unable to subscribe to the events query client for some reason.
> 2. The events query client closed for some reason.
> 
> These may be due to network issues or some unexpected error or whatever but these are naturally possibilities. If either of these reasons do happen we are already treating them as errors but in order for the retry logic to work using a channel makes sense.
> 
> I am a little unsure what you mean by retry in the same way but treat them as errors - it sounds like that would just be the same thing to me?

Errors are received through the error channel and those trigger a Sleep and a retry. However when the error channel is closed OnError returns and the SDK panics. A first thought was just to eliminate that return so the execution continues on to the same Sleep and retry logic. workFn() is called every loop and errCh created anew anyway.

The current design is the SDK is to take care of connection details and do its best to ensure there's always a connection to a node. We should keep this as much as possible to keep using the SDK as simple as we can.

Either:

- SDK is modified to remove the panic and return an error to the gateway when retry limit reached. Then the gateway can make a new SDK and try again.
- SDK is modified to allow unlimited retries.

At the moment we should pick the quickest one that solves the issue so we can get our gateway on testnet as soon as possible. The second option was quicker so I went with that.

There are other discussions to have around who reconnects, how to support using multiple nodes and how configurable the SDK should be but for now we want an MVP ASAP.

> 
> > Do we need a retryLimit at all? At the moment it does not appear we do, is there a future scenario where we might?
> 
> The `retryLimit` is essentially there to stop an infinite loop occurring if say a connection to the full-node providing the on-chain data a gateway is faulty or the node goes offline.

From a gateway perspective we want an infinite loop. SDK only allows for specifying a single node and it should do its best to always be connected to that node. 

If we don't have an infinite loop the gateway will have to handle connection failures and re-create the SDK. That's reasonable too, but shifts some work to gateway devs.

> By setting the `eventsBytesRetryLimit` to `^uint64(0)` and removing the `if retryLimit > 0` check you will get the same result - infinitely retry (essentially). As soon as you have a successful connection to an events query client for 10 seconds your `retryCount` resets so you can again fail and retry to connect over and over.

If we do this `retryCount >= retryLimit` will be true on the first pass and return. We'd need another tweak.

> @ezeike Are you finding that your connections are failing a lot and you are reaching the 10 retry limit often? If so I think investigating what is causing that could help us figure out a better solution than an infinite loop.

I saw the error when I didn't have localnet running. Moving forward though we can't have the SDK panic when a node connection is lost or not available.

h5law:
I'll try keep this one shorter @ezeike.

> Errors are received through the error channel and those trigger a Sleep and a retry. However when the error channel is closed OnError returns and the SDK panics. A first thought was just to eliminate that return so the execution continues on to the same Sleep and retry logic. workFn() is called every loop and errCh created anew anyway.
> The current design is the SDK is to take care of connection details and do its best to ensure there's always a connection to a node. We should keep this as much as possible to keep using the SDK as simple as we can.
> Either:
>   SDK is modified to remove the panic and return an error to the gateway when retry limit reached. Then the gateway can make a new SDK and try again.
>   SDK is modified to allow unlimited retries.
>   At the moment we should pick the quickest one that solves the issue so we can get our gateway on testnet as soon as possible. The second option was quicker so I went with that.
> There are other discussions to have around who reconnects, how to support using multiple nodes and how configurable the SDK should be but for now we want an MVP ASAP.

Though about adding this

```go
if retryPanicErr := recover(); retryPanicErr != nil {
	retryErrorMsg := if retryPanicErr.Error()
	if strings.Contains(retryErrorMsg, ".goPublishEvents should never reach this spot:") {
		restartSDK() // or however you do this I am unfamiliar but rest is correct
	}
}
```

This means we can still have the limit - fixes your problem of needing and infinite loop as is very minimal code change quite simple infact.

> If we do this retryCount >= retryLimit will be true on the first pass and return. We'd need another tweak.

If we don't go with the above solution we will count down instead of count up if that makes sense as we will use the max uint64 value counting down is the only real option - again minimal changes.

ezeike:
IMO the proper approach here is to refactor replay client to remove the panics that are present in the package.

As a consequence there is now no way to limit the number of connection attempts the replay client makes.

Ideally we give an error channel to replay client to signal to the caller that it is giving up but that is a larger PR and would require more complexity in our gateway.

We (the backend team) have decided we'd like to the panics and move forward with testing as-is. We can re-evaluate the error channel at a later date if the need becomes apparent.

If we were to implement an error channel (and a config option for retryLimit) we'd need to modify these three clients:

BlockClient, DelegationClient, TxClient 

their associated supplier functions, and both gateways (appgate and grove gateway) to handle the config and extra state management.

pkg/client/events/replay_client.go
```diff
@@ -285,10 +294,8 @@ func (rClient *replayClient[T]) newMapEventsBytesToTFn(errCh chan<- error) func(
 				return *new(T), true
 			}
 
-			panic(fmt.Sprintf(
-				"unexpected error deserialising event: %v; eventBz: %s",
-				err, string(eventBz),
-			))
+			logger.Error().Err(err).Msgf("unexpected error deserialising event")
```
ezeike, line 297: How can we handle this error without a panic?

pkg/client/events/replay_client.go
```diff
@@ -21,6 +21,7 @@ const (
 	// eventsBytesRetryLimit is the maximum number of times to attempt to
 	// re-establish the events query bytes subscription when the events bytes
 	// observable returns an error or closes.
+	// TODO to make this a customisable parameter in the appgateserver and relayminer config files.
```
Olshansk, line 24:
```suggestion
	// TODO_TECHDEBT: to make this a customizable parameter in the appgateserver and relayminer config files.
```

pkg/client/events/replay_client.go
```diff
@@ -117,7 +118,16 @@ func NewEventsReplayClient[T any](
 	}
 
 	// Concurrently publish events to the observable emitted by replayObsCache.
-	go rClient.goPublishEvents(ctx)
+	go func() {
```
Olshansk, line 121: How confident are you that this is the correct approach?

Going to delegate to @bryanchriswhite who has more context, but reply client was designed and intended to use `publishEvents` directly.

pkg/client/events/replay_client.go
Olshansk: Can you:

1. Explain what causes the panic?
2. Provide an example of how it can be reproduced?

Olshansk (changes requested):

pkg/client/events/replay_client.go
ezeike: I don't believe it is, the correct approach is with returning an error channel of some sort so SDK consumer can handle it. This was a compromise for short term solution with minimal changes.

pkg/client/events/replay_client.go
ezeike: With the current code base nothing causes this panic but a future change could introduce one that would cause our gateway to panic and shutdown, or continually panic and shut down. It's a possibility I would like to look at removing, an error deserializing an event shouldn't shut down our gateway.

pkg/client/events/options.go
```diff
@@ -9,3 +9,18 @@ func WithDialer(dialer client.Dialer) client.EventsQueryClientOption {
...
 	}
 }
+
+// WithConnRetryLimit returns an option function which sets the number
+// of times the replay client should retry in the event that it encounters
+// an error or its connection is interrupted.
+// If connRetryLimit is < 0, it will retry indefinitely.
+func WithConnRetryLimit[T any](limit int) client.EventsReplayClientOption[T] {
+	return func(client client.EventsReplayClient[T]) {
+		// Ignore the zero value because limit may be provided via a partially
```
Olshansk, line 19: @bryanchriswhite When you work on updating one governance parameter at a time, taking this pattern into consideration will be critical!

pkg/retry/retry_test.go
```diff
@@ -335,3 +335,100 @@ func TestOnError_RetryCountResetTimeout(t *testing.T) {
 		require.Contains(t, line, expectedPrefix)
 	}
 }
+
+// assert that a negative retry limit continually calls workFn
+func TestOnError_NegativeRetryLimit(t *testing.T) {
+	t.Skip("TODO_TECHDEBT: this test should pass but contains a race condition around the logOutput buffer")
```
Olshansk, line 341: @bryanchriswhite @ezeike So is the decision to fix this or merge it in as is?

Olshansk (changes requested):
A couple small comments / questions but also need to validate w/ E2E tests. The more recent [pull request template](https://github.com/pokt-network/poktroll/blob/main/.github/pull_request_template.md) as more details.

pkg/retry/retry_test.go
ezeike: Merge in as is. Tests confirmed passing locally without the -race flag.

bryanchriswhite: Let's skip any tests which will fail with the `-race` flag present as we should be using it in CI and in general to mitigate adding race conditions.

Olshansk: @bryanchriswhite I know you participated in contributing this PR, but could you also review and approve it assuming it looked good to you?

@ezeike Could you please review [1] to run E2E tests.

[1] https://github.com/pokt-network/poktroll/blob/main/.github/pull_request_template.md

pkg/client/interface.go
```diff
@@ -229,6 +229,15 @@ type TxClientOption func(TxClient)
 // SupplierClientOption defines a function type that modifies the SupplierClient.
 type SupplierClientOption func(SupplierClient)
 
+// DelegationClientOption defines a function type that modifies the BlockClient.
```
bryanchriswhite, line 232:
```suggestion
// DelegationClientOption defines a function type that modifies the DelegationClient.
```

pkg/retry/retry_test.go
```diff
@@ -332,6 +339,105 @@ func TestOnError_RetryCountResetTimeout(t *testing.T) {
...
+				time.Sleep(retryResetTimeout)
+				errCh <- testErr
+			}()
+		} else {
+			errCh <- testErr
+		}
+
+		// Increment the invocation count atomically
+		atomic.AddInt32(&testFnCallCount, 1)
+		return errCh
```
bryanchriswhite, line 386:
```suggestion
		// Increment the invocation count atomically
		count := atomic.AddInt32(&testFnCallCount, 1) - 1
		if count == int32(retryLimit) {
			go func() {
				time.Sleep(retryResetTimeout)
				errCh <- testErr
			}()
		} else {
			errCh <- testErr
		}
		return errCh
```

bryanchriswhite (approved):
**Preemptively approving**: one last, small suggestion: https://github.com/pokt-network/poktroll/pull/442#discussion_r1576500971

pkg/retry/retry_test.go
```diff
@@ -332,6 +339,103 @@ func TestOnError_RetryCountResetTimeout(t *testing.T) {
...
 	)
 	for _, line := range logOutputLines {
-		require.Contains(t, line, expectedPrefix)
+		require.Contains(t, line, expectedErrMsg)
+	}
+}
+
+// assert that a negative retry limit continually calls workFn
+func TestOnError_NegativeRetryLimit(t *testing.T) {
+	t.Skip("TODO_TECHDEBT: this test should pass but contains a race condition around the logOutput buffer")
```
Olshansk, line 348:
```suggestion
	t.Skip("TODO_TECHDEBT(@bryanchriswhite): this test should pass but contains a race condition around the logOutput buffer")
```

Olshansk (changes requested):
@ezeike

1. Please merge with main
2. See bryan's comment
3. Make sure E2E tests pass on DevNet

Will approve after ☝️

bryanchriswhite:
I am unable to reproduce the failing E2E test locally. :thinking:

Olshansk:
@ezeike @bryanchriswhite Could one of you please tend to the remaining `TODO_IN_THIS_` comments?

Olshansk (approved):
Approving now that I see E2E tests passing in https://github.com/pokt-network/poktroll/pull/442/commits/0693ca19c6abb48120515880fd04350d6ea568f9

Issue #440, opened 2024-03-19
[SDK] Method to Obtain Latest Block Prior to a Block Event


ezeike:
## Objective

Modify the SDK to fetch the latest block upon connection to the pocket node.

## Origin Document

In order for gateways to send relays the SDK requires the latest block height. Currently the latest block height is only available once a new block has been created and the SDK receives it through the block client.

The effect is gateways cannot send relays until a new block has been created.

## Goals

- Enable the SDK to obtain block data upon connection, rather than solely waiting for a new block.

## Deliverables

- [ ] PR containing enhancement to fetch block data on initial connection to node.

## General deliverables

- [ ] **Comments**: Add/update TODOs and comments alongside the source code so it is easier to follow.
- [ ] **Testing**: Add new tests (unit and/or E2E) to the test suite.
- [ ] **Documentation**: Update architectural or development READMEs; use [mermaid](https://mermaid-js.github.io/mermaid/) diagrams where appropriate.

---

**Creator**: @ezeike 
**Co-Owners**: @red-0ne

ezeike:
![image](https://github.com/pokt-network/poktroll/assets/12902351/0a417b46-5d10-41ed-8d61-c068fb75a30f)

Olshansk:
@ezeike Reminder to also update the GitHub Issue metadata. I updated it on this one to streamline things.

![Screenshot 2024-03-19 at 8 07 53 PM](https://github.com/pokt-network/poktroll/assets/1892194/8fcffe7e-d8aa-45bd-816b-c7096ff8b84b)

Olshansk:
Needed for gateways when block times become large. Another ~ 4-6 hours of work before finishing.

okdas:
@red-0ne Can we also add a metric (gauge) - something like `blockchain_height`, and we update this value whenever we get information about the new block on appgateserver and relayminer?

That way, we will know what block there is from our off-chain actors' Point of View. E.g. if our off-chain actor connects to the outdated full node or something - the operator could set an alert on that and troubleshoot the issue.

Olshansk:
Super awesome idea @okdas. + 💯

Issue #425, opened 2024-03-12
[Tests] Flaky TxClient tests


red-0ne:
## Objective

Have a non-flaky `TxClient_SignAndBroadcast_Succeeds` that doesn't arbitrarily fail.

Running `make itest 5 10 ./pkg/client/tx/... -- -run TxClient_SignAndBroadcast_Succeeds` should have a deterministic outcome.

## Origin Document

This specific test is arbitrarily failing making unit and CI tests failing too often with false positives and requiring us to run them again.

![image](https://github.com/pokt-network/poktroll/assets/231488/fa2bc74f-99d6-4d62-b802-87986af20525)

## Goals

- Restructure the `TxClient_SignAndBroadcast_Succeeds` test and make it have a deterministic outcome
-  Confirm whether the issue is with the test itself or with the code being testsd.

## Deliverables

- [ ] A single PR fixing the `TxClient_SignAndBroadcast_Succeeds` test on `pkg/client/tx/client_test.go`.
- [ ] Add comments explaining why it's happening and how it is solved.

## Non-goals / Non-deliverables

- Refactor the tested code
- Complete rewrite of the test

## General deliverables

- [ ] **Comments**: Add/update TODOs and comments alongside the source code so it is easier to follow.
- [ ] **Testing**: Add new tests (unit and/or E2E) to the test suite.
- [ ] **Documentation**: Update architectural or development READMEs; use [mermaid](https://mermaid-js.github.io/mermaid/) diagrams where appropriate.

---

**Creator**: @red-0ne 
**Co-Owners**: @Olshansk @bryanchriswhite

red-0ne:
My investigation led me to think that the issue is that `err, ok := <-errCh` in [pkg/client/tx/client_test.go#L137](https://github.com/pokt-network/poktroll/blob/feat/proof-validation/pkg/client/tx/client_test.go#L137) fails to emit making the test to timeout.

This happens because `txResultsBzPublishCh <- either.Success[[]byte](rpcResultBz)` in [pkg/client/tx/client_test.go#L132](https://github.com/pokt-network/poktroll/blob/feat/proof-validation/pkg/client/tx/client_test.go#L132) is blocked or late to receive its value, which in turn blocks closing the `txErrCh` in [pkg/client/tx/client.go#L423](https://github.com/pokt-network/poktroll/blob/main/pkg/client/tx/client.go#L423)

Olshansk:
[...]

PR #444, opened 2024-03-25
[BlockClient] Refactor BlockClient to fetch latest block at init

red-0ne:
## Summary

This PR introduces a cometClient that fetches the latest block when the BlockClient is initialized, enabling it to acquire knowledge about the last committed block immediately upon initialization.

The purpose of this is to prevent the need to wait for a new block to be committed before obtaining this information. This is necessary because the WebSocket connection only notifies about blocks committed after the subscription is made.

## Issue

In non-localnet environments, block times will be significant, making it necessary to obtain this block information before a new block is committed.

image

- #440 
- #425 

## Type of change

Select one or more:

- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

## Sanity Checklist


pkg/client/block/client.go
```diff
@@ -58,20 +73,78 @@ type blockClient struct {
...
+	}
+
+	initialBlock := &cometBlockEvent{}
+	initialBlock.Data.Value.Block = queryBlockResult.Block
+	initialBlock.Data.Value.BlockID = queryBlockResult.BlockID
+
+	b.latestBlockMu.Lock()
+	defer b.latestBlockMu.Unlock()
+
+	if b.latestBlock != nil {
```
ezeike, line 138: This test needs to be ==

pkg/sdk/deps_builder.go
```diff
@@ -31,12 +32,17 @@ func (sdk *poktrollSDK) buildDeps(
 	// Supply the logger
 	deps = depinject.Configs(deps, depinject.Supply(polylog.Ctx(ctx)))
 
+	cosmosClient, err := cosmosclient.NewClientFromNode(config.QueryNodeUrl.Host)
```
ezeike, line 35:
```suggestion
	cosmosClient, err := cosmosclient.NewClientFromNode(config.QueryNodeUrl.String())
```

If only .Host is passed the cosmosclient will default the scheme to http

pkg/client/block/client.go
red-0ne: Good catch

pkg/client/block/client.go
```diff
@@ -58,20 +73,80 @@ type blockClient struct {
...
-// LatestsNBlocks returns the last n blocks observed by the BockClient.
+// LastNBlocks returns the last n blocks observed by the BlockClient.
 func (b *blockClient) LastNBlocks(ctx context.Context, n int) []client.Block {
 	return b.eventsReplayClient.LastNEvents(ctx, n)
 }
 
+// LastBlock returns the last blocks observed by the BlockClient.
+func (b *blockClient) LastBlock() client.Block {
+	b.latestBlockMu.Lock()
+	defer b.latestBlockMu.Unlock()
```
ezeike, line 98: This lock doesn't protect b.latestBlock against concurrent access.

We can either:
- Copy the block before return
- Since the return value is only used for getting the height and hash, we could maintain those as fields in the block client

testutil/testclient/testblock/client.go
```diff
@@ -104,6 +104,7 @@ func NewAnyTimeLastNBlocksBlockClient(
 	// returns the mock block.
 	blockClientMock := mockclient.NewMockBlockClient(ctrl)
 	blockClientMock.EXPECT().LastNBlocks(gomock.Any(), gomock.Any()).Return([]client.Block{blockMock}).AnyTimes()
```
Olshansk, line 106: Can we remove this given that we added the one below or do we need both?

pkg/sdk/deps_builder.go
```diff
@@ -31,12 +32,17 @@ func (sdk *poktrollSDK) buildDeps(
 	// Supply the logger
 	deps = depinject.Configs(deps, depinject.Supply(polylog.Ctx(ctx)))
 
+	cosmosClient, err := cosmosclient.NewClientFromNode(config.QueryNodeUrl.String())
```
Olshansk, line 35: We call it `cometClient` in `NewBlockClient` and `cosmosClient` here.

Can you pick the one correct name and use the same thing in both places?

pkg/deps/config/suppliers.go
```diff
@@ -77,14 +77,22 @@ func NewSupplyEventsQueryClientFn(queryNodeRPCURL *url.URL) SupplierFn {
...
+func NewSupplyBlockClientFn(queryNodeRPCURL *url.URL) SupplierFn {
 	return func(
 		ctx context.Context,
 		deps depinject.Config,
 		_ *cobra.Command,
 	) (depinject.Config, error) {
+
+		// Create a cosmos client from the queryNodeRPCURL used by the block client
+		// to initialize the block client by querying the latest block.
+		cosmosClient, err := cosmosclient.NewClientFromNode(queryNodeRPCURL.String())
```
Olshansk, line 89: Ditto about naming of `cosmosClient` vs `cometClient` vs `cometBlockClient`. Pick one and be consistent

pkg/client/interface.go
```diff
@@ -146,6 +146,8 @@ type BlockClient interface {
 	// LastNBlocks returns the latest N blocks that have been committed to
 	// the chain.
 	LastNBlocks(context.Context, int) []Block
+	// LastBlock returns the latest block that have been committed to the chain.
```
Olshansk, line 149:
```suggestion
	// LastBlock returns the latest block that has been committed on-chain.
```

pkg/client/block/client.go
```diff
@@ -46,7 +49,19 @@ func NewBlockClient(
 	if err != nil {
 		return nil, err
 	}
-	return &blockClient{eventsReplayClient: client}, nil
+
+	bClient := &blockClient{
```
Olshansk, line 53: If we rename `blockClient` to `blockReplayClient`, we can name the local variable `blockClient`

pkg/client/block/client.go
```diff
@@ -58,20 +73,80 @@ type blockClient struct {
...
+	return b.latestBlock
+}
+
 // Close closes the underlying websocket connection for the EventsQueryClient
 // and closes all downstream connections.
 func (b *blockClient) Close() {
 	b.eventsReplayClient.Close()
 }
+
+// getLatestBlocks listens to the EventsReplayClient's EventsSequence and updates
```
Olshansk, line 108: Shold we rename to `goListenForLatestBlock`

1. It's just one block, not `s`
2. It's intended to run as a goroutine. Make sure to update the comments
3. It doesn't just get, it constantly listens for it.

pkg/client/block/client.go
```diff
@@ -58,20 +73,80 @@ type blockClient struct {
...
+	latestBlockCh := b.eventsReplayClient.EventsSequence(ctx).Subscribe(ctx).Ch()
+	for block := range latestBlockCh {
+		b.latestBlockMu.Lock()
+		b.latestBlock = block
+		b.latestBlockMu.Unlock()
+	}
+}
+
+// getInitialBlock fetches the initial block from the chain and sets it as the
+// latest block observed by the blockClient.
```
Olshansk, line 120:
```suggestion
// getInitialBlock fetches the latest committed on-chain block at the time that the client starts up.
// It updates the internal state of the blockClient for future retrieval.
```

pkg/client/block/client.go
```diff
@@ -58,20 +73,80 @@ type blockClient struct {
...
+// This is necessary to ensure that the blockClient has the latest block when
```
Olshansk, line 121:
```suggestion
// This is necessary to ensure that the most recent block is available to the blockClient when
```

pkg/client/block/client.go
```diff
@@ -58,20 +73,80 @@ type blockClient struct {
...
+// it is first created.
+// If b.latestBlock is already set by the getLatestBlocks goroutine, then this
+// function skips setting the initial block.
```
Olshansk, line 124: Why?

pkg/client/block/client.go
```diff
@@ -58,20 +73,80 @@ type blockClient struct {
...
+	initialBlock := &cometBlockEvent{}
+	initialBlock.Data.Value.Block = queryBlockResult.Block
+	initialBlock.Data.Value.BlockID = queryBlockResult.BlockID
+
+	b.latestBlockMu.Lock()
+	defer b.latestBlockMu.Unlock()
+
+	// If the latest block is not already set by the getLatestBlocks goroutine,
+	// then set it to the initial block.
+	if b.latestBlock == nil {
```
Olshansk, line 140: Should we not make the first line:

`if b.latestBlock != nil: return`?

pkg/client/block/client.go
```diff
@@ -58,20 +73,80 @@ type blockClient struct {
...
+		b.latestBlock = initialBlock
+	}
+
+	return nil
+}
+
+// cometBlockClient is an interface that defines the Block method for the comet
+// client. This is used to mock the comet client in the block package.
+
+type cometBlockClient interface {
```
Olshansk, line 150: Why don't we just use `CometRPC` from `client/cometbft.go` directly?

pkg/client/block/client.go
```diff
@@ -58,20 +73,80 @@ type blockClient struct {
...
+	// If the latest block is not already set by the getLatestBlocks goroutine,
+	// then set it to the initial block.
+	if b.latestBlock == nil {
+		b.latestBlock = initialBlock
+	}
+
+	return nil
+}
+
+// cometBlockClient is an interface that defines the Block method for the comet
```
Olshansk, line 147: 1. "hiding" interfaces like this at the bottom of a file isn't great for discoverability.
2. Putting (1) aside, I think we can use `CometRPC` directly, no?

pkg/client/block/client_test.go
```diff
@@ -139,3 +152,18 @@ type testBlockEventValueStruct struct {
 	Block   *types.Block  `json:"block"`
 	BlockID types.BlockID `json:"block_id"`
 }
+
+type testCometClient struct {
```
Olshansk, line 156: If we use `clitestutil "github.com/cosmos/cosmos-sdk/testutil/cli"` and not create a new interface, we can just use `clitestutil.NewMockCometRPC` directly and avoid the creation of new types.

Olshansk (changes requested):

bryanchriswhite (changes requested):
Thanks @red-0ne - nice one! :raised_hands: 

I collected my feedback into a branch & opened it as #453; PTAL.

testutil/testclient/testblock/client.go
red-0ne: We are not using `LastNBlocks` anywhere, I think it's safe to remove.

pkg/client/block/client.go
red-0ne: If the subscription updates the latest block before the initial query, then there is no need to re-assign it again, since the subscription may come with a newer block (if the client comes with a late reply)

red-0ne:
Merged @bryanchriswhite's PR #453 changes.

pkg/client/block/client.go
```diff
@@ -46,32 +50,142 @@ func NewBlockClient(
...
+		return nil, err
+	}
+
+	blockClient.forEachBlockEvent(ctx)
+
+	if err := blockClient.getInitialBlock(ctx); err != nil {
+		return nil, err
+	}
+
+	return blockClient, nil
```
Olshansk, line 74:
```suggestion
	return blockClient.getInitialBlock(ctx)
```

pkg/client/block/client.go
```diff
@@ -46,32 +50,142 @@ func NewBlockClient(
...
 
-// blockClient is a wrapper around an EventsReplayClient that implements the
+// blockReplayClient is a wrapper around an EventsReplayClient that implements the
 // BlockClient interface for use with cosmos-sdk networks.
-type blockClient struct {
+type blockReplayClient struct {
+	// queryClient is the RPC client that is used to query for the initial block
+	// upon blockClient construction. The result of this query is only used if it
+	// returns before the eventsReplayClient receives its first event.
+	queryClient cometclient.CometRPC
```
Olshansk, line 83: Should we rename this to `oneTimeUseQueryClient` or `onStartQueryClient` to be ultra explicit?

pkg/client/block/client.go
```diff
@@ -46,32 +50,142 @@ func NewBlockClient(
...
+		}
+
+		blockResult := cometBlockResult(*queryBlockResult)
+		blockQueryResultCh <- &blockResult
+	}()
+
+	return errCh
+}
+
+// cometBlockResult is a non-alias of the comet ResultBlock type that implements the client.Block interface.
```
Olshansk, line 179: #PUC on why we need it

pkg/client/block/client.go
```diff
@@ -46,32 +50,142 @@ func NewBlockClient(
...
+// TODO_DISCUSS_IN_THIS_COMMIT: should this be moved somewhere else?
```
Olshansk, line 180: Let's move it into its on file

pkg/client/block/client.go
```diff
@@ -46,32 +50,142 @@ func NewBlockClient(
...
+type cometBlockResult coretypes.ResultBlock
+
+func (cbr *cometBlockResult) Height() int64 {
+	return cbr.Block.Height
+}
+
+// TODO_CONSIDERATION: should this return an array instead of a slice
```
Olshansk, line 187: Remove. It's fine

Olshansk (changes requested):
Beginning to look very simple!

pkg/client/block/block_result.go
```diff
@@ -0,0 +1,18 @@
+package block
+
+import (
+	coretypes "github.com/cometbft/cometbft/rpc/core/types"
+)
+
+// cometBlockResult is a non-alias of the comet ResultBlock type that implements
+// the client.Block interface used across the code base which unifies the access
```
Olshansk, line 8:
```suggestion
// the client.Block interface used across the codebase which unifies the access
```

pkg/client/block/block_result.go
```diff
@@ -0,0 +1,18 @@
...
+// to height and hash of a block across different block clients.
```
Olshansk, line 9:
```suggestion
// to a block's height and hash across different block clients.
```

pkg/client/block/client.go
```diff
@@ -36,42 +39,147 @@ func NewBlockClient(
...
 	}
-	return &blockClient{eventsReplayClient: client}, nil
+
+	blockClient.forEachBlockEvent(ctx, latestBlockPublishCh)
+
+	if err := blockClient.getInitialBlock(ctx, latestBlockPublishCh); err != nil {
+		return nil, err
+	}
+
+	return blockClient, nil
```
Olshansk, line 75:
```suggestion
	return blockClient.getInitialBlock(ctx, latestBlockPublishCh)
```

pkg/client/block/client.go
```diff
@@ -36,42 +39,147 @@ func NewBlockClient(
...
 }
 
-// blockClient is a wrapper around an EventsReplayClient that implements the
+// blockReplayClient is a wrapper around an EventsReplayClient that implements the
 // BlockClient interface for use with cosmos-sdk networks.
-type blockClient struct {
+type blockReplayClient struct {
+	// onStartQueryClient is the RPC client that is used to query for the initial block
+	// upon blockClient construction. The result of this query is only used if it
```
Olshansk, line 82:
```suggestion
	// upon blockReplayClient construction. The result of this query is only used if it
```

pkg/client/block/client.go
```diff
@@ -36,42 +39,147 @@ func NewBlockClient(
...
+	// ReplayObservable#Last() is guaranteed to return at least one element.
+	return b.latestBlockReplayObs.Last(ctx, 1)[0]
 }
 
 // Close closes the underlying websocket connection for the EventsQueryClient
 // and closes all downstream connections.
-func (b *blockClient) Close() {
+func (b *blockReplayClient) Close() {
 	b.eventsReplayClient.Close()
+	//close(b.latestBlockPublishCh)
```
Olshansk, line 117:
```suggestion
```

pkg/client/block/client.go
```diff
@@ -36,42 +39,147 @@ func NewBlockClient(
...
+	b.close()
+}
+
+// forEachBlockEvent asynchronously observes block event notifications from the
+// EventsReplayClient's EventsSequence observable & publishes each to latestBlockPublishCh.
+func (b *blockReplayClient) forEachBlockEvent(
```
Olshansk, line 123: Wdyt of this instead?

```suggestion
func (b *blockReplayClient) asyncForwardBlockEvent(
```

pkg/client/block/client.go
```diff
@@ -36,42 +39,147 @@ func NewBlockClient(
...
+
+	// Query the latest block asynchronously.
+	queryErrCh := b.queryLatestBlock(ctx, blockQueryResultCh)
+
+	// Wait for either the latest block query response, error, or the first block
+	// event to arrive & use whichever occurs first or return an error.
+	//
+	// NB: #latestBlockReplayObs is a proxy for the events sequence observable
+	// because it is guaranteed to be notified on block events in #goForEachBLockEvent().
+	var initialBlock client.Block
```
Olshansk, line 153: 1. goForEachBLockEvent no longer ecists
2. This comment is adding confusion rather than clarity for me

```suggestion
	var initialBlock client.Block
```

pkg/client/block/client.go
```diff
@@ -36,42 +39,147 @@ func NewBlockClient(
...
+	case err := <-queryErrCh:
+		return err
+	}
+
+	// Publish the fastest result as the initial block.
+	latestBlockPublishCh <- initialBlock
+	return nil
+}
+
+// queryLatestBlock constructs a comet RPC block client & asynchronously queries for
```
Olshansk, line 167: `comet RPC block client` Is this still true?

pkg/client/block/client.go
```diff
@@ -36,42 +39,147 @@ func NewBlockClient(
...
+// the latest block. It returns an error channel which may be sent a block query error.
+// It is *NOT* intended to be called in a goroutine.
+func (b *blockReplayClient) queryLatestBlock(ctx context.Context, blockQueryResultCh chan<- client.Block) <-chan error {
```
Olshansk, line 170: I think this is an example where named return params will actually make it easier and you can remove the make.

```suggestion
func (b *blockReplayClient) queryLatestBlock(
	ctx context.Context, blockQueryResultCh chan<- client.Block,
) (errCh <-chan error) {
```

Olshansk (changes requested):

pkg/client/block/client.go
red-0ne: We need to return a `blockClient, err` tuple

pkg/client/block/client.go
red-0ne: The created `errCh` is a send/receive, while a receive-only channel is returned.

pkg/client/block/client.go
```diff
@@ -36,42 +39,148 @@ func NewBlockClient(
...
 	// These enable the EventsReplayClient to correctly map the raw event bytes
 	// to Block objects and to correctly return a BlockReplayObservable
 	eventsReplayClient client.EventsReplayClient[client.Block]
+
+	// latestBlockReplayObs is a replay observable that combines blocks observed by
+	// the block query client & the events replay client. It is the "canonical"
+	// source of block notifications for blockClient.
+	latestBlockReplayObs observable.ReplayObservable[client.Block]
+
+	// close is a function that cancels the context of the blockClient.
```
Olshansk, line 98: s/blockClient/blockReplayClient

Please make sure there are no other lingering instances of this as well.

Olshansk (changes requested):

testutil/testclient/testeventsquery/client.go
```diff
@@ -43,9 +43,10 @@ func NewOneTimeEventsQuery(
 ) *mockclient.MockEventsQueryClient {
 	t.Helper()
 	ctrl := gomock.NewController(t)
+	ctx, _ = context.WithCancel(ctx)
```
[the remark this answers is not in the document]
red-0ne: `cancelCtx` feels like the cancel function itself. Will rename to `cancellableCtx`

pkg/observable/channel/replay.go
```diff
@@ -44,208 +36,145 @@ func NewReplayObservable[V any](
...
-	ctx, cancel := context.WithCancel(ctx)
-	removeAndCancel := func(toRemove observable.Observer[V]) {
-		ro.observerManager.remove(toRemove)
-		cancel()
+	// If the replay buffer does not have enough values, wait for the source observable
+	// to emit enough values to satisfy the request.
+	ch := ro.bufferingObsvbl.Subscribe(ctx).Ch()
+	// Initialize latestValues with the values in the replay buffer in case the
+	// source observable is closed or the context is canceled before it has a chance
+	// to emit any values.
```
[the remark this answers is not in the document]
red-0ne: This is about the case where the observable never emits in between the time of subscription and cancellation. I'll reword it to the following:
```
before it has a chance to emit an updated buffer of values.
```

pkg/observable/channel/replay_test.go
```diff
@@ -107,9 +159,11 @@ func TestReplayObservable_Last_Full_ReplayBuffer(t *testing.T) {
 			name:             "n < replayBufferSize",
 			replayBufferSize: 5,
 			lastN:            3,
-			// the replay buffer is not full so Last should return values
-			// starting from the first published value.
-			expectedValues: values[:3], // []int{1, 2, 3},
+			// the replay buffer has enough values to return to Last, it should return
+			// the last n values in the replay buffer.
+			// TODO_DISCUSS: Shouldn't Last order the values by having the most recent
```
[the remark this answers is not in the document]
red-0ne: This was a forgotten comment before the change about having LIFO. Will update the results in the comments.

And remove the TODO_DISCUSS.

pkg/observable/channel/replay.go
```diff
@@ -44,208 +36,145 @@ func NewReplayObservable[V any](
...
+// It replays the values stored in the replay buffer before emitting new values.
+func (ro *replayObservable[V]) Subscribe(ctx context.Context) observable.Observer[V] {
+	obs, ch := NewObservable[V]()
+	ctx, cancel := context.WithCancel(ctx)
 
-	ro.observerManager.add(observer)
+	go func() {
+		// Replay the values stored in the buffer form the oldest to the newest.
+		ro.replayBufferMu.RLock()
+		for i := len(ro.replayBuffer) - 1; i >= 0; i-- {
```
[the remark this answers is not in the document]
red-0ne: `Subscribe` needs to emit in `FIFO` as it represents the emission of values as they come.

`Last` needs to return the last values in `LIFO`.

The choice of storing the **buffer** with `LIFO` was only motivated by the simplicity of `Last` method implementation.

testutil/testclient/testeventsquery/client.go
```diff
@@ -35,18 +35,18 @@ func NewLocalnetClient(t *testing.T, opts ...client.EventsQueryClientOption) cli
 // of which is set to the publish channel of the events bytes observable publish
 // channel.
 func NewOneTimeEventsQuery(
-	ctx context.Context,
+	cancellableCtx context.Context,
```
Olshansk, line 38: Doesn't it become a cancellableCtx but the provided one a plain context?

pkg/observable/interface.go
```diff
@@ -11,7 +11,8 @@ import "context"
 // to new observers, before publishing new values to observers.
 type ReplayObservable[V any] interface {
 	Observable[V]
-	// Last synchronously returns the last n values from the replay buffer.
+	// Last synchronously returns the last n values from the replay buffer ordered
+	// from the most recent to the oldest.
```
Olshansk, line 15: Can you call out explicitly that this is LIFO.

Similarly, can you call out explicitly that subscribe is FIFO.

ENginers will understand.

pkg/observable/channel/replay.go
```diff
@@ -61,8 +61,9 @@ func ToReplayObservable[V any](
 
 // Last synchronously returns the last n values from the replay buffer.
 // It blocks until n values have been accumulated or its context is canceled,
-// in which case it returns the values accumulated so far.
-// The values returned are ordered from newest to oldest.
+// If it is cancelled before n values are accumulated, it returns all the available
```
Olshansk, line 64:
```suggestion
// If it is canceled before n values are accumulated, it returns all the available
```

Olshansk (approved):
A couple small #PUC comments but feel free to merge after.

PR #450, opened 2024-03-28
[RelayMiner] Handle gzip compression

okdas:
## Summary

MissKitty42 ran into an issue with RelayMiner. Whenever AppGate Server sends relay to their RelayMiner, the response is an unexpected binary sequence instead of json.

This appears to be related to gzip compression: https://discord.com/channels/553741558869131266/1217918754801975398/1222944716828704898

## Type of change

Select one or more:

- [x] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

## Sanity Checklist


pkg/relayer/proxy/proxy_test.go
```diff
@@ -516,6 +518,66 @@ func TestRelayerProxy_Relays(t *testing.T) {
 			expectedErrCode: -32000,
 			expectedErrMsg:  "session expired", // Relay rejected by the supplier
 		},
+		// How do I do this? :D
```
okdas, line 521: @red-0ne I'm not familiar with relay miner testing suite, what's the best way to test gzip compressed response?

pkg/relayer/proxy/relay_builders.go
```diff
@@ -10,42 +10,45 @@ import (
 
 // newRelayRequest builds a RelayRequest from an http.Request.
 func (sync *synchronousRPCServer) newRelayRequest(request *http.Request) (*types.RelayRequest, error) {
-	requestBz, err := io.ReadAll(request.Body)
+	requestBody, err := io.ReadAll(request.Body)
 	if err != nil {
 		return nil, err
 	}
+	defer request.Body.Close()
```
okdas, line 17: @red-0ne Claude 3 mentioned `Body.Close()` never happens, and I also couldn't find that in the code, so I added it here. Could it be a reason for a goroutine leak? Though the tests seem to pass, I want to check in with you to make sure we're not closing the body twice.

pkg/relayer/proxy/proxy_test.go
okdas: Nvm, we decided test will be added in a subsequent pr.

pkg/relayer/proxy/relay_builders.go
Olshansk: It seems dangerous to me to to a `.Close` inside of a helper where the request gets passed in.

I feel like we should do it in `ServeHTTP` inside of `pkg/appgateserver/server.go`.

pkg/relayer/proxy/relay_builders.go
```diff
@@ -10,42 +10,45 @@ import (
...
+// newRelayResponse builds a RelayResponse from a response body reader and a SessionHeader.
 // It also signs the RelayResponse and assigns it to RelayResponse.Meta.SupplierSignature.
-// The response's Body is passed directly into the RelayResponse.Payload field.
+// The response body is passed directly into the RelayResponse.Payload field.
 func (sync *synchronousRPCServer) newRelayResponse(
-	response *http.Response,
+	responseBody io.ReadCloser,
 	sessionHeader *sessiontypes.SessionHeader,
 ) (*types.RelayResponse, error) {
+	defer responseBody.Close()
```
Olshansk, line 37: Ditto. I don't think we shold be closing things inside of helpers. It creates side effects.

pkg/relayer/proxy/synchronous.go
```diff
@@ -249,14 +250,28 @@ func (sync *synchronousRPCServer) serveHTTP(
...
 	}
 
+	// Check if the response is gzip-compressed
+	// TODO: Add other algorithms, or an alternative would be to switch to http
+	// client that manages all low-level HTTP decisions for us, something like
+	// https://github.com/imroc/req, https://github.com/valyala/fasthttp or
+	// https://github.com/go-resty/resty
+	var responseBody io.ReadCloser = httpResponse.Body
+	if httpResponse.Header.Get("Content-Encoding") == "gzip" {
+		responseBody, err = gzip.NewReader(httpResponse.Body)
```
Olshansk, line 260: Can you create a helper that takes `httReponse` and returns the correct one? It'll be easier to add new ones in the future.

Olshansk (changes requested):

pkg/relayer/proxy/synchronous.go
okdas: Great idea!

Olshansk (approved):
So much cleaner after the changes!

Issue #459, opened 2024-04-05
[Config] Simplify RelayMiner config


red-0ne:
## Objective

Simplify the `RelayMiner`'s `relay_miner_config.yaml`, providing operators with a more intuitive config file.

## Origin Document

Private net external testers expressed confusion w.r.t. how the `RelayMiner`s are configured.
![image](https://github.com/pokt-network/poktroll/assets/231488/d22df552-9c20-43ef-a8f9-dae2f405a1a4)

## Goals

- Have a more intuitive `relay_miner_config.yaml` reflecting the `RelayMiner`'s setup
- Clear-up `RelayMiner`'s confusion w.r.t. the `proxies` section and how it's linked to the `suppliers` section.

## Deliverables

- [ ] A signle PR changing the `RelayMiner` configuration parser to handle the following structure:
https://gist.github.com/okdas/c20a2c0d778f4c291c8388a386b52d20

## Non-goals / Non-deliverables

- Change how the `RelayMiner` sets up its servers
- Change the structure resulting from parsing the config file

## General deliverables

- [ ] **Comments**: Add/update TODOs and comments alongside the source code so it is easier to follow.
- [ ] **Testing**: Add new tests (unit and/or E2E) to the test suite.
- [ ] **Documentation**: Update architectural or development READMEs; use [mermaid](https://mermaid-js.github.io/mermaid/) diagrams where appropriate.

---

**Creator**: @red-0ne 
**Co-Owners**: @okdas

red-0ne:
Closed as per #477

PR #477, opened 2024-04-11
[Config] feat: Simplify relay miner config

red-0ne:
## Summary

Simplify the relay miner config by removing the proxy section:
* Rewrites the `relay_miner_configs_reader`.
* Keeps the same resulting struct that's handled by the `RelayMiner`.
* Update config reader tests and remove the no-longer applicable tests.
* Update proxy tests.
* Update the documentation to reflect the new config structure.
* Update the `RelayRequest` server handler to support multiple services within the same host.

## Issue

- #459 

## Type of change

Select one or more:

- [ ] Bug fix
- [x] Code health or cleanup
- [x] Documentation
- [ ] Other (specify)

## Testing

## Sanity Checklist


localnet/kubernetes/values-relayminer-2.yaml
```diff
@@ -2,10 +2,9 @@ config:
   signing_key_name: supplier2
```
Olshansk, line 2: _tl;dr Can we simplify it? Let's discuss..._
---

@red-0ne As I'm looking at this, I have a bunch of questions:

1. Why do we need `server_type`?
	2. Is the `server_typr` the same as `backend_url`?
3. Why to we need `listen_address`?
	4. 	Is `listen_address` going to be the same as `publicly_exposed_endpoints` on localhost w/o the scheme?
5. There is redundancy in ports, schemes and URLs. 

You don't have to answer these questions, but it's confusing because:
1. I thought we aligned on remove `server_type`
2. `listen_address` feels redundant.

I want to emphasize that it's not something we should change, but it's still not blatantly obvious.

I look at https://pkg.go.dev/net/http#Request and found that:

```go
// For client requests, Host optionally overrides the Host
// header to send. If empty, the Request.Write method uses
// the value of URL.Host. Host may contain an international
// domain name.
Host [string](https://pkg.go.dev/builtin#string)
```	
And:

```go
// For client requests, the URL's Host specifies the server to
// connect to, while the Request's Host field optionally
// specifies the Host header value to send in the HTTP
// request.
URL *[url](https://pkg.go.dev/net/url).[URL](https://pkg.go.dev/net/url#URL)
```

And from our implementation:

```go
URL:  &url.URL{Scheme: scheme, Host: proxiedServices[proxyServeName].ListenAddress},
Host: proxiedServices[proxyServeName].Suppliers[serviceId].PubliclyExposedEndpoints[0]
```

Under what circumstance would we not be using `[0]`?

Olshansk (changes requested):
@red-0ne Just started a review and have some questions I think we should discuss before diving deeper./

localnet/kubernetes/values-relayminer-2.yaml
red-0ne: `listen_address` is needed and indicates on which address the relay miner will be listening to relays on.
`server_type` is removed now and `listen_address` changed to `listen_url`, we infer the `server_type` from the url scheme component.

As of `Host` and `Url`, this is how its is intended to work. The request's host is different from the url's to exercise how the `RelayMiner` accepts requests forwarded from reverse proxies and situations where the `listen_url` address is not publicly exposed.

> Under what circumstance would we not be using [0]?

This is a test scenario where we exercise requests sent to the first `PubliclyExposedEndpoints` and handled by the server listening at `ListenAddress`.

testutil/testproxy/relayerproxy.go
```diff
@@ -304,8 +304,8 @@ func MarshalAndSend(
 		Header: http.Header{
 			"Content-Type": []string{"application/json"},
 		},
-		URL:  &url.URL{Scheme: scheme, Host: proxiedServices[proxyServeName].Host},
-		Host: proxiedServices[proxyServeName].Suppliers[serviceId].Hosts[0],
+		URL:  &url.URL{Scheme: scheme, Host: proxiedServices[proxyServeName].ListenAddress},
+		Host: proxiedServices[proxyServeName].Suppliers[serviceId].PubliclyExposedEndpoints[0],
```
Olshansk, line 308: When do we ever use not `[0]`?

pkg/relayer/config/types.go
```diff
@@ -96,11 +86,11 @@ type RelayMinerPocketNodeConfig struct {
 type RelayMinerProxyConfig struct {
 	// ProxyName is the name of the proxy server, used to identify it in the config
 	ProxyName string
-	// Type is the transport protocol used by the proxy server like (http, https, etc.)
-	Type ProxyType
-	// Host is the host on which the proxy server will listen for incoming
+	// ServerType is the transport protocol used by the proxy server like (http, https, etc.)
+	ServerType ServerType
+	// ListenAddress is the host on which the proxy server will listen for incoming
```
Olshansk, line 91: Should this be ListenURL?

pkg/relayer/config/types.go
```diff
@@ -121,12 +111,12 @@ type RelayMinerMetricsConfig struct {
 type RelayMinerSupplierConfig struct {
 	// ServiceId is the serviceId corresponding to the current configuration.
 	ServiceId string
-	// Type is the transport protocol used by the supplier, it must match the
+	// ServerType is the transport protocol used by the supplier, it must match the
 	// type of the proxy it is associated with.
-	Type ProxyType
-	// Hosts is a list of hosts advertised on-chain by the supplier, the corresponding
+	ServerType ServerType
```
Olshansk, line 116: Can we remove this?

pkg/relayer/config/types.go
```diff
@@ -2,10 +2,10 @@ package config
 
 import "net/url"
 
-type ProxyType int
+type ServerType int
```
Olshansk, line 5: Is this still necessary?

pkg/relayer/proxy/synchronous.go
```diff
@@ -101,6 +103,28 @@ func (sync *synchronousRPCServer) ServeHTTP(writer http.ResponseWriter, request
...
 	defer request.Body.Close()
 
+	sync.logger.Debug().Msg("serving synchronous relay request")
+
+	// Extract the relay request from the request body.
+	sync.logger.Debug().Msg("extracting relay request from request body")
+	relayRequest, err := sync.newRelayRequest(request)
+	if err != nil {
+		proxyName := sync.proxyConfig.ProxyName
+		sync.replyWithError(ctx, []byte{}, writer, proxyName, unknownServiceId, err)
```
Olshansk, line 113: 1. Should we not pass the service id of the request even when there's an error?
2. I think even an empty string would cause less confusion in the case of an error

docusaurus/docs/operate/configs/supplier_staking_config.md
```diff
@@ -98,15 +98,16 @@ For example, it must match the regex `^[a-zA-Z0-9_-]{1,8}$`, and spaces are disa
...
+to the Pocket Network. Each `endpoint` object consists of a `publicly_exposed_url`
+and a `rpc_type`.
 
-##### `url`
+##### `publicly_exposed_url`
 
 _`Required`_
 
-The `url` defines the endpoint for sending `RelayRequests` from the Pocket Network's `Gateways` and `Applications`. Typically, this endpoint is provided by a RelayMiner, which acts as a proxy. This setup means that the specified URL directs to the RelayMiner's endpoint, which in turn, routes the requests to the designated service node.
+The `publicly_exposed_url` defines the endpoint for sending `RelayRequests` from the Pocket Network's `Gateways` and `Applications`. Typically, this endpoint is provided by a RelayMiner, which acts as a proxy. This setup means that the specified URL directs to the RelayMiner's endpoint, which in turn, routes the requests to the designated service node.
```
Olshansk, line 108: `provided by a RelayMiner, which acts as a proxy. This setup means that the specified URL directs to the RelayMiner's endpoint,`

It's a bit confusing that "it's provided by a RelayMiner` and then "directs to a RelayMine`.

Could you try rewording this a bit?

docusaurus/docs/operate/configs/supplier_staking_config.md
```diff
@@ -98,15 +98,16 @@ For example, it must match the regex `^[a-zA-Z0-9_-]{1,8}$`, and spaces are disa
...
 
-- **Example**: `https://etherium-relayminer1.relayminers.com:443` demonstrates how a RelayMiner endpoint, typically offered by the RelayMiner, proxies requests to an Etherium backend data node.
+- **Example**: `https://etherium-relayminer1.relayminers.com:443` demonstrates how a RelayMiner endpoint, typically offered by the RelayMiner, proxies requests to an Ethereum backend data node.
```
Olshansk, line 110: 1. Try to avoid using `RelayMiner` twice so close together.
2. Can you explain how this related to the on-stake supplier endpoint so it's even clearer?

localnet/kubernetes/values-relayminer-1.yaml
```diff
@@ -2,10 +2,8 @@ config:
...
   suppliers:
     - service_id: anvil
-      type: http
       service_config:
-        url: http://anvil:8547/
-      proxy_names:
-        - http-proxy
-      hosts:
-        - tcp://relayminer1:8545
+        backend_url: http://anvil:8547/
```
Olshansk, line 6: Thoughts on renaming this to `internal_service_url`?

cc @okdas

Olshansk (changes requested):
@red-0ne Not a complete review, but I had some thoughts / points and wanted to publish a partial review.

testutil/testproxy/relayerproxy.go
red-0ne: https://github.com/pokt-network/poktroll/pull/477#discussion_r1562461950

pkg/relayer/config/types.go
red-0ne: Not in the context of this struct. The `YAMLRelayMinerSupplierConfig` which is the exact representation of the config file has `ListenUrl` and embeds the listening address and the server type in its `<server_type>://<listen_address>`.

While `RelayMinerProxyConfig` get the parsed version, with the extracted `ServerType` and `ListenAddress` components to be able to start a server and listen to the provided address.

We are effectively changing the config file structure but not the `RelayMiner` logic that uses it. This is where having the both structs came handy! Thank to your's and Bryan's reviews back then :)

pkg/relayer/config/types.go
red-0ne: This is still needed by the relay miner, in fact we just removed the `server_type` entry from the config to embed it into the `listen_url` but not getting rid of that info.

pkg/relayer/config/types.go
red-0ne: https://github.com/pokt-network/poktroll/pull/477#discussion_r1563826905

pkg/relayer/proxy/synchronous.go
red-0ne: 1. In that code branch we were not able to parse the `RelayRequest` to extract the serviceId from it.
2. Will pass an empty string :+1:

pkg/relayer/config/proxy_http_config_parser.go
```diff
@@ -39,21 +36,21 @@ func (proxyConfig *RelayMinerProxyConfig) parseHTTPProxyConfig(
...
-		return ErrRelayMinerConfigInvalidSupplier.Wrap("empty supplier url")
+	// Check if the supplier backend url is empty
+	if len(yamlSupplierServiceConfig.BackendUrl) == 0 {
+		return ErrRelayMinerConfigInvalidSupplier.Wrap("empty supplier backend url")
 	}
 
-	// Check if the supplier url is a valid URL
-	supplierServiceUrl, err := url.Parse(yamlSupplierServiceConfig.Url)
+	// Check if the supplier backend url is a valid URL
+	supplierServiceBackendUrl, err := url.Parse(yamlSupplierServiceConfig.BackendUrl)
```
Olshansk, line 45: Love the naming

pkg/relayer/config/proxy_http_config_parser.go
```diff
@@ -1,34 +1,31 @@
 package config
 
-import (
-	"fmt"
-	"net/url"
-)
+import "net/url"
 
 // parseHTTPProxyConfig populates the proxy fields of the target structure that
 // are relevant to the "http" type in the proxy section of the config file.
```
Olshansk, line 6: Do we have a proxy section in the config file?

pkg/relayer/config/proxy_http_config_parser.go
```diff
@@ -1,34 +1,31 @@
...
+	yamlSupplierConfig YAMLRelayMinerSupplierConfig,
 ) error {
-	// Check if the proxy host is a valid URL.
-	// Since `yamlProxyConfig.Host` is a string representing the host, we need to
-	// prepend it with the "http://" scheme to make it a valid URL; we end up
-	// using the `Host` field of the resulting `url.URL` struct, so the prepended
-	// scheme is irrelevant.
-	proxyUrl, err := url.Parse(fmt.Sprintf("http://%s", yamlProxyConfig.Host))
+	// Check if the proxy listen address is a valid URL.
+	// Since `yamlProxyConfig.ListenAddress` is a string representing the host,
```
Olshansk, line 12: Should `yamlProxyConfig` be `yamlSupplierConfig`

pkg/relayer/config/proxy_http_config_parser.go
```diff
@@ -1,34 +1,31 @@
...
-			"invalid proxy host %s",
+			"invalid proxy listen address %s",
 			err.Error(),
 		)
 	}
 
-	if proxyUrl.Host == "" {
-		return ErrRelayMinerConfigInvalidProxy.Wrap("empty proxy host")
+	if listenUrl.Host == "" {
+		return ErrRelayMinerConfigInvalidProxy.Wrap("empty proxy listen address")
```
Olshansk, line 25: Should this be url instead of address?

pkg/relayer/proxy/proxy_test.go
```diff
@@ -658,12 +658,12 @@ func sendRequestWithDifferentSession(
...
 ) (errCode int32, errorMessage string) {
-	// Use secondaryService instead of service1 so the session IDs don't match
+	// Use a block height that generates a different session ID
 	req := testproxy.GenerateRelayRequest(
 		test,
 		appPrivateKey,
-		secondaryService,
-		blockHeight,
+		defaultService,
+		blockHeight+sessionkeeper.GetSessionGracePeriodBlockCount(),
```
Olshansk, line 666: Can you declare a local height variable right below the comment so it's easier to read & follow.?

Olshansk (changes requested):
@red-0ne I did another partial review and noticed some outdated comments / references.

Here is my ask:
1. Look through the comments, configs, etc and update:
1.1 Update outdated docs / comments
1.2 Revisit if we should still keep the name `proxy` in some of the unxposed config files. Not saying we should change it but consider it.

2. @okdas PTAL at this PR before EOD.

3. @red-0ne I'll pull the updated branch and do a final review (and push my own nits/edits) tomorrow hopefully before I get on the plane.

localnet/kubernetes/values-relayminer-1.yaml
okdas: @Olshansk tbh I like backend batter. In my mental model, it makes more sense.

localnet/kubernetes/values-relayminer-2.yaml
```diff
@@ -2,10 +2,8 @@ config:
...
-      type: http
       service_config:
-        url: http://anvil:8547/
-      proxy_names:
-        - http-proxy
-      hosts:
-        - tcp://relayminer2:8545
+        backend_url: http://anvil:8547/
+        publicly_exposed_endpoints:
+        - relayminer2:8545
```
okdas, line 8: @red-0ne we don't need to have a port in `publicly_exposed_endpoints`, because the port is already specified in `listen_url`.

okdas (changes requested):
This looks amazing, I just want to make sure we don't mention ports and schemas in `publicly_exposed_endpoints`, because they are specified in `listen_url`. Let me know if you need help adjusting the examples/localnet configs!

pkg/relayer/config/proxies_config_hydrator.go
```diff
@@ -1,58 +1,59 @@
...
+	yamlSupplierConfigs []YAMLRelayMinerSupplierConfig,
 ) error {
 	// At least one proxy is required
-	if len(yamlProxyConfigs) == 0 {
-		return ErrRelayMinerConfigInvalidProxy.Wrap("no proxies provided")
+	if len(yamlSupplierConfigs) == 0 {
+		return ErrRelayMinerConfigInvalidSupplier.Wrap("no suppliers provided")
 	}
 
 	relayMinerConfig.Proxies = make(map[string]*RelayMinerProxyConfig)
```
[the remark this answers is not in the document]
red-0ne: What about replacing it with `Server`?
We would have:
* `RelayMinerProxyConfig` -> `RelayMinerServerConfig`
* `relayMinerConfig.Proxies` -> `relayMinerConfig.Servers`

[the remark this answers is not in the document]
red-0ne: Will update the codebase, comments and docs we think it's better naming (cc @okdas , @bryanchriswhite)

localnet/poktrolld/config/relayminer_config.yaml
```diff
@@ -1,22 +1,16 @@
...
-    proxy_names:
-      - http-proxy
-    hosts:
+      backend_url: http://anvil:8547/
+      publicly_exposed_endpoints:
       - tcp://relayminers:8545
-metrics:
-  enabled: true
-  addr: :9090
+    listen_url: http//0.0.0.0:8545
```
okdas, line 16:
```suggestion
    listen_url: http://0.0.0.0:8545
```

okdas (approved):
LGTM! Everything worked on my machine, too! I found a typo in a config that isn't really used, but everything else looks great.

Thank you for making configs easier to use!

localnet/poktrolld/config/relayminer_config.yaml
```diff
@@ -1,22 +1,16 @@
...
   - service_id: anvil
-    type: http
     service_config:
-      url: http://anvil:8547/
-    proxy_names:
-      - http-proxy
-    hosts:
+      backend_url: http://anvil:8547/
+      publicly_exposed_endpoints:
       - tcp://relayminers:8545
```
okdas, line 15: I just realized that this should probably just be `relayminers`.
```suggestion
      - relayminers
```

pkg/relayer/config/types.go
Olshansk: Going to update it myself but good opportunity to #PUC.

localnet/poktrolld/config/relayminer_config.yaml
Olshansk: 1. Probably or definitely?
2. Why?

okdas: > Probably or definitely

Definitely

> Why

Because the port is specified in the `listen_url` instead.

Olshansk (approved):
@red-0ne Finally finished my edits and reviews. 

Overall takeaways:
1. The config simplification is good, but I anticipate there will be more.
2. We have to make the docs more user friendly.
3. When we make major refactors like this, need to keep the "macro view" of the source code in mind along with the "macro view".
4. You are right that the concept of the "proxy" still exists, but there were a handful of functions/types that were still tied to the old approach.
5. Run `git diff c7a3cdfdb015e0c662a4d5bad7fb8c0bfa0161c1` to see the changes I pushed today.
6. PTAL a look at (5), run devnet tests and let's push it through!

Issue #476, opened 2024-04-11
[Rings] Wait for redelegation grace periods to elapse before updating rings


bryanchriswhite:
## Objective

While application re-delegation events may be observed within a session, which MUST update on-chain state, this SHOULD NOT effect the authority of the gateways until after the corresponding grace period has elapsed. This would become problematic during proof verification because 1) proofs are verified after the session (and some grace period) and 2) proof verification includes validating ring signatures, which therefore constrains the timing of re-delegation effectiveness (see: [Redelegation Event Delays](https://www.notion.so/buildwithgrove/Redelegation-Event-Delays-d7d94882dbb24621a0958bb3410e998f)).

 I.e. gateways which had authority to sign for an application which was delegated to it **at the beginning of the current session** should retain that authority until after the respective proof window has closed.

### [Redelegation Event Delays](https://www.notion.so/buildwithgrove/Redelegation-Event-Delays-d7d94882dbb24621a0958bb3410e998f)

![image](https://github.com/pokt-network/poktroll/assets/600733/ddfad193-6c5d-4e8c-9d46-88cffc8f3830)

## Goals

- Ensure alignment of ring updates for on- & off-chain actors.
- Ensure proofs for sessions where the relay request was signed for by a delegate gateway are valid even when the respective application undelegates in that session.

## Deliverables

- [ ] A PR that updates the ring cache, the delegate and undelegate keepers to handle delayed undelegations, archival of old delegations and pruning of expired delegations.

## Non-goals / Non-deliverables

- Have a more fine grained `RingCache` client.

## General deliverables

- [ ] **Comments**: Add/update TODOs and comments alongside the source code so it is easier to follow.
- [ ] **Testing**: Add new tests (unit and/or E2E) to the test suite.
- [ ] **Makefile**: Add new targets to the Makefile to make the new functionality easier to use.
- [ ] **Documentation**: Update architectural or development READMEs; use [mermaid](https://mermaid-js.github.io/mermaid/) diagrams where appropriate.

---

**Creator**: @bryanchriswhite 
**Co-Owners**: @red-0ne

Olshansk:
Going to reword what @bryanchriswhite said from my POV just so it's clearer to me.

Assume the following:

1. `App` delegated to `Gateway` at start of Session (`height=9`)
2. Session started at `height=10`
3. Session will end at `height=20`
4. `Gateway` signed a `relay` on behalf of `App` at `heigh=12`
[...]

PR #518, opened 2024-05-06
[Application] feat: Add past delegation proofs

red-0ne:
## Summary

This PR adds support for old delegations verification by:
* Update Application proto to add undelegations map
* Adding a map of session end block heights to undelegations updated at undelegation time
* EndBlocker procedure that prunes old delegations past a retention time
* Pass in target block height to the ring client to generate the right gateway list at the provided block height
* Fix ring cache invalidation
* Fix the block hash used to generate the proof
* Fix account querier returning account with nil pub keys (accounts created at genesis time)
* Remove unused pubkey client code

*1129 LOC out of 1314 are autogenerated pulsar code*

## Issue

- #476 

## Type of change

Select one or more:

- [x] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

## Sanity Checklist


Olshansk:
@red-0ne Confirming that this is not ready for initial review yet. Right?

red-0ne:
> @red-0ne Confirming that this is not ready for initial review yet. Right?

It is, just lacking E2E tests from the previous PR to be ported to this one

x/application/keeper/prune_undelegations.go
```diff
@@ -0,0 +1,37 @@
...
+	currentHeight := ctx.BlockHeight()
+	// Calculate the block height at which undelegations should be pruned
+	undelegationRetentionBlocks := sessionkeeper.GetSessionGracePeriodBlockCount() +
+		(sessionkeeper.NumBlocksPerSession * undelegationRetentionSessions)
+
+	// Iterate over all applications and prune undelegations that are older than
+	// the retention period.
+	for _, application := range k.GetAllApplications(ctx) {
+		for undelegationsHeight := range application.Undelegations {
+			if undelegationsHeight < uint64(currentHeight-undelegationRetentionBlocks) {
```
Olshansk: Should this not be > ?

x/application/keeper/msg_server_undelegate_from_gateway.go
```diff
@@ -49,6 +51,34 @@ func (k msgServer) UndelegateFromGateway(ctx context.Context, msg *types.MsgUnde
 	// Remove the gateway from the application's delegatee gateway public keys
 	foundApp.DelegateeGatewayAddresses = append(foundApp.DelegateeGatewayAddresses[:foundIdx], foundApp.DelegateeGatewayAddresses[foundIdx+1:]...)
 
+	sdkCtx := sdk.UnwrapSDKContext(ctx)
+	currentBlock := sdkCtx.BlockHeight()
+	sessionEndHeight := uint64(sessionkeeper.GetSessionEndBlockHeight(currentBlock))
+
+	// TODO_INVESTIGATE: When there is no undelegation, the undelegations map is nil
```
Olshansk: There are two cases: 

1. Prod
- Search for where we initialize it and make sure we provide a default value
2. Tests
- Search for all `.Application{` in the code ane make sure it's initialized

proto/poktroll/application/application.proto
```diff
@@ -15,4 +15,11 @@ message Application {
   cosmos.base.v1beta1.Coin stake = 2; // The total amount of uPOKT the application has staked
   repeated poktroll.shared.ApplicationServiceConfig service_configs = 3; // The list of services this appliccation is configured to request service for
   repeated string delegatee_gateway_addresses = 4 [(cosmos_proto.scalar) = "cosmos.AddressString", (gogoproto.nullable) = false]; // The Bech32 encoded addresses for all delegatee Gateways, in a non-nullable slice
+  map<uint64, UndelegationFromAppToGatewayEvent> undelegations = 5 [(gogoproto.nullable) = false]; // The map of session end heights to undelegations from Gateway, in a non-nullable map
 }
+
+// UndelegationFromAppToGatewayEvent defines the type used to store an undelegation
+// event from an application to a gateway that happened at a specific session end height.
+message UndelegationFromAppToGatewayEvent {
```
Olshansk: Do not use `Event` unelss it's a real event that we emit. It is a specific primitive in the Cosmos SDK.

x/application/keeper/msg_server_undelegate_from_gateway.go
```diff
@@ -49,6 +51,34 @@ func (k msgServer) UndelegateFromGateway(ctx context.Context, msg *types.MsgUnde
 	// Remove the gateway from the application's delegatee gateway public keys
 	foundApp.DelegateeGatewayAddresses = append(foundApp.DelegateeGatewayAddresses[:foundIdx], foundApp.DelegateeGatewayAddresses[foundIdx+1:]...)
 
+	sdkCtx := sdk.UnwrapSDKContext(ctx)
```
Olshansk, line 55: Let's move all of this business logic into a new helper.

x/application/keeper/msg_server_undelegate_from_gateway.go
```diff
@@ -49,6 +51,34 @@ func (k msgServer) UndelegateFromGateway(ctx context.Context, msg *types.MsgUnde
...
+	}
+
+	// Add the gateway to the undelegated gateways list if it's not already there.
+	if !slices.Contains(undelegationsAtBlock.UndelegatedGateways, msg.GatewayAddress) {
+		undelegationsAtBlock.UndelegatedGateways = append(
+			undelegationsAtBlock.UndelegatedGateways,
+			msg.GatewayAddress,
+		)
+
+		foundApp.Undelegations[sessionEndHeight] = undelegationsAtBlock
```
Olshansk: I don't think this line is necessary since slices are pointer. Can you remove and confirm?

x/application/keeper/msg_server_undelegate_from_gateway_test.go
```diff
@@ -92,6 +92,9 @@ func TestMsgServer_UndelegateFromGateway_SuccessfullyUndelegate(t *testing.T) {
 	_, err = srv.UndelegateFromGateway(ctx, undelegateMsg)
```
Olshansk, line 92: I know you said we don't need any new unit/integration tests, but I still feel like a unique test where we do stuff:

1. Undelegate mid session
2. Redelegate mid session
3. Check when we're halfway into the retention period (still there)
4. Check when we're past the retention period (shouldn't be there)

I STRONGLY feel there are at least 3-4 "unit" tests that need to be added for very cryptic logic like this.

If you disagree or need help with framing them, lmk

x/application/keeper/prune_undelegations.go
Olshansk: I'm changing it, but just wanted to call out that the following is what helped me identify this potential bug:

1. Extra local variables
2. Better naming

Other than that, regardless of whether I'm right or wrong on this, this is why we need unit/integration (not E2E tests) DEDICATED to this.

Olshansk (changes requested):
@red-0ne Please do the following:

1. See my comments. Even if they're outdated (based on code), PTAL at all of them
2. Review my commit in [`82abcc8` (#518)](https://github.com/pokt-network/poktroll/pull/518/commits/82abcc89b15069daff73d38e08ec30b45f8bd6b5). Note the comments I made and the var changes I made.
3. Let's make the same changes as in (2) everywhere across the PR
4. Please respond to my comments in (1)
5. Re-request the review (even if it's still not complete) once I can start taking a look at what we have.

Other than that, love how much simpler this is!

x/application/keeper/prune_undelegations.go
red-0ne: I still believe it is `<`. We are pruning undelegations that are **older** than `pruningBlockHeight` which means, if they have a **smaller** block height than `pruningBlockHeight` then they should be deleted.

x/application/keeper/msg_server_undelegate_from_gateway.go
red-0ne: Will try this out :+1:

x/application/keeper/prune_undelegations.go
Olshansk: Okay, agreed. Let's change it to `if undelegationSessionEndHeight < pruningBlockHeight  then delete`, Thoughts?

Even though you are right, the fact that a test didn't fail for this still gives me "some" points though ;)

x/application/keeper/msg_server_undelegate_from_gateway.go
red-0ne: Seems that it's still needed

x/application/keeper/msg_server_undelegate_from_gateway_test.go
```diff
@@ -272,3 +281,225 @@ func TestMsgServer_UndelegateFromGateway_SuccessfullyUndelegateFromUnstakedGatew
...
+	// Verify that the the pending undelegation map no longer contains the
+	// sessionEndHeight key.
+	require.Empty(t, app.PendingUndelegations[sessionEndHeight])
+
+	// Verify that the reconstructed delegatee gateway list includes the reundelegated gateway
+	gatewayAddressesAfterPruning := rings.GetRingAddressesAtBlock(&app, sdkCtx.BlockHeight())
+	require.Contains(t, gatewayAddressesAfterPruning, gatewayAddrToRedelegate)
+}
+
+// delegateThenUndelegate factors delegating an application to two gateways,
```
Olshansk, line 436: factors?

x/application/keeper/msg_server_undelegate_from_gateway_test.go
```diff
@@ -272,3 +281,225 @@ func TestMsgServer_UndelegateFromGateway_SuccessfullyUndelegateFromUnstakedGatew
...
+// then undelegates the application from one of the gateways to be used in
+// multiple tests.
+func delegateThenUndelegate(
```
Olshansk, line 439: Please update the comments and the name of the function to reflect all the "side effects" of this.

It does A LOT like creating an app, staking it, etc...

x/application/keeper/msg_server_undelegate_from_gateway_test.go
```diff
@@ -272,3 +281,225 @@ func TestMsgServer_UndelegateFromGateway_SuccessfullyUndelegateFromUnstakedGatew
...
+	)
+
+	// Verify that the other gateways are still delegated to the application.
+	require.Contains(t, app.DelegateeGatewayAddresses, delegatedGatewayAddr)
+
+	// Verify that the reconstructed delegatee gateway list include the undelegated gateway.
+	gatewayAddresses := rings.GetRingAddressesAtBlock(&app, sdkCtx.BlockHeight())
+	require.Contains(t, gatewayAddresses, gatewayAddrToUndelegate)
+
+	// Increment the block height to the next session and  run the pruning
```
Olshansk, line 323:
```suggestion
	// Increment the block height to the next session and run the pruning
```

x/application/keeper/msg_server_undelegate_from_gateway_test.go
```diff
@@ -272,3 +281,225 @@ func TestMsgServer_UndelegateFromGateway_SuccessfullyUndelegateFromUnstakedGatew
...
+
+	// Get the updated application state.
+	app, isAppFound = k.GetApplication(sdkCtx, app.Address)
+	require.True(t, isAppFound)
+	require.NotNil(t, app)
+
+	// Verify that when queried for the next session the reconstructed delegatee
+	// gateway list does include the undelegated gateway.
+	nextSessionGatewayAddresses := rings.GetRingAddressesAtBlock(&app, nextSessionStartHeight)
+	require.NotContains(t, nextSessionGatewayAddresses, gatewayAddrToUndelegate)
```
Olshansk, line 337: Can you please add a few more steps to make sure the same thing applies when you take the grace period into consideration?

x/application/keeper/msg_server_undelegate_from_gateway_test.go
```diff
@@ -272,3 +281,225 @@ func TestMsgServer_UndelegateFromGateway_SuccessfullyUndelegateFromUnstakedGatew
...
+	require.True(t, isAppFound)
+	require.NotNil(t, foundApp)
+
+	return sdkCtx, foundApp, gatewayAddrs
+}
+
+func getPruningBlockHeight(currentHeight int64) (pruningHeihgt int64) {
+	nextSessionStartHeight := int64(sessionkeeper.GetSessionEndBlockHeight(currentHeight) + 1)
+
+	numBlocksUndelegationRetention := sessionkeeper.GetSessionGracePeriodBlockCount() +
```
Olshansk, line 501: Isn't there a function / helper for this in the actual source code?

x/application/keeper/msg_server_undelegate_from_gateway_test.go
```diff
@@ -272,3 +281,225 @@ func TestMsgServer_UndelegateFromGateway_SuccessfullyUndelegateFromUnstakedGatew
...
+}
+
+func TestMsgServer_UndelegateFromGateway_RedelegationAfterUndelegationAtTheSameSessionNumber(t *testing.T) {
+	k, ctx := keepertest.ApplicationKeeper(t)
+	srv := keeper.NewMsgServerImpl(k)
+
+	undelegationHeight := int64(1)
+	sdkCtx, app, gatewayAddrs := delegateThenUndelegate(ctx, t, srv, k, undelegationHeight)
+
+	gatewayAddrToRedelegate := gatewayAddrs[0]
```
Olshansk, line 380: There's a lot of implicit knowledge here that index 0 is the one that was undelegated above.

Can you change the helper function to return `ctx, app, delegatedToAddr, pendingUndelegateToAddr` so it's VERY clear and explicit.

x/application/keeper/msg_server_undelegate_from_gateway_test.go
```diff
@@ -272,3 +281,225 @@ func TestMsgServer_UndelegateFromGateway_SuccessfullyUndelegateFromUnstakedGatew
...
+	require.Contains(t, app.DelegateeGatewayAddresses, gatewayAddrToRedelegate)
+
+	// Verify that the the pending undelegation map no longer contains the
+	// sessionEndHeight key.
+	require.Empty(t, app.PendingUndelegations[sessionEndHeight])
+
+	// Verify that the reconstructed delegatee gateway list includes the reundelegated gateway
+	gatewayAddressesAfterPruning := rings.GetRingAddressesAtBlock(&app, sdkCtx.BlockHeight())
+	require.Contains(t, gatewayAddressesAfterPruning, gatewayAddrToRedelegate)
+}
```
Olshansk, line 434: These tests are PERFECT!

pkg/crypto/rings/client.go
```diff
@@ -194,3 +206,52 @@ func (rc *ringClient) addressesToPubKeys(
...
+			}
+
+			activeDelegationsAtHeight = append(activeDelegationsAtHeight, gatewayAddress)
+			// Mark the gateway address as added to avoid duplicates.
+			addedDelegations[gatewayAddress] = true
+		}
+
+	}
+
+	// add app address twice to make the ring size of minimum 2
```
Olshansk, line 246: #PUC. We only add the app address once here, so it is used twice in total when generating the ring. However, saying twice here makes it misleading.

Olshansk (changes requested):

x/application/keeper/msg_server_undelegate_from_gateway_test.go
```diff
@@ -92,6 +95,9 @@ func TestMsgServer_UndelegateFromGateway_SuccessfullyUndelegate(t *testing.T) {
 	_, err = srv.UndelegateFromGateway(ctx, undelegateMsg)
 	require.NoError(t, err)
 
+	sdkCtx.WithBlockHeight(4)
```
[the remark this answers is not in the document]
red-0ne: Thanks for pointing this out. Setting the height and pruning undelegations are no longer needed in this test, since it's only asserting on `EventRedelegation` event emission and `DelegateeGatewayAddresses` being updated (which is no longer delayed)

x/application/keeper/msg_server_undelegate_from_gateway_test.go
red-0ne: Will rephrase the comment.

x/application/keeper/msg_server_undelegate_from_gateway_test.go
red-0ne: Added.

pkg/crypto/rings/client.go
red-0ne: Took this case out of the `GetRingAddressesAtBlock` since this function deals with delegatee gateways addresses.
Also made the first addition of the appAddress closer to the second (and to the comment)

pkg/crypto/rings/client.go
```diff
@@ -194,3 +206,52 @@ func (rc *ringClient) addressesToPubKeys(
...
 	return pubKeys, nil
 }
+
+// GetRingAddressesAtBlock returns the active gateway delegations for the given
+// application and target block height while accounting for the pending undelegations.
+// The ring addresses slice is reconstructed by adding back the past delegated gateways
+// that have been undelegated after the target session end height.
+func GetRingAddressesAtBlock(app *apptypes.Application, blockHeight int64) []string {
+	// Get the target session end height at which we want to get the active delegations.
+	targetSessionEndHeight := uint64(sessionkeeper.GetSessionEndBlockHeight(blockHeight))
```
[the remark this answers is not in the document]
Olshansk: Does this implicitly assume that `GetNumBlocksUndelegationRetention` > `SessionGracePeriod`?

x/application/keeper/msg_server_undelegate_from_gateway_test.go
```diff
@@ -332,24 +324,40 @@ func TestMsgServer_UndelegateFromGateway_DelegationIsActiveUntilNextSession(t *t
...
+	sdkCtx = sdkCtx.WithBlockHeight(afterSessionGracePeriodHeight)
+	k.EndBlockerPruneAppToGatewayPendingUndelegation(sdkCtx)
+
+	// Verify that when queried for a block height past the tested session's grace period,
+	// the reconstructed delegatee gateway list does not include the undelegated gateway.
+	pastGracePeriodGatewayAddresses := rings.GetRingAddressesAtBlock(&app, afterSessionGracePeriodHeight)
+	require.NotContains(t, pastGracePeriodGatewayAddresses, pendingUndelegateToAddr)
+
+	// Verify that when queried for a block height corresponding to the past session
+	// that has its grace period elapsed, the reconstructed delegatee gateway list
```
Olshansk, line 344: `that has its grace period elapsed` - I don't fully understand this. What am I missing about the explanation?

x/application/keeper/msg_server_undelegate_from_gateway_test.go
```diff
@@ -332,24 +324,40 @@ func TestMsgServer_UndelegateFromGateway_DelegationIsActiveUntilNextSession(t *t
...
+	// does include the undelegated gateway.
+	pastGracePeriodGatewayAddresses = rings.GetRingAddressesAtBlock(&app, int64(sessionEndHeight))
```
Olshansk, line 346: A bit confusing that the variable here and above has the same name but they mean different things - What am I missing?

Olshansk (changes requested):

x/application/keeper/msg_server_undelegate_from_gateway.go
```diff
@@ -49,21 +52,57 @@ func (k msgServer) UndelegateFromGateway(ctx context.Context, msg *types.MsgUnde
...
+	// Add the gateway address to the undelegated gateways list if it's not already there.
+	if !slices.Contains(undelegatingGatewayListAtBlock.GatewayAddresses, gatewayAddress) {
+		undelegatingGatewayListAtBlock.GatewayAddresses = append(
+			undelegatingGatewayListAtBlock.GatewayAddresses,
+			gatewayAddress,
+		)
+		app.PendingUndelegations[sessionEndHeight] = undelegatingGatewayListAtBlock
+	} else {
+		k.logger.Warn(fmt.Sprintf(
+			"Application undelegating (again) from gateway it's already undelegating from with address [%s]",
```
bryanchriswhite, line 103: Is it worth printing the application address as well?

Wdyt about making it an `Info()` log instead of a `Warn()`? My thinking is that it's not something a poktrolld operator needs to be concerned about.

proto/poktroll/application/application.proto
```diff
@@ -15,4 +15,16 @@ message Application {
   cosmos.base.v1beta1.Coin stake = 2; // The total amount of uPOKT the application has staked
   repeated poktroll.shared.ApplicationServiceConfig service_configs = 3; // The list of services this appliccation is configured to request service for
   repeated string delegatee_gateway_addresses = 4 [(cosmos_proto.scalar) = "cosmos.AddressString", (gogoproto.nullable) = false]; // The Bech32 encoded addresses for all delegatee Gateways, in a non-nullable slice
```
bryanchriswhite, line 17: This may be out of scope for this PR but I feel the need to call out that I think this field is named incorrectly. In my view, the `application` is the delegatee and the `gateway` is the delegate; delegatees delegate authority to their respective delegates.

x/application/keeper/msg_server_undelegate_from_gateway_test.go
```diff
@@ -272,3 +275,249 @@ func TestMsgServer_UndelegateFromGateway_SuccessfullyUndelegateFromUnstakedGatew
...
+
+	// Verify that the gateway is added to the pending undelegation list with the
+	// right sessionEndHeight as the map key.
+	sessionEndHeight := uint64(sessionkeeper.GetSessionEndBlockHeight(undelegationHeight))
+	require.Contains(t,
+		app.PendingUndelegations[sessionEndHeight].GatewayAddresses,
+		pendingUndelegateToAddr,
+	)
+
+	// Verify that the other gateways are still delegated to the application.
```
bryanchriswhite, line 308:
```suggestion
	// Verify that the application is still delegating to other gateways.
```

x/application/keeper/msg_server_undelegate_from_gateway_test.go
```diff
@@ -272,3 +275,249 @@ func TestMsgServer_UndelegateFromGateway_SuccessfullyUndelegateFromUnstakedGatew
...
+	require.Contains(t, app.DelegateeGatewayAddresses, delegatedToAddr)
+
+	// Verify that the reconstructed delegatee gateway list include the undelegated gateway.
```
bryanchriswhite, line 311:
```suggestion
	// Verify that the reconstructed delegatee gateway list includes the undelegated gateway.
```

x/application/keeper/msg_server_undelegate_from_gateway_test.go
```diff
@@ -272,3 +275,249 @@ func TestMsgServer_UndelegateFromGateway_SuccessfullyUndelegateFromUnstakedGatew
 	require.Equal(t, appAddr, foundApp.Address)
 	require.Equal(t, 0, len(foundApp.DelegateeGatewayAddresses))
 }
+
+func TestMsgServer_UndelegateFromGateway_DelegationIsActiveUntilNextSession(t *testing.T) {
+	k, ctx := keepertest.ApplicationKeeper(t)
+	srv := keeper.NewMsgServerImpl(k)
+
+	undelegationHeight := int64(1)
+	sdkCtx, app, delegatedToAddr, pendingUndelegateToAddr :=
```
bryanchriswhite, line 284: NIT:
```suggestion
	sdkCtx, app, delegatedToAddr, pendingUndelegateFromAddr :=
```

bryanchriswhite (changes requested):
Nice one @red-0ne! :sunglasses: 

I have to take a break on this for now but I'm almost to the end:
image

Great comments! :muscle:

pkg/crypto/rings/client.go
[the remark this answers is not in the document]
red-0ne: Yes, pruning is guarenteed to be > `SessionGracePeriod`
```go
func GetNumBlocksUndelegationRetention() int64 {
	return sessionkeeper.GetSessionGracePeriodBlockCount() +
		(sessionkeeper.NumBlocksPerSession * NumSessionsAppToGatewayUndelegationRetention)
}
```

x/application/keeper/msg_server_undelegate_from_gateway_test.go
```diff
@@ -272,3 +275,249 @@ func TestMsgServer_UndelegateFromGateway_SuccessfullyUndelegateFromUnstakedGatew
 	require.Equal(t, appAddr, foundApp.Address)
 	require.Equal(t, 0, len(foundApp.DelegateeGatewayAddresses))
 }
+
+func TestMsgServer_UndelegateFromGateway_DelegationIsActiveUntilNextSession(t *testing.T) {
```
bryanchriswhite, line 279: I think a comment here with numbered steps explaining what's happening would be helpful.

x/application/keeper/msg_server_undelegate_from_gateway_test.go
bryanchriswhite:
```suggestion
	// for which the grace period has elapsed, the reconstructed delegatee gateway list
```

x/application/keeper/msg_server_undelegate_from_gateway_test.go
```diff
@@ -272,3 +275,249 @@ func TestMsgServer_UndelegateFromGateway_SuccessfullyUndelegateFromUnstakedGatew
...
+
+	// Verify that the gateway is also present in the pending undelegation list with the
+	// right sessionEndHeight as the map key.
+	sessionEndHeight := uint64(sessionkeeper.GetSessionEndBlockHeight(undelegationHeight))
+	require.Contains(t,
+		app.PendingUndelegations[sessionEndHeight].GatewayAddresses,
+		gatewayAddrToRedelegate,
+	)
+
+	// Verify that the reconstructed delegatee gateway list include the redelegated gateway.
```
bryanchriswhite, line 416:
```suggestion
	// Verify that the reconstructed delegatee gateway list includes the redelegated gateway.
```

x/application/keeper/msg_server_undelegate_from_gateway_test.go
```diff
@@ -272,3 +275,249 @@ func TestMsgServer_UndelegateFromGateway_SuccessfullyUndelegateFromUnstakedGatew
...
+	require.NotNil(t, app)
+
+	// Verify that the application is still delegated to the gateway
+	require.Contains(t, app.DelegateeGatewayAddresses, gatewayAddrToRedelegate)
+
+	// Verify that the the pending undelegation map no longer contains the
+	// sessionEndHeight key.
+	require.Empty(t, app.PendingUndelegations[sessionEndHeight])
+
+	// Verify that the reconstructed delegatee gateway list includes the reundelegated gateway
```
bryanchriswhite, line 438:
```suggestion
	// Verify that the reconstructed delegatee gateway list includes the redelegated gateway
```

x/application/keeper/msg_server_undelegate_from_gateway_test.go
```diff
@@ -272,3 +275,249 @@ func TestMsgServer_UndelegateFromGateway_SuccessfullyUndelegateFromUnstakedGatew
...
+				Service: &sharedtypes.Service{Id: "svc1"},
+			},
+		},
+	}
+	_, err := srv.StakeApplication(ctx, stakeMsg)
+	require.NoError(t, err)
+
+	// Generate gateway addresses, mock the gateways being staked then delegate the
+	// application to the gateways.
+	delegatedToAddr = sample.AccAddress()
```
bryanchriswhite, line 477: NIT: Wdyt about `delegateAddr? (https://github.com/pokt-network/poktroll/pull/518/files#r1599675821)

x/application/keeper/msg_server_undelegate_from_gateway_test.go
```diff
@@ -272,3 +275,249 @@ func TestMsgServer_UndelegateFromGateway_SuccessfullyUndelegateFromUnstakedGatew
...
+	pastGracePeriodGatewayAddresses = rings.GetRingAddressesAtBlock(&app, int64(sessionEndHeight))
+	require.Contains(t, pastGracePeriodGatewayAddresses, pendingUndelegateToAddr)
+}
+
+func TestMsgServer_UndelegateFromGateway_DelegationIsPrunedAfterRetentionPeriod(t *testing.T) {
+	k, ctx := keepertest.ApplicationKeeper(t)
+	srv := keeper.NewMsgServerImpl(k)
+
+	undelegationHeight := int64(1)
+	sdkCtx, app, delegatedToAddr, pendingUndelegateToAddr :=
```
bryanchriswhite, line 355: NIT: Wdyt about `delegateAddr? (https://github.com/pokt-network/poktroll/pull/518/files#r1599675821)

x/application/keeper/prune_undelegations.go
```diff
@@ -0,0 +1,50 @@
...
+// and prunes app to gateway undelegations that have exceeded the retention delay.
+func (k Keeper) EndBlockerPruneAppToGatewayPendingUndelegation(ctx sdk.Context) error {
+	currentHeight := ctx.BlockHeight()
+
+	// Calculate the block height at which undelegations should be pruned
+	numBlocksUndelegationRetention := GetNumBlocksUndelegationRetention()
+	if currentHeight <= numBlocksUndelegationRetention {
+		return nil
+	}
+	pruningBlockHeight := uint64(currentHeight - numBlocksUndelegationRetention)
```
bryanchriswhite, line 27: The name confused me for a minute here. I was expecting it to be used with `<=` below. It seems to me it may be better described as `earliestUndelegationRetentionHeight` or `firstUnprunedUndelegationHeight` or something like that.

x/application/keeper/prune_undelegations.go
```diff
@@ -0,0 +1,50 @@
...
+		}
+
+		k.SetApplication(ctx, application)
+	}
+
+	return nil
+}
+
+// GetNumBlocksUndelegationRetention returns the number of blocks undelegations
+// should be kept before being pruned.
```
bryanchriswhite, line 46:
```suggestion
// GetNumBlocksUndelegationRetention returns the number of blocks for which
// undelegations should be kept before being pruned.
```

pkg/crypto/rings/client.go
```diff
@@ -142,34 +154,46 @@ func (rc *ringClient) VerifyRelayRequestSignature(
 	return nil
 }
 
-// getDelegatedPubKeysForAddress returns the gateway public keys an application
-// delegated the ability to sign relay requests on its behalf.
-func (rc *ringClient) getDelegatedPubKeysForAddress(
+// getRingPubKeysForAddress returns the public keys corresponding to a ring
+// It is a slice consisting of the application's public key and the public keys
```
bryanchriswhite, line 158:
```suggestion
// It returns a slice consisting of the application's public key and the public keys
```

x/application/keeper/msg_server_undelegate_from_gateway_test.go
red-0ne: Good catch! It should be `gatewayAddressesBeforeUndelegation`

pkg/crypto/rings/client.go
```diff
@@ -142,34 +154,46 @@ func (rc *ringClient) VerifyRelayRequestSignature(
...
-		// requires AT LEAST two pubKeys. If the Application has not delegated
-		// to any gateways, the app's own address needs to be used twice to
-		// create a ring. This is not a huge issue but an improvement should
-		// be investigated in the future.
-		ringAddresses = append(ringAddresses, appAddress)
-	} else {
-		// add the delegatee gateway addresses
-		ringAddresses = append(ringAddresses, app.DelegateeGatewayAddresses...)
+
+	// If there is no delegateeGatewayAddresses, add app address a second time to
```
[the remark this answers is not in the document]
red-0ne: Will remove.

pkg/relayer/session/proof.go
```diff
@@ -52,27 +52,37 @@ func (rs *relayerSessionsManager) submitProofs(
 // at which a proof can be submitted for the given session, then emits the session
 // **at that moment**.
 func (rs *relayerSessionsManager) mapWaitForEarliestSubmitProofHeight(
```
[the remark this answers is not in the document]
red-0ne: Let's deal with this in #521

proto/poktroll/application/application.proto
red-0ne: We discussed this, I think you are right.

cc @Olshansk

Olshansk (approved):
Please see my last two comments but pre-emtively approving.

bryanchriswhite (approved):
:100:

PR #520, opened 2024-05-08
[Genesis Config] Set MaxValidators to 1

Olshansk:
Added the following change and comment in `config.yml`. Should be self-explanatory.

```yaml
    # TODO_BLOCKER(@Olshansk): Figure out what this should be on Shannon
    # re-genesis. We're setting it to 1 for Alpha TestNet https://github.com/pokt-network/poktroll/pull/1 so Grove
    # maintains the only validator until Alpha TestNet https://github.com/pokt-network/poktroll/pull/2.
    max_validators: 1
```

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -88,6 +88,10 @@ genesis:
     staking:
       params:
         bond_denom: upokt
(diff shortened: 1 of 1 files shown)


okdas (approved):
This will be helpful.

Issue #517, opened 2024-05-03
[Protocol] Add a `NumBlocksPerSession` governance parameter


Olshansk:
## Objective

Implementation of `NumBlocksPerSession`without hacky tech debt.

## Origin Document

There are many comments and temporary workaround in the code. the following is just one of them:

```go
const (
	// TODO_BLOCKER: Remove direct usage of these constants in helper functions
	// when they will be replaced by governance params
	NumBlocksPerSession = 4
```

## Goals

- Resolve a lot of TECHDEBT related to the number of blocks per session
- Implement some of the core macro utility of the protocol

## Deliverables

**_Reminder: This can be multiple PRs_** 

- [x] Add governance parameters for `sessionkeeper.NumBlocksPerSession`
    (#530)
- [x] Techdebt - search for, identify, update and cleanup all comments / code / functions related to documented tech debt.
  -  #545 
  - #546
  - #558
  - #538 
  - #550 
  - #555 
  - #556 
  - #557
  - #559 
  - #560
- [x] Business logic - implement the actual thing
- [ ] Testing
	- [x] 	Ensure we have at least 1 E2E test
	- [ ] 	Ensure we have at least 1 integration (in memory) test
- [ ] Makefile targets to:
	- [x] Modify / update the parameters
	- [x] Run the relevant E2E tests
- [ ] Docs in docusurus to to minimally document this new parameter

## Non-goals / Non-deliverables

- Exhaustively testing all edge cases and variations of changing this value

---

**Creator**: @Olshansk 
**Co-Owners**: @bryanchriswhite

PR #530, opened 2024-05-14
[Session] add `num_blocks_per_session` param

bryanchriswhite:
## Summary

Adds `num_blocks_per_session` param to the session module and removes & updates obsoleted usages of `sessionkeeper.NumBlocksPerSession` constant.

## Issue

- #517 

## Type of change

Select one or more:

- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

## Sanity Checklist


tools/scripts/params/tokenomics_compute_units_to_tokens_multiplier.json
```diff
@@ -5,7 +5,7 @@
         "@type": "/poktroll.tokenomics.MsgUpdateParam",
         "authority": "pokt10d07y265gmmuvt4z0w9aw880jnsr700j8yv32t",
         "name": "compute_units_to_tokens_multiplier",
-        "as_int64": "69"
+        "as_int64": "42"
```
Olshansk, line 8: Makes sense to me 🤓

x/session/types/params.go
```diff
@@ -2,7 +2,15 @@ package types
 
 import paramtypes "github.com/cosmos/cosmos-sdk/x/params/types"
 
-var _ paramtypes.ParamSet = (*Params)(nil)
+const (
+	DefaultNumBlocksPerSession = 4
+	NameNumBlocksPerSession    = "num_blocks_per_session"
```
Olshansk, line 7: What are your thoughts on `ParamNumBlocksPerSession` instead of `NameNumBlocksPerSession`?

This would apply everywhere.

x/application/types/params.go
```diff
@@ -6,6 +6,8 @@ import (
 	paramtypes "github.com/cosmos/cosmos-sdk/x/params/types"
 )
 
+const NameMaxDelegatedGateways = "max_delegated_gateways"
```
Olshansk, line 9: Some of the other params have a default beside the name. Why don't we have one here?

x/session/types/params.go
```diff
@@ -20,11 +30,36 @@ func DefaultParams() Params {
...
+// ValidateBasic does a sanity check on the provided params.
+func (params *Params) ValidateBasic() error {
+	if err := ValidateNumBlocksPerSession(params.NumBlocksPerSession); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// ValidateNumBlocksPerSession validates the MinRelayDifficultyBits param
```
Olshansk, line 52:
```suggestion
// ValidateNumBlocksPerSession validates the NumBlocksPerSession param
```

x/session/types/params.go
```diff
@@ -20,11 +30,36 @@ func DefaultParams() Params {
...
+// NB: The argument is an interface type to satisfy the ParamSetPair function signature.
+func ValidateNumBlocksPerSession(v interface{}) error {
+	numBlocksPerSession, ok := v.(uint64)
+	if !ok {
+		return ErrSessionParamInvalid.Wrapf("invalid parameter type: %T", v)
+	}
+
+	if numBlocksPerSession < 0 {
```
Olshansk, line 60:
```suggestion
	if numBlocksPerSession < 1 {
```

Olshansk, line 60: Overkill to add unit tests for this?

Olshansk (changes requested):

x/application/types/params.go
bryanchriswhite: We do, it's just declared in the `var` group below. I think that's the result of an ignite scaffolding template. When I added `num_blocks_per_session` manually, I grouped constants together.

x/session/types/params.go
bryanchriswhite: Nice catch - thanks! :raised_hands:

x/session/types/params.go
bryanchriswhite: I'm down with that.

Olshansk (approved):
This is setting some really great standards that others will just copy-pasta to add more params. Love it!

okdas:
@bryanchriswhite I saw this is blocked due to e2e test failing on CI. Looking at the output of the test, it seems like the reason is the accounts were not fully initialized due to send transactions not working correctly. It appears we've recently introduced some fees. Here's the error:
```
raw_log: ‘insufficient fees; got:  required: 2000upokt: insufficient fee’
```
I'll see how to address that, and I'll probably end up pushing a fix in your PR.

okdas:
We started having this issue because the `minimum-gas-prices` was set to `"0.01upokt"` (I put it there while changing this config on TestNet).

On LocalNet, we have this set to `0upokt`. Originally, I started fixing the tests to account for the new fee, but it would take more time than I originally thought, so I just rolled back that configuration on DevNets keeping it on TestNet. 

[E2E tests mostly pass](https://github.com/pokt-network/poktroll/actions/runs/9174064847/job/25252596546). There's this e2e test error though, which seems to be related to that PR:
Screenshot 2024-05-21 at 4 44 07 PM

@bryanchriswhite is this error expected?

bryanchriswhite:
Thanks for investigating and for the context @okdas!

> [E2E tests mostly pass](https://github.com/pokt-network/poktroll/actions/runs/9174064847/job/25252596546). There's this e2e test error though, which seems to be related to that PR: Screenshot 2024-05-21 at 4 44 07 PM
> 
> @bryanchriswhite is this error expected?

This is not expected; however, it makes sense if the DevNet state is not reset between E2E test runs and a previous run failed after altering DevNet state. The test attempts to reset all module params to their default values after each scenario but depending on how the test fails, it seems, this may not happen as expected. :thinking: 

**It might be better to add the "unauthorized" key to the config.yaml instead of adding it to the keyring during the test**. I've noticed this also seems to cause errors locally when running `make acc_initialize_pubkeys` after running E2E tests; although, I don't think it actually causes any actual issue.

After re-building and -running CI, it's passing. :+1: Thanks!

bryanchriswhite:
> It might be better to add the "unauthorized" key to the config.yaml instead of adding it to the keyring during the test.

Done in 6c6ec62b4aa3c18488c28b62a69d004361585f80.

bryanchriswhite:
:rotating_light: DO NOT delete this branch until #538 has changed its base branch to main! :rotating_light:

PR #536, opened 2024-05-16
[Bug] Ring address ordering makes valid subsets beings rejected

red-0ne:
## Summary

This PR fixes the `RelayRequest` signature verification process that fails when `delegateeGatewayAddresses` ordering differs at on-chain verification time.

## Issue

The `RelayRequest` verification step checks that the ring addresses used to sign the request are contained within the expected ring at proof verification time. For this we use `ringSig.Ring().Equal(expectedRing)`.

The actual implementation of `Equal` [1] imposes that the request ring pubkeys must be in the **same order** as the expected one.

Using `Equal` in the following `(requestRingAddrs, expectedRingAddrs)` examples would yield undesired results in some cases:
```
* Equals(AB, ABC) == true  // desired
* Equals(BC, ABC) == false // undesired
* Equals(AC, ABC) == false // undesired
```
Given our delegation and undelegation use cases with their dynamic aspect, we want the `(BC, ABC)` and `(AC, ABC)` cases to yield `true`. 

[1] https://github.com/noot/ring-go/blob/master/ring.go#L23

## Type of change

Select one or more:

- [x] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

## Sanity Checklist

### DIFF
--- a/pkg/crypto/rings/client.go
+++ b/pkg/crypto/rings/client.go
@@ -127,7 +127,11 @@ func (rc *ringClient) VerifyRelayRequestSignature(
 	// Get the ring for the application address of the relay request.
 	sessionEndHeight := sessionHeader.GetSessionEndBlockHeight()
 	appAddress := sessionHeader.GetApplicationAddress()
-	expectedAppRingPoints, err := rc.getRingPointsForAddressAtHeight(ctx, appAddress, sessionEndHeight)
+	expectedRelayRingPointsForApp, err := rc.getRingPointsForAddressAtHeight(
+		ctx,
+		appAddress,
+		sessionEndHeight,
+	)
 	if err != nil {
 		return ErrRingClientInvalidRelayRequest.Wrapf(
 			"error getting ring for application address %s: %v", appAddress, err,
@@ -136,7 +140,7 @@ func (rc *ringClient) VerifyRelayRequestSignature(
 
 	// Check that the expected ring signature points map contains the public keys
 	// in the relay request's ring signature.
-	if !ringPointsContain(expectedAppRingPoints, relayRequestRingSig) {
+	if !ringPointsContain(expectedRelayRingPointsForApp, relayRequestRingSig) {
 		return ErrRingClientInvalidRelayRequestSignature.Wrapf(
 			"ring signature in the relay request does not match the expected one for the app %s", appAddress,
 		)
@@ -220,7 +224,9 @@ func (rc *ringClient) addressesToPubKeys(
 }
 
 // getRingPointsForAddressAtHeight returns a map of the ring points for the given
-// application
+// application at a specific height. It takes into account the application itself
+// as well as all the addresses it delegated to. It returns a map of encoded
+// ring points to Point objects (i.e. public keys).
 func (rc *ringClient) getRingPointsForAddressAtHeight(
 	ctx context.Context,
 	appAddress string,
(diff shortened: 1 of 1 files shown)


pkg/crypto/rings/client.go
```diff
@@ -258,3 +287,18 @@ func GetRingAddressesAtBlock(app *apptypes.Application, blockHeight int64) []str
...
 	return activeDelegationsAtHeight
 }
+
+// ringPointsContain checks if the given ring points map contains the public keys
+// in the given ring signature.
+func ringPointsContain(
+	ringPoints map[string]ringtypes.Point,
+	signature *ring.RingSig,
+) bool {
+	for _, point := range signature.PublicKeys() {
```
Olshansk, line 297:
```suggestion
	for _, publicKey := range ringSig.PublicKeys() {
```

pkg/crypto/rings/client.go
```diff
@@ -258,3 +287,18 @@ func GetRingAddressesAtBlock(app *apptypes.Application, blockHeight int64) []str
 
 	return activeDelegationsAtHeight
 }
+
+// ringPointsContain checks if the given ring points map contains the public keys
+// in the given ring signature.
+func ringPointsContain(
+	ringPoints map[string]ringtypes.Point,
+	signature *ring.RingSig,
```
Olshansk, line 295:
```suggestion
	ringSig *ring.RingSig,
```

pkg/crypto/rings/client.go
```diff
@@ -217,6 +219,33 @@ func (rc *ringClient) addressesToPubKeys(
 	return pubKeys, nil
 }
 
+// getRingPointsForAddressAtHeight returns a map of the ring points for the given
+// application
```
Olshansk, line 223:
```suggestion
// application at a specific height. It takes into account the application itself
// as well as all the addresses it delegated to. It returns a map of encoded
// ring points to Point objects (i.e. public keys).
```

pkg/crypto/rings/client.go
```diff
@@ -126,15 +127,16 @@ func (rc *ringClient) VerifyRelayRequestSignature(
 	// Get the ring for the application address of the relay request.
 	sessionEndHeight := sessionHeader.GetSessionEndBlockHeight()
 	appAddress := sessionHeader.GetApplicationAddress()
-	expectedAppRing, err := rc.GetRingForAddressAtHeight(ctx, appAddress, sessionEndHeight)
+	expectedAppRingPoints, err := rc.getRingPointsForAddressAtHeight(ctx, appAddress, sessionEndHeight)
```
Olshansk, line 130: Wdyt of this name instead: `expectedRelayRingPointsForApp`

pkg/crypto/rings/client.go
```diff
@@ -217,6 +219,33 @@ func (rc *ringClient) addressesToPubKeys(
...
+
+	// Get the points on the secp256k1 curve for the public keys in the ring.
+	points, err := pointsFromPublicKeys(ringPubKeys...)
+	if err != nil {
+		return nil, err
+	}
+
+	ringPoints := make(map[string]ringtypes.Point, len(points))
+	for _, point := range points {
+		ringPoints[string(point.Encode())] = point
```
Olshansk, line 242: It's not super clear to me what `string(point.Encode())`. Is it like a hash or something else?

Trying to think if there's a better name to make this clear.

Olshansk (changes requested):
left some comments with suggested namings since it was a little confusing for me so long after looking at this quote. PTAL

Olshansk (approved):

Issue #531, opened 2024-05-14
[Load Testing] Add CLI flag to adapt load testing to Dev/Test Net


red-0ne:
## Objective

Be able to load test DevNet and TestNet with existing staked `Gateway`s and `Supplier`s that are not controlled by the tester.

## Origin Document

![image](https://github.com/pokt-network/poktroll/assets/231488/d0205a8d-b148-4772-993a-7e1e756c8a99)

The current load testing flow involves scaling `Supplier`s and `Gateway`s by staking the accounts corresponding to the pre-provisioned off-chain actors.

This flow is not applicable to persistent chains like in DevNet andTestNets where these actors already exist on-chain and out of control of the tester.

This issue is blocking #136 

## Goals

- Be able to test the already staked `Supplier`s and `Gateway`s on DevNet and TestNet

## Deliverables

- [ ] A PR that adds `--persistent-net` flag to the load testing framework that makes it skip staking `Gateway`s and `Supplier`s

## Non-goals / Non-deliverables

- Complete rewrite to the current load testing framework to comply with the above requirement.

## General deliverables

- [ ] **Comments**: Add/update TODOs and comments alongside the source code so it is easier to follow.
- [ ] **Testing**: Add new tests (unit and/or E2E) to the test suite.
- [ ] **Makefile**: Add new targets to the Makefile to make the new functionality easier to use.
- [ ] **Documentation**: Update architectural or development READMEs; use [mermaid](https://mermaid-js.github.io/mermaid/) diagrams where appropriate.

---

**Creator**: @red-0ne 
**Co-Owners**: @bryanchriswhite

PR #540, opened 2024-05-18
[Testing] Stress / load test ephemeral & persistent chains

red-0ne:
## Summary

Make the necessary adjustments to the load test framework to be able to run on persistent chains.

## Issue

The load testing framework was not able to run without scaling `Gateways` and `Suppliers`. Which is not the goal for testing persistent chains like DevNet and TestNet.

- #531 

## Type of change

Select one or more:

- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

## Sanity Checklist


load-testing/config/load_test_manifest_reader.go
```diff
@@ -11,15 +11,23 @@ import (
 type ProvisionedActorConfig struct {
 	// KeyName is the **name** of the key in the keyring to be used by the given actor.
 	KeyName string `yaml:"key_name"`
+	// Address is the address of the actor, which is used to identify already staked
```
Olshansk, line 14: What does persistent chains mean? If applicable, #PUC

load-testing/config/load_test_manifest_reader.go
```diff
@@ -11,15 +11,23 @@ import (
...
+	// actors in the network in persistent chains.
+	Address string `yaml:"address"`
```
Olshansk, line 16: Why can't we use `keyName` to determine the the address?

load-testing/tests/relays_stress.feature
```diff
@@ -5,13 +5,9 @@ Feature: Loading gateway server with relays
     And a rate of "1" relay requests per second is sent per application
```
Olshansk, line 5: 1. Why was the granularity of this test resolved?
2. If (1) is related to us relying on a "persistent chain" is there a way to have two tests:
- High granularity on ephemeral chain
- Low granularity on persistent chain

load-testing/config/load_test_manifest_reader.go
Olshansk, line 14: What does persistent chain mean? If applicable, #PUC

load-testing/config/load_test_manifest_reader.go
Olshansk, line 16: Why can't we use `keyName` to determine the the address?

1. We already have helpers for this.
2. I feel like this creates two sources of truth.
3. I feel like we're getting into a weird scenario where there's implicit institutional knowledge

load-testing/config/load_test_manifest_reader.go
```diff
@@ -11,15 +11,23 @@ import (
...
 }
 
 // LoadTestManifestYAML is the struct which the load test manifest is deserialized into.
 // It contains the list of suppliers and gateways that the load test expects to be pre-provisioned.
 type LoadTestManifestYAML struct {
-	Suppliers []ProvisionedActorConfig `yaml:"suppliers"`
-	Gateways  []ProvisionedActorConfig `yaml:"gateways"`
+	// PersistentChain is a flag that indicates whether the test is expected to be
+	// run on localnet or long living chains (i.e. TestNet/DevNet).
+	PersistentChain bool                     `yaml:"persistent_chain"`
```
Olshansk, line 26: I'm open to having a discussion about this because it'll involve updating the whole PR, but I'd rather we switch the whole thing from `PersistentChain` to `EphemeralChain`.

- I think the default is that chains are persistent.
- When I read this, I was confused by what a persistent chain is.
- I'd argue that "ephemeral chain" is a lot more clear than "persistent chain" and self explanatory of the alternatives.

Olshansk, line 26: Add `Is` prefix to w/e the final variable ends up being.

load-testing/config/load_test_manifest_reader.go
```diff
@@ -72,5 +78,23 @@ func ParseLoadTestManifest(manifestContent []byte) (*LoadTestManifestYAML, error
 		}
 	}
 
+	if parsedManifest.PersistentChain {
```
Olshansk, line 81: For ephemeral chains (non persistent chains), we need to explain why we don't configure suppliers.

#PUC

load-testing/tests/init_load_test.go
```diff
@@ -11,12 +11,19 @@ import (
...
+	flagManifestFilePath string = "../../loadtest_manifest.yaml"
+	logger               polylog.Logger
 )
 
 func init() {
 	flag.StringVar(&flagLogLevel, "log-level", "", "Specifies the log level for the runner")
+	flag.StringVar(
+		&flagManifestFilePath,
+		"manifest",
+		"../../loadtest_manifest.yaml",
```
Olshansk, line 24: You assigned a default but not using it.

1. Move the default to a constant
2. Don't assign the default to the var
3. Use the constant here

load-testing/tests/relays_stress_helpers_test.go
```diff
@@ -48,9 +48,10 @@ import (
 //
 // TODO_UPNEXT(@red-One, @bryanchriswhite) move to a new file.
 type actorLoadTestIncrementPlans struct {
-	apps      actorLoadTestIncrementPlan
-	gateways  actorLoadTestIncrementPlan
-	suppliers actorLoadTestIncrementPlan
+	apps            actorLoadTestIncrementPlan
+	gateways        actorLoadTestIncrementPlan
+	suppliers       actorLoadTestIncrementPlan
+	persistentChain bool
```
Olshansk, line 54: `isEphemeralChain` would work better

loadtest_manifest.yaml
```diff
@@ -1,6 +1,10 @@
 # List of pre-provisioned suppliers used for load testing
 # The number of pre-provisioned **LocalNet** actors are managed in
 # localnet_config.yaml by the respective actors `count` property.
+persistent_chain: false
```
Olshansk, line 4: `is_ephemeral_chain: true` IMO sounds better.

loadtest_manifest.yaml
```diff
@@ -22,4 +26,15 @@ gateways:
   - key_name: gateway2
     exposed_url: http://localhost:42080
   - key_name: gateway3
-    exposed_url: http://localhost:42081
\ No newline at end of file
+    exposed_url: http://localhost:42081
+
+# Persistent chain manifest example:
```
Olshansk, line 31: Why can't we enable this in this PR?

load-testing/tests/relays_stress_helpers_test.go
```diff
@@ -115,22 +116,33 @@ func (s *relaysSuite) initFundingAccount(fundingAccountKeyName string) {
...
 
-	for _, gateway := range provisionedActors.Gateways {
-		s.gatewayUrls[gateway.KeyName] = gateway.ExposedUrl
+	s.persistentChain = loadTestManifest.PersistentChain
+
+	for _, gateway := range loadTestManifest.Gateways {
+		// In the case of persistent chain, the test has no entry for the gateway
+		// and no keyName is provided. The gateway address is used to identify the
+		// gateway in the network.
+		if s.persistentChain {
```
Olshansk, line 134: 1. The code is going to get really messy with all of these ifs all over the place
2. There is already an implicit assumption that we should only define `Address` or `KeyName` but not both
3. I think this means we may need:
- Mutiple structs
- Multiple functions

It gets messy when you star doing this (and then adding another bool, and another)

load-testing/tests/relays_stress_helpers_test.go
```diff
@@ -457,6 +482,30 @@ func (s *relaysSuite) mapStakingInfoWhenStakingAndDelegatingNewApps(
...
+		// The application is created with the keyName formatted as "app-%d",
+		// starting from 1.
+		application := s.createApplicationAccount(i+1, appFundingAmount)
+		// Add a bank.MsgSend message to fund the application.
+		s.addPendingFundMsg(application.accAddress, sdk.NewCoins(application.amountToStake))
+
+		applications = append(applications, application)
+	}
+
+	if s.persistentChain {
```
Olshansk, line 505: Anywhere we return early on `persistentChain` (or potentially epehmeralChain), we need to add a comment why.

Olshansk (changes requested):
DId a first round of reviews but think we need to iterate a bit.

1. I think `persistent_chain` should be renamed to `is_ephemeral_chain` which would require negating all the ifs & updating lots of comments.

2. Anytime we return early based on the value on the bool, we need to #PUC explaining why.

3. The `keyName` / `address` feels like there's an implicit `oneof` that's not enforced by comments, or the reader, or an explanation, etc... I think there's an opportunity to create structures for it, but am not sure what the simplest path forward is. The goal is to remove the implicit institutional knowledge added in the PR.

load-testing/tests/relays_stress.feature
red-0ne: Not sure if I'm understanding this right.

The RPS is not related to the type of chain tested, this number could be changed to any integer we want.

Obviously we can increase the rate to have a higher one, but since we are parsing it as integer we currently can't go below.

We can offer a lower RPS by either:
1. Parse the input number as a float
2. Change the step to something like `a rate of "<int>" relay requests per "<int> seconds is sent per application`

WDYT?

load-testing/config/load_test_manifest_reader.go
red-0ne: We can't use the `keyName` in that case because the `keyName->address` association comes from the fact that we have that key present the test environment's `keyRing`. Which is not the case for TestNet where `Gateways` are owned and operated by other's and we don't have their keys in our kyring.

I agress that there's no explicit indications on whether to provide `address`es or `keyNames` and feel like comments are not enough.

An alternative would be to have different manifests for each case. It still remains that the `is_ephemeral` flag is the one that dictates whether the test targets a `LocalNet` or a `TestNet`. Rather than the filename.

We may also have dedicated manifest parsers for clarity too.

load-testing/config/load_test_manifest_reader.go
red-0ne: I prefer `ephemeral chain` too, I'll update the code.

loadtest_manifest.yaml
red-0ne: Since the current manifest is used to run the test. The `TestNet` and `LocalNet` scenarios are mutually exclusive, so only one of them could be uncommented/activated.

As mentioned in previous comments, I suggest having a second manifest that exercises testing non-ephemeral chains. We'll indicate which one to trigger by pointing to the corresponding manifest file.

WDYT?

load-testing/tests/relays_stress_helpers_test.go
red-0ne: We considered having multiple functions but they will be almost duplicated (except for what's in these conditionals).

I propose to keep but document them. And refactor if/when we introduce more functionality in the future.

load-testing/tests/init_load_test.go
bryanchriswhite: @Olshansk @red-0ne I would argue that it's be more idiomatic to construct the path with `filepath.Join`. This would also be OS-agnostic.

Olshansk: I can't argue with what @bryanchriswhite is arguing for.

Olshansk:
@red-0ne Please re-request review and update PR title when this is ready.

red-0ne:
* Separated (non-)ephemeral config parsing validation.
* Now we have different manifest for ephemeral and non-ephemeral tests.
* Added Makefile targets for each of the ephemeral and non-ephemeral tests.
* Moved the manifest from the root directory to `load-testing`

Possible improvement:

Be able to specify the test plan (feature files) to execute from the manifest file

load-testing/config/load_test_manifest_reader.go
```diff
@@ -11,15 +11,23 @@ import (
...
 	ExposedUrl string `yaml:"exposed_url"`
 }
 
 // LoadTestManifestYAML is the struct which the load test manifest is deserialized into.
 // It contains the list of suppliers and gateways that the load test expects to be pre-provisioned.
 type LoadTestManifestYAML struct {
-	Suppliers []ProvisionedActorConfig `yaml:"suppliers"`
-	Gateways  []ProvisionedActorConfig `yaml:"gateways"`
+	// IsEphemeralChain is a flag that indicates whether the test is expected to be
+	// run on localnet or long living chains (i.e. TestNet/DevNet).
```
Olshansk, line 25:
```suggestion
	// IsEphemeralChain is a flag that indicates whether the test is expected to
	// run on LocaLnet or a long-living remote chain (i.e. TestNet/DevNet).
```

load-testing/config/load_test_manifest_reader.go
```diff
@@ -36,41 +44,95 @@ func ParseLoadTestManifest(manifestContent []byte) (*LoadTestManifestYAML, error
...
+		if _, err := url.Parse(gateway.ExposedUrl); err != nil {
+			return nil, ErrEphemeralChainLoadTestInvalidManifest.Wrapf("invalid supplier server url: %s", err)
+		}
 	}
 
-	for _, supplier := range parsedManifest.Suppliers {
-		if supplier.KeyName == "" {
-			return nil, ErrLoadTestInvalidManifest.Wrap("empty supplier key name")
+	for _, supplier := range manifest.Suppliers {
+		if len(supplier.KeyName) == 0 {
```
Olshansk, line 86: Why don't we check for `supplier.Address`?

load-testing/config/load_test_manifest_reader.go
```diff
@@ -36,41 +44,95 @@ func ParseLoadTestManifest(manifestContent []byte) (*LoadTestManifestYAML, error
...
+	if len(manifest.ServiceId) == 0 {
+		return nil, ErrEphemeralChainLoadTestInvalidManifest.Wrap("empty service id")
+	}
+
+	for _, gateway := range manifest.Gateways {
+		if len(gateway.KeyName) == 0 {
+			return nil, ErrEphemeralChainLoadTestInvalidManifest.Wrap("empty gateway key name")
+		}
+
+		if len(gateway.Address) > 0 {
```
Olshansk, line 72: #PUC why this is forbidden

load-testing/config/load_test_manifest_reader.go
```diff
@@ -36,41 +44,95 @@ func ParseLoadTestManifest(manifestContent []byte) (*LoadTestManifestYAML, error
...
-			return nil, ErrLoadTestInvalidManifest.Wrap("empty gateway key name")
+	return manifest, nil
+}
+
+func validatedNonEphemeralChainManifest(manifest *LoadTestManifestYAML) (*LoadTestManifestYAML, error) {
+	if len(manifest.Gateways) == 0 {
+		return nil, ErrNonEphemeralChainLoadTestInvalidManifest.Wrap("empty gateways entry")
+	}
+
+	if len(manifest.Suppliers) > 0 {
```
Olshansk, line 107: #PUC for forbidden configs

load-testing/config/load_test_manifest_reader.go
```diff
@@ -36,41 +44,95 @@ func ParseLoadTestManifest(manifestContent []byte) (*LoadTestManifestYAML, error
...
+	if len(manifest.TestNetNode) == 0 {
+		return nil, ErrNonEphemeralChainLoadTestInvalidManifest.Wrap("empty testnet node url")
+	}
+
+	if len(manifest.ServiceId) == 0 {
+		return nil, ErrNonEphemeralChainLoadTestInvalidManifest.Wrap("empty service id")
+	}
+
+	for _, gateway := range manifest.Gateways {
+		if len(gateway.KeyName) > 0 {
```
Olshansk, line 120: #PUC why

load-testing/config/load_test_manifest_reader.go
Olshansk: Can you #PUC with a a short TODO_IMPROVE, and link to this thread (https://github.com/pokt-network/poktroll/pull/540/files#r1607291985) as well?

load-testing/tests/relays_stress_helpers_test.go
```diff
@@ -115,22 +117,36 @@ func (s *relaysSuite) initFundingAccount(fundingAccountKeyName string) {
...
-		s.gatewayUrls[gateway.KeyName] = gateway.ExposedUrl
+	loadTestManifest, err := config.ParseLoadTestManifest(loadTestManifestContent)
+	require.NoError(s, err)
+
+	s.isEphemeralChain = loadTestManifest.IsEphemeralChain
+
+	for _, gateway := range loadTestManifest.Gateways {
+		gatewayId := gateway.Address
+		// In the case of non-ephemeral chain load testing, the test uses the
+		// gateway's keyName as the gatewayId.
```
Olshansk, line 138: #PUC why

Olshansk (approved):
PTAL at the NITs I left related to the comments, but not blocking and LGTM.

Love how this turned out after the first iteration!

load-testing/tests/relays_stress_helpers_test.go
```diff
@@ -517,29 +513,34 @@ func (s *relaysSuite) sendFundAvailableActorsTx(
...
 
 	// Fund accounts for **all** suppliers that will be used over the duration of the test.
-	for i := int64(0); i < plans.suppliers.maxActorCount; i++ {
-		// It is assumed that the suppliers keyNames are sequential, starting from 1
-		// and that the keyName is formatted as "supplier%d".
-		keyName := fmt.Sprintf("supplier%d", i+1)
-		supplier := s.addActor(keyName, supplierStakeAmount)
+	suppliersAdded := int64(0)
+	for _, supplierAddress := range s.availableSupplierAddresses {
+		if suppliersAdded > plans.suppliers.maxActorCount {
```
Olshansk, line 518: Not a blocker, but if we start counting at 0, we reach the max when it's `>=`, not `>`.

Ditto elsewhere.

load-testing/tests/relays_stress_helpers_test.go
```diff
@@ -760,17 +761,16 @@ func (plan *actorLoadTestIncrementPlan) shouldIncrementSupplierCount(
...
-	accAddress, err := keyRecord.GetAddress()
+// address provided in the corresponding provisioned actors slice.
+func (s *relaysSuite) addActor(actorAddress string, actorStakeAmount sdk.Coin) *accountInfo {
+	accAddress, err := sdk.AccAddressFromBech32(actorAddress)
+	keyRecord, err := s.txContext.GetKeyring().KeyByAddress(accAddress)
 	require.NoError(s, err)
+	require.NotNil(s, keyRecord)
 
 	return &accountInfo{
 		accAddress:    accAddress,
```
Olshansk, line 772: Do we still need both `accAddress` and `address` or is there a way to just keep one?

load-testing/tests/relays_stress_test.go
```diff
@@ -77,8 +77,8 @@ var (
 	// maxConcurrentRequestLimit is the maximum number of concurrent requests that can be made.
 	// By default, it is set to the number of logical CPUs available to the process.
 	maxConcurrentRequestLimit = runtime.GOMAXPROCS(0)
-	// fundingAccountKeyName is the key name of the account used to fund other accounts.
-	fundingAccountKeyName = "pnf"
+	// fundingAccountAddress is the key name of the account used to fund other accounts.
+	fundingAccountAddress = "pokt1eeeksh2tvkh7wzmfrljnhw4wrhs55lcuvmekkw"
```
Olshansk, line 81:
```suggestion
	fundingAccountAddress = "pokt1eeeksh2tvkh7wzmfrljnhw4wrhs55lcuvmekkw" // address for pnf account
```

load-testing/tests/relays_stress_test.go
```diff
@@ -217,8 +225,8 @@ type relaysSuite struct {
 
 // accountInfo contains the account info needed to build and send transactions.
 type accountInfo struct {
-	// keyName is the key name of the account available in the keyring used by the test.
-	keyName       string
+	// The address of the account available in the keyring used by the test.
+	address       string
 	accAddress    sdk.AccAddress
```
Olshansk, line 230: We have `address` and `accAddress`. Do we need both?

Olshansk:
@red-0ne Love the simplification, so much clearer now!

Only big remaining question is whether we need both `address` and `accAddress` in `accountInfo`.

I know I can pull the PR and figure it out myself but just asking to expedite things.

load-testing/tests/relays_stress_test.go
red-0ne: The reason why `accAddress` is kept is that `bank.NewMsgSend` and  `KeyRing.GetKeyByAddress` require `AccAddress` structs.

We can create them on the fly though. I'll give their removal a try since we can derive one from the other.

red-0ne:
@Olshansk , `accAddress` removed successfully

Olshansk (approved):
:shipit:

PR #538, opened 2024-05-17
[Relayminer] Query for on-chain session param `num_blocks_per_session`

bryanchriswhite:
## Summary

Refactors the relayminer to query for the current `num_blocks_per_session` **each (or even multiple) time(s) it's used**. This is a naive implementation to just get the relayminer using the on-chain param instead of the const that's still defined in the session keeper pkg but which will be removed in forthcoming work.

## Next Steps
2. Implement `ModuleParamsClient[P any]` (see: [notion - ModuleParamsClient](https://www.notion.so/buildwithgrove/ModuleParamsClient-4af05dff9a0b4a22b157d2d92f2b262f)).
1. Refactor `sessionQuerier` to use `ModuleParamsClient[sessiontypes.Params]` to avoid excessive querying.
1. ...

## Issue

- #517 

## Type of change

Select one or more:

- [x] New feature, functionality or library
- [ ] Bug fix
- [x] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

**Documentation changes** (only if making doc changes)
- [ ] `make docusaurus_start`; only needed if you make doc changes

**Local Testing** (only if making code changes)
- [x] **Unit Tests**: `make go_develop_and_test`
- [x] **LocalNet E2E Tests**: `make test_e2e`
- See [quickstart guide](https://dev.poktroll.com/developer_guide/quickstart) for instructions

**PR Testing** (only if making code changes)
- [x] **DevNet E2E Tests**: Add the `devnet-test-e2e` label to the PR.
    - **THIS IS VERY EXPENSIVE**, so only do it after all the reviews are complete.
    - Optionally run `make trigger_ci` if you want to re-trigger tests without any code changes
    - If tests fail, try re-running failed tests only using the GitHub UI as shown [here](https://github.com/pokt-network/poktroll/assets/1892194/607984e9-0615-4569-9452-4c730190c1d2)

## Sanity Checklist

- [x] I have tested my changes using the available tooling
- [x] I have commented my code
- [x] I have performed a self-review of my own code; both comments & source code
- [x] I create and reference any new tickets, if applicable
- [x] I have left TODOs throughout the codebase, if applicable

----

| :rocket: This description was created by [Ellipsis](https://www.ellipsis.dev) for commit bf24e6c6d0e5199387598c891eacdf4c02d35e6b  | 
|--------|

### Summary:
[...]

PR #546, opened 2024-05-23
refactor: move `num_blocks_per_session` param to shared module

bryanchriswhite:
## Summary

Move `num_blocks_per_session` param from the `session` module to the `shared` module.

## Issue

- #517 

## Type of change

Select one or more:

- [ ] Bug fix
- [x] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

## Sanity Checklist

----

| :rocket: This description was created by [Ellipsis](https://www.ellipsis.dev) for commit aff8853b83378d8bdbc41df624c63689f0b5ffa1  | 
|--------|

### Summary:
Refactor `num_blocks_per_session` parameter from `session` to `shared` module, updating all relevant files and configurations.

**Key points**:
- Moved `num_blocks_per_session` from `session` to `shared` module.
- Updated Go and protobuf files to reflect new parameter location.
- Adjusted tests and configuration to align with updated parameter handling.
- Updated references in `e2e/tests/reset_params_test.go`, `pkg/relayer/session/proof.go`, and `x/proof/keeper/msg_server_submit_proof.go`.
- Added handling for `sharedtypes.ModuleName` in test setup.
- Adjusted session proof handling to use new parameter location.

----
Generated with :heart: by [ellipsis.dev](https://www.ellipsis.dev)


e2e/tests/reset_params_test.go
```diff
@@ -86,6 +87,13 @@ func (s *suite) msgUpdateParamsToDefaultsAny(moduleName string) *codectypes.Any
 				Params:    tokenomicstypes.DefaultParams(),
 			},
 		)
+	case sharedtypes.ModuleName:
```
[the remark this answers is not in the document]
bryanchriswhite: This is test code used to reset module params to their default values in-between scenarios. It has nothing to do with moving the actual parameter.

pkg/relayer/session/proof.go
```diff
@@ -94,7 +98,11 @@ func (rs *relayerSessionsManager) waitForEarliestSubmitProofsHeightAndGeneratePr
 	// submitProofWindowStartBlock is the block that will have its hash used as the
 	// source of entropy for all the session trees in that batch, waiting for it to
 	// be received before proceeding.
-	submitProofWindowStartBlock := rs.waitForBlock(ctx, submitProofsWindowStartHeight)
+	sessionPathBlock := rs.waitForBlock(ctx, sessionGracePeriodEndHeight)
```
red-0ne, line 101: Not sure this wait is effective, `supplierClient.CreateClaim` already waits for the tx to be committed.

I also think we need a better definition of what should be the `sessionPathBlock` and from which block it needs to be derived from. I believe it should be in terms of `submitProofWindowStartHeight`.

red-0ne (commented):
Glad to see this refactor.

I left a couple of comments about the `sessionPathBlock`. Let's make sure to have this 100% right even before having the submission windows.

pkg/relayer/session/proof.go
```diff
@@ -77,11 +77,15 @@ func (rs *relayerSessionsManager) waitForEarliestSubmitProofsHeightAndGeneratePr
...
-	// TODO_TECHDEBT(@red-0ne): Centralize the business logic that involves taking
+	sessionEndHeight := sessionTrees[0].GetSessionHeader().GetSessionEndBlockHeight()
+
+	sessionGracePeriodEndHeight := sessionkeeper.GetSessionGracePeriodBlockCount() + sessionEndHeight
+
+	// TODO_TECHDEBT(#516): Centralize the business logic that involves taking
 	// into account the heights, windows and grace periods into helper functions.
-	submitProofsWindowStartHeight := createClaimHeight + sessionkeeper.GetSessionGracePeriodBlockCount()
-	// TODO_BLOCKER: query the on-chain governance parameter once available.
+	// TODO_BLOCKER(#516): The proof submission window SHOULD NOT overlap with the claim window.
```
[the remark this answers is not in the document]
Olshansk: Seems like it uses `sessionGracePeriodEndHeight + 1`.

Olshansk (approved):
Aside from the comments left by @red-0ne, this LGTM!

pkg/relayer/session/proof.go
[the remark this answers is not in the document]
bryanchriswhite: This section is updated again in #538, I'll incorporate this feedback there:
image

pkg/relayer/session/proof.go
bryanchriswhite: Nice observation! I see your point. :thinking: 

It's especially interesting because this change is what fixed the session E2E test that was failing due to invalid proof paths. I'm going to do some more investigating and come back with more info.

bryanchriswhite: For additional context: The issue this is trying to resolve is that we only get a reference to the block we're waiting for (since we threw out the block query client). Since the block whose hash is used for the proof path seed comes earlier than the block we're actually waiting for here (proof window begin), we need another way get that earlier block hash.

bryanchriswhite: @red-0ne and I had a discussion about this, we think we understand why this is working and have a plan to make it more explicit. 

### Background

The reason we're seeing the "old blocks" in the range is because of an improvement that we made a while back, which populates the `ReplayClient#EventsSequence()` return replay observable's replay buffer, starting from the time of  `ReplayClient` construction (as opposed to starting from when `#EventSequence()` is called). These "old blocks" are observed by the range because they're in the replay buffer, which is emitted by new replay observable subscribers, prior to emitting real-time notifications (not from the replay buffered).

The size of the replay observable's replay buffer sets a limit on how many past blocks (in this case) are observed by `#EventsSequence()` callers (and derivatives thereof; e.g. `#CommittedBlocksSequence`), before being "caught up" and notified of in real-time.

### Proposal
_(In another PR)_

Make `#CommittedBlocksSequence()` and `#EventsSequence()` receive an argument (`replayBufferStartOffset`) which limits the earliest replay buffer notification emitted. This also applies the the analogous `DelegationClient` method.

In this case, the new `replayBufferStartOffset` MUST be large enough such that it ensures that the path block height for the earliest session in the session trees map. In order to satisfy this constraint, in the event that the earliest session path height block was not observed (e.g. cold start w/ old persisted session trees), we also need to add support in the block client to pre-populate the replay buffer by querying with a `BlockQueryClient`, starting from the earliest late session end height.

---

Another bug we exposed is how `mapBlockToSessionsToClaim()` is currently emitting on-time sessions, in the event that late sessions are present. As currently written, on-time sessions would be emitted to observers one block after late sessions. This map is intended to categorize late vs. on-time sessions by grouping them into separate downstream notifications; however, altering the timing of those notifications was not intended. 

To correct for this we need to (re-)re-write the `mapBlockToSessionsToClaim()` `Map()` as a `ForEach()`. @red-0ne had expressed it this way in [a previous iteration](https://github.com/pokt-network/poktroll/blob/7c69ed8aa792d685bfaa0758050c7f625434001a/pkg/relayer/session/session.go#L90).

red-0ne: All settled!

Olshansk (approved):

Issue #494, opened 2024-04-25
[Utility] Per-chain Relays To Token Multiplier


Olshansk:
## Objective

Feature parity with RTTM from Morse.

## Origin Document

- #437
- [Pocket Forum PIP33](https://forum.pokt.network/t/pip33-unleashing-the-potential-of-pocket-as-universal-rpc-provider/4797/8)
- https://github.com/pokt-network/pocket-core/pull/1580

## Goals

- Achieve RTTM parity w/ Morse

## Deliverables

- [ ] **Research Morse**: Study & understand how RTTM works in Morse
- [ ] **Governance**: Add the RTTM governance parameter
- [ ] **Implement**: Implement RTTM in Shannon
- [ ] **Integrate**: Make sure the RTTM values are probability capture in the Sparse Merkle Trie
- [ ] **CLI**:  Update the `poktrolld` (if necessary) to support updating the RTTM governance parameter
- [ ] **Test**: Add unit, integration and E2E tests
- [ ] **Tooling**: Add `Makefile` targets to easily trigger (on LocalNet) and test (e2e) modifying / updating / testing different RTTM values
- [ ] **Document**: Add a new page to dev.poktroll.com explaining how RTTM works and how to use it
- [ ] **Comments**: Add/update TODOs and comments alongside the source code so it is easier to follow.

## Non-goals / Non-deliverables

- Expanding beyond what Morse has for RTTM (except integration into the SMT)

## Estimated Days of Work

**2 days**

_Disclaimer_: This is the total projected number of estimated hours to completion & merge. The owner of this tickets is expected to use this GitHub issue to communicate with the core protocol team along the way, with update & feedback for each deliverable throughout the duration of this work._

---

**Creator**: @Olshansk 
**Co-Owners**: @moatus

rBurgett:
@Olshansk I'd be happy to take this.

Olshansk:
@rBurgett Sounds great!

Few questions / suggestions (some redundancy from our offline conversations but posting here for posterity)

1. **Q1**: When do you think you'll be able to start?
2. **S1**: Please start with a draft PR and cc the team (@bryanchriswhite, @Olshansk, @red-0ne) before diving too deep into the implementation, tests, docs, etc...
3. **S2**: Make sure to look into `ignite` to see if there is any scaffolding opportunity (less work, standards, etc...)

CCing @moatus for visibility.

rBurgett:
[...]

PR #552, opened 2024-05-27
[Utility] Feat: Chain-specific compute unit to tokens multipliers

rBurgett:
## Summary

- Added a `compute_units_per_relay` param to new services
- Updated `SettleSessionAccounting()` to compute units per relay along with the general compute units to tokens multipliers
- Added new tests.
- Updated existing tests
- Added comments where necessary

## Issue
* #494 

## Type of change

Select one or more:

- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

## Sanity Checklist

## Summary by CodeRabbit

- **New Features**
  - Added a `ComputeUnitsPerRelay` field to services, enabling configuration of compute units required per relay.
  - Enhanced the `AddService` command to allow specifying compute units per relay.
  - Introduced a new method to retrieve compute units per relay from application service configurations.

- **Bug Fixes**
  - Improved error handling with a new error for cases where applications lack service configurations.

- **Tests**
  - Added test cases to verify the correct handling of compute units per relay in various scenarios.
  - Updated existing tests to include the new compute units per relay field and ensure compatibility with new features.


proto/poktroll/tokenomics/params.proto
```diff
@@ -15,6 +15,9 @@ message Params {
   // TODO_DOCUMENT(@Olshansk): Make sure to document the units of this parameter (or the map) once finalized.
   uint64 compute_units_to_tokens_multiplier = 1 [(gogoproto.jsontag) = "compute_units_to_tokens_multiplier", (gogoproto.moretags) = "yaml:\"compute_units_to_tokens_multiplier\""];
 
+  // The chain-specific amount of upokt that a compute unit should translate to when settling a session.
+  map<string, uint64> chain_compute_units_to_tokens_multipliers = 2 [(gogoproto.jsontag) = "chain_compute_units_to_tokens_multipliers", (gogoproto.moretags) = "yaml:\"chain_compute_units_to_tokens_multipliers\""];
```
Olshansk, line 19:
```suggestion
  // The service-specific amount of upokt that a compute unit should translate to when settling a session.
  map<string, uint64> service_compute_units_to_tokens_multipliers = 2 [(gogoproto.jsontag) = "service_compute_units_to_tokens_multipliers", (gogoproto.moretags) = "yaml:\"service_compute_units_to_tokens_multipliers\""];
```

proto/poktroll/tokenomics/tx.proto
```diff
@@ -55,6 +55,10 @@ message MsgUpdateParam {
     string as_string = 3 [(gogoproto.jsontag) = "as_string"];
     int64 as_int64 = 6 [(gogoproto.jsontag) = "as_int64"];
     bytes as_bytes = 7 [(gogoproto.jsontag) = "as_bytes"];
+    // cannot add a map<string, int64> here
```
Olshansk, line 58:
```suggestion
    // TODO_IN_THIS_PR: Discuss the issue below
    // cannot add a map<string, int64> here
```

Olshansk: This automatically fails the build. See `Makefile` for more helpers.

proto/poktroll/tokenomics/tx.proto
```diff
@@ -55,6 +55,10 @@ message MsgUpdateParam {
...
+    // do we want to create a custom message type
+    // or just store the chain-specific multipliers
```
Olshansk, line 60: Add a type called:

```proto
message ServiceComputeUnitsToTokenMultiplers {
	map<string, int64> multipliers = 1;
}
```

x/tokenomics/keeper/settle_session_accounting.go
```diff
@@ -103,7 +103,17 @@ func (k Keeper) SettleSessionAccounting(
 	logger.Info(fmt.Sprintf("About to start settling claim for %d compute units", claimComputeUnits))
 
 	// Calculate the amount of tokens to mint & burn
-	settlementAmt = k.getCoinFromComputeUnits(ctx, root)
+	serviceID := sessionHeader.Service.Id
+	multiplier, found := k.GetChainMuliplier(ctx, serviceID)
+	if found {
+		settlementAmt = k.getCoinFromComputeUnitsByChainMultiplier(root, multiplier)
+	} else {
+		// ToDo: should this be an error rather than a warning? should chain-specific multipliers be mandatory?
```
Olshansk, line 111: 1. Use the appropriate TODOs: https://github.com/pokt-network/poktroll/blob/main/Makefile#L456
2. We also have this if you want a reminder to avoid merging the code in: https://github.com/pokt-network/poktroll/blob/6f2fe5b6ecb2db74572ecc32e8fbd6e3b9dc2461/.github/workflows/reviewdog.yml#L11

x/tokenomics/keeper/settle_session_accounting_test.go
```diff
@@ -47,8 +47,8 @@ func TestSettleSessionAccounting_HandleAppGoingIntoDebt(t *testing.T) {
 		SessionHeader: &sessiontypes.SessionHeader{
 			ApplicationAddress: app.Address,
 			Service: &sharedtypes.Service{
-				Id:   "svc1",
-				Name: "svcName1",
+				Id:   "0021",
```
Olshansk, line 50: Let's revert this for now.

We haven't made a decision on service IDs in shannon, but there's an opportunity to make them semantic (e.g. `ETH`), JWTs, structures, etc.... 

We'll discuss this in the future, but let's not port over legacy yet.

Leaving one comment but ditto elsewhere.

x/tokenomics/types/errors.go
```diff
@@ -22,4 +22,5 @@ var (
 	ErrTokenomicsApplicationNewStakeInvalid = sdkerrors.Register(ModuleName, 1115, "application stake cannot be reduced to a -ve amount")
 	ErrTokenomicsParamNameInvalid           = sdkerrors.Register(ModuleName, 1116, "the provided param name is invalid")
 	ErrTokenomicsParamInvalid               = sdkerrors.Register(ModuleName, 1117, "the provided param is invalid")
+	ErrTokenomicsChainMultiplierNotFound    = sdkerrors.Register(ModuleName, 1118, "multiplier for chain ID not found")
```
Olshansk, line 25: s/chain/service

x/tokenomics/types/message_update_param.go
```diff
@@ -47,6 +47,9 @@ func (msg *MsgUpdateParam) ValidateBasic() error {
 	switch msg.Name {
 	case ParamComputeUnitsToTokensMultiplier:
 		return msg.paramTypeIsInt64()
+	case ParamChainComputeUnitsToTokensMultiplier:
+		// see note in tx.proto
```
Olshansk, line 51: 1. Make sure to have TODO_IN_THIS_PR if this comment is temporary
2. #PUC (Please Update Comment) of what we want to see in tx.proto
3. Consider providing full relative path to tx.proto to simplify it for future devs

Olshansk (changes requested):
@rBurgett great start!

In the future, add `TODO_IN_THIS_PR_DISCUSS: ...` so we can keep all the conversations in separate threads in line.

I did an initial quick review (not in depth) to unblock you, and will go deeper during the the next one. I think the answer below and in some comments should provide the context you need.

> Should new parameter be stored as a map<string, int64> or as a string (e.g. 0021=42,0001=42) and then parsed when used? Currently, update messages expect one of string, int64, or bytes and a map is not valid within oneof unless wrapped in a new message. Tests will not pass until I implement the update functionality.

Replied in comments.

> The issue uses the term "chain" but as far as I can find, the project refers to what used to be chains as "services" so should the parameter be called service-specific rather than chain-specific?

Yes.

> I added necessary validation on chain (service) IDs. I could not find any other place in the codebase that checks service IDs. Currently, I check if all characters are uppercase hex with a length of at least four. That covers all current supported chains in Pocket, but 

Can you add a `TODO_BLOCKER_BETA: Discuss and align on what service IDs should be like 

> I don't know what exact requirements are expected or if they have changed with Shannon. In tests, there are chain IDs with non-hex characters. If we would enforce hex, I'd need to update those tests as well.

This is still an open design space in Shannon.

Now is not the right time to dive into it (it will be in a month or so), but I anticipate they will either be semantic strings (e.g. ETH), or a serailized structure that contains metadata.

> Is this meant to replace the compute_units_to_tokens_multiplier param? Or, will we continue supporting a general/default cuttm? Deprecating the general compute_units_to_tokens_multiplier will require updates to many tests.

Yea, let's get rid of the global constant!

proto/poktroll/tokenomics/tx.proto
rBurgett: @Olshansk do we want the property named `ServiceComputeUnitsToTokenMultiplers` or `ServiceComputeUnitsToTokensMultiplers`? The previous property was called `compute_units_to_tokens_multiplier` with a plural `tokens` so I have thus far used that convention. Either works for me, I just want to be consistent.

rBurgett: @Olshansk do we want the tokenomics param to use this type as well or should I just use this for update transactions? Currently, the `service_compute_units_to_tokens_multipliers` param is a raw `map<string, int64>`.

Olshansk: This should be the end state:

```proto
message ComputeUnitsToTokensMultipliers {
	map<string, int64> service_compute_units_to_tokens_map = 1;
}

message Params {
	ComputeUnitsToTokensMultipliers compute_units_to_tokens_multipliers = 1;
}
```

I don't think `MsgUpdateParam` will be able to handle this though so CCing @bryanchriswhite for thoughts / input.


--- 

@rBurgett I don't fully understand the second question. Can you reiterate?

rBurgett: @Olshansk yeah, `MsgUpdateParam` did not like that. You answered my questions with your comment above. I should be good to go. Thanks.

bryanchriswhite: @Olshansk It's not clear to me why you're suggesting to create a new type to hold the map. It seems to me that `Params` can have a map field for this purpose. However, if we do end up using a new type it MUST have the `option (gogoproto.equal = true);` option because `Params` has it and the generated code will not compile otherwise.

Olshansk: @rBurgett If we can avoid the extra wrapper and have the `map` as a raw primtive in `Params`, that does make more sense. I don't recall how/why we went down the other path.

@bryanchriswhite How would the `Update` message be structure if we have a `map` in `Params`?

rBurgett: The issue was the use of `oneof`  within the update message. A map cannot be used there. A raw map is fine in the tokenomics params. But, for the update transaction it needs to be something that can be used in `oneof`  in order to fit the current setup. So, we could serialize it as a string or wrap the map in a message, which you said was ideal. As of now, my testing with using a new message type in the update tx has not worked, with the final serialization failing with an error because it does not implement Message from the cosmos sdk. That is what I have been trying to figure out at the moment.

bryanchriswhite: I see the issue now with `MsgUpdateParam`. :thinking: Using a new type to encapsulate the map would be valid a workaround for this, although it might be a bit verbose.

```protobuf
// proto/poktroll/tokenomics/tx.proto
message MsgUpdateParam {
  ...
  oneof as_type {
    string as_string = 3 [(gogoproto.jsontag) = "as_string"];
    int64 as_int64 = 6 [(gogoproto.jsontag) = "as_int64"];
    bytes as_bytes = 7 [(gogoproto.jsontag) = "as_bytes"];
    StringInt64Map as_string_int64_map = 8 [(gogoproto.jsontag) = "as_string_int64_map"];
  }
}

message StringInt64Map {
  map<string, int64> entries = 1;
}
```

I don't see any reason why we wouldn't be able to continue to extend support for new types of params in this way.

rBurgett: @bryanchriswhite for the moment, I have it implemented as @Olshansk suggested above, although I like the idea of generalizing it as you suggest. It is already verbose, so I don't see that as an issue. I am happy to implement it however you folks prefer.

bryanchriswhite: I would argue that if the reason for changing the data structure like this is to support `MsgUpdateParam`, then the additional types that result should be describing "generic" types that can be used in `as_type`. This would be as opposed to having use-case specific types, like @Olshansk's example.

I also think that these new types should be defined in the `shared` module's `params.proto` so that they can be reused between modules.

Olshansk: `StringInt64Map` lgtm

rBurgett:
@Olshansk I updated my PR moving away from the chain-specific multiplier to a service-based compute units to relay which is combined with the compute units to tokens multiplier when generating rewards. I updated the PR description with my changes. The tests are still all over the place on my computer. I can run the tests individually and they all pass, but when I run them all together with the make file they randomly fail. The e2e tests fail as well, although I am not sure if that is because of my changes or because of the same reasons the other tests randomly fail. I am currently looking at the e2e tests to try and figure out what is going on, but the main functionality has been pushed up and all of the regular tests pass so feel free to look it over and let me know if it is implemented as you expected based on the updated requirements. Thank you!

On a related note, I still cannot get the e2e tests working. I am not sure how to debug it. I thought that I would try running the localnet in order to make sure that works, but that doesn't work either. Following the quickstart guide, I have never been able to get the localnet up since beginning work on this and I don't have the slightest idea how to debug it due to my lack of familiarity with the tools. All attempts so far have been fruitless. The Tilt ui lists everything as "Update pending" except for an error on `protocol-dashboards-label` that says: `The connection to the server 127.0.0.1:37189 was refused - did you specify the right host or port? Build Failed: Command "kubectl label configmap protocol-dashboards grafana_dashboard=1 --overwrite" failed: exit status 1`.

x/shared/types/service.go
```diff
@@ -0,0 +1,10 @@
+package types
+
+// NewService creates a new Service instance
```
Olshansk, line 3: 1. Do you think this is really necessary? Seems like an unnecessary abstraction (for now) given how it maps 1:1.
2. If we do decide to go down this path, add a `TODO_TECHDEBT` so we adopt it everywhere else for consistency (not in this PR)

x/tokenomics/keeper/settle_session_accounting.go
```diff
@@ -184,10 +183,10 @@ func (k Keeper) SettleSessionAccounting(
 	return nil
 }
 
-func (k Keeper) getCoinFromComputeUnits(ctx context.Context, root smt.MerkleRoot) sdk.Coin {
+func (k Keeper) getCoinFromComputeUnits(ctx context.Context, root smt.MerkleRoot, computeUnitsPerRelay uint64) sdk.Coin {
 	// Retrieve the existing tokenomics params
 	params := k.GetParams(ctx)
 
-	upokt := math.NewInt(int64(root.Sum() * params.ComputeUnitsToTokensMultiplier))
+	upokt := math.NewInt(int64(root.Sum() * computeUnitsPerRelay * params.ComputeUnitsToTokensMultiplier))
```
Olshansk, line 190:
```suggestion
        claimedComputeUnits := root.Sum()
	upokt := math.NewInt(int64(claimedComputeUnits * computeUnitsPerRelay * params.ComputeUnitsToTokensMultiplier))
```

x/service/types/message_add_service_test.go
```diff
@@ -36,11 +36,18 @@ func TestMsgAddService_ValidateBasic(t *testing.T) {
 				Service: sharedtypes.Service{Id: "svc1"}, // Name intentionally omitted
 			},
 			expectedErr: ErrServiceMissingName,
+		}, {
+			desc: "valid service supplier address - no compute units per relay",
```
Olshansk, line 40:
```suggestion
			desc: "valid service supplier address - zero compute units per relay",
```

x/service/types/message_add_service.go
```diff
@@ -30,5 +34,16 @@ func (msg *MsgAddService) ValidateBasic() error {
...
 	}
+	if err := ValidateComputeUnitsPerRelay(msg.Service.ComputeUnitsPerRelay); err != nil {
+		return err
+	}
+	return nil
+}
+
+// ValidateComputeUnitsPerRelay makes sure the compute units per relay is a valid value
+func ValidateComputeUnitsPerRelay(computeUnitsPerRelay uint64) error {
+	if computeUnitsPerRelay == 0 {
```
Olshansk, line 45: Wdyt of adding an upper bound to avoid potential overflows w/ all the multiplication?


If agreed, let's:
1. Add a local constant and set it to like 2^16
2. Add a unit test
3. Validate here

x/tokenomics/keeper/settle_session_accounting.go
Olshansk, line 190: TODO_MAINNET: Use `math.BigInt` to avoid potential overflows

x/service/module/tx_add_service_test.go
```diff
@@ -113,11 +125,22 @@ func TestCLI_AddService(t *testing.T) {
...
+					test.service.Name,
+					strconv.FormatUint(test.service.ComputeUnitsPerRelay, 10),
+					fmt.Sprintf("--%s=%s", flags.FlagFrom, test.supplierAddress),
+				}
+			} else {
+				args = []string{
+					test.service.Id,
+					test.service.Name,
+					fmt.Sprintf("--%s=%s", flags.FlagFrom, test.supplierAddress),
+				}
```
Olshansk, line 143: Wdyt of this?

```go
argsAndFlags := []string{
	test.service.Id,
	test.service.Name,
}
if test.service.ComputeUnitsPerRelay > 0 {
	argsAndFlags := append(args, strconv.FormatUint(test.service.ComputeUnitsPerRelay, 10),)
}
argsAndFlags = append(args, fmt.Sprintf("--%s=%s", flags.FlagFrom, test.supplierAddress))
```

Olshansk (changes requested):

Olshansk:
@rBurgett 

1. Did an initial review - PTAL at the comments.
2. Please merge w/ main and update the branch
3. General direction LGTM
4. In the future, please re-request review when you need another look

Re tests: Can you open up a new GitHub issue / bug and provide more details on running E2E and uni tests from main on your machine, we can debug there and cc others.

x/service/module/tx_add_service.go
```diff
@@ -15,18 +16,27 @@ var _ = strconv.Itoa(0)
...
 			serviceNameStr := args[1]
 
+			computeUnitsPerRelay := types.DefaultComputeUnitsPerRelay
+			// if compute units per relay argument is provided
+			if len(args) > 2 {
+				computeUnitsPerRelay, err = strconv.ParseUint(args[2], 10, 64)
+				if err != nil {
+					return types.ErrServiceInvalidComputUnitsPerRelay.Wrapf("unable to parse as uint64: %s", args[2])
+				}
+			}
```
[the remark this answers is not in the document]
rBurgett: @Olshansk from what I see, the loggers are tied to the keepers. In this context, I don't know how to access the appropriate logger. I looked in all of the other CLI handlers and I do not see any previous examples of printing logs from this context. Do you just want a simple `PrintLn` here? Or, would `clientCtx.PrintString()` be appropriate? Or, is there another way to get access to a logger from outside of a keeper?

x/shared/types/service.go
rBurgett: @Olshansk the problem right now is that every time we create a new service instance, we create a literal. Going forward, that makes it very difficult to make updates and identify issues. By using a constructor, it allows us to easily track usages and if we change how we construct them it immediately and clearly breaks the parts of the code that require updating. Constructing literals masks issues because Go automatically adds default values to any property not specified when constructing the literal (and that is part of why Go best practices say to use constructor functions rather than literals). I added the `service.go` file here because everywhere else that we have constructor functions for protobuf messages we have a go file right next to the generated proto file (e.g. `params.pb.go` has a params constructor function in `params.go`). Rather than remove the Service constructor, I would suggest we update every place where we construct a new Service instance to use it. I would be happy to do that and I think it would make the code much more robust. As of now, any of those constructed services in tests have a default `0` value for compute units per relay and that could cause unexpected behavior in the future. If you don't want the constructor here, I could move it into service rather than shared, but that would break with the convention I mentioned above.

x/service/types/message_add_service.go
rBurgett: Sounds good to me.

x/tokenomics/keeper/settle_session_accounting.go
rBurgett: @Olshansk are you saying that I should add that as a comment?

x/service/module/tx_add_service_test.go
```diff
@@ -50,14 +51,16 @@ func TestCLI_AddService(t *testing.T) {
 	// Wait for a new block to be committed
 	require.NoError(t, net.WaitForNextBlock())
 
-	// Prepare two valid services
+	// Prepare three valid services
```
Olshansk, line 54: Where's the third one?

x/service/types/message_add_service.go
```diff
@@ -2,19 +2,24 @@ package types
 
 import (
 	sdk "github.com/cosmos/cosmos-sdk/types"
-
 	"github.com/pokt-network/poktroll/x/shared/types"
 )
 
+const (
+	DefaultComputeUnitsPerRelay uint64 = 1
+	ComputeUnitsPerRelayMax     uint64 = 2 ^ 16
```
Olshansk, line 10: #PUC (Please Update Comment) why we added this

x/tokenomics/keeper/settle_session_accounting.go
```diff
@@ -184,10 +183,11 @@ func (k Keeper) SettleSessionAccounting(
 	return nil
 }
 
-func (k Keeper) getCoinFromComputeUnits(ctx context.Context, root smt.MerkleRoot) sdk.Coin {
+func (k Keeper) getCoinFromComputeUnits(ctx context.Context, root smt.MerkleRoot, computeUnitsPerRelay uint64) sdk.Coin {
```
Olshansk, line 186: NIT: thoughts on renaming this to "computeUnitsToCoins"?

Olshansk (changes requested):
In terms of a draft. LGTM so far.

@bryanchriswhite recently added [1] for everything we need to make it work E2E (with helpers, unit tests, etc..). Can you PTAL?

[1] https://dev.poktroll.com/develop/developer_guide/adding_params

x/service/module/tx_add_service_test.go
rBurgett: Oops, I'll remove that.

x/tokenomics/keeper/settle_session_accounting.go
rBurgett: That would be confusing. It is the compute_units_per_relay param from the service, so imo the name is appropriate.

x/service/module/tx_add_service_test.go
```diff
@@ -81,6 +84,15 @@ func TestCLI_AddService(t *testing.T) {
 			supplierAddress: account.Address.String(),
 			service:         svc1,
 		},
+		{
+			desc:            "valid - add new service without compute units per relay",
```
Olshansk, line 88: Shouldn't we make this invalid?

Olshansk: Looked at the coce. 

Please update test description to explain that it's valid because it defaults to the global value.

x/service/types/message_add_service.go
```diff
@@ -2,19 +2,25 @@ package types
 
 import (
 	sdk "github.com/cosmos/cosmos-sdk/types"
-
 	"github.com/pokt-network/poktroll/x/shared/types"
 )
 
+const (
+	DefaultComputeUnitsPerRelay uint64 = 1
+	// ComputeUnitsPerRelayMax is the maximum allowed compute_units_per_relay value when adding or updating a service
```
Olshansk, line 10:
```suggestion
	// ComputeUnitsPerRelayMax is the maximum allowed compute_units_per_relay value when adding or updating a service.
	// TODO_MAINNET: The reason we have a maximum is to account for potential integer overflows. This is
	// something that needs to be revisited or reconsidered prior to mainnet.
```

x/tokenomics/keeper/settle_session_accounting.go
Olshansk: 1. But this function is on the receiver, not on the service.
2. There is no service in context (not the receiver, not the module, etc...)
3. This function is designed in a "functional" manner: provide computeUnits and get coins. We're not "getting it from the service"

x/shared/types/service.go
Olshansk: Makes sense to me. Please create a GitHub Issues capturing this and add a TODO_TECHDEBT to make sure we adopt it elsewhere.

`TODO_TECHDEBT(#XXX): ...`

Olshansk (changes requested):
ANother few comments.

Still missing:
1. Details from https://dev.poktroll.com/develop/developer_guide/adding_params
2. E2E test

x/tokenomics/keeper/settle_session_accounting.go
rBurgett: @Olshansk I can rename it if you'd like, but I don't get why the compute units per relay would be named compute units to coins. It isn't compute units for coins, it's compute units per relay. The compute units to tokens multiplier tokenomics param defines the coin for compute units.

Olshansk: How about `computeUnitsPerRelayToNumCoins` and we add a `godoc` with a further explanation?

rBurgett: @Olshansk ohhh, you are talking about the function name. The parameter was highlighted, so I thought your comment was about the `computeUnitsPerRelay` parameter. I couldn't figure out why you wanted to name it something completely separate from what it was. I'll change the function name as you suggest.

x/shared/types/service.go
Olshansk: @rBurgett Quick feedback on large block of texts like ☝️ 

I appreciate the thoughtful feedback, but just looking at it is a lof of **cognitive load for a reviewer**.

Requests:
- Try to keep it shorter (if possible)
- Use bullet points (if possible)
- Put yourself in the shoes of the person doing the review

🙏

x/tokenomics/keeper/settle_session_accounting.go
Olshansk: > The parameter was highlighted

That's just hot GitHub works: highlighting the addition.

> I'll change the function name as you suggest.

👍

Olshansk:
@rBurgett What do you think your ETA on finishing this is?

1. Trying to understand if it'll make it into Alpha TestNet #2
   - Assume (hypothetically) hard cutoff is next Friday to be MERGED in
   - This implies code complete by Monday at the latest
3. **Just looking for a confidence level between 1 - 10 provided**:
 - 2.1 Other priorities you have
 - 2.2 The open feedback on this ticket

rBurgett:
@Olshansk the only two remaining things I know are (unless something got lost in all these GitHub comments):

1) the documentation update
2) the e2e test.

I can update documentation, but until I can run a localnet and e2e tests, I can't add a new e2e test. I'm still trying to figure out what is wrong there, but it's a losing battle so far.

If I could successfully run a localnet and e2e tests, for sure I could have it ready by Monday. Otherwise, I can't say.

Olshansk:
@rBurgett 

>  I can't add a new e2e test. I'm still trying to figure out what is wrong there, but it's a losing battle so far.

**Comments:**
1. See the video I embedded in [Quickstart](https://dev.poktroll.com/develop/developer_guide/quickstart) for reference.
2. I'm not going to be reviewing any PRs for the next 3 weeks, so someone else from the team (cc @bryanchriswhite @red-0ne) will have to help out. 
3. In my comment [last week](https://github.com/pokt-network/poktroll/pull/552#issuecomment-2163725397), we talked about creating a new github issue dedicated to this to unblock you. Have you had a chance to create it?

Screenshot 2024-06-20 at 2 05 55 PM

**Call to action:**

1. Please confirm you've seen this
2. 👍 ?
3. Please create a new issue w/ details

rBurgett:
@Olshansk 
1. I have watched the video, but the services do not start up like in the video or the text instructions.
2. Thanks for the heads up.
3. I'll create that issue now.

Olshansk:
@okdas Please see @rBurgett's comment here: https://github.com/pokt-network/poktroll/pull/552#issuecomment-2181658737

I won't be around to provide support on this next week.

rBurgett:
@Olshansk I created an issue. There was all sorts of stuff in the issue template, but it's not my project so I'm not sure what all that was for, I just wanted to put up an issue. I put up what little information I have.

Olshansk:
@rBurgett Do you think you'll still be picking up / finishing off this work?

rBurgett:
@Olshansk you bet. I have learned my way all around the tests and I'm adding the new e2e test and will push the full thing up by Monday at the latest.

rBurgett:
@Olshansk I updated all necessary tests to work with compute units per relay on services. I updated the settlement e2e test to check the compute units per relay which will multiply to the expected reward. That test already checks the reward amount during settlement, so it was not necessary to create another test.

The developer guide says to add any necessary items to the makefile, but it is not clear to me if there is anything that I need to add since we are not updating a keeper's params. I see commands for staking apps and updating top level params, among others. But, there are no makefile commands for updating a service registered within an application (e.g. a command to update the name of a service registered in an application). If there is a makefile command that I need to add, I will need some direction on what you would like to see.

okdas:
I just noticed the PR is open from the repo under `rBurgett` account, not `pokt-network` org, so we won't be able to run e2e tests automatically. So we will have two options:
- Once the PR has been reviewed, we can re-open the PR from a different branch (under `pokt-network/poktroll` repo). That way, we can make sure the changes pass the gates. 
- Run e2e and go tests locally and bypass the gates/checks. I can perform this step if you don't object @Olshansk.

rBurgett:
@Olshansk I'm waiting on one of you for a response. I wrote in my comment a week ago the status of the PR. If you need something else, please let me know. I want to have this finished.

x/service/module/tx_add_service.go
```diff
@@ -1,6 +1,7 @@
 package service
```
Olshansk, line 1: Make sure to review this doc to make sure nothing is missed: https://dev.poktroll.com/develop/developer_guide/adding_params

Olshansk (changes requested):

x/service/module/tx_add_service.go
rBurgett: @Olshansk I did review that document, but I can't see anything that I need to update. There are no docs about services within applications for me to edit and there are no make scripts for updating services either. If I missed them, please let me know.

Olshansk: I didn't do an exhaustive comparison of everything in the document and this PR but here is just one example:

Screenshot 2024-07-19 at 4 18 46 PM

**Required changes:**
1. Add a makefile target that executes an update transaction to update the param you introduced

**Why?**
1. Serve as an example for others
2. This is something we'll need VERY SHORTLY as we start testing it
3. As the person who implemented this feature, it'll be easier/faster for you than whoever picks it up from here
4. It's another way to validate in a real environment (i.e. LocalNet) that the CLI works in conjunction with the real environment.

**Does that make sense?**

I can look through the rest of the document and see if there's anything else.

Olshansk (changes requested):

x/tokenomics/keeper/keeper_settle_pending_claims_test.go
```diff
@@ -59,12 +59,19 @@ func (s *TestSuite) SetupTest() {
...
+		1,
+	)
+
 	// Prepare a claim that can be inserted
 	s.claim = prooftypes.Claim{
 		SupplierAddress: supplierAddr,
 		SessionHeader: &sessiontypes.SessionHeader{
 			ApplicationAddress:      appAddr,
-			Service:                 &sharedtypes.Service{Id: testServiceId},
+			Service:                 &sharedtypes.Service{Id: service.Id},
```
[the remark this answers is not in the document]
rBurgett: @Olshansk the session header is supposed to be `the minimal amount of data required to hydrate & retrieve all data relevant to the session`. That is why only the service ID is included in the header. That is enough to hydrate the service when necessary. When we settle session accounting, we get the full service from the application itself. If we include the data in the tests even though it won't be available in production, it risks people using the header to get service data that isn't available in real life.

x/tokenomics/keeper/settle_session_accounting_test.go
```diff
@@ -52,8 +64,7 @@ func TestSettleSessionAccounting_HandleAppGoingIntoDebt(t *testing.T) {
 		SessionHeader: &sessiontypes.SessionHeader{
 			ApplicationAddress: app.Address,
 			Service: &sharedtypes.Service{
```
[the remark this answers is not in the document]
rBurgett: @Olshansk the session header is `the minimal amount of data required to hydrate & retrieve all data relevant to the session`.

For reference:
https://github.com/pokt-network/poktroll/blob/main/proto/poktroll/session/session.proto#L15

x/service/module/tx_add_service.go
```diff
@@ -1,6 +1,7 @@
...
 
 import (
```
Olshansk, line 3:
```suggestion
// TODO_BETA: Add `UpdateService` or modify `AddService` to `UpsertService` to allow service owners
// to update parameters of existing services. This will requiring updating `proto/poktroll/service/tx.proto` and
// all downstream cod paths.
import (
```

Olshansk (approved):
LGTM! I'll follow up with a few NITs in a followup PR but let's get this in!

Issue #516, opened 2024-05-03
[Protocol] `Claim` & `Proof` Open & Close Windows


Olshansk:
## Objective

Implementation `Claim` & `Proof` windows without hacky tech debt.

## Origin Document

There are many comments and temporary work around in the code. the following is just one of them:

```go
// TODO_BLOCKER: query the on-chain governance parameter once available.
// `* 3` is just a random factor Olshansky added for now to make sure expiration
// doesn't happen immediately after a session's grace period is complete.
submitProofWindowEndHeight := sessionkeeper.GetSessionGracePeriodBlockCount() * 3
```

## Goals

- Resolve a lot of TECHDEBT related to claim / proof windows
- Set the (but not implement) foundation for [probabilistic proofs](https://github.com/pokt-network/pocket-core/blob/staging/docs/proposals/probabilistic_proofs.md)
- Implement some of the core macro utility of the protocol

## Deliverables

**_Reminder: This can be multiple PRs_** 

- [x] Add governance parameters for the following:
	- [x] `sessionkeeper.ClaimOpenWindowMinNumBlocks` - The minimum number of blocks (after a session ends) when a claim can be sent. Should be functionally independent of session grace period, but they will be "tied at the hip" in the implementation.
  		- #561 
	- [x] `sessionkeeper.ClaimCloseWindowMaxNumBlocks` - The max number of blocks after a session ends when claims are no longer accepted. It must strictly greater than when a claim window opens.
  	  	- #568 
	- [x] `sessionkeeper.ProofOpenWindowMinNumBlocks` - The min number of blocks (after a session ends) when a proof can be sent. It MUST be after the claim window closes.
		- #569 
	- [x] `sessionkeeper.SubmitProofCloseWindowNumBlocks` - The max number of block after a session ends. It must be strictly greater than when a proof window opens.
		- #570 
- [ ] Techdebt - search for, identify, update and cleanup all comments / code / functions related to documented tech debt.
- [ ] Business logic - implement the actual thing
	- #565 
	- #566
	- #592
	- #588
- [ ] Testing
	- [x] 	Ensure we have at least 1 E2E test for each of the governance parameters above
	- [x] 	Ensure we have at least 1 integration (in memory) test for each of the governance parameter sabove
- [ ] Add a `TODO_BLOCKER` where appropriate for the non-goals below
- [ ] Makefile targets to:
	- [x] Modify / update the parameters
	- [ ] Run the relevant E2E tests
[...]

PR #561, opened 2024-05-28
[Claim/proof] feat: add `claim_window_open_offset_blocks` shared module param

bryanchriswhite:
## Summary

- Adds `ClaimWindowOpenOffsetBlocks` shared module param
- Integrates it into the relayminer
- Integrates it into the proof module's on-chain verification

## Issue

- #516

## Type of change

Select one or more:

- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

## Sanity Checklist

### DIFF
--- a/x/shared/types/params_test.go
+++ b/x/shared/types/params_test.go
@@ -40,3 +40,33 @@ func TestParams_ValidateNumBlocksPerSession(t *testing.T) {
 		})
 	}
 }
+
+func TestParams_ValidateClaimWindowOpenOffsetBlocks(t *testing.T) {
+	tests := []struct {
+		desc                        string
+		claimWindowOpenOffsetBlocks any
+		err                         error
+	}{
+		{
+			desc:                        "invalid type",
+			claimWindowOpenOffsetBlocks: "invalid",
+			err:                         ErrSharedParamInvalid.Wrapf("invalid parameter type: %T", "invalid"),
+		},
+		{
+			desc:                        "valid ClaimWindowOpenOffsetBlocks",
+			claimWindowOpenOffsetBlocks: uint64(4),
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.desc, func(t *testing.T) {
+			err := ValidateClaimWindowOpenOffsetBlocks(tt.claimWindowOpenOffsetBlocks)
+			if tt.err != nil {
+				require.Error(t, err)
+				require.Contains(t, err.Error(), tt.err.Error())
+			} else {
+				require.NoError(t, err)
+			}
+		})
+	}
+}


proto/poktroll/shared/params.proto
```diff
@@ -12,4 +12,5 @@ message Params {
   option (gogoproto.equal) = true;
 
   uint64 num_blocks_per_session = 1 [(gogoproto.jsontag) = "num_blocks_per_session"];
+  uint64 claim_window_open_offset_blocks = 2 [(gogoproto.jsontag) = "claim_window_open_offset_blocks"];
```
Olshansk, line 15: #PUC on what this is, what its for, how its used.

(e.g. number of blocks after session ends ..)

x/shared/types/params.go
```diff
@@ -65,3 +74,14 @@ func ValidateNumBlocksPerSession(v interface{}) error {
 
 	return nil
 }
+
+// ValidateClaimWindowOpenOffsetBlocks validates the ClaimWindowOpenOffsetBlocks param
+// NB: The argument is an interface type to satisfy the ParamSetPair function signature.
+func ValidateClaimWindowOpenOffsetBlocks(v interface{}) error {
```
Olshansk, line 80: Let's ensure it's greater than zero and add a test.

Olshansk (changes requested):

x/shared/types/params.go
bryanchriswhite: I could imagine adding a test for the error case, where the input is not a uint64 but I don't think it's totally necessary. `ValidateClaimWindowOpenOffsetBlocks()` is only called in two places:

image
> Fig. 1. Usages.

image
> Fig 2. Usage in `shared.msgServer#UpdateParam()`, where it is guaranteed to be a uint64.

image
> Fig 3. Usage in `sharedtypes.Params#ParamSetPairs()`, which is called by the cosmos-sdk framework and seems to be related to migrations somehow.

Olshansk: Whether it's an explicit check or a unit test, either works.

proto/poktroll/shared/params.proto
red-0ne: Shouldn't we start using this in https://github.com/pokt-network/poktroll/blob/issues/516/feat/claim-window-open-offset-blocks/pkg/relayer/session/claim.go#L94

Olshansk: ❓

bryanchriswhite: @red-0ne It's in the pipeline: https://github.com/pokt-network/poktroll/pull/565/files#diff-78aa6b89c55aadb85f27b5f9af9cd4b4264dc1389e64b9ada24fae6f99cc9791R93. This PR is just adding the `ClaimWindowOpenOffsetBlocks` param.

red-0ne (commented):
Shouldn't we have a method in the `sharedquerier` client to query this param?

Olshansk (approved):

PR #562, opened 2024-05-28
[LoadTest] Switch funding account to key name instead of address

okdas:
## Summary

This change allows to run the load test against TestNet the same way we run it against LocalNet.

## Type of change

Select one or more:

- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

## Sanity Checklist

### DIFF
--- a/load-testing/loadtest_manifest.yaml
+++ b/load-testing/loadtest_manifest.yaml
@@ -6,6 +6,9 @@ is_ephemeral_chain: false
 testnet_node: https://testnet-validated-validator-rpc.poktroll.com
 # The service ID to request relays from.
 service_id: "0021"
+# The address of the account that will be used to fund the the application accounts
+# so that they can stake on the network.
+funding_account_address: pokt14eh973xt99s7edugnyvd5d5u50d6j0hysw2vsm
 # In non-ephemeral chains, the gateways are identified by their address.
 gateways:
   # address is the bech32 pokt address of the gateway.
--- a/load-testing/loadtest_manifest.yaml
+++ b/load-testing/loadtest_manifest.yaml
@@ -8,7 +8,7 @@ testnet_node: https://testnet-validated-validator-rpc.poktroll.com
 service_id: "0021"
 # The address of the account that will be used to fund the the application accounts
 # so that they can stake on the network.
-funding_account_address: pokt14eh973xt99s7edugnyvd5d5u50d6j0hysw2vsm
+funding_account_address: pokt14eh973xt99s7edugnyvd5d5u50d6j0hysw2vsm # address for pnf account
 # In non-ephemeral chains, the gateways are identified by their address.
 gateways:
   # address is the bech32 pokt address of the gateway.
--- a/load-testing/localnet_loadtest_manifest.yaml
+++ b/load-testing/localnet_loadtest_manifest.yaml
@@ -6,7 +6,7 @@ is_ephemeral_chain: true
 service_id: anvil
 # The address of the account that will be used to fund the the application,
 # gateway and supplier accounts so that they can stake on the network.
-funding_account_address: pokt1eeeksh2tvkh7wzmfrljnhw4wrhs55lcuvmekkw
+funding_account_address: pokt1eeeksh2tvkh7wzmfrljnhw4wrhs55lcuvmekkw # address for pnf account
 
 # List of pre-provisioned suppliers used for load testing.
 # Thse suppliers will be progressively staked during the load test, according
--- a/load-testing/localnet_loadtest_manifest.yaml
+++ b/load-testing/localnet_loadtest_manifest.yaml
@@ -4,6 +4,9 @@
 is_ephemeral_chain: true
 # The service ID to use for the load test.
 service_id: anvil
+# The address of the account that will be used to fund the the application,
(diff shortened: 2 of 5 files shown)


load-testing/tests/relays_stress_helpers_test.go
```diff
@@ -102,17 +102,17 @@ func (s *relaysSuite) setupTxEventListeners() {
 }
 
 // initFundingAccount initializes the account that will be funding the onchain actors.
-func (s *relaysSuite) initFundingAccount(fundingAccountAddress string) {
+func (s *relaysSuite) initFundingAccount(fundingAccountKeyName string) {
 	// The funding account record should already exist in the keyring.
-	accAddress, err := sdk.AccAddressFromBech32(fundingAccountAddress)
+	fundingAccountKeyRecord, err := s.txContext.GetKeyring().Key(fundingAccountKeyName) //KeyByAddress(accAddress)
```
red-0ne, line 107:
```suggestion
	fundingAccountKeyRecord, err := s.txContext.GetKeyring().Key(fundingAccountKeyName)
```

load-testing/tests/relays_stress_test.go
```diff
@@ -77,8 +77,8 @@ var (
 	// maxConcurrentRequestLimit is the maximum number of concurrent requests that can be made.
 	// By default, it is set to the number of logical CPUs available to the process.
 	maxConcurrentRequestLimit = runtime.GOMAXPROCS(0)
-	// fundingAccountAddress is the key name of the account used to fund other accounts.
-	fundingAccountAddress = "pokt1eeeksh2tvkh7wzmfrljnhw4wrhs55lcuvmekkw" // address for pnf account
+	// fundingAccountKeyName is the key name of the account used to fund other accounts.
+	fundingAccountKeyName = "pnf" // address for pnf account
```
red-0ne, line 81:
```suggestion
	fundingAccountKeyName = "pnf" // key name for pnf account
```

red-0ne (approved):
Thanks for putting in the PR!

There's a couple of minor changes, but preemptively approving :+1:

load-testing/tests/relays_stress_test.go
```diff
@@ -349,7 +349,7 @@ func (s *relaysSuite) LocalnetIsRunning() {
 	s.setupTxEventListeners()
 
 	// Initialize the funding account.
-	s.initFundingAccount(fundingAccountAddress)
+	s.initFundingAccount(fundingAccountKeyName)
```
Olshansk, line 352: @red-0ne I would push back on this. 

An address is the same across every environment and a single source of truth.

Anyone can import a private key with a different name.

IMO we should update all environments to use addresses instead of key names

Olshansk (changes requested):

load-testing/tests/relays_stress_test.go
okdas: > Anyone can import a private key with a different name.

Anyone can also change the code to replace hard-coded addresses? Not sure what problem we're solving by using addresses in code. We can make address an environment variable as a workaround for this test.

Olshansk: - keynames are subjective (can be different on different machines for the same address)
- addresses are objective (there is no room for discrepancy, confusion or cognitive overload)

Getting ahead of the conversations in the future

red-0ne: The issue comes from the fact that different environments may have different funding addresses (pnf address was different in LocalNet and TestNet).

I agree that a keyname may reference different addresses, that in the future we'd even have to change it to `faucet` or similar and implicit changes to the keyname (e.g. regenesis, homedir change, config.yaml...) may even be harder to debug.

At the same time I understand @okdas 's concerns about changing the code just to run a TestNet load test.

While having this address in an environment variable might solve the issue, I would rather prefer having the manifest as a source of truth and introduce a `funding_account_address` entry and remove the hardcoded variable.

load-testing/loadtest_manifest.yaml
```diff
@@ -6,6 +6,9 @@ is_ephemeral_chain: false
 testnet_node: https://testnet-validated-validator-rpc.poktroll.com
 # The service ID to request relays from.
 service_id: "0021"
+# The address of the account that will be used to fund the the application accounts
+# so that they can stake on the network.
+funding_account_address: pokt14eh973xt99s7edugnyvd5d5u50d6j0hysw2vsm
```
Olshansk, line 11: #PUC if this is `faucet` or `pnf` as an extra layer of info

load-testing/localnet_loadtest_manifest.yaml
```diff
@@ -4,6 +4,9 @@
 is_ephemeral_chain: true
 # The service ID to use for the load test.
 service_id: anvil
+# The address of the account that will be used to fund the the application,
+# gateway and supplier accounts so that they can stake on the network.
+funding_account_address: pokt1eeeksh2tvkh7wzmfrljnhw4wrhs55lcuvmekkw
```
Olshansk, line 9: Ditto

Olshansk (approved):
Couple of small comments but love the final solution!

PR #563, opened 2024-05-28
[Ring] Replace `noot/ring-go` with `pokt-network/ring-go`

Olshansk:
Replace the external ring package with our own fork.


red-0ne (approved):
Glad to see that we're using an internally maintained `ring-go`!

Issue #528, opened 2024-05-13
[RelayMiner] RelayMiner Lean Client


RawthiL:
## Objective

Enable a `N:1 Supplier:RelayMiner` relation, i.e. connect multiple Suppliers to the same RelayMiner process.

This is required for mid-large node runners operations and economical constraints.

## Origin Document

[This comment](https://github.com/pokt-network/poktroll/pull/524/files#diff-d93c4dd4583de84916aab1b03b2819c4d53aff7aa4e269af16648ac1c75b7c06R157) refers the a `N:1 Supplier:RelayMiner` relation as "not okay"  when it is fundamental for mid-large node runners.

Without this feature the hardware costs of these operations will be at least 6 times higher due to hardware requirements.

More context [at this discord thread](https://discord.com/channels/553741558869131266/1234943674903953529/1239636373825257654).

Related Lean Pocket data:
- Morse Issue: https://github.com/pokt-network/pocket-core/issues/1437
- Forum discussion of original problem that lead to LeanPOKT: https://forum.pokt.network/t/optimizing-the-v0-client-to-save-the-network-a-ton-of-money/2510
- Morse Implementation
	- https://github.com/pokt-network/pocket-core/pull/1463
	- https://github.com/pokt-network/pocket-core/pull/1488

## Goals

- Allow the connection of multiple Suppliers to a single RelayMiner
- A single RelayMiner process should be at least able to handle 700 Supliers to match LeanPOKT capabilities
- Reduce the costs for RelayMiner operators that support multiple Suppliers

## Deliverables

- [ ] A PR to enable the RelayMiner process to have multiple signing `Supplier`s:
  - Update the config parser code (`pkg/relayer/config`) to:
    - Allow `signing_key_names` array in the `suppliers` entries.
    - Have a top level `default_signing_key_names` that act as a default value for the `suppliers` entries that lack `signing_key_names`
  - Update the `RelayRequestMetadata` proto (`proto/poktroll/relay.proto`) to add the `supplier_address` field
  - Update the `relayerProxy` struct (`pkg/relayer/proxy`) to:
    - Remove the `rp.signingKeyName` entry.
    -  Use `rp.serverConfigs.SupplierConfigsMap` as a source of truth for the available signing keys
    - Check that `RelayRequestMetadata.SupplierAddress` is effectively present in `rp.serverConfigs.SupplierConfigsMap[ServiceId].SigningKeyNames` and serving the requested `ServiceId`
[...]

PR #567, opened 2024-05-30
[RelayMiner] LeanClient

okdas:
## Summary

Adds an ability to run many suppliers on one relayminer

## Issue

- #528 

## Type of change

Select one or more:

- [ ] Bug fix
- [ ] Code health or cleanup
- [x] Documentation
- [ ] Other (specify)

## Testing

- [x] `make docusaurus_start`; only needed if you make doc changes

## Sanity Checklist


pkg/relayer/cmd/cmd.go
```diff
@@ -202,6 +202,7 @@ func setupRelayerDependencies(
 		config.NewSupplyRingCacheFn(),
 		supplyTxFactory,
 		supplyTxContext,
+		// TODO_IN_THIS_PR: should we have many txclients - one for each key?
```
okdas, line 205: @red-0ne what do you think we should do about proof/claim code? I've started going down the path of changing `supplierClient` and `txClient` to support multiple `signingKeyNames`, but that's a lot of changes and I'm not sure if that's the right thing to do. I'm thinking we can add an address to `sessionTree`, that way we know which client/supplier to reference when submitting transactions. Or maybe there should be multiple `sessionTree`s for each address? So many questions! :D 

Handling relays seems straightforward enough so far.

Olshansk: Without having full context or diving into the code, IMO the best approach would be:

- supplierClientMap: map (signing address) -> supplierClient
- txClientMap: map (signing address) -> txClient

Seems like a good `SOLID` design.

red-0ne: We definitely need to add the supplier address in the `SessionTree` struct [1]. We will then group messages by supplier when claiming [2] and proving [3]

We can have a `supplierClientMap` and a `txClientMap` and then match the C&P tx signers. This will involve changing what is supplied with `depinject` [4]. It will also create a websocket connection for each txClient.

Another approach would be to have multiple signers in the `SupplierClient` and `txClient` and will have to change the tx listener [5] to catch transactions for all the suppliers.

[1] https://github.com/pokt-network/poktroll/blob/main/pkg/relayer/session/sessiontree.go#L21
[2] https://github.com/pokt-network/poktroll/blob/main/pkg/relayer/session/claim.go#L156
[3] https://github.com/pokt-network/poktroll/blob/main/pkg/relayer/session/proof.go#L190
[4] https://github.com/pokt-network/poktroll/blob/main/pkg/relayer/cmd/cmd.go#L206-L207
[5] https://github.com/pokt-network/poktroll/blob/main/pkg/client/tx/client.go#L172

Olshansk: > Another approach would be to have multiple signers in the SupplierClient and txClient and will have to change the tx listener [5] to catch transactions for all the suppliers.

👎 This doesn't follow good encapsulation principles IMO.

> We can have a supplierClientMap and a txClientMap and then match the C&P tx signers. This will involve changing what is supplied with depinject [4]. It will also create a websocket connection for each txClient.

👍  Seems simple, extendible, easier to debug, etc...

What @red-0ne said

pkg/relayer/session/session.go
```diff
@@ -123,17 +132,22 @@ func (rs *relayerSessionsManager) Start(ctx context.Context) {
...
-	rs.relayObs.UnsubscribeAll()
+	for _, v := range rs.relayObs {
+		v.UnsubscribeAll()
+	}
 }
 
 // SessionsToClaim returns an observable that notifies when sessions are ready to be claimed.
 func (rs *relayerSessionsManager) InsertRelays(relays relayer.MinedRelaysObservable) {
+	// TODO_IN_THIS_PR: potentially use a map[keyname/address]relayObs instead
+	// TODO_IN_THIS_COMMIT: figure out how to get an address from relays, or add an argument to `InsertRelays`.
```
okdas, line 143: @red-0ne what do you think the best way to get the address here?

Olshansk: Here is my thought procs:

1. `relayerSessionsManager` is a receiver. I wonder what's inside of that?
2. I went to `sesison.go` and found:

```go
type relayerSessionsManager struct {
	logger polylog.Logger

	relayObs relayer.MinedRelaysObservable

	// sessionsToClaimObs notifies about sessions that are ready to be claimed.
	sessionsToClaimObs observable.Observable[[]relayer.SessionTree]

	// sessionTrees is a map of block heights pointing to a map of SessionTrees
	// indexed by their sessionId.
	// The block height index is used to know when the sessions contained in the entry should be closed,
	// this helps to avoid iterating over all sessionsTrees to check if they are ready to be closed.
	sessionsTrees   sessionsTreesMap
	sessionsTreesMu *sync.Mutex

	// blockClient is used to get the notifications of committed blocks.
	blockClient client.BlockClient

	// supplierClient is used to create claims and submit proofs for sessions.
	supplierClient client.SupplierClient

	// pendingTxMu is used to prevent concurrent txs with the same sequence number.
	pendingTxMu sync.Mutex

	// storesDirectory points to a path on disk where KVStore data files are created.
	storesDirectory string

	// sharedQueryClient is used to query shared module parameters.
	sharedQueryClient client.SharedQueryClient
}
```
3. I found a bunch of clients such as `sharedQueryClient client.SharedQueryClient`
4. I started wondering what other clients exist so I clicked into `client.SharedQueryClient`
5. That took me to `interface.go`
6. I did a ctrl-f for `AccountQueryClient` and found this:
```go
// AccountQueryClient defines an interface that enables the querying of the
// on-chain account information
type AccountQueryClient interface {
	// GetAccount queries the chain for the details of the account provided
	GetAccount(ctx context.Context, address string) (cosmostypes.AccountI, error)

	// GetPubKeyFromAddress returns the public key of the given address.
	GetPubKeyFromAddress(ctx context.Context, address string) (cryptotypes.PubKey, error)
}
```
7. My intuition is telling me that I just need to update the `relayerSessionsManager` to inject an instance of `AccountQueryClient`.
8. I would look at how this is done in other part of the codebase and copy-pasta

@okdas Does that help?

red-0ne: Not sure what you want to achieve here, but I don't think it will be useful to try to split the relays by `Supplier` there.

You can notice that this pipe has a dead end at `mapAddMinedRelayToSessionTree`. Which is just adding `Relay`s to its corresponding `SMT`.

A better place where you can group `SessionTree`s by `Supplier` would be at `newMapClainSessionsFn`.

There [1] you could have `sessionClaims` to be a `map[supplierAddress][]*relayer.SessionClaim` instead of `[]*relayer.SessionClaim`

And do (error handling is omitted)
```go
for _, session := range sessionTrees {
  address := session.supplierAddress
  sessionClaims[address] = append(sessionClaims[address], relayer.SessionClaim)
}

for supplierAddress, supplierClaims := sessionClaims {
  rs.supplierClientMap[supplierAddress].CreateClaims(ctx, supplierclaims)
}
```

You do the same thing for proofs [2]

[1] https://github.com/pokt-network/poktroll/blob/main/pkg/relayer/session/claim.go#L155
[2] https://github.com/pokt-network/poktroll/blob/main/pkg/relayer/session/proof.go#L151

okdas: @red-0ne thanks - that's precisely what I planned to do (there's a todo in that section of the file). I originally thought we'd also need to group the observables, but it doesn't seem to be the case. Thank you! I think I'm getting really close now and will send a PR for a review before my EOD.

pkg/client/supplier/client.go
```diff
@@ -41,7 +41,7 @@ func NewSupplierClient(
 
 	if err := depinject.Inject(
 		deps,
-		&sClient.txClient,
+		&sClient.txClients,
```
okdas, line 44: @red-0ne I know we talked about not trying to make depinject happy for now, but I think I'm stuck with this issue now:
```
can't resolve type github.com/pokt-network/poktroll/pkg/client/client.TxClientMap for github.com/pokt-network/poktroll/pkg/client/supplier.NewSupplierClient (/Users/dk/pocket/poktroll/pkg/client/supplier/client.go:42):
	while resolving:
		client.TxClientMap for github.com/pokt-network/poktroll/pkg/client/supplier.NewSupplierClient (/Users/dk/pocket/poktroll/pkg/client/supplier/client.go:42)
```

The supplier client should have the same issue (they are very look alike). What do you think I might have missed? Thanks!

okdas:
@Olshansk @red-0ne E2E tests are green. I think this PR is ready for the first pass of reviews. Please let me know what you think and if I need to change anything.

I know the go tests are not fixed yet, and this is what I'm going to be focusing on next.

Thanks!

pkg/client/supplier/client.go
```diff
@@ -143,6 +143,10 @@ func (sClient *supplierClient) CreateClaims(
 	return <-errCh
 }
 
+func (sClient *supplierClient) Address() *cosmostypes.AccAddress {
```
Olshansk, line 146: 1. Is this being used anywhere?
2. godoc please

pkg/relayer/cmd/cmd.go
```diff
@@ -254,6 +254,7 @@ func supplyTxContext(
 	deps depinject.Config,
 	_ *cobra.Command,
 ) (depinject.Config, error) {
+	// TODO_IN_THIS_PR: do we need new context per key?
```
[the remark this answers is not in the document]
Olshansk: Let's change it to a TODO_CONSIDERATION

pkg/relayer/proxy/relay_verifier.go
```diff
@@ -75,7 +78,9 @@ func (rp *relayerProxy) VerifyRelayRequest(
 
 	// Check if the relayRequest is allowed to be served by the relayer proxy.
 	for _, supplier := range session.Suppliers {
-		if supplier.Address == rp.supplierAddress {
+		// TODO_IN_THIS_PR: does that make sense?
```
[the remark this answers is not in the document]
Olshansk: Yes, it's great!

pkg/relayer/proxy/relay_signer.go
```diff
@@ -10,9 +10,9 @@ import (
 // TODO_TECHDEBT(@red-0ne): This method should be moved out of the RelayerProxy interface
 // that should not be responsible for signing relay responses.
 // See https://github.com/pokt-network/poktroll/issues/160 for a better design.
-func (rp *relayerProxy) SignRelayResponse(relayResponse *types.RelayResponse) error {
+func (rp *relayerProxy) SignRelayResponse(relayResponse *types.RelayResponse, supplierAddr string) error {
 	// create a simple signer for the request
-	signer := signer.NewSimpleSigner(rp.keyring, rp.signingKeyName)
+	signer := signer.NewSimpleSigner(rp.keyring, rp.supplierAddresses[supplierAddr])
```
Olshansk, line 15: Let's make sure `supplierAddr` is present to avoid an accidental panic.

pkg/relayer/proxy/proxy.go
```diff
@@ -25,10 +25,10 @@ var _ relayer.RelayerProxy = (*relayerProxy)(nil)
 type relayerProxy struct {
 	logger polylog.Logger
 
-	// signingKeyName is the supplier's key name in the Cosmos's keybase. It is used along with the keyring to
+	// signingKeyNames are the supplier's key name in the Cosmos's keybase. They are used along with the keyring to
 	// get the supplier address and sign the relay responses.
```
Olshansk, line 29:
```suggestion
	// signingKeyNames are the supplier key name in the Cosmos's keybase.
	// They are used along with the keyring to get the supplier address and
	// sign the relay responses.
```

pkg/relayer/proxy/proxy.go
```diff
@@ -63,8 +63,10 @@ type relayerProxy struct {
 	// ringCache is used to obtain and store the ring for the application.
 	ringCache crypto.RingCache
 
-	// supplierAddress is the address of the supplier that the relayer proxy is running for.
-	supplierAddress string
+	// supplierAddresses are the address of the suppliers that the relayer proxy is running for.
+	// We use this in relay verification to check if we should process the relay, and to
+	// resolve which signing key name to use for signing.
+	supplierAddresses map[string]string
```
Olshansk, line 69: 1. Rename to `XtoYMap`

pkg/relayer/proxy/proxy.go
```diff
@@ -63,8 +63,10 @@ type relayerProxy struct {
 	// ringCache is used to obtain and store the ring for the application.
 	ringCache crypto.RingCache
 
-	// supplierAddress is the address of the supplier that the relayer proxy is running for.
-	supplierAddress string
+	// supplierAddresses are the address of the suppliers that the relayer proxy is running for.
```
Olshansk, line 66: Explain what the key and values are please.

proto/poktroll/service/relay.proto
```diff
@@ -22,6 +23,10 @@ message RelayRequestMetadata {
   // application has delegated to. The signature is made using the ring of the
   // application in both cases.
   bytes signature = 2;
+
+  // The supplier address the relay is sent to. It is being used on the RelayMiner to 
+  // route to the correct supplier.
```
Olshansk, line 28: This is a little weird.

I'm not opposed to it, but we're changing one of the most fundamental structures in the protocol (that'll be in every single leaf) solely for the reason of `It is being used on the RelayMiner to  route to the correct supplier`.

@red-0ne Do you have thoughts here?
@okdas can you add a `TODO_IN_THIS_PR: Re-evaluate` so we don't forget.

Olshansk: If we do keep this, we'll also need to update the on-chain logic.

We'll need to verifythat the signer of the response equals to the public key of the address here.

If this is not done, we're going to find ourselves in a place where weird shit can happen between the metadata and the signatures.

localnet/kubernetes/values-relayminer-1.yaml
```diff
@@ -1,5 +1,5 @@
 config:
-  signing_key_name: supplier1
+  default_signing_key_names: [supplier1]
```
Olshansk, line 2: Why is `default_signing_key_names` an array? I expect ONE default.

NO need for an explanation here in comments, let's either

1. Update the documentation (#PUC)
2. Change it

pkg/client/interface.go
```diff
@@ -53,6 +53,14 @@ type SupplierClient interface {
 		ctx context.Context,
 		sessionProofs []*relayer.SessionProof,
 	) error
+	// Address returns the address of the SupplierClient.
+	Address() *cosmostypes.AccAddress
+}
+
+// SupplierClientMap so we can depinject many supplier clients
+// Map key is an address
```
Olshansk, line 61:
```suggestion
// SupplierClientMap is a helper struct needed to depinject multiple supplier clients.
// The inner structure maps a supplier address to a list of supplier clients for that address.
```

pkg/relayer/config/supplier_hydrator.go
```diff
@@ -38,6 +38,10 @@ func (supplierConfig *RelayMinerSupplierConfig) HydrateSupplier(
 		)
 	}
 
+	// Don't check the SigningKeyNames here - we'll copy the keys from the root config in `HydrateSuppliers`
+	// if this list is empty.
```
Olshansk, line 42:
```suggestion
	// Intentionally not verifying SigningKeyNames here.
	// Wwe'll copy the keys from the root config in `HydrateSuppliers` if this list is empty.
```

pkg/relayer/config/supplier_hydrator.go
```diff
@@ -38,6 +38,10 @@ func (supplierConfig *RelayMinerSupplierConfig) HydrateSupplier(
...
+	supplierConfig.SigningKeyNames = yamlSupplierConfig.SigningKeyNames
```
Olshansk, line 43: Please mention where the second part is done.

pkg/relayer/config/types.go
```diff
@@ -132,6 +133,10 @@ type RelayMinerSupplierConfig struct {
 	// Other supplier types may embed other fields in the future. eg. "https" may
 	// embed a TLS config.
 	ServiceConfig *RelayMinerSupplierServiceConfig
+
+	// SigningKeyNames: a list of key names that can accept relays for that supplier.
```
Olshansk, line 137: I still don't understand why it's a list. #PUC

pkg/relayer/config/suppliers_config_hydrator.go
```diff
@@ -13,6 +13,22 @@ func (relayMinerConfig *RelayMinerConfig) HydrateSuppliers(
 			return err
 		}
 
+		// If SigningKeyNames are not specified for this supplierConfig, we want
```
Olshansk, line 16: nice

pkg/relayer/session/sessiontree.go
```diff
@@ -256,3 +262,8 @@ func (st *sessionTree) StartClaiming() error {
 	st.isClaiming = true
 	return nil
 }
+
+// SupplierAddress
```
Olshansk, line 266: #PUC

pkg/relayer/session/claim.go
```diff
@@ -155,17 +155,24 @@ func (rs *relayerSessionsManager) newMapClaimSessionsFn(
...
+				SessionHeader:   session.GetSessionHeader(),
+				SupplierAddress: *session.SupplierAddress(),
 			})
 		}
 
-		if err := rs.supplierClient.CreateClaims(ctx, sessionClaims); err != nil {
-			failedCreateClaimsSessionsPublishCh <- sessionTrees
-			return either.Error[[]relayer.SessionTree](err), false
+		// Create claims for each supplier address in `sessionTrees`.
+		for k, _ := range sessionClaims {
```
Olshansk, line 171: s/k/supplierAddr

pkg/relayer/session/proof.go
```diff
@@ -148,18 +148,24 @@ func (rs *relayerSessionsManager) newMapProveSessionsFn(
...
+				SupplierAddress: *session.SupplierAddress(),
 			})
 		}
 
-		// SubmitProof ensures on-chain proof inclusion so we can safely prune the tree.
-		if err := rs.supplierClient.SubmitProofs(ctx, sessionProofs); err != nil {
-			failedSubmitProofSessionsCh <- sessionTrees
-			return either.Error[[]relayer.SessionTree](err), false
+		// Submit proofs for each supplier address in `sessionTrees`.
+		for k, _ := range sessionProofs {
```
Olshansk, line 163: Ditto

Olshansk (changes requested):
Overall, looks pretty good. Only few things:

1. We've updated the RelayRequestMetadata and there's no verification anywhere for a pretty important structure. @red-0ne would love to get your thoughts here.

2. Will we be adding an E2E tests for this?

3. NEed to update comments a bit on why the default is a list.

pkg/client/supplier/client.go
okdas: 1. [Yes](https://github.com/pokt-network/poktroll/blob/63c67681d80079327153b5348bac9ee35569423a/pkg/relayer/cmd/cmd.go). This allows us to use account address in the supplier client map instead of a key name. If we were to use the key name here, we'd need to resolve it to the address on multiple places.
2. On it!

proto/poktroll/service/relay.proto
okdas: Two ways I can think of to solve that:
1. Remove the supplier address from the protobuf and rely on a different mechanism to select the supplier. I suggest using HTTP headers. Here's the example on the `shannon-sdk` side:
Screenshot 2024-06-06 at 2 48 07 PM
And relayminer:
Screenshot 2024-06-06 at 2 48 24 PM

2. Keep the supplier address in protobuf. We can introduce a proto that would be included in `RelayRequest` but would not be included in `SessionTree`. I am not sure exactly how yet, but there must be a way.

Also, a related question. That address that is being sent along with a relay is basically a recommendation to which supplier to route the request to. RelayMiner could potentially use a different supplier, that is also in the session, to serve the request instead. Correct?

go.mod
```diff
@@ -42,7 +45,7 @@ require (
 	github.com/grpc-ecosystem/grpc-gateway/v2 v2.18.1
 	github.com/hashicorp/go-metrics v0.5.2
 	github.com/pokt-network/ring-go v0.1.0
-	github.com/pokt-network/shannon-sdk v0.0.0-20240605120544-ff4fc44e8d6c
+	github.com/pokt-network/shannon-sdk v0.0.0-20240606213254-42787a73b552
```
okdas, line 48: Now that `shannon-sdk` is in a separate repo, I needed to bump the commit hash to the one that uses the new protobuf (from `poktroll` repo). Cross dependencies are hard. 
Opened a PR in `shannon-sdk` repo: https://github.com/pokt-network/shannon-sdk/pull/3/files

pkg/relayer/config/types.go
okdas: @Olshansk I pushed docusarus documentation update - please let me know if it's not enough and we should explain/repeat this in the code as well.

localnet/kubernetes/values-relayminer-1.yaml
okdas: Updated the docs.

proto/poktroll/service/relay.proto
red-0ne: As discussed in [1], adding the supplier's address to the `RelayRequest.Meta` seemed to be the most reliable way to route requests.

> We'll need to verifythat the signer of the response equals to the public key of the address here.

Absolutely! The relay miner also needs to check that the `RelayRequest.SupplierAddress` matches a supplier that it manages.

@okdas, Adding a header filed in the request to identify the supplier seems like an implicit protocol requirement that lives outside of the protobuf. I prefer keep everything explicit and contained in the `RelayRequest` protobuf definition.

[1] https://github.com/pokt-network/poktroll/issues/528#issuecomment-2111043405

okdas: @red-0ne after sleeping on this thought, I 100% agree. I added a check on a RelayMiner - if the supplier address is missing on the incoming request - it'll throw an error.

The other solution would be to adopt Morse ideology, where one virtual HTTP server/domain can serve requests only from one signing key (key1 -> supplier1.relayminers.com, key2 -> supplier2.relayminers.com). That wouldn't require any changes to the on-chain logic but would be an additional step in hydrating the configs on RelayMiner, and perhaps some checks. I personally don't see issues with that, and this design is community comfortable with.

docusaurus/docs/operate/configs/relayminer_config.md
```diff
@@ -282,6 +303,41 @@ a service provided by the `Supplier` and served by the `RelayMiner` instance.
 It MUST match the `Service.Id` specified by the supplier when staking for the
 service.
 
+### `signing_key_names`
+
+_`Required`_ if `default_signing_key_names` is empty.
+
+This option specifies the list of signing key names specific to a supplier. If a supplier does not provide its own `signing_key_names`, the `RelayMiner` will use the `default_signing_key_names`.
```
Olshansk, line 310:
```suggestion
This option specifies the list of signing key names specific to a supplier.
If a supplier does not provide its own `signing_key_names`, the `RelayMiner` will use the `default_signing_key_names`.
```

docusaurus/docs/operate/configs/relayminer_config.md
```diff
@@ -282,6 +303,41 @@ a service provided by the `Supplier` and served by the `RelayMiner` instance.
...
+
+#### Example Configuration
+
+```yaml
+default_signing_key_names:
+  - supplier_key2
+  - supplier_key3
+suppliers:
+  # This supplier will only run for `supplier_key1` and `supplier_key4` keys
+  - service_id: supplier1
```
Olshansk, line 320: I'm a bit confused:

1. Can you avoid using `supplier` for `service_id`?
2. Do you mean `This supplier will ...` or `This relayminder will ...`

In either case, can you try to put yourself in the shoes of someone who has never thought about this and make it really obvious how this works? Even I'm confused...

Maybe add a mermaid diagram if its applicable.

docusaurus/docs/operate/configs/relayminer_config.md
```diff
@@ -282,6 +303,41 @@ a service provided by the `Supplier` and served by the `RelayMiner` instance.
...
+```
+#### How It Works
+
+- If `signing_key_names` are specified for a supplier, the `RelayMiner` will use these keys instead of the `default_signing_key_names`.
+- This allows each supplier to have a unique set of keys for signing transactions, providing flexibility and better key management.
+
+#### Benefits
+
+- **Granular Control**: Allows each supplier to have its own specific keys, enhancing security and management.
+- **Fallback Mechanism**: If signing_key_names are not specified, the RelayMiner falls back to `default_signing_key_names`, ensuring continuity of operations.
```
Olshansk, line 339:
```suggestion
- **Fallback Mechanism**: If `signing_key_names` are not specified, the RelayMiner falls back to `default_signing_key_names`, ensuring continuity of operations.
```

docusaurus/docs/operate/configs/relayminer_config.md
```diff
@@ -154,10 +178,7 @@ flowchart
     S2 -.- RM4
 ```
 
-TODO(#528): It is not currently possible to have a single `RelayMiner` instance
-running with multiple `signing_key_name`s as this would involve more complex logic
-and/or configuration to determine which key to use, especially in the case of
-`Supplier`s that have overlapping services provided.
+TODO(#528): Currently, it is not possible to run a single RelayMiner instance with multiple `signing_key_names` due to the complexity in determining which key to use for each supplier, especially if suppliers have overlapping services. This feature will be available once issue #528 is resolved.
```
Olshansk, line 181: Aren't you resolving this?

pkg/deps/config/suppliers.go
```diff
@@ -412,3 +446,20 @@ func NewSupplySharedQueryClientFn() SupplierFn {
...
 }
+
+// newSupplyTxClientFn returns a new depinject.Config which is supplied with
+// the given deps and the new TxClient.
+func newSupplyTxClientsFn(ctx context.Context, deps depinject.Config, signingKeyName string) (depinject.Config, error) {
+	txClient, err := tx.NewTxClient(
+		ctx,
+		deps,
+		tx.WithSigningKeyName(signingKeyName),
+		// TODO_TECHDEBT: populate this from some config.
```
Olshansk, line 457: Can you move the TECHDEBT to where we define the variable?

pkg/relayer/cmd/cmd.go
```diff
@@ -254,6 +252,7 @@ func supplyTxContext(
 	deps depinject.Config,
 	_ *cobra.Command,
 ) (depinject.Config, error) {
+	// TODO_CONSIDERATION: do we need new context per key?
```
Olshansk, line 255: Either remove, or provide more details.

.github/workflows-helpers/run-e2e-test-job-template.yaml
```diff
@@ -10,6 +10,68 @@ spec:
       labels:
         pokt.network/purpose: e2e-tests
     spec:
+      initContainers:
```
Olshansk, line 13: This is new and separate from lean client.

1. Can we add some comments on why we added this?
2. Is there a way to put this in a source file and import?

docusaurus/docs/operate/configs/relayminer_config.md
```diff
@@ -104,42 +111,59 @@ and `supplier` specific sections and configurations.
...
 smt_store_path: <string>
 ```
 
-### `signing_key_name`
+### `default_signing_key_names`
 
-_`Required`_
+_`Required`_ if `suppliers.*.signing_key_names` is not specified. 
+
+This configuration option specifies a list of key names from the keyring that the `RelayMiner` will use to sign transactions. These keys are used to derive the public key and the corresponding address, which are essential for interacting with the Pocket network. Each key name listed here must be present in the keyring used to start the `RelayMiner` instance.
```
Olshansk, line 122: Multiline (80-120) plaease

Olshansk, line 122:
```suggestion
This configuration option specifies a list of key names from the keyring that the `RelayMiner` will use to sign transactions. These key names are used to derive the public key and the corresponding address, which are essential for interacting with POKT network. Each key name listed here must be present in the keyring used to start the `RelayMiner` instance.
```

docusaurus/docs/operate/configs/relayminer_config.md
```diff
@@ -104,42 +111,59 @@ and `supplier` specific sections and configurations.
...
+```
 
-The name of the key that will be used to sign transactions, derive the public key
-and the corresponding address. This key name MUST be present in the keyring that is used
-to start the `RelayMiner` instance.
+In this example, `keyname1` and `keyname2` are the names of keys in the keyring that will be used to sign transactions.
+
+#### How It Works
+
+- If a supplier does not specify its own `signing_key_names`, the supplier will use the keys listed in `default_signing_key_names`.
```
Olshansk, line 135: Consider adding a flow chart to make this visual: https://mermaid.js.org/syntax/flowchart.html

I bet you can ask ChatGPT to create one based on the text as a starting point.

docusaurus/docs/operate/configs/relayminer_config.md
```diff
@@ -104,42 +111,59 @@ and `supplier` specific sections and configurations.
...
+- You can specify more than one key, allowing the `RelayMiner` to operate multiple suppliers with different keys/addresses.
+- This setup is useful for running multiple RelayMiner instances across different regions or for redundancy.
```
Olshansk, line 137: This seems like a benefit, not a how it works.

job.yaml
```diff
@@ -0,0 +1,123 @@
...
+  ttlSecondsAfterFinished: 120
+  template:
+    metadata:
+      labels:
+        pokt.network/purpose: e2e-tests
+    spec:
+      initContainers:
+        - name: check-services
+          image: ruby:3.0
+          command: ["/bin/sh"]
```
Olshansk, line 16: This script feels like a copy-paste of the other one.

Again:
1. Need comments on why we need this and what's happening
2. Can we put it in a source file?

Olshansk (changes requested):

job.yaml
okdas: This was not supposed to be committed, my bad.

red-0ne (changes requested):
Awesome to see this feature near completeness.

I've left some change requests but the overall structure has the shape it should have.

x/proof/keeper/msg_server_submit_proof_test.go
```diff
@@ -1117,9 +1118,12 @@ func newEmptySessionTree(
 		_ = os.RemoveAll(testSessionTreeStoreDir)
 	})
 
+	AccAddress := cosmostypes.MustAccAddressFromBech32(supplierAddr)
```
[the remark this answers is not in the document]
okdas: No good reason, changing this.

docusaurus/docs/operate/configs/relayminer_config.md
okdas: 1. 👍
2. Reworded as "RelayMiner will only provide service for `llm-model` using `supplier_key1` and `supplier_key4` keys."

.github/workflows-helpers/run-e2e-test-job-template.yaml
okdas: I've noticed some intermittent issues with e2e tests, started investigating and this was the result. There are ways on how to optimize this and move the code out, but I'll add a todo for myself to clean this up later.

docusaurus/docs/operate/configs/relayminer_config.md
okdas: I've restructured the documentation page - I think it is more user-friendly now. Also includes the diagram.

pkg/relayer/config/relayminer_configs_reader_test.go
```diff
@@ -90,6 +90,66 @@ func Test_ParseRelayMinerConfigs(t *testing.T) {
 				},
 			},
 		},
+		{
+			desc: "valid: relay miner config with signing key configured on supplier level",
```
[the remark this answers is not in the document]
okdas: Great idea - added the test.

red-0ne (approved):
This looks good to me. :rocket:

okdas:
Seems like the only e2e test that fails is the same flaky `TestFeatures/Session_Namespace/Supplier_completes_claim/proof_lifecycle_for_a_valid_session`.

x/proof/keeper/msg_server_submit_proof.go
```diff
@@ -167,6 +167,12 @@ func (k msgServer) SubmitProof(ctx context.Context, msg *types.MsgSubmitProof) (
 	}
 	logger.Info("successfully validated relay request")
 
+	// Make sure tha the supplier address in the proof matches the one in the relay request.
```
Olshansk, line 170:
```suggestion
	// Make sure that the supplier address in the proof matches the one in the relay request.
```

docusaurus/docs/operate/configs/relayminer_config.md
```diff
@@ -409,11 +331,119 @@ for the same supplier service.
...
+        H[service_id: etherium-mainnet]
+        H --> I[signing_key_names]
+        I --> J[Not Specified]
+    end
+
+    subgraph Effective_Signing_Keys
+        direction TB
+        K[llm-model] --> L[keyname3]
+        K --> M[keyname4]
+        N[Supplier 2] --> O[keyname1]
```
Olshansk, line 399:
```suggestion
        N[etherium-mainnet] --> O[keyname1]
```

Olshansk (approved):
Couple of NITs but LGTM. Please merge with main

okdas:
Merging with one flaky test still failing.

Screenshot 2024-06-14 at 5 05 26 PM

PR #577, opened 2024-06-04
[TODO] Team's Iteration16 TODO preparation

Olshansk:
Go through ALMOST all the TODOs and and to the following:

- Add an owner where applicable
- Mark things as `TODO_BLOCKER` if they need to be done in iteration #16
- Mark things as `TODO_BETA` if they are needed for the beta launch
- Mark things as `TODO_MAINNET` if they are needed for mainnet

This enables generating a TODO list like so:

```
grep -r "TODO_BLOCKER(@Olshansk)" . | sed 's/^/- [ ] `/' | sed 's/:/`:/'
```

### DIFF
--- a/.github/workflows/main-build.yml
+++ b/.github/workflows/main-build.yml
@@ -30,7 +30,7 @@ jobs:
 
       # This is commented out because proto generated code is now tracked in the
       # repo so this step is skipped.
-      # TODO_IMPROVE: Have a step that regenerates the proto files and ensures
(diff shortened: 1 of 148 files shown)


red-0ne (approved):
LGTM! Let's merge it.

okdas (approved):
Thank you for going through all the TODOs.



I double-checked to-dos assigned to me - they do make sense, but a few seem already resolved. I'll clean them up in #587.

Issue #402, opened 2024-02-29
[Proof] Claim & proof distribution validation


bryanchriswhite:
## Objective

The proof module SHOULD validate that claim and proof messages are submitted after a minimum delay, in blocks, after the respective session has ended to mitigate congestion caused by many (potentially all) suppliers submitting claims/proofs at the same time.

## Origin Document

![image](https://github.com/pokt-network/poktroll/assets/600733/11abb71f-e017-4e0c-ac4b-b83fdf908815)

![image](https://github.com/pokt-network/poktroll/assets/600733/03169008-3723-459a-8b11-1fad7008bab9)

## Goals

- Define gov. params in the proof module sufficient to support a "global" delay/offset in claim/proof the message distribution validation implementation.

See the application and service modules for examples of modules with params:
- Application: `max_delegated_gateways`
  - `proto/poktroll/application/params.proto`
  - `x/application/types/genesis_test.go`
  - `x/application/types/params.go`
  - `x/application/keeper/msg_server_delegate_to_gateway.go` (NB: `k.GetParams()`)
  - `x/application/keeper/msg_update_params_test.go` (application module may not be a good example; however, it seems like there should be tests that exercise updates for all params)

## Deliverables

- [ ] ~~A PR which adds `claim_msg_distribution_block_offset` to the proof module.~~ (duplicates #516 )
- [ ] ~~A PR which adds `proof_msg_distribution_block_offset` to the proof module.~~ (duplicates #516)
- [ ] ~~A PR which implements claim/proof message distribution validation using only these gov. params (omitting pseudo-random factor) & corresponding off-chain pkg changes.~~ (duplicates #516)
- [x] A PR which adds the pseudo-random factor to claim/proof message distribution validation (last block hash?).
  - Keeper has no API to query for past blocks (& hashes) by default (i.e. depends "block hash store").
  - Simplify "block hash store" to only store relevant blocks?

## Non-goals / Non-deliverables

- Introducing new gov. params other than those outlined in this issue.

## General deliverables

- [ ] **Comments**: Add/update TODOs and comments alongside the source code so it is easier to follow.
- [ ] **Testing**: Add new tests (unit and/or E2E) to the test suite.
- [ ] **Makefile**: Add new targets to the Makefile to make the new functionality easier to use.
- [ ] **Documentation**: Update architectural or development READMEs; use [mermaid](https://mermaid-js.github.io/mermaid/) diagrams where appropriate.

---

**Creator**: @bryanchriswhite 
[...]

PR #616, opened 2024-06-18
[Code Health] refactor: block query client interface

bryanchriswhite:
## Summary

Define a new interface which is a subset of `cometclient.CometRPC` and only contains members that poktroll depends on.

## Issue

We were depending on a cometbft interface which was much larger than we need. This came while working on #402 in an integration test. Defining our own interface for this purpose is more  "SOLID".

- #402 

## Type of change

Select one or more:

- [ ] Bug fix
- [x] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

## Sanity Checklist


pkg/client/interface.go
```diff
@@ -297,3 +298,10 @@ type SharedQueryClient interface {
 	// the session that includes queryHeight opens.
 	GetProofWindowOpenHeight(ctx context.Context, queryHeight int64) (int64, error)
 }
+
+// BlockQueryClient defines an interface that enables the querying of
```
Olshansk, line 302: LGTM. No issues. Just calling out something below but not looking for a response.


Looks like we're already diverging from [1] w.r.t Naming (e.g. `BlockClient` vs `BlockQueryClient`)

I know:
- It has different uses
- It has different contexts
- Functionality differs

Just calling it out.

cc @adshmh @red-0ne 

[1] https://github.com/pokt-network/shannon-sdk/blob/d6cb20a34a2e50ea26f196f7c671e60a66bbc015/block.go

Olshansk (approved):

pkg/client/interface.go
red-0ne: Just note that `shannon-sdk`'s block client and `poktroll`'s block client will diverge in functionality.

While the former will mainly be used to query block heights, the later wants the full block to get block events and txs.

So they are definitely not interchangeable.

red-0ne (approved):

bryanchriswhite:
E2E tests passed locally but failed to run in CI:
image

PR #619, opened 2024-06-18
[Relayminer] off-chain claim/proof message distribution

bryanchriswhite:
## Summary

Implement claim/proof message distribution in the relayminer.

## Issue

Prepares the relayminer for forthcoming on-chain claim/proof message distribution validation.

- #402

## Type of change

Select one or more:

- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

## Sanity Checklist


pkg/client/query/sharedquerier.go
```diff
@@ -105,3 +108,66 @@ func (sq *sharedQuerier) GetSessionGracePeriodEndHeight(
...
+
+// GetEarliestClaimCommitHeight returns the earliest block height at which a claim
+// for the session that includes queryHeight can be committed for a given supplier.
+//
+// TODO_TECHDEBT(#543): We don't really want to have to query the params for every method call.
+// Once `ModuleParamsClient` is implemented, use its replay observable's `#Last()` method
+// to get the most recently (asynchronously) observed (and cached) value.
+// TODO_BLOCKER(@bryanchriswhite, #543): We also don't really want to use the current value of the params.
+// Instead, we should be using the value that the params had for the session which includes queryHeight.
+func (sq *sharedQuerier) GetEarliestClaimCommitHeight(ctx context.Context, queryHeight int64, supplierAddr string) (int64, error) {
```
Olshansk, line 120: OPTIONAL NIT: multiple method header

x/shared/session.go
```diff
@@ -106,3 +107,45 @@ func GetProofWindowCloseHeight(sharedParams *sharedtypes.Params, queryHeight int
...
+	sharedParams *sharedtypes.Params,
+	queryHeight int64,
+	claimWindowOpenBlockHash []byte,
+	supplierAddr string,
+) int64 {
+	claimWindowOpenHeight := GetClaimWindowOpenHeight(sharedParams, queryHeight)
+
+	// Generate a deterministic random (non-negative) int64, seeded by the claim
+	// window open block hash and the supplier address.
+	randomNumber := poktrand.SeededInt63(claimWindowOpenBlockHash, []byte(supplierAddr))
```
Olshansk, line 124: A claim is already ted to a single supplier. WHy do we need to add the other variable for entropy as well? I feel like we can reduce a lot of variables being passed around if so.

x/shared/session.go
```diff
@@ -106,3 +107,45 @@ func GetProofWindowCloseHeight(sharedParams *sharedtypes.Params, queryHeight int
...
+
+	distributionWindowSizeBlocks := sharedParams.ClaimWindowCloseOffsetBlocks - sharedParams.ClaimWindowOpenOffsetBlocks
```
Olshansk, line 126: OPTIONAL NIT: Should this be its own helper?

x/shared/session.go
```diff
@@ -106,3 +107,45 @@ func GetProofWindowCloseHeight(sharedParams *sharedtypes.Params, queryHeight int
...
+	randCreateClaimHeightOffset := randomNumber % int64(distributionWindowSizeBlocks)
+
+	return claimWindowOpenHeight + randCreateClaimHeightOffset
+}
+
+// GetEarliestProofCommitHeight returns the earliest block height at which a proof
+// for the session that includes queryHeight can be committed for a given supplier
+// and the passed sharedParams.
+func GetEarliestProofCommitHeight(
```
Olshansk, line 135: Have you put thought into making this be part of the keeper and just having it query sharedParams?

I'm thinking of a simpler less "functional" approach

Olshansk (dismissed):

x/shared/session.go
bryanchriswhite: I'm not 100% sure I followed the question.

The rationale here is that `claimWindowOpenBlockHash` will be common to all sessions (app/supplier pairings) with the same session number; therefore, in order for this offset to have the intended effect of distributing message commit heights across their respective windows, it must also consider additional entropy which is either unique to suppliers or applications. If we choose the former (supplier), relayminers can continue to batch claim and proof messages on a per session number basis, which would no longer be possible in the other case. If you're suggesting to use the claim itself as a source of entropy for this, I think that would create the same issue. I think the fact that relayminers can batch claims/proofs is basically all upside.

x/shared/session.go
```diff
@@ -106,3 +107,45 @@ func GetProofWindowCloseHeight(sharedParams *sharedtypes.Params, queryHeight int
...
 		int64(sharedParams.GetProofWindowCloseOffsetBlocks())
 }
+
+// GetEarliestClaimCommitHeight returns the earliest block height at which a claim
+// for the session that includes queryHeight can be committed for a given supplier
+// and the passed sharedParams.
+func GetEarliestClaimCommitHeight(
+	sharedParams *sharedtypes.Params,
+	queryHeight int64,
+	claimWindowOpenBlockHash []byte,
```
Olshansk, line 117: Would it not be easier to just give the `queryHeight` and get `claimWindowOpenBlockHash` internally given that we already have access to `sharedParams`?

x/shared/session.go
Olshansk: Got it. 

I interpreted it as as `claimHash` which (IMO) should work even though you said `I think that would create the same issu`.

Now that I understand the reasoning, it makes sense. Contemplating if another name would make it clearer, but I think it's on me for misreading.

x/shared/session.go
bryanchriswhite: This static function has no way to query for the claim window open block such that it can get a reference to its hash.

x/shared/session.go
bryanchriswhite: TL;DR, there is no appropriate object to make these static functions  members of.

Like the other helpers in this package, this static function is intended to be the source of truth (identity) for "earliest proof commit height". These functions are used both on and off chain, so while we could move them into the shared keeper package, that would result in the relay miner's dependency try including on-chain code. This would likely bloat the relayminer binary size.

For the sake of completeness, I also considered moving these functions to become methods of `sharedtypes.Param` but this does't feel right to me either.

pkg/client/query/sharedquerier.go
bryanchriswhite: I'm not sure what the ask is here. Can you elaborate?

pkg/client/interface.go
```diff
@@ -294,9 +294,15 @@ type SharedQueryClient interface {
 	// GetClaimWindowOpenHeight returns the block height at which the claim window of
 	// the session that includes queryHeight opens.
 	GetClaimWindowOpenHeight(ctx context.Context, queryHeight int64) (int64, error)
+	// GetEarliestClaimCommitHeight returns the earliest block height at which a claim
+	// for the session that includes queryHeight can be committed for a given supplier.
```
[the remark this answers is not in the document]
bryanchriswhite: I would push back a bit here. We have consistently used `CreateClaim` and `SubmitProof` in the context of the messages; however, I see the claim/proof window and commit offsets/heights as separate from the messages themselves. I would argue that when it comes to the claim/proof window and commit height offsets, we've been consistently more concise (e.g. `ClaimWindowOpen...`). In my view, `ClaimCommit...` is more consistent in this context and has greater semantic density.

x/shared/session.go
```diff
@@ -106,3 +107,52 @@ func GetProofWindowCloseHeight(sharedParams *sharedtypes.Params, queryHeight int
...
+	supplierAddr string,
+) int64 {
+	claimWindowOpenHeight := GetClaimWindowOpenHeight(sharedParams, queryHeight)
+
+	// Generate a deterministic random (non-negative) int64, seeded by the claim
+	// window open block hash and the supplier address.
+	randomNumber := poktrand.SeededInt63(claimWindowOpenBlockHash, []byte(supplierAddr))
+
+	distributionWindowSizeBlocks := GetClaimWindowSizeBlocks(sharedParams)
+	randCreateClaimHeightOffset := randomNumber % int64(distributionWindowSizeBlocks)
```
red-0ne, line 127: Since `distributionWindowSizeBlocks` ranges from window open to window close, we may end up with a `randCreateClaimHeightOffset` that is too close (1 block before `WindowCloseHeight`).

I believe `distributionWindowSizeBlocks` should be something like:
`distributionWindowSizeBlocks := GetClaimWindowSizeBlocks(sharedParams) - someSafetyBlocks`

x/shared/session.go
```diff
@@ -106,3 +107,52 @@ func GetProofWindowCloseHeight(sharedParams *sharedtypes.Params, queryHeight int
...
+	proofWindowOpenBlockHash []byte,
+	supplierAddr string,
+) int64 {
+	proofWindowOpenHeight := GetProofWindowOpenHeight(sharedParams, queryHeight)
+
+	// Generate a deterministic random (non-negative) int64, seeded by the proof
+	// window open block hash and the supplier address.
+	randomNumber := poktrand.SeededInt63(proofWindowOpenBlockHash, []byte(supplierAddr))
+
+	distributionWindowSizeBlocks := sharedParams.ProofWindowCloseOffsetBlocks - sharedParams.ProofWindowOpenOffsetBlocks
```
red-0ne, line 154: We could have a `GetProofWindowSizeBlocks` just like for claims.

x/shared/session_test.go
```diff
@@ -0,0 +1,81 @@
...
+		require.NoError(t, err)
+
+		expected := test()
+
+		// Ensure consecutive calls are deterministic.
+		for deterministicIdx := 0; deterministicIdx < 1000; deterministicIdx++ {
+			require.Equalf(t, expected, test(), "on call number %d", deterministicIdx)
+		}
+	}
+}
```
red-0ne, line 81: Would be great to have a test that ensures that `GetEarliestXXXCommitHeight` always ends up in the `XXXWindow` by instrumenting the `poktrand.SeedInt63`

red-0ne (commented):
Have left some comments but it's great seeing this coming into place :relieved:

x/shared/session_test.go
bryanchriswhite: I'm not sure what you're referring to with respect to instrumenting `poktrand.SeededInt63()` but I added a test that uses a large sample size to try to be exhaustive.

pkg/client/query/sharedquerier.go
```diff
@@ -105,3 +108,66 @@ func (sq *sharedQuerier) GetSessionGracePeriodEndHeight(
...
+	if err != nil {
+		return 0, err
+	}
+
+	// NB: Byte slice representation of block hashes don't need to be normalized.
+	claimWindowOpenBlockHash := claimWindowOpenBlock.BlockID.Hash.Bytes()
+
+	return shared.GetEarliestClaimCommitHeight(
+		sharedParams,
+		queryHeight,
```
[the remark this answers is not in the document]
bryanchriswhite: The query height is not in the same session number as the proof window open height - 1 so there shouldn't be any such block that can be passed and serve both functions.

red-0ne (approved):
All good from my side! Thanks for addressing the change requests.

x/shared/session_test.go
```diff
@@ -0,0 +1,233 @@
...
+		// NB: sample concurrently to save time.
+		go func() {
+			// Randomize queryHeight, proofWindowOpenBlockHash, and supplierAddr.
+			queryHeight := rand.Int63()
+			supplierAddr := sample.AccAddress()
+			var proofWindowOpenBlockHash [32]byte
+			_, err := rand.Read(proofWindowOpenBlockHash[:])
+
+			if !assert.NoError(t, err) {
+				t.Log("cancelling")
```
bryanchriswhite, line 105:
```suggestion
```

Issue #489, opened 2024-04-24
[Utility] Introduce an unbonding period after an actor unstakes


Olshansk:
## Objective

Design & implement a framework of how to handle on-chain `Unbonding` after an actor submits a valid `Unstake` transaction.

## Origin Documents

- [Cosmos SDK documentation on unbonding](https://docs.cosmos.network/v0.50/build/modules/staking)
- [Cosmos explanation of unbonding in the context of validators](https://docs.cosmos.network/v0.46/modules/staking/02_state_transitions.html)
- Morse ([pocket-core](https://github.com/pokt-network/pocket-core)) has a parameter called `UnstakingTime` which can be found [here](https://github.com/search?q=repo%3Apokt-network%2Fpocket-core%20pos%2FUnstakingTime&type=code)
- Unstaking has been discussed on the Pocket Forums as can be found [here](https://forum.pokt.network/t/pup-31-reduce-unstakingtime-due-to-current-market-instability/4026)

## Goals

- `Unbonding` definition: a state during which an actor operates normally, after unstaking, before their escrowed stake is returned and they are removed as active actors
- At a minimum, have parity with unbonding periods from Morse
- Evaluate options for how on-chain actor unbonding could be implemented
- Design, document and implement unbonding of Applications, Suppliers & Gateways
- Create a pattern reusable for all actors

## Deliverables

*IMPORTANT NOTE TO THE OWNER: The deliverables below are not a complete list of everything that needs to be done, but rather a starting point to help. The research, which is part of the deliverables, should drive most of the design decisions.*

1. `Research`
    - [ ]  Put together a reference doc with bullet points, links, highlights, learnings capturing the research below(< 1 page total
    - `Other Projects`
        - [ ]  Investigate how actor (e.g. validator) unbonding is designed and implemented in the Cosmos SDK
        - [ ]  Investigate how at least one other project does (non-validator) actor unbonding in the Cosmos ecosystem
    - `Morse`
        - [ ]  Investigate how actor unbonding currently works in Morse
2. `Before implementation`
    - [ ]  Prepare a 1-pager “mini spec” notion / docs / markdown and get feedback & review before kicking off the implementation
3. `Implementation requirements`
    - [ ]  Unbonding should finish  automatically `N` blocks from the moment the actor submits the unstaking transaction
    - [ ]  `N` should have a unique governance parameter for every actor: Application, Supplier, Gateway
[...]

PR #666, opened 2024-07-08
[Supplier] Implement unbonding period

red-0ne:
## Summary

This PR adds an unbonding period when a `Supplier` submits an unstaking transaction.

* Adds an `unbondingHeight` to the `Supplier` proto.
* Resets the `unbondingHeight` when a supplier re-stakes during that period.
* Adds a stateless function that calculates the `unbondingHeight` for a given height.
* Prevents unstake transactions of `Supplier`s that are already in an unbonding period.
* Adds an `EndBlocker` routine to effectively unbond `Supplier`s that reached their unbonding height.
* Keeps the `Supplier`s normally functioning during the unbonding period.

## Issue

`Supplier`s that submit an `unstake` transaction have to wait for a certain time before getting their staked funds back.

- #489 

## Type of change

Select one or more:

- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

## Sanity Checklist

## Summary by CodeRabbit

- **New Features**
  - Introduced a mechanism to handle supplier unstaking with checks for previous unstaking actions and updated session end height parameters.
  - Added `EndBlockerUnbondSuppliers` function to manage supplier unbonding at session end.
  - Implemented new methods to determine supplier unbonding status and activity based on specific conditions.
  - Created a function to calculate the unbonding height for suppliers based on input parameters.

- **Bug Fixes**
  - Improved unstaking logic to prevent immediate supplier removal.

- **Tests**
  - Enhanced test coverage for scenarios like canceling unbonding if restaked and failure if not in the unbonding period.

- **Chores**
  - Added new error handling and validation logic related to supplier functionalities.


x/supplier/keeper/unbond_suppliers.go
```diff
@@ -0,0 +1,63 @@
...
+				ctx, types.ModuleName, supplierAddress, []sdk.Coin{*supplier.Stake},
+			); err != nil {
+				logger.Error(fmt.Sprintf(
+					"could not send %v coins from %s module to %s account due to %v",
+					supplier.Stake, supplierAddress, types.ModuleName, err,
+				))
+				return err
+			}
+
+			// Update the Supplier in the store
```
bryanchriswhite, line 56:
```suggestion
			// Remove the Supplier from the store
```

x/supplier/types/errors.go
```diff
@@ -16,4 +16,6 @@ var (
 	ErrSupplierInvalidSessionId          = sdkerrors.Register(ModuleName, 1107, "invalid session ID")
 	ErrSupplierInvalidService            = sdkerrors.Register(ModuleName, 1108, "invalid service in supplier")
 	ErrSupplierInvalidSessionEndHeight   = sdkerrors.Register(ModuleName, 1109, "invalid session ending height")
+	ErrSupplierUnbonding                 = sdkerrors.Register(ModuleName, 1110, "supplier is in unbonding period")
```
bryanchriswhite, line 19: Are we sure about "unbonding" as opposed to "unstaking"? My understanding was that the "bond" nomenclature was an extension of the chemistry analogy of the "atom" token in the cosmos-hub network. I think here, it is less precisely defined.

x/supplier/types/params.go
```diff
@@ -1,8 +1,16 @@
...
+	paramtypes "github.com/cosmos/cosmos-sdk/x/params/types"
+)
 
-var _ paramtypes.ParamSet = (*Params)(nil)
+var (
+	_ paramtypes.ParamSet = (*Params)(nil)
+
+	KeySupplierUnbondingPeriodBlocks            = []byte("SupplierUnbondingPeriodBlocks")
+	ParamSupplierUnbondingPeriodBlocks          = "supplier_unbonding_period_blocks"
+	DefaultSupplierUnbondingPeriodBlocks uint64 = 4 // TODO_MAINNET: Determine the default value.
```
bryanchriswhite, line 12: Maybe it makes sense to define that constant here.

x/supplier/types/params.go
```diff
@@ -21,10 +31,37 @@ func DefaultParams() Params {
...
+// ValidateSupplierUnbondingPeriodBlocks validates the SupplierUnbondingPeriodBlocks
+// governance parameter.
+// NB: The argument is an interface type to satisfy the ParamSetPair function signature.
+func ValidateSupplierUnbondingPeriodBlocks(v interface{}) error {
+	supplierUnbondingPeriodBlocks, ok := v.(uint64)
+	if !ok {
+		return ErrSupplierParamsInvalid.Wrapf("invalid parameter type: %T", v)
+	}
+
+	if supplierUnbondingPeriodBlocks <= 0 {
```
bryanchriswhite, line 62: TL;DR, the const & `=` over `<=` as a `uint64` cannot be less than 0.

```suggestion
	if supplierUnbondingPeriodBlocks = SupplierNotUnstakingPeriodBlocks {
```

bryanchriswhite (changes requested):
Nice one @red-0ne! :raised_hands:  :rocket:

x/supplier/keeper/msg_server_unstake_supplier_test.go
```diff
@@ -54,19 +36,87 @@ func TestMsgServer_UnstakeSupplier_Success(t *testing.T) {
...
+
+	// Make sure the supplier entered the unbonding period
+	foundSupplier, isSupplierFound = k.GetSupplier(ctx, supplierAddr)
+	require.True(t, isSupplierFound)
+	require.Greater(t, foundSupplier.UnstakeCommitSessionEndHeight, int64(0))
+
+	unbondingHeight := foundSupplier.UnstakeCommitSessionEndHeight + int64(k.GetParams(ctx).SupplierUnbondingPeriodBlocks)
+
+	// Stake the supplier again
+	stakeMsg = createStakeMsg(supplierAddr, initialStake+1)
```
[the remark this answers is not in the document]
red-0ne: Yes it is to avoid hitting the "Same staking amount" error.

I don't fully grasp why upstaking is needed. But if it's to avoid spamming/quickly-changing the supplier info, then we should keep it for staking cancellation too.

Otherwise we'd circumvent the upstaking by allowing suppliers to unstake->cancel (unless we check that they did not change any param).

x/supplier/keeper/unbond_suppliers.go
```diff
@@ -0,0 +1,63 @@
...
+	// Only process unbonding at the end of the session.
+	if currentHeight != sessionEndHeight {
+		return nil
+	}
+
+	logger := k.Logger().With("method", "UnbondSupplier")
+
+	// Iterate over all suppliers and unbond suppliers that have finished the unbonding period.
+	// TODO_IMPROVE: Use an index to iterate over suppliers that have initiated the
+	// unbonding action instead of iterating over all suppliers.
```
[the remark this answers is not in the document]
red-0ne: Yes we could. There are  a lot of places where indexes would improve look-ups (e.g. undelegation pruning)

We should discuss this in terms of scope and timelines.

x/supplier/types/errors.go
red-0ne: Are you suggesting to fully replace the `Unbonding` term by `Unstaking`?

x/supplier/types/params.go
red-0ne: I've used `SupplierNotUnstaking` instead of `SupplierNotUnstakingPeriodBlocks` as it makes more sense in other places where it's used.

x/supplier/types/errors.go
bryanchriswhite: I suppose so. :sweat_smile:

bryanchriswhite (approved):
:ok_hand: Preemptively approving. I still think we should do a find and replace like`s/unbonding/unstaking/g`, and there was one import alias I think should be renamed. Otherwise this LGTM. :fire::fire::fire: :fire_engine:

x/supplier/types/errors.go
red-0ne: @Olshansk, any opinion on this?

Olshansk: https://docs.cosmos.network/v0.47/build/modules/staking

The way I see it is that `Stake` / `Unstake` is an action (verb) triggered by an end user.

I realize that `Bond` / `Unbond` can also be a verb, but I see it as more of a status. See [cosmos/staking/v1beta1/staking.proto](https://github.com/cosmos/cosmos-sdk/blob/main/x/staking/proto/cosmos/staking/v1beta1/staking.proto)

Screenshot 2024-07-15 at 1 35 11 PM

Olshansk (changes requested):
@red-0ne I'd like to review this PR before we merge it so just marking as a cross for reference for now.

x/supplier/types/errors.go
bryanchriswhite: Another thing worth considering is that we're not even using the terms "stake" and "staking" in the same sense as conventional cosmos-sdk conceives of it as in their picture, it's part of DPoS, which is not the case for us. I would argue that this accumulates in favor of avoiding these ambiguous terms in our unique context, where we've already appropriated and defined "stake" and "staking".

red-0ne:
@bryanchriswhite , I changed the unbonding period to be in terms of `ProofWindowEndHeight` instead of a gov. param.

re-requesting your review.

x/supplier/types/errors.go
Olshansk: @bryanchriswhite

1. In cosmos, the `x/saking` module is strictly for validators so I was using it as a reference.
2. I hear your argument and don't feel ultra strongly about it, but am still not convinced after doing some research.
3. See Ethereum & Solana docs for reference below where "staked assets" have "unbonding periods" after an "unstaking event".

Lmk if you still feel like we should rename it.

--- 

Source 1 (ethereum): https://support.bitpanda.com/hc/en-us/articles/8324987568284-Ethereum-Staking

Screenshot 2024-07-16 at 8 14 12 PM

Source 2 (solana): https://docs.blockdaemon.com/docs/solana-specific-info

Screenshot 2024-07-16 at 8 14 41 PM

testutil/keeper/supplier.go
```diff
@@ -36,6 +37,9 @@ func SupplierKeeper(t testing.TB) (keeper.Keeper, context.Context) {
 	stateStore.MountStoreWithDB(storeKey, storetypes.StoreTypeIAVL, db)
 	require.NoError(t, stateStore.LoadLatestVersion())
 
+	logger := log.NewTestLogger(t)
+	ctx := sdk.NewContext(stateStore, cmtproto.Header{}, false, logger)
```
Olshansk, line 41: isn't this `sdkCtx` too?

x/session/keeper/session_hydrator.go
```diff
@@ -167,6 +168,14 @@ func (k Keeper) hydrateSessionSuppliers(ctx context.Context, sh *sessionHydrator
 
 	candidateSuppliers := make([]*sharedtypes.Supplier, 0)
 	for _, s := range suppliers {
+		// Exclude suppliers that are in an unbonding period.
```
Olshansk, line 171:
```suggestion
		// Exclude suppliers that are actively unbonding.
```

x/session/keeper/session_hydrator.go
```diff
@@ -167,6 +168,14 @@ func (k Keeper) hydrateSessionSuppliers(ctx context.Context, sh *sessionHydrator
...
+		// TODO_TECHDEBT: Suppliers that stake mid-session SHOULD NOT be included
```
Olshansk, line 172: Do you have a ticket number for this?

proto/poktroll/shared/supplier.proto
```diff
@@ -13,5 +13,8 @@ message Supplier {
   string address = 1 [(cosmos_proto.scalar) = "cosmos.AddressString"]; // The Bech32 address of the supplier using cosmos' ScalarDescriptor to ensure deterministic encoding
   cosmos.base.v1beta1.Coin stake = 2; // The total amount of uPOKT the supplier has staked
   repeated SupplierServiceConfig services = 3; // The service configs this supplier can support
+  // The session end height corresponding the the unstake tx commit height.
+  // If the supplier is not unstaking, this value will be 0.
+  uint64 unstake_commit_session_end_height = 4;
```
Olshansk, line 18:
```suggestion
  // The session end height at which an actively unbonding supplier is unbonds its stake.
  // If the supplier did not unstake, this value will be 0.
  uint64 unstake_session_end_height = 4;
```

x/supplier/types/params.go
```diff
@@ -2,7 +2,10 @@ package types
 
 import paramtypes "github.com/cosmos/cosmos-sdk/x/params/types"
 
-var _ paramtypes.ParamSet = (*Params)(nil)
+var (
+	_                    paramtypes.ParamSet = (*Params)(nil)
+	SupplierNotUnstaking uint64              = 0
```
Olshansk, line 7:
```suggestion
	// The value of `unstake_session_end_height` if the supplier is not actively in the unbonding period.
	SupplierNotUnstakingHeight uint64              = 0
```

x/supplier/module/abci.go
```diff
@@ -0,0 +1,16 @@
...
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	"github.com/pokt-network/poktroll/x/supplier/keeper"
+)
+
+// EndBlocker is called every block and handles supplier related updates.
+func EndBlocker(ctx sdk.Context, k keeper.Keeper) error {
+	if err := k.EndBlockerUnbondSupplier(ctx); err != nil {
```
Olshansk, line 11:
```suggestion
	if err := k.EndBlockerUnbondSuppliers(ctx); err != nil {
```

x/supplier/keeper/unbond_suppliers.go
```diff
@@ -0,0 +1,71 @@
...
+	"fmt"
+
+	cosmostypes "github.com/cosmos/cosmos-sdk/types"
+	"github.com/pokt-network/poktroll/x/shared"
+	sharedtypes "github.com/pokt-network/poktroll/x/shared/types"
+	"github.com/pokt-network/poktroll/x/supplier/types"
+)
+
+// EndBlockerUnbondSupplier unbonds suppliers that have finished the unbonding period.
+func (k Keeper) EndBlockerUnbondSupplier(ctx context.Context) error {
```
Olshansk, line 14:
```suggestion
// EndBlockerUnbondSuppliers unbonds suppliers that have finished the unbonding period.
func (k Keeper) EndBlockerUnbondSuppliers(ctx context.Context) error {
```

x/supplier/keeper/msg_server_unstake_supplier.go
```diff
@@ -7,6 +7,7 @@ import (
 	sdk "github.com/cosmos/cosmos-sdk/types"
```
Olshansk, line 7: In #489 we describe the unbonding time governance parameter.

Is that added here and I can't find it or will it be added in a future PR?

Olshansk (changes requested):
Partial review. Will continue tomorrow.

e2e/tests/init_test.go
```diff
@@ -566,6 +589,50 @@ func (s *suite) validateAmountChange(prevAmount, currAmount int, expectedAmountC
...
+		"--output=json",
+	}
+
+	res, err := s.pocketd.RunCommandOnHostWithRetry("", numQueryRetries, args...)
+	require.NoError(s, err, "error getting supplier module params")
+
+	var resp sharedtypes.QueryParamsResponse
+	responseBz := []byte(strings.TrimSpace(res.Stdout))
+	s.cdc.MustUnmarshalJSON(responseBz, &resp)
+	windowCloseHeight := shared.GetProofWindowCloseHeight(
```
Olshansk, line 629: Shouldn't we get this from the supplier field?

e2e/tests/session_steps_test.go
```diff
@@ -251,6 +251,35 @@ func (s *suite) waitForNewBlockEvent(
 	}
 }
 
+// waitForBlockHeight waits for a NewBlock event to be observed whose height is
+// greater than or equal to the target height.
+func (s *suite) waitForBlockHeight(targetHeight int64) {
```
Olshansk, line 256: Why not reuse `waitForNewBlockEvent(` inside where the `isEventMatchFn` checks the target height?

x/session/keeper/session_hydrator.go
```diff
@@ -167,6 +168,14 @@ func (k Keeper) hydrateSessionSuppliers(ctx context.Context, sh *sessionHydrator
...
+		// in the current session's suppliers list and must wait until the next one.
+		if s.UnstakeCommitSessionEndHeight != suppliertypes.SupplierNotUnstaking &&
```
Olshansk, line 174: can we create a `func (s *Supplier) isUnbonding() bool` function as a helper?

x/supplier/keeper/msg_server_unstake_supplier.go
```diff
@@ -37,23 +38,25 @@ func (k msgServer) UnstakeSupplier(
...
-		return nil, err
-	}
-
-	// Send the coins from the supplier pool back to the supplier
-	err = k.bankKeeper.SendCoinsFromModuleToAccount(ctx, types.ModuleName, supplierAddress, []sdk.Coin{*supplier.Stake})
-	if err != nil {
-		logger.Error(fmt.Sprintf("could not send %v coins from %s module to %s account due to %v", supplier.Stake, supplierAddress, types.ModuleName, err))
-		return nil, err
+	// Check if the supplier has already initiated the unstake action.
+	if supplier.UnstakeCommitSessionEndHeight != types.SupplierNotUnstaking {
```
Olshansk, line 42: The `s.IsUnbonding()` helper cuold be reused here.

x/supplier/keeper/msg_server_unstake_supplier.go
```diff
@@ -37,23 +38,25 @@ func (k msgServer) UnstakeSupplier(
...
+	sharedParams := k.sharedKeeper.GetParams(ctx)
+
+	// Mark the supplier as unstaking by recording the height at which it should stop
+	// providing service.
+	// The supplier MUST continue to provide service until the end of the current
+	// session. I.e., on-chain sessions' suppliers list MUST NOT change mid-session.
+	// Removing it right away could have undesired effects on the network
+	// (e.g. a session with less than the minimum or 0 number of suppliers,
+	// off-chain actors that need to listen to session supplier's change mid-session, etc).
+	supplier.UnstakeCommitSessionEndHeight = uint64(shared.GetSessionEndHeight(&sharedParams, currentHeight))
```
Olshansk, line 58: This should be based on a governance param based on the original ticket.

x/supplier/keeper/unbond_suppliers.go
```diff
@@ -0,0 +1,71 @@
...
+			k.RemoveSupplier(ctx, supplierAddress.String())
+			logger.Info(fmt.Sprintf("Successfully removed the supplier: %+v", supplier))
+		}
+	}
+
+	return nil
+}
+
+// GetSupplierUnbondingHeight returns the height at which the supplier can be unbonded.
+func (k Keeper) GetSupplierUnbondingHeight(
```
Olshansk, line 61: See the ticket as to how this should be based on a governance param

x/supplier/keeper/unbond_suppliers.go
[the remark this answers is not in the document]
Olshansk: #PUC w/ reference to example PR.

x/supplier/keeper/unbond_suppliers.go
```diff
@@ -0,0 +1,71 @@
...
+	currentHeight := sdkCtx.BlockHeight()
+
+	logger := k.Logger().With("method", "UnbondSupplier")
+
+	// Iterate over all suppliers and unbond suppliers that have finished the unbonding period.
+	// TODO_IMPROVE: Use an index to iterate over suppliers that have initiated the
+	// unbonding action instead of iterating over all suppliers.
+	for _, supplier := range k.GetAllSuppliers(ctx) {
+		// Ignore suppliers that have not initiated the unbonding action.
+		if supplier.UnstakeCommitSessionEndHeight == types.SupplierNotUnstaking {
```
Olshansk, line 25: Another example where a helper could be reused.

x/supplier/keeper/unbond_suppliers.go
```diff
@@ -0,0 +1,71 @@
...
+			continue
+		}
+
+		unbondingHeight := k.GetSupplierUnbondingHeight(ctx, &supplier)
+
+		if unbondingHeight <= currentHeight {
```
Olshansk, line 31: 1. Shouldn't this be >= ?
2. IF I'm correct, do we not have integration tests for this?

Olshansk (changes requested):

x/supplier/keeper/unbond_suppliers.go
red-0ne: @Olshansk , I added it but then removed it after realising that unbonding should not happen before the `Supplier`'s `ProofWindowEndHeight`.

We can have a gov param that extends this height, making it an extension to `ProofWindowEndHeight`, I'll put it back if you think that we need that extension.

x/supplier/types/errors.go
Olshansk: @red-0ne I talked to Bryan about this. Can you just update the docs next to the enum/types making it very very clear what (un)staking and (un)bonding is?

e2e/tests/init_test.go
```diff
@@ -566,6 +589,50 @@ func (s *suite) validateAmountChange(prevAmount, currAmount int, expectedAmountC
...
+
+	args := []string{
+		"query",
+		"shared",
+		"params",
+		"--output=json",
+	}
+
+	res, err := s.pocketd.RunCommandOnHostWithRetry("", numQueryRetries, args...)
+	require.NoError(s, err, "error getting supplier module params")
```
bryanchriswhite, line 624:
```suggestion
	require.NoError(s, err, "error getting shared module params")
```

x/session/keeper/session_hydrator.go
bryanchriswhite:
```suggestion
		// TODO_TECHDEBT(#695): Suppliers that stake mid-session SHOULD NOT be included
```

x/session/keeper/session_hydrator.go
bryanchriswhite: It would have to take a height argument.

```
func (s *Supplier) isUnbonding(queryHeight int64) bool
```

e2e/tests/init_test.go
```diff
@@ -566,6 +589,50 @@ func (s *suite) validateAmountChange(prevAmount, currAmount int, expectedAmountC
...
+
+	var resp sharedtypes.QueryParamsResponse
+	responseBz := []byte(strings.TrimSpace(res.Stdout))
+	s.cdc.MustUnmarshalJSON(responseBz, &resp)
+	windowCloseHeight := shared.GetProofWindowCloseHeight(
+		&resp.Params,
+		int64(supplier.UnstakeCommitSessionEndHeight),
+	)
```
bryanchriswhite, line 632: ~~Wdyt about moving this into a new static function like `shared.GetUnbondedHeight()` or something?~~

Now seeing the keeper method, left a comment there.

x/supplier/keeper/unbond_suppliers.go
bryanchriswhite: I think this should be refactored to a static function in the shared pkg alongside the window height helpers.

x/supplier/keeper/msg_server_unstake_supplier_test.go
```diff
@@ -54,19 +36,85 @@ func TestMsgServer_UnstakeSupplier_Success(t *testing.T) {
...
+	require.NoError(t, err)
+
+	// Make sure the supplier is no longer in the unbonding period
+	foundSupplier, isSupplierFound = k.GetSupplier(ctx, supplierAddr)
+	require.True(t, isSupplierFound)
+	require.Equal(t, foundSupplier.UnstakeCommitSessionEndHeight, types.SupplierNotUnstaking)
+
+	ctx = keepertest.SetBlockHeight(ctx, int64(unbondingHeight))
+
+	// Run the endblocker to unbond the supplier
```
bryanchriswhite, line 108:
```suggestion
	// Run the endblocker, the supplier should not be unbonded.
```

x/supplier/keeper/unbond_suppliers.go
```diff
@@ -0,0 +1,71 @@
...
+				return err
+			}
+
+			// Send the coins from the supplier pool back to the supplier.
+			if err = k.bankKeeper.SendCoinsFromModuleToAccount(
+				ctx, types.ModuleName, supplierAddress, []cosmostypes.Coin{*supplier.Stake},
+			); err != nil {
+				logger.Error(fmt.Sprintf(
+					"could not send %v coins from %s module to %s account due to %v",
+					supplier.Stake, supplierAddress, types.ModuleName, err,
```
bryanchriswhite, line 46:
```suggestion
					"could not send %s coins from %s module to %s account due to %v",
					supplier.Stake.String(), supplierAddress, types.ModuleName, err,
```

x/supplier/types/params.go
bryanchriswhite:
```suggestion
	// SupplierNotUnstakingHeight is the value of `unstake_session_end_height` if the
	// supplier is not actively in the unbonding period.
	SupplierNotUnstakingHeight uint64              = 0
```

bryanchriswhite (changes requested):

e2e/tests/init_test.go
red-0ne: The supplier has the `SessionEndHeight` of the block at which the `MsgUnstake` tx was committed. I'll try storing the `UnbondingHeight` to see if its better API.

e2e/tests/session_steps_test.go
red-0ne: IIRC it's because `isEventMatchFn` receives `abci.Event`s from `newBlockEvent.Data.Value.ResultFinalizeBlock.Events` which do not contain the block height.

Block height is in `newBlockEvent.Data.Value.Block.Header.Height`.

There's probably room for a refactor but this will affect what `isEventMatchFn` receives.

x/session/keeper/session_hydrator.go
red-0ne: @bryanchriswhite, I think there's a nuance here that needs to be considered given the value of `UnstakeSessionEndHeight`:
* `0`: never unstaked
* `!= 0 && < currentSessionEnd`: initiated an unstake but still active (serving relays)
* `> currentSessionEnd`: is unbonding and no longer active (serving relays)

I believe `IsUnbonding` should only indicate if the supplier initiated an unstake, and not whether it's sill active or not. This would maximize the reusablility of the method.

Given that, it's signature should be `IsUnbonding() bool` and should not take the height as argument.

We could still have a `IsActive(queryHeight) bool` that Indicates whether it should be included in sessions or not.

x/supplier/keeper/msg_server_unstake_supplier.go
red-0ne: We are not storing when the unbonding should be effective, but rather when the `Supplier` becomes inactive.

The unbonding height is derived from this one at the `EndBlocker` level.

x/supplier/keeper/unbond_suppliers.go
red-0ne: > Shouldn't this be >= ?

We unbond when the unbonding height is anterior (less) than the current height.

> IF I'm correct, do we not have integration tests for this?

We do [x/supplier/keeper/msg_server_unstake_supplier_test.go#L54](https://github.com/pokt-network/poktroll/pull/666/files#diff-bd629e76fe1cba24cc427beb9d08642a2b767cb67f1573e2c2a502cf8e9dd618R54)

x/shared/types/supplier.go
```diff
@@ -0,0 +1,13 @@
...
+// SupplierNotUnstaking is the value of `unstake_session_end_height` if the
+// supplier is not actively in the unbonding period.
+const SupplierNotUnstaking uint64 = 0
+
+func (s *Supplier) IsUnbonding() bool {
+	return s.UnstakeSessionEndHeight != SupplierNotUnstaking
+}
+
+func (s *Supplier) IsActive(queryHeight int64) bool {
+	return !s.IsUnbonding() || uint64(queryHeight) <= s.UnstakeSessionEndHeight
```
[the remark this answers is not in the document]
red-0ne: Notice that `IsActive` checks whether the `queryHeight` is anterior to the `Supplier`'s **`UnstakeSessionEndHeight`** (i.e. The session end height that contains the **unstake** tx).

While the `unbond_suppliers.go` check is comparing the `currentHeight` to the **`SupplierUnbondingHeight`** (which **is not the same** as `UnstakeSessionEndHeight`)

x/supplier/keeper/unbond_suppliers.go
red-0ne: > Should we use `!supplier.IsActive()` here?

No, because `IsActive` returns `true` for the `Supplier`'s `[stake, unstake]` interval. While here, we check against the `[unstake, proof_window_close]` interval.

x/shared/types/supplier.go
```diff
@@ -0,0 +1,20 @@
+package types
+
+// SupplierNotUnstaking is the value of `unstake_session_end_height` if the
+// supplier is not actively in the unbonding period.
+const SupplierNotUnstaking uint64 = 0
+
+// IsUnbonding returns whether the supplier has submitted an unstake message,
```
Olshansk, line 7:
```suggestion
// IsUnbonding returns true if the supplier is actively unbonding.
// It determines if the supplier has submitted an unstake message,
```

x/shared/types/supplier.go
```diff
@@ -0,0 +1,20 @@
...
+// in which case the supplier has an UnstakeSessionEndHeight set.
+func (s *Supplier) IsUnbonding() bool {
+	return s.UnstakeSessionEndHeight != SupplierNotUnstaking
+}
+
+// IsActive returns whether the supplier is allowed to serve requests at the
```
Olshansk, line 13: Great comment!

e2e/tests/init_test.go
```diff
@@ -442,6 +443,24 @@ func (s *suite) AModuleEndBlockEventIsBroadcast(module, eventType string) {
...
+
+	supplier := s.getSupplierInfo(accName)
+	require.True(s, supplier.IsUnbonding())
+}
+
+func (s *suite) TheUserWaitsForUnbondingPeriodToFinish(accName string) {
+	_, ok := accNameToSupplierMap[accName]
+	require.True(s, ok, "supplier %s not found", accName)
+
+	unbondingHeight := s.getSupplierUnbondingHeight(accName)
```
Olshansk, line 460: Great and simple!

x/shared/supplier.go
```diff
@@ -0,0 +1,12 @@
...
+
+import (
+	sharedtypes "github.com/pokt-network/poktroll/x/shared/types"
+)
+
+func GetSupplierUnbondingHeight(
+	sharedParams *sharedtypes.Params,
+	supplier *sharedtypes.Supplier,
+) int64 {
+	return GetProofWindowCloseHeight(sharedParams, int64(supplier.UnstakeSessionEndHeight))
```
Olshansk, line 11:
```suggestion
	// TODO_UPNEXT(@red-0ne): Add a governance parameter called `supplier_unbonding_period`
	// equal to the number of blocks required to unbond. The value should enforce (when being updated)
	// to be after proof window close height and should still round to the end of the nearest session.
	return GetProofWindowCloseHeight(sharedParams, int64(supplier.UnstakeSessionEndHeight))
```

x/supplier/keeper/unbond_suppliers.go
```diff
@@ -0,0 +1,68 @@
...
+	for _, supplier := range k.GetAllSuppliers(ctx) {
+		// Ignore suppliers that have not initiated the unbonding action.
+		if !supplier.IsUnbonding() {
+			continue
+		}
+
+		unbondingHeight := k.GetSupplierUnbondingHeight(ctx, &supplier)
+
+		// If the unbonding height is older (less) than the current height, unbond the supplier.
+		if unbondingHeight <= currentHeight {
```
Olshansk, line 32:
```suggestion
		// If the unbonding height is ahead of the current height, the supplier stays in the unbonding state.
		if unbondingHeight > currentHeight {
			continue
		}
```

x/shared/types/supplier.go
```diff
@@ -0,0 +1,20 @@
...
+// given query height.
+// A supplier that has not submitted an unstake message is always active.
+// A supplier that has submitted an unstake message is active until the end of
+// the session containing the height at which unstake message was submitted.
+func (s *Supplier) IsActive(queryHeight int64) bool {
+	return !s.IsUnbonding() || uint64(queryHeight) <= s.UnstakeSessionEndHeight
```
Olshansk, line 19:
```suggestion
	// TODO_UPNEXT(@red-0ne): When introducing a governance parameter for unbonding, this will also need
	// to be updated to reflect sessions when the supplier is not active.
	return !s.IsUnbonding() || uint64(queryHeight) <= s.UnstakeSessionEndHeight
```

Olshansk (changes requested):

x/shared/supplier.go
```diff
@@ -0,0 +1,17 @@
+package shared
+
+import (
+	sharedtypes "github.com/pokt-network/poktroll/x/shared/types"
+)
+
+// GetProofWindowCloseHeight returns the height at which the given supplier finishes unbonding.
```
Olshansk, line 7:
```suggestion
// GetSupplierUnbondingHeight returns the height at which the given supplier finishes unbonding.
```

Olshansk (approved):

PR #709, opened 2024-07-26
[Bug] Prevent commit txs to be submitted outside close windows

red-0ne:
## Summary

This PR reduces `distributionWindowSizeBlocks` by `closeOffsetBlocks` to prevent `Supplier`s to have their `EarliestSupplier{Claim/Proof}CommitHeight` to be committed too late such that the `windowVerification` rejects it due to `ErrProof{Claim/Proof}OutsideOfWindow`.

## Issue

image

## Type of change

Select one or more:

- [x] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

## Sanity Checklist

## Summary by CodeRabbit

- **New Features**
	- Introduced a buffer mechanism for supplier claim and proof submission windows to enhance reliability.
	- Added a constant to improve the calculation of distribution window sizes, allowing suppliers more time for submissions.

- **Bug Fixes**
	- Adjusted parameters in testing to extend the duration of claim and proof windows, potentially improving outcomes related to supplier commitments.

### DIFF
--- a/x/proof/keeper/session.go
+++ b/x/proof/keeper/session.go
@@ -109,6 +109,8 @@ func (k Keeper) validateClaimWindow(
 	// TODO_BLOCKER(@bryanchriswhite, @red-0ne): Enforce an additional "latest
 	// supplier claim/proof commit offset" such that all suppliers have the same
 	// "supplier claim/proof commit window" size.
+	// This also prevent suppliers to end-up with only 1 block window to submit claims
+	// when EarliestSupplierClaimCommitHeight is too close to ClaimWindowCloseHeight.
 	// See: https://github.com/pokt-network/poktroll/pull/620/files#r1656548473.
 	if currentHeight < earliestClaimCommitHeight {
 		return types.ErrProofClaimOutsideOfWindow.Wrapf(
--- a/x/shared/session.go
+++ b/x/shared/session.go
@@ -5,6 +5,10 @@ import (
 	sharedtypes "github.com/pokt-network/poktroll/x/shared/types"
 )
 
+// closeBlockOffset is the offset to subtract from the earliest supplier commit
+// height to ensure that the supplier has enough time to commit the claim/proof.
+const closeBlockOffset = 1
+
 // TODO_DOCUMENT(@bryanchriswhite): Move this into the documentation: https://github.com/pokt-network/poktroll/pull/571#discussion_r1630923625
 
 // GetSessionStartHeight returns the block height at which the session containing
@@ -118,7 +122,7 @@ func GetEarliestSupplierClaimCommitHeight(
 	// window open block hash and the supplier address.
 	randomNumber := poktrand.SeededInt63(claimWindowOpenBlockHash, []byte(supplierAddr))
 
-	distributionWindowSizeBlocks := sharedParams.GetClaimWindowCloseOffsetBlocks()
+	distributionWindowSizeBlocks := sharedParams.GetClaimWindowCloseOffsetBlocks() - closeBlockOffset
 	randCreateClaimHeightOffset := randomNumber % int64(distributionWindowSizeBlocks)
 
 	return claimWindowOpenHeight + randCreateClaimHeightOffset
@@ -139,7 +143,7 @@ func GetEarliestSupplierProofCommitHeight(
 	// window open block hash and the supplier address.
 	randomNumber := poktrand.SeededInt63(proofWindowOpenBlockHash, []byte(supplierAddr))
 
-	distributionWindowSizeBlocks := sharedParams.GetProofWindowCloseOffsetBlocks()
+	distributionWindowSizeBlocks := sharedParams.GetProofWindowCloseOffsetBlocks() - closeBlockOffset
 	randCreateProofHeightOffset := randomNumber % int64(distributionWindowSizeBlocks)
 
 	return proofWindowOpenHeight + randCreateProofHeightOffset
--- a/x/shared/session.go
+++ b/x/shared/session.go
@@ -122,8 +122,8 @@ func GetEarliestSupplierClaimCommitHeight(
 	// window open block hash and the supplier address.
 	randomNumber := poktrand.SeededInt63(claimWindowOpenBlockHash, []byte(supplierAddr))
 
-	distributionWindowSizeBlocks := sharedParams.GetClaimWindowCloseOffsetBlocks() - closeBlockOffset
-	randCreateClaimHeightOffset := randomNumber % int64(distributionWindowSizeBlocks)
+	claimWindowNumBlocks := sharedParams.GetClaimWindowCloseOffsetBlocks() - closeBlockOffset
+	randCreateClaimHeightOffset := randomNumber % int64(claimWindowNumBlocks)
 
 	return claimWindowOpenHeight + randCreateClaimHeightOffset
 }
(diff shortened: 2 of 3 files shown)


x/shared/session.go
```diff
@@ -5,6 +5,10 @@ import (
 	sharedtypes "github.com/pokt-network/poktroll/x/shared/types"
 )
 
+// closeBlockOffset is the offset to subtract from the earliest supplier commit
+// height to ensure that the supplier has enough time to commit the claim/proof.
+const closeBlockOffset = 1
```
Olshansk, line 10: The PR description did not make it clear that you're introducing a NEW parameter. 

Haven't thought through everything yet - but the fact that we have like 4 params + grace periods and now we need to add more offsets on top of offsets seems like a code smell.

x/shared/session.go
```diff
@@ -118,7 +122,7 @@ func GetEarliestSupplierClaimCommitHeight(
 	// window open block hash and the supplier address.
 	randomNumber := poktrand.SeededInt63(claimWindowOpenBlockHash, []byte(supplierAddr))
 
-	distributionWindowSizeBlocks := sharedParams.GetClaimWindowCloseOffsetBlocks()
+	distributionWindowSizeBlocks := sharedParams.GetClaimWindowCloseOffsetBlocks() - closeBlockOffset
```
Olshansk, line 125: s/distributionWindowSizeBlocks/claimWindowNumBlocks

x/shared/session.go
```diff
@@ -139,7 +143,7 @@ func GetEarliestSupplierProofCommitHeight(
 	// window open block hash and the supplier address.
 	randomNumber := poktrand.SeededInt63(proofWindowOpenBlockHash, []byte(supplierAddr))
 
-	distributionWindowSizeBlocks := sharedParams.GetProofWindowCloseOffsetBlocks()
+	distributionWindowSizeBlocks := sharedParams.GetProofWindowCloseOffsetBlocks() - closeBlockOffset
```
Olshansk, line 146: s/distributionWindowSizeBlocks/proofWindowNuMBlocks

x/shared/session.go
```diff
@@ -5,6 +5,10 @@ import (
 	sharedtypes "github.com/pokt-network/poktroll/x/shared/types"
```
Olshansk, line 5: @red-0ne Do we have a mermaid diagram anywhere that shows the windows & params?

I'd advocate to have it be done now because even I'm personally beginning to get confused and lose track of what's going on.

x/shared/session.go
Olshansk, line 10: This seems like the wrong approach. 

Your goal (in the PR description) is to avoid having the supplier be late.

The on-chain protocol (under x/) sets the rules. 
The off-chain libraries have to follow the protocol.

This laws (protocol) don't bend to servers (web2).

If we have an offset like this, it has to go off-chain.

x/shared/session_test.go
```diff
@@ -168,9 +168,9 @@ func TestClaimProofWindows(t *testing.T) {
 			sharedParams: sharedtypes.Params{
```
Olshansk, line 168: These sorts of bugs are where putting together a real unit test goes a really long way.

Olshansk (changes requested):
@red-0ne I know you weren't planning on it, but I'm going to increase the scope of this bug fix because it's really critical that this is clear, documented, tested and done correctly.

x/shared/session.go
red-0ne: This doesn't necessarily need to be a gov param and the actual value should be just fine for most of the cases (we could even inline it).

I really don't see an easier way to handle this edge case. Which is, just taking into account the nature of transactions submission that may not be committed at the height of the submission (for reasons like mempool congestion, network latency...)

red-0ne: Maybe the PR description is missing nuance.

We are trying to avoid that the protocol imposes an `EarliestSupplierCommitHeight` that is too constraining, making its chances to submit a failing (out of window) transaction non-negiligible.

On a `[10, 20]` commit window it basically prevents the protocol from imposing to the supplier to submit his tx at block `19` or later.

x/shared/session_test.go
red-0ne: This bug could not be observed by a unit test as its causes are mostly timing and latency.

x/shared/session.go
Olshansk, line 10: I ran a small program:

```
func main() {
	claimOpenHeight := 10
	claimWindow := 5
	for i := range 20 {
		result := claimOpenHeight + (i % claimWindow)
		fmt.Println(result)
	}
}
```

Which outputted things in `[10, 14]` INCLUSIVE.

Am I correct in understanding that the way this is currently working is that sometimes the supplier only has 1 block between "earliest claim commit height" and "claim window close" during which they have to do this?

If this is the case, then:
1. We can't and should not bandaid this but rather have a proper solution
2. If my understanding of (1) is not correct, please elaborate and help me understand

Olshansk (changes requested):

x/shared/session.go
```diff
@@ -118,8 +122,8 @@ func GetEarliestSupplierClaimCommitHeight(
 	// window open block hash and the supplier address.
 	randomNumber := poktrand.SeededInt63(claimWindowOpenBlockHash, []byte(supplierAddr))
 
-	distributionWindowSizeBlocks := sharedParams.GetClaimWindowCloseOffsetBlocks()
-	randCreateClaimHeightOffset := randomNumber % int64(distributionWindowSizeBlocks)
+	claimWindowNumBlocks := sharedParams.GetClaimWindowCloseOffsetBlocks() - closeBlockOffset
+	randCreateClaimHeightOffset := randomNumber % int64(claimWindowNumBlocks)
 
 	return claimWindowOpenHeight + randCreateClaimHeightOffset
```
Olshansk, line 128:
```suggestion
	// TODO_POST_MAINNET: Determine if we need to enforce claim distribution in order to avoid
	// all claims or proofs being accidently submitted at the same time.
	// claimWindowNumBlocks := sharedParams.GetClaimWindowCloseOffsetBlocks() - closeBlockOffset
	// randCreateClaimHeightOffset := randomNumber % int64(claimWindowNumBlocks)
	// return claimWindowOpenHeight + randCreateClaimHeightOffset
	return claimWindowOpenHeight
```

x/shared/session.go
Olshansk: Rought notes from convo with @red-0ne :

10: theoretical earliest
11: potential earliest
12: potential earliest
13: potential earliest
14: potential earliest
15: absolute latest

Situation: SUpplier 10:
Earliest: 14
Aboslute latest: 15

Assume 1 min block times.

If supplier downtime (e.g. server is down) for 61 seconds -> loses all money.

x/proof/keeper/session.go
```diff
@@ -109,6 +109,8 @@ func (k Keeper) validateClaimWindow(
 	// TODO_BLOCKER(@bryanchriswhite, @red-0ne): Enforce an additional "latest
```
Olshansk, line 109: Remove this comment

x/shared/session.go
```diff
@@ -118,8 +122,8 @@ func GetEarliestSupplierClaimCommitHeight(
 	// window open block hash and the supplier address.
 	randomNumber := poktrand.SeededInt63(claimWindowOpenBlockHash, []byte(supplierAddr))
 
-	distributionWindowSizeBlocks := sharedParams.GetClaimWindowCloseOffsetBlocks()
-	randCreateClaimHeightOffset := randomNumber % int64(distributionWindowSizeBlocks)
+	claimWindowNumBlocks := sharedParams.GetClaimWindowCloseOffsetBlocks() - closeBlockOffset
```
Olshansk, line 125: Remove `GetEarliestSupplierClaimCommitHeight` everywhere

x/shared/session.go
Olshansk, line 10: Remove

Olshansk (changes requested):
Discussed a bunch of stuff offline. Let's overhaul this thing!

red-0ne:
Closing in favor of #712

Issue #447, opened 2024-03-25
[RelayMiner] Add an ability to test RelayMiner configuration without application


okdas:
## Objective

The current Morse mainnet client has a feature to send relays without application signature. We should add this functionality to RelayMiner: an ability to send relay to RelayMiner withough providing meta information (session, signature, etc) so operators can verify their RelayMiner configuration.

## Origin Document

Requests from the community:
- [Discord](https://discord.com/channels/553741558869131266/1217918754801975398/1221908766765875240);

First mentioned in [internal communication](https://discord.com/channels/824324475256438814/1138895490331705354/1219810632841429113).

In the current Morse client, the feature is enabled by adding `--simulateRelay` argument. In our implementation, we can consider adding this to the relayminer config file.

## Goals

- Operators can configure their RelayMiners to send requests to the backend service to make sure applications will get expected service.

## General deliverables

- [ ] **Comments**: Add/update TODOs and comments alongside the source code so it is easier to follow.
- [ ] **Testing**: Add new tests (unit and/or E2E) to the test suite.
- [ ] **Makefile**: Add new targets to the Makefile to make the new functionality easier to use.
- [ ] **Documentation**: Update architectural or development READMEs; use [mermaid](https://mermaid-js.github.io/mermaid/) diagrams where appropriate.

---

**Creator**: @okdas
**Co-Owners**: [OPTIONAL - GitHub handle of co-owner(s)]

Olshansk:
@okdas to provide a link to `pocket-core` and an example of how it works on Morse.

okdas:
In `pocket-core` guidelines, we recommend node runners to [start the process](https://github.com/pokt-network/docs/blob/d128654cb088b5b680e51cdf1589dacec6e7311c/content/node/setup.md?plain=1#L145) with `--simulateRelay` prior to staking, so they can test before going live. That way, they would avoid being ignored by the gateway (or getting jailed/slashed - though I don't think this is functionality is turned on).

This argument then [adds a new http route](https://github.com/pokt-network/pocket-core/blob/staging/app/cmd/rpc/server.go#L23) where the relay simulation request can be sent to. For example, this is how I used to do that:

Fuse:
```
curl -X POST --data '{"relay_network_id":"0005","payload":{"data":"{\"jsonrpc\":\"2.0\",\"method\":\"eth_blockNumber\",\"params\":[],\"id\":0}","method":"POST","path":"","headers":{}}}'  http://localhost:8081/v1/client/sim
```

[...]

PR #744, opened 2024-08-17
[RelayMiner]: add `proxy.Ping(...)` capability to test connectivity between relay servers and backend URLs

eddyzags:
## Summary

This PR adds the capability to test the connectivity between the Relay Servers and the Backend URLs in two ways.

1. Safeguard at Startup:
For every `suppliers.[].service_config.backend_url` referenced as input inside the Relay Miner Configuration file, the Relay Proxy will verify wether the network connection between the targeted `backend_url` and the relayerminer process is functioning properly. If one or more connections aren't possible, the relay miner won't be able to start.
 
2. Configurable Ping HTTP server:
The Relay Miner process will listen for incoming request to synchronously test the connectivity of every referenced `suppliers.[].service_config.backend_url`. If one or more backend URLs aren't reachable, the incoming request will fail.

Based on the `serverConfig.ServerType` (Example: HTTP), each Server Type will implement their own logic to implement to test the connectivity. 

## Issue

- https://github.com/pokt-network/poktroll/issues/447

## Type of change

Select one or more:

- [X] New feature, functionality or library
- [ ] Bug fix
- [ ] Code health or cleanup
- [X] Documentation
- [ ] Other (specify)

## Testing

- [X] `make docusaurus_start`; only needed if you make doc changes

- [X] **Unit Tests**: `make go_develop_and_test`

## Sanity Checklist

- [X] I have tested my changes using the available tooling
- [X] I have commented my code
- [X] I have performed a self-review of my own code; both comments & source code
- [X] I create and reference any new tickets, if applicable
- [X] I have left TODOs throughout the codebase, if applicable

## Summary by CodeRabbit

- **New Features**
	- Introduced a new configuration section for the `ping` functionality, allowing users to test backend connectivity within the relay miner's setup.
	- Added methods to handle ping requests, enhancing health check capabilities for relay servers.

- **Bug Fixes**
	- Improved error handling during the server startup process if any relay server is unreachable.

- **Tests**
	- Added tests for the new ping functionality to ensure operational integrity and reliability of the relay miner.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -81,38 +81,9 @@ endif
 .PHONY: install_ci_deps
 install_ci_deps: ## Installs `mockgen` and other go tools
 	go install "github.com/golang/mock/mockgen@v1.6.0" && mockgen --version
-	go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.60.3 && golangci-lint --version
+	go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.59.1 && golangci-lint --version
 	go install golang.org/x/tools/cmd/goimports@latest
 	go install github.com/mikefarah/yq/v4@latest
-
-.PHONY: install_cosmovisor
-install_cosmovisor: ## Installs `cosmovisor`
-	go install cosmossdk.io/tools/cosmovisor/cmd/cosmovisor@v1.6.0 && cosmovisor version --cosmovisor-only
-
-.PHONY: cosmovisor_cross_compile
-cosmovisor_cross_compile: # Installs multiple cosmovisor binaries for different platforms (used by Dockerfile.release)
-	@COSMOVISOR_VERSION="v1.6.0"; \
-	PLATFORMS="linux/amd64 linux/arm64"; \
-	mkdir -p ./tmp; \
-	echo "Fetching Cosmovisor source..."; \
-	temp_dir=$$(mktemp -d); \
-	cd $$temp_dir; \
-	go mod init temp; \
-	go get cosmossdk.io/tools/cosmovisor/cmd/cosmovisor@$$COSMOVISOR_VERSION; \
-	for platform in $$PLATFORMS; do \
-		OS=$${platform%/*}; \
-		ARCH=$${platform#*/}; \
-		echo "Compiling for $$OS/$$ARCH..."; \
-		GOOS=$$OS GOARCH=$$ARCH go build -o $(CURDIR)/tmp/cosmovisor-$$OS-$$ARCH cosmossdk.io/tools/cosmovisor/cmd/cosmovisor; \
-	done; \
-	cd $(CURDIR); \
-	rm -rf $$temp_dir; \
-	echo "Compilation complete. Binaries are in ./tmp/"; \
-	ls -l ./tmp/cosmovisor-*
-
-.PHONY: cosmovisor_clean
-cosmovisor_clean:
-	rm -f ./tmp/cosmovisor-*
 
 ########################
 ### Makefile Helpers ###
@@ -131,6 +102,167 @@ list: ## List all make targets
 .DEFAULT_GOAL := help
 help: ## Prints all the targets in all the Makefiles
 	@grep -h -E '^[a-zA-Z0-9_-]+:.*?## .*$$' $(MAKEFILE_LIST) | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-60s\033[0m %s\n", $$1, $$2}'
+
+##############
+### Checks ###
+##############
+
+# TODO_DOCUMENT: All of the `check_` helpers can be installed differently depending
+# on the user's OS and environment.
+# NB: For mac users, you may need to install with the proper linkers: https://github.com/golang/go/issues/65940
+
+.PHONY: check_go_version
+# Internal helper target - check go version
+check_go_version:
+	@# Extract the version number from the `go version` command.
+	@GO_VERSION=$$(go version | cut -d " " -f 3 | cut -c 3-) && \
+	MAJOR_VERSION=$$(echo $$GO_VERSION | cut -d "." -f 1) && \
+	MINOR_VERSION=$$(echo $$GO_VERSION | cut -d "." -f 2) && \
+	\
+	if [ "$$MAJOR_VERSION" -ne 1 ] || [ "$$MINOR_VERSION" -le 20 ] ; then \
+		echo "Invalid Go version. Expected 1.21.x or newer but found $$GO_VERSION"; \
+		exit 1; \
+	fi
+
+.PHONY: check_ignite_version
+# Internal helper target - check ignite version
+check_ignite_version:
+	@version=$$(ignite version 2>/dev/null | grep 'Ignite CLI version:' | awk '{print $$4}') ; \
+	if [ "$$(printf "v28\n$$version" | sort -V | head -n1)" != "v28" ]; then \
+		echo "Error: Version $$version is less than v28. Exiting with error." ; \
+		exit 1 ; \
+	fi
+
+.PHONY: check_mockgen
+# Internal helper target- Check if mockgen is installed
+check_mockgen:
+	{ \
+	if ( ! ( command -v mockgen >/dev/null )); then \
+		echo "Seems like you don't have `mockgen` installed. Please visit https://github.com/golang/mock#installation and follow the instructions to install `mockgen` before continuing"; \
+		exit 1; \
+	fi; \
+	}
+
+
+.PHONY: check_act
+# Internal helper target - check if `act` is installed
+check_act:
+	{ \
+	if ( ! ( command -v act >/dev/null )); then \
+		echo "Seems like you don't have `act` installed. Please visit https://github.com/nektos/act before continuing"; \
+		exit 1; \
+	fi; \
+	}
+
+.PHONY: check_gh
+# Internal helper target - check if `gh` is installed
+check_gh:
+	{ \
+	if ( ! ( command -v gh >/dev/null )); then \
+		echo "Seems like you don't have `gh` installed. Please visit https://cli.github.com/ before continuing"; \
+		exit 1; \
+	fi; \
+	}
+
+.PHONY: check_docker
+# Internal helper target - check if docker is installed
+check_docker:
+	{ \
+	if ( ! ( command -v docker >/dev/null && (docker compose version >/dev/null || command -v docker-compose >/dev/null) )); then \
+		echo "Seems like you don't have Docker or docker-compose installed. Make sure you review build/localnet/README.md and docs/development/README.md  before continuing"; \
+		exit 1; \
+	fi; \
+	}
+.PHONY: check_kind
+# Internal helper target - check if kind is installed
+check_kind:
+	@if ! command -v kind >/dev/null 2>&1; then \
+		echo "kind is not installed. Make sure you review build/localnet/README.md and docs/development/README.md  before continuing"; \
+		exit 1; \
+	fi
+
+.PHONY: check_docker_ps
+ ## Internal helper target - checks if Docker is running
+check_docker_ps: check_docker
+	@echo "Checking if Docker is running..."
+	@docker ps > /dev/null 2>&1 || (echo "Docker is not running. Please start Docker and try again."; exit 1)
+
+.PHONY: check_kind_context
+## Internal helper target - checks if the kind-kind context exists and is set
+check_kind_context: check_kind
+	@if ! kubectl config get-contexts | grep -q 'kind-kind'; then \
+		echo "kind-kind context does not exist. Please create it or switch to it."; \
+		exit 1; \
+	fi
+	@if ! kubectl config current-context | grep -q 'kind-kind'; then \
+		echo "kind-kind context is not currently set. Use 'kubectl config use-context kind-kind' to set it."; \
+		exit 1; \
+	fi
+
+
+.PHONY: check_godoc
+# Internal helper target - check if godoc is installed
+check_godoc:
+	{ \
+	if ( ! ( command -v godoc >/dev/null )); then \
+		echo "Seems like you don't have godoc installed. Make sure you install it via 'go install golang.org/x/tools/cmd/godoc@latest' before continuing"; \
+		exit 1; \
+	fi; \
+	}
+
+.PHONY: check_npm
+# Internal helper target - check if npm is installed
+check_npm:
+	{ \
+	if ( ! ( command -v npm >/dev/null )); then \
+		echo "Seems like you don't have npm installed. Make sure you install it before continuing"; \
+		exit 1; \
+	fi; \
+	}
+
+.PHONY: check_jq
+# Internal helper target - check if jq is installed
+check_jq:
+	{ \
+	if ( ! ( command -v jq >/dev/null )); then \
+		echo "Seems like you don't have jq installed. Make sure you install it before continuing"; \
+		exit 1; \
+	fi; \
+	}
+
+.PHONY: check_yq
+# Internal helper target - check if `yq` is installed
+check_yq:
+	{ \
+	if ( ! ( command -v yq >/dev/null )); then \
+		echo "Seems like you don't have `yq` installed. Make sure you install it before continuing"; \
+		exit 1; \
+	fi; \
+	}
+
+.PHONY: check_node
+# Internal helper target - check if node is installed
+check_node:
+	{ \
+	if ( ! ( command -v node >/dev/null )); then \
+		echo "Seems like you don't have node installed. Make sure you install it before continuing"; \
+		exit 1; \
+	fi; \
+	}
+
+.PHONY: check_proto_unstable_marshalers
+check_proto_unstable_marshalers: ## Check that all protobuf files have the `stable_marshalers_all` option set to true.
+	go run ./tools/scripts/protocheck/cmd unstable
+
+.PHONY: fix_proto_unstable_marshalers
+fix_proto_unstable_marshalers: ## Ensure the `stable_marshaler_all` option is present on all protobuf files.
+	go run ./tools/scripts/protocheck/cmd unstable --fix
+	${MAKE} proto_regen
+
+
+.PHONY: warn_destructive
+warn_destructive: ## Print WARNING to the user
+	@echo "This is a destructive action that will affect docker resources outside the scope of this repo!"
 
 #######################
 ### Proto  Helpers ####
@@ -182,6 +314,75 @@ docker_wipe: check_docker warn_destructive prompt_user ## [WARNING] Remove all t
 	docker images -q | xargs -r -I {} docker rmi {}
 	docker volume ls -q | xargs -r -I {} docker volume rm {}
 
+########################
+### Localnet Helpers ###
+########################
+
+.PHONY: localnet_up
+localnet_up: check_docker_ps check_kind_context proto_regen localnet_regenesis ## Starts up a clean localnet
+	tilt up
+
+.PHONY: localnet_up_quick
+localnet_up_quick: check_docker_ps check_kind_context ## Starts up a localnet without regenerating fixtures
+	tilt up
+
+.PHONY: localnet_down
+localnet_down: ## Delete resources created by localnet
+	tilt down
+
+.PHONY: localnet_regenesis
+localnet_regenesis: check_yq warn_message_acc_initialize_pubkeys ## Regenerate the localnet genesis file
+# NOTE: intentionally not using --home <dir> flag to avoid overwriting the test keyring
+	@echo "Initializing chain..."
+	@set -e
+	@ignite chain init --skip-proto
+	AUTH_CONTENT=$$(cat ./tools/scripts/authz/dao_genesis_authorizations.json | jq -r tostring); \
+	$(SED) -i -E 's!^(\s*)"authorization": (\[\]|null)!\1"authorization": '$$AUTH_CONTENT'!' ${HOME}/.poktroll/config/genesis.json;
+
+	@cp -r ${HOME}/.poktroll/keyring-test $(POKTROLLD_HOME)
+	@cp -r ${HOME}/.poktroll/config $(POKTROLLD_HOME)/
+
+.PHONE: localnet_relayminer1_ping
+localnet_relayminer1_ping:
+	@echo "Pinging relayminer 1..."
+	@curl -X GET localhost:7001 || (echo "Failed to ping relayminer1. Make sure your localnet environment or your relayminer pod is up and running"; exit 1)
+	@echo "OK"
+
+.PHONE: localnet_relayminer2_ping
+localnet_relayminer2_ping:
+	@echo "Pinging relayminer 2..."
+	@curl -X GET localhost:7002 || (echo "Failed to ping relayminer2. Make sure your localnet environment or your relayminer pod is up and running"; exit 1)
+	@echo "OK"
+
+.PHONE: localnet_relayminer3_ping
+localnet_relayminer3_ping:
+	@echo "Pinging relayminer 3..."
+	@curl -X GET localhost:7003 || (echo "Failed to ping relayminer3. Make sure your localnet environment or your relayminer pod is up and running"; exit 1)
+	@echo "OK"
+
+.PHONY: send_relay_sovereign_app_JSONRPC
+send_relay_sovereign_app_JSONRPC: # Send a JSONRPC relay through the AppGateServer as a sovereign application
+	curl -X POST -H "Content-Type: application/json" \
+	--data '{"jsonrpc":"2.0","method":"eth_blockNumber","params":[],"id":1}' \
+	$(APPGATE_SERVER)/anvil
+
+.PHONY: send_relay_delegating_app_JSONRPC
+send_relay_delegating_app_JSONRPC: # Send a relay through the gateway as an application that's delegating to this gateway
+	@appAddr=$$(poktrolld keys show app1 -a) && \
+	curl -X POST -H "Content-Type: application/json" \
+	--data '{"jsonrpc":"2.0","method":"eth_blockNumber","params":[],"id":1}' \
+	$(GATEWAY_URL)/anvil?applicationAddr=$$appAddr
+
+.PHONY: send_relay_sovereign_app_REST
+send_relay_sovereign_app_REST: # Send a REST relay through the AppGateServer as a sovereign application
+	curl -X POST -H "Content-Type: application/json" \
+	--data '{"model": "qwen:0.5b", "stream": false, "messages": [{"role": "user", "content":"count from 1 to 10"}]}' \
+	$(APPGATE_SERVER)/ollama/api/chat
+
+.PHONY: cosmovisor_start_node
+cosmovisor_start_node: # Starts the node using cosmovisor that waits for an upgrade plan
+	bash tools/scripts/upgrades/cosmovisor-start-node.sh
+
 ###############
 ### Linting ###
 ###############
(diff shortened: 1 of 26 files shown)


pkg/relayer/cmd/cmd.go
```diff
@@ -134,6 +135,17 @@ func runRelayer(cmd *cobra.Command, _ []string) error {
 		}
 	}
 
+	if relayMinerConfig.Ping.Enabled {
```
Olshansk, line 138: 👌

pkg/relayer/config/types.go
```diff
@@ -83,6 +89,12 @@ type RelayMinerConfig struct {
 	Pprof                  *RelayMinerPprofConfig
 	Servers                map[string]*RelayMinerServerConfig
 	SmtStorePath           string
+	Ping                   *RelayMinerPingConfig
+}
+
+type RelayMinerPingConfig struct {
```
Olshansk, line 95: NIT: Please add godoc

pkg/relayer/proxy/proxy.go
```diff
@@ -139,14 +139,18 @@ func (rp *relayerProxy) Start(ctx context.Context) error {
...
-
 	// Start the ring cache.
 	rp.ringCache.Start(ctx)
 
 	startGroup, ctx := errgroup.WithContext(ctx)
 
 	for _, relayServer := range rp.servers {
 		server := relayServer // create a new variable scoped to the anonymous function
+
+		if err := server.Ping(ctx); err != nil {
```
Olshansk, line 150:
```suggestion
		// Ensure that each of the backing data notes respond to a ping request (at the least)
		// before continuing operation.
		if err := server.Ping(ctx); err != nil {
```

pkg/relayer/proxy/proxy.go
```diff
@@ -187,3 +191,25 @@ func (rp *relayerProxy) validateConfig() error {
 
 	return nil
 }
+
+// Ping tests the connectivity between all the managed relay servers and their respective backend URLs.
+func (rp *relayerProxy) Ping(ctx context.Context) []error {
```
Olshansk, line 196:
```suggestion
// PingAll tests the connectivity between all the managed relay servers and their respective backend URLs.
func (rp *relayerProxy) PingAll(ctx context.Context) []error {
```

pkg/relayer/proxy/proxy.go
```diff
@@ -187,3 +191,25 @@ func (rp *relayerProxy) validateConfig() error {
...
+	var i int
+	for _, srv := range rp.servers {
+		if err := srv.Ping(ctx); err != nil {
+			rp.logger.Error().Err(err).
+				Msg("an unexpected error occured while pinging backend URL")
+			errs = append(errs, err)
+		}
+
+		i++
+	}
```
Olshansk, line 208:
```suggestion
	for i, srv := range rp.servers {
		if err := srv.Ping(ctx); err != nil {
			rp.logger.Error().Err(err).
				Msg("an unexpected error occured while pinging backend URL")
			errs = append(errs, err)
		}
	}
```

pkg/relayer/relayminer.go
```diff
@@ -134,3 +134,22 @@ func (rel *relayMiner) ServePprof(ctx context.Context, addr string) error {
 
 	return nil
 }
+
+func (rel *relayMiner) ServePing(ctx context.Context, ln net.Listener) error {
```
Olshansk, line 138: NIT: Please add godoc

pkg/relayer/relayminer.go
```diff
@@ -134,3 +134,22 @@ func (rel *relayMiner) ServePprof(ctx context.Context, addr string) error {
...
+	go func() {
```
Olshansk, line 139:
```suggestion
	// Start a long-lived goroutine that starts an HTTP server responding to ping request.
	// A single ping request on the relay server broadcasts a ping to all backing services / data nodes.
	go func() {
```

testutil/testrelayer/proxy.go
```diff
@@ -33,5 +33,36 @@ func NewMockOneTimeRelayerProxy(
...
+	relayerProxyMock.EXPECT().
+		ServedRelays().
+		Return(returnedRelaysObs).
+		Times(1)
+	relayerProxyMock.EXPECT().
+		Ping(gomock.Eq(ctx)).
+		Times(1)
+
 	return relayerProxyMock
 }
```
Olshansk, line 68: Two requests:

- [ ] Can you make sure to add documentation of this configuration in `poktrolld/config/relayminer_config.yaml`
- [ ] Can you add a target in our `Makefile` that runs a `ping` on a `LocalNet`?

pkg/relayer/proxy/synchronous.go
```diff
@@ -92,6 +93,26 @@ func (sync *synchronousRPCServer) Stop(ctx context.Context) error {
...
+		resp, err := c.Head(supplierCfg.ServiceConfig.BackendUrl.String())
+		if err != nil {
+			return err
+		}
+		_ = resp.Body.Close()
+
+		if resp.StatusCode >= http.StatusInternalServerError {
+			return errors.New("ping failed")
+		}
+
```
Olshansk, line 110:
```suggestion
```

pkg/relayer/relayminer_test.go
```diff
@@ -57,3 +60,69 @@ func TestRelayMiner_StartAndStop(t *testing.T) {
 	err = relayminer.Stop(ctx)
 	require.NoError(t, err)
 }
+
+func TestRelayMiner_Ping(t *testing.T) {
+	srObs, _ := channel.NewObservable[*servicetypes.Relay]()
```
Olshansk, line 65: Great job leveraging this!

Olshansk (changes requested):
@eddyzags For a first contribution, this is super impressive!! 💪 

Left some comments but great job otherwise. 🙌 

Let's finish it off and lmk if you're interested in contributing more as we get closer to BETA.

pkg/relayer/proxy/proxy.go
eddyzags: 100%

pkg/relayer/proxy/proxy.go
eddyzags: Nice catch. I forgot to do some cleanup here. 🙏🏾

testutil/testrelayer/proxy.go
eddyzags: Yes, it makes sense to me. I am on it.

eddyzags:
Hey @Olshansk  👋🏾 ! Thanks for taking the time to review the code and the feedback.

> Let's finish it off

Yes let's tackle that. I will make changes to the latest comments by Monday at the latest.

I am also working on the forward feature ([see issue comment](https://github.com/pokt-network/poktroll/issues/447#issuecomment-2217352174)). I'll open a PR as soon as I can.


> and lmk if you're interested in contributing more as we get closer to BETA.

And yes, I am definitely interested in contributing more 🚀

Olshansk:
@eddyzags Sounds great! Please re-request a review when you're ready.

Olshansk:
@eddyzags Just FYI: https://github.com/pokt-network/poktroll/pull/744#issuecomment-2326902768

Olshansk:
@eddyzags Another bump

eddyzags:
Hey @Olshansk 

I've just added the last requested modifications:
* Document Localnet configuration file
* Add Localnet helpers to Makefile to ping Relayer Miners

As the Ping safeguard implementation won't start the Relay Miner pod if the suppliers aren't reachable, I had to clean the Relay Miner's Helm value files when starting the Localnet environment based on the constraints below:
  
1. If `ollama.enabled=true` & `rest.enabled=false` in `poktroll/localnet_config.yaml`
   Use a Relayer Miner configuration that defines an Anvil & Ollama supplier as a value file for the `relayerminer` Helm Chart.
2. If `ollama.enabled=false` & `rest.enabled=true` in `poktroll/localnet_config.yaml`
   Use a Relay Miner configuration that defines an Anvil & Rest supplier as a value file for the `relayeminer` Helm Chart.
3. if `ollama.enabled=true` & `rest.enabled=true` in `poktroll/localnet_config.yaml`
   Use a Relay Miner configuration that defines an Anvil, Rest, & Ollama supplier as a value file for the Helm Chart

In other words, this new logic forces us to reference the available suppliers in the Relay Miner's configuration.
After some tests, the Relay Miner pod starts correctly only if I activate the deployment of an Ollama and Rest supplier (`ollama.enabled=true` & `rest.enabled=true` in `localnet_config.yaml`). But for the other scenarios, the `relayminer` pod remains in `CrashLoopBackoff` state.
The Relayer Miner pod isn't failing because of the Ping safeguard, but because the Relay Miner isn't managing the associated service endpoint ([see error instruction here](https://github.com/pokt-network/poktroll/blob/fbac82f151f8e9ecf4bd5ddb0590191d934137de/pkg/relayer/proxy/server_builder.go#L88-L92)).

Error message:
```
failed to start relay miner: service endpoint http://relayminer1:8545 not handled by the relay miner: service endpoint not handled by relayer proxy
```
The Relay miner configuration seems to expects all the suppliers list (Ollama, Anvi & Rest) even if the supplier's container isn't deployed. If this is true, the Relay Miner won't be able to start with the Ping safeguard if one supplier isn't deployed.
I would appreciate some guidance here 🙏🏾 . Is there a link with the fact that we define the supplier list in the `genesis.json`?

Olshansk:
@eddyzags Appreciate the detailed feedback. 

Re review - I will take a look next week but hopefully @bryanchriswhite can help out as well.

> failed to start relay miner: service endpoint http://relayminer1:8545 not handled by the relay miner: service endpoint not handled by relayer proxy

Re this error. I'll take a deeper look next week but in the meantime, can you try:
1. Removing `localnet_config.yaml`
2. Merging with main
3. Re-running localnet

This might be something historical we've already solved

Makefile
```diff
@@ -81,39 +81,10 @@ endif
 .PHONY: install_ci_deps
 install_ci_deps: ## Installs `mockgen` and other go tools
 	go install "github.com/golang/mock/mockgen@v1.6.0" && mockgen --version
-	go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.60.3 && golangci-lint --version
+	go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.59.1 && golangci-lint --version
 	go install golang.org/x/tools/cmd/goimports@latest
 	go install github.com/mikefarah/yq/v4@latest
 
-.PHONY: install_cosmovisor
```
Olshansk, line 88: Why was this removed?

Olshansk (changes requested):
@eddyzags Please merge with `main` and will re-review then.

pkg/relayer/relayminer_test.go
```diff
@@ -57,3 +60,69 @@ func TestRelayMiner_StartAndStop(t *testing.T) {
...
+	servedRelaysObs := relayer.RelaysObservable(srObs)
+
+	mrObs, _ := channel.NewObservable[*relayer.MinedRelay]()
+	minedRelaysObs := relayer.MinedRelaysObservable(mrObs)
```
bryanchriswhite, line 69:
```suggestion
	// servedRelaysObs is NEVER published to. It exists to satisfy test mocks.
	srObs, _ := channel.NewObservable[*servicetypes.Relay]()
	servedRelaysObs := relayer.RelaysObservable(srObs)

	// minedRelaysObs is NEVER published to. It exists to satisfy test mocks.
	mrObs, _ := channel.NewObservable[*relayer.MinedRelay]()
	minedRelaysObs := relayer.MinedRelaysObservable(mrObs)
```

pkg/relayer/relayminer_test.go
```diff
@@ -57,3 +60,69 @@ func TestRelayMiner_StartAndStop(t *testing.T) {
...
+	err = relayminer.Start(ctx)
+	require.NoError(t, err)
+
+	time.Sleep(time.Millisecond)
+
+	filename := "/tmp/relayerminer.ping.sock"
+
+	ln, err := net.Listen("unix", filename)
+	require.NoError(t, err)
+	defer os.Remove(filename)
```
bryanchriswhite, line 107: Neat - love it! :star_struck:

pkg/relayer/relayminer_test.go
```diff
@@ -57,3 +60,69 @@ func TestRelayMiner_StartAndStop(t *testing.T) {
...
+
+	err = relayminer.ServePing(ctx, ln)
+	require.NoError(t, err)
+
+	time.Sleep(time.Millisecond)
+
+	c := http.Client{
```
bryanchriswhite, line 114: Same as above:

```suggestion
	httpClient := http.Client{
```

pkg/relayer/relayminer_test.go
```diff
@@ -57,3 +60,69 @@ func TestRelayMiner_StartAndStop(t *testing.T) {
...
+		Transport: &http.Transport{
+			DialContext: func(ctx context.Context, network, addr string) (net.Conn, error) {
+				return net.Dial(ln.Addr().Network(), ln.Addr().String())
+			},
+		},
```
bryanchriswhite, line 119: 1. Can we please extract the `*http.Transport` value (i.e. keep the `&` here) to a `var`?
2. Are we sure that it's necessary to provide a `DialContext` func? If so, please add a comment explaining.

pkg/relayer/relayminer_test.go
```diff
@@ -57,3 +60,69 @@ func TestRelayMiner_StartAndStop(t *testing.T) {
...
+	}
+	require.NoError(t, err)
+
+	_, err = c.Get("http://unix")
+	require.NoError(t, err)
```
bryanchriswhite, line 124: #PUC (please update comment). It's not clear to me what's going on here and what this assertion implies.

I was expecting to see an assertion regarding the contents of the socket file or something to ensure that the http request made it through.

testutil/testrelayer/proxy.go
```diff
@@ -33,5 +33,36 @@ func NewMockOneTimeRelayerProxy(
...
+	relayerProxyMock.EXPECT().
+		Start(gomock.Eq(ctx)).
+		Times(1)
+	relayerProxyMock.EXPECT().
+		Stop(gomock.Eq(ctx)).
+		Times(1)
+	relayerProxyMock.EXPECT().
+		ServedRelays().
+		Return(returnedRelaysObs).
+		Times(1)
```
bryanchriswhite, line 62: I think this can this be simplified to:

```suggestion
	relayerProxyMock := NewMockOneTimeRelayerProxy(ctx, t, returnedRelaysObs relayer.RelaysObservable)
```

bryanchriswhite (changes requested):
Super cool feature @eddyzags :sunglasses: - thanks for the contribution! :rocket: :raised_hands: 

(And apologies for the review delay :sweat_smile:)

pkg/relayer/proxy/proxy.go
```diff
@@ -187,3 +193,22 @@ func (rp *relayerProxy) validateConfig() error {
...
+				Msg("an unexpected error occured while pinging backend URL")
+			errs = append(errs, err)
+		}
+	}
+
+	if len(errs) > 0 {
+		return errs
+	}
+
+	return nil
```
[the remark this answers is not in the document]
bryanchriswhite: I'm also wondering if we really need to log the error in addition to returning it. Could you please double-check whether and where the returned error is printed? If, for example, the supplier sees the ping error when starting the relayminer binary, then I don't think we need to log the error.

Olshansk:
@eddyzags Do you still plan on finishing off this work?

Makefile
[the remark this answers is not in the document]
eddyzags: No, the changes were not intentional. The only added modifications are the additional Makefile rules to ping the relay miner 1, 2 and 3. Let me fix this

Makefile
eddyzags: Not intentional. Let me fix this

Tiltfile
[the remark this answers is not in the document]
eddyzags: Ditto

docusaurus/docs/operate/configs/relayminer_config.md
```diff
@@ -173,6 +174,21 @@ pprof:
 
 You can learn how to use that endpoint on the [Performance Troubleshooting](../../develop/developer_guide/performance_troubleshooting.md) page.
 
+### `ping`
+
+Configures a `ping` server to test the connectivity of every backend URLs. If
+all the backend URLs are reachable, the endpoint returns a 200 HTTP
+Code. Otherwise, if one or more backend URLs aren't reachable, the service
+returns an 500 HTTP Internal server error.
```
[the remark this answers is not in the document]
eddyzags: Indeed, I think those status codes are more appropriate.
I am on it 👍🏾

localnet/kubernetes/values-relayminer-1-all.yaml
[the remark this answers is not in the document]
eddyzags: Yes, these changes were intentional and required for this PR.

In the main branch, you can see that Tilt is [setting the Kubernetes Values for the Relayminer's configuration (1, 2 & 3)](https://github.com/pokt-network/poktroll/blob/e2df9fdae03e9e5356114df0f1e766cfbdd8efdb/Tiltfile#L259) where all the suppliers are defined ([see here for Relayminer 1](https://github.com/pokt-network/poktroll/blob/main/localnet/kubernetes/values-relayminer-1.yaml)) whether the `localnet_config_default` ([see here](https://github.com/pokt-network/poktroll/blob/e2df9fdae03e9e5356114df0f1e766cfbdd8efdb/Tiltfile#L28-L82)) actually deploys the Kubernetes pod or not. This is a problem because these PR modifications prevent the Relayminer application from starting if one or more Backend URLs aren't reachable.

I needed to find a way to:

```
Pseudocode:

* If Title `localnet_config` enables Ollama, rest, and anvil deployment
    * Use a Relayminer configuration where Ollama, Rest, and Anvil are defined as suppliers.
* If Title `localnet_config` enables Ollama, Rest. But disabled Anvil
    * Use a Relayminer configuration where Ollama and Rest are defined as suppliers. But not Anvil
* If Title `localnet_config` enables Ollama. But disabled Rest and Anvil
    * Use a Relayminer configuration where Ollama supplier is defined as a supplier. But not Rest, Anvil.
, etc...
```

Maybe there is some better way to define the according suppliers based on the `localnet_config dynamically`. I am open to suggestions.

localnet/kubernetes/values-relayminer-common.yaml
```diff
@@ -11,3 +11,6 @@ config:
   pprof:
     enabled: true
     addr: localhost:6060
+  ping:
+    enabled: true
+    addr: localhost:8081
```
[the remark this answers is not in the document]
eddyzags: > Do we need to add something to the deployment and/or service, and port forwards in tilt to expose this locally?

Yes, we must add port forwardings to the Relayminer Kubernetes resources. This is done [here](https://github.com/eddyzags/poktroll/blob/d3bb424b55e6109e75b900dde2994d278a0a0266/Tiltfile#L272)

> How is this file used?

This file (`localnet/kubernetes/values-relayminer-common.yaml`) is used by Tilt as a common configuration for the Relayminer 1, 2 & 3.

Every time we start a local net, Tilt is responsible for defining the common relay miner configuration for us. Therefore, we are enabling the ping feature for all Relayminers

pkg/relayer/cmd/cmd.go
```diff
@@ -134,6 +135,17 @@ func runRelayer(cmd *cobra.Command, _ []string) error {
...
+		ln, err := net.Listen("tcp", relayMinerConfig.Ping.Addr)
```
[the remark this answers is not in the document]
eddyzags: If the `Ping.Addr` isn't formatted properly, the `net` package will return a detailed message error for us. ([see here](https://github.com/golang/go/blob/5123f38e050c5ee7130d459ea247d998a838b5a1/src/net/ipsock.go#L156-L218))

pkg/relayer/cmd/cmd.go
```diff
@@ -134,6 +135,17 @@ func runRelayer(cmd *cobra.Command, _ []string) error {
...
+		if err != nil {
+			return fmt.Errorf("failed to listen ping server: %w", err)
+		}
+
+		if err := relayMiner.ServePing(ctx, ln); err != nil {
+			return fmt.Errorf("failed to start ping server: %w", err)
+		}
```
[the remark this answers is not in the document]
eddyzags: I'm sorry, but I'm not sure I understand. Do you want me to extract the `net.Listen(...)` and `relayMiner.ServePing(...)` instructions into another function, such as `func setupPingServer(addr string) error,` in the same package? If yes, I would love to know why as well. Two instructions don't seem too verbose.

pkg/relayer/proxy/proxy.go
```diff
@@ -139,14 +139,20 @@ func (rp *relayerProxy) Start(ctx context.Context) error {
 	if err := rp.BuildProvidedServices(ctx); err != nil {
 		return err
 	}
-
```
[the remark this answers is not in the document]
eddyzags: Oops! Agree. Will do

pkg/relayer/proxy/proxy.go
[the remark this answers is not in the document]
eddyzags: > Could you please double-check whether and where the returned error is printed?

The log error is only printed here.

> If, for example, the supplier sees the ping error when starting the relayminer binary, then I don't think we need to log the error.

I'm sorry, but I'm not sure I understand what you mean by "if the supplier sees the ping error." Do you mean if the supplier logs an error, we don't need to log in the relayminer program?

Keep in mind that there are two separate scenarios/processes here:
1. Safeguard at startup
    If one or more backend URLs aren't reachable, we stop the relayminer application and log an error message.
2. Synchrounous HTTP call to ping all backend URLs (suppliers)
    If one or more backend URLs aren't reachable, we return 500 Internal service errors and log an error message.
 
The second scenario allows the relayminer users to check whether the connectivity between the supplier and the relay miner is still established at any given time. So I think we must log an error log in both cases in my opinion (if something goes wrong)

pkg/relayer/proxy/synchronous.go
```diff
@@ -91,6 +92,25 @@ func (sync *synchronousRPCServer) Stop(ctx context.Context) error {
...
+		c := &http.Client{Timeout: 2 * time.Second}
+
+		resp, err := c.Head(supplierCfg.ServiceConfig.BackendUrl.String())
+		if err != nil {
+			return err
+		}
+		_ = resp.Body.Close()
+
+		if resp.StatusCode >= http.StatusInternalServerError {
+			return errors.New("ping failed")
```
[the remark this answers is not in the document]
eddyzags: Agree with that. On it 👍🏾

pkg/relayer/relayminer.go
```diff
@@ -134,3 +134,28 @@ func (rel *relayMiner) ServePprof(ctx context.Context, addr string) error {
...
+// backend URLs).
+func (rel *relayMiner) ServePing(ctx context.Context, ln net.Listener) error {
+	// Start a long-lived goroutine that starts an HTTP server responding to
+	// ping requests. A single ping request on the relay server broadcasts a
+	// ping to all backing services/data nodes.
+	go func() {
+		if err := http.Serve(ln, http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
+			rel.logger.Debug().Msg("pinging relay servers...")
+
+			if errs := rel.relayerProxy.PingAll(ctx); errs != nil {
```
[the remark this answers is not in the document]
eddyzags: Yes, this is a good point. Thanks for the suggestion 👍🏾

pkg/relayer/relayminer_test.go
eddyzags: > Are we sure that it's necessary to provide a DialContext func? If so, please add a comment explaining.

Yes, this is required to instantiated an unencrypted TCP connection.

pkg/relayer/relayminer_test.go
eddyzags: Indeed, nice catch.
I'll add the assertion based on the HTTP status code. The ping endpoint must return an HTTP 204 status code; otherwise, the test will fail.

docusaurus/docs/operate/configs/relayminer_config.md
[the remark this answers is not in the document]
eddyzags: I'm not so sure about `503 Service Unavailable` in the end. The Relayminer server was ready to handle the request, but subsequent calls are not possible. Therefore, there is an internal error in my opinion.

Tiltfile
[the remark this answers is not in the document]
eddyzags: see here https://github.com/pokt-network/poktroll/pull/744#discussion_r1835350571

docusaurus/docs/operate/configs/relayminer_config.md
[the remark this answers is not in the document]
bryanchriswhite: Could you elaborate  on "The Relayminer server was ready to handle the request, subsequent calls are not possible" I'm not sure I follow?

Do you think [504 Gateway Timeout](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/504) is more appropriate?

My thinking is that 500 is quite generic and we should consider using a more specific status if one applies.

pkg/relayer/proxy/synchronous.go
```diff
@@ -91,6 +92,25 @@ func (sync *synchronousRPCServer) Stop(ctx context.Context) error {
...
 }
 
+// Ping tries to dial the suppliers backend URLs to test the connection.
+func (sync *synchronousRPCServer) Ping(ctx context.Context) error {
+	for _, supplierCfg := range sync.serverConfig.SupplierConfigsMap {
+		c := &http.Client{Timeout: 2 * time.Second}
+
+		resp, err := c.Head(supplierCfg.ServiceConfig.BackendUrl.String())
+		if err != nil {
+			return err
```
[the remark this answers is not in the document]
bryanchriswhite:
```suggestion
			return ErrRelayerProxySupplierNotReachable.Wrapf(
				"endpoint URL %q err: %s",
				endpointUrl, err,
			)
				
```

pkg/relayer/proxy/synchronous.go
```diff
@@ -91,6 +91,25 @@ func (sync *synchronousRPCServer) Stop(ctx context.Context) error {
 	return sync.server.Shutdown(ctx)
 }
 
+// Ping tries to dial the suppliers backend URLs to test the connection.
+func (sync *synchronousRPCServer) Ping(ctx context.Context) error {
+	for _, supplierCfg := range sync.serverConfig.SupplierConfigsMap {
+		httpClient := &http.Client{Timeout: 2 * time.Second}
```
bryanchriswhite, line 97:
```suggestion
		httpClient := &http.Client{Timeout: 2 * time.Second}
		endpointUrl := supplierCfg.ServiceConfig.BackendUrl.String()
```

pkg/relayer/config/types.go
```diff
@@ -24,6 +24,14 @@ type YAMLRelayMinerConfig struct {
 	Pprof                  YAMLRelayMinerPprofConfig      `yaml:"pprof"`
 	SmtStorePath           string                         `yaml:"smt_store_path"`
 	Suppliers              []YAMLRelayMinerSupplierConfig `yaml:"suppliers"`
+	Ping                   YAMLRelayMinerPingConfig       `yaml:"ping"`
+}
+
+// YAMLRelayMinerPingConfig represents the configuration to expose a ping server.
+type YAMLRelayMinerPingConfig struct {
+	Enabled bool `yaml:"enabled"`
+	// Addr is the address to bind to (format: 'hostname:port')
```
bryanchriswhite, line 33:
```suggestion
	// Addr is the address to bind to (format: 'hostname:port' where 'hostname' can be a DNS name or an IP)
```

pkg/relayer/proxy/proxy.go
```diff
@@ -187,3 +195,22 @@ func (rp *relayerProxy) validateConfig() error {
...
+
+	for _, srv := range rp.servers {
+		if err := srv.Ping(ctx); err != nil {
+			err = errors.Join(err, err)
+		}
+	}
+
+	if err != nil {
+		rp.logger.Error().Err(err).
+			Msg("an unexpected error occured while pinging backend URL")
```
bryanchriswhite, line 211: I don't think an error message is necessary here, especially if we add more detail elsewhere as suggested.
```suggestion
		rp.logger.Error().Err(err).Send()
```

pkg/relayer/relayminer_test.go
```diff
@@ -57,3 +60,72 @@ func TestRelayMiner_StartAndStop(t *testing.T) {
 	err = relayminer.Stop(ctx)
 	require.NoError(t, err)
 }
+
+func TestRelayMiner_Ping(t *testing.T) {
```
bryanchriswhite, line 64: Is this all the test coverage that there is or did I miss something?

I would love to see coverage over a few more scenarios:

1. Multi-supplier relayminer success.
2. Relayminer with multiple endpoint, including a DNS name, ipv4 address and ipv6 address success
    Mocking the DNS name is pretty straightforward with the [go-mockdns](https://github.com/foxcpp/go-mockdns) package.
3. Multi-endpoint partial success/failure on startup (i.e. some endpoints should be unreachable on startup)
4. Multi-endpoint partial success/failure post-startup (i.e. some endpoints should become unreachable after startup)

Consider refactoring this test code into a [test suite](https://github.com/stretchr/testify?tab=readme-ov-file#suite-package) to de-duplicate common code by moving it into methods on the suite and storing case-specific test state on the suite struct (per test case).

pkg/relayer/relayminer_test.go
bryanchriswhite: Digging in: when I uncomment the http client transport configuration I get the following error:

```
    relayminer_test.go:125: 
        	Error Trace:	/home/bwhite/Projects/pokt/poktroll/pkg/relayer/relayminer_test.go:125
        	Error:      	Received unexpected error:
        	            	Get "http://unix": dial tcp: lookup unix: no such host
        	Test:       	TestRelayMiner_Ping
```

Which makes sense given that our listener is a unix socket. :sweat_smile: 

```suggestion
		// Override transport configuration to adapt the http client to the unix socker listener.
		Transport: &http.Transport{
			DialContext: func(ctx context.Context, network, addr string) (net.Conn, error) {
				return net.Dial(ln.Addr().Network(), ln.Addr().String())
			},
		},
```

bryanchriswhite (changes requested):

Tiltfile
```diff
@@ -250,21 +250,32 @@ helm_resource(
...
             "--values=./localnet/kubernetes/values-relayminer-common.yaml",
             "--values=./localnet/kubernetes/values-relayminer-" + str(actor_number) + ".yaml",
             "--set=metrics.serviceMonitor.enabled=" + str(localnet_config["observability"]["enabled"]),
             "--set=development.delve.enabled=" + str(localnet_config["relayminers"]["delve"]["enabled"]),
             "--set=logLevel=" + str(localnet_config["relayminers"]["logs"]["level"]),
             "--set=image.repository=poktrolld",
-        ],
+    ]
+
+    if localnet_config["rest"]["enabled"]:
```
okdas, line 264: Why do we need multiple configuration files, and not just one?

docusaurus/docs/operate/configs/relayminer_config.md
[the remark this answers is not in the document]
eddyzags: > My thinking is that 500 is quite generic and we should consider using a more specific status if one applies.

I agree. 500 can be generic, and we must use a more specific status code. I also think we shouldn't use 500 because that implies that the server (in this case, `relayminer`) is completely unusable. But that isn't entirely the case. AFAIK if one supplier isn't reachable anymore, the other configured ones can still function, and the relayminer can continue to mine requests for them.

> Could you elaborate on "The Relayminer server was ready to handle the request, subsequent calls are not possible" I'm not sure I follow?

In my mind, there are three actors here:
1. The client to the `/ping` endpoint. (could be `curl -x GET <addr>:<port>/ping`)
2. The server (`relayminer`)
3. The third-party service (`supplier`)

What I meant by _"The Relayminer server was ready/able to handle requests"_ is the interaction between 1 & 2. And what I meant by the _"subsequent calls are not possible"_ is the interaction between 2 & 3. I hope that makes sense.

My problem with `503 Service Unavailable` is that it implies a temporary state. But we are not so sure about that. The supplier might never come back online.
So yes, I think `504 Gateway Timeout` is more appropriate if the context timeout. But if the the third-party (`supplier`) return an unexpected response, we must use `503 Bad Gateway` IMO. wdyt?

pkg/relayer/cmd/cmd.go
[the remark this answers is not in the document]
eddyzags: Thanks for your explanation Bryan, I understand now.

IIRC, I went this way because it was more practical for me to test the `ServePing(...)` when accepting a `net.Listener` interface. The implementation of `net.Listener` could be anything ("TCP" connection on the production code, "Unix domain" connection on the unit test).
But I can manage something to stay consistent and follow that single responsibility principle, yes.

pkg/relayer/cmd/cmd.go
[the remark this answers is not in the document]
eddyzags: Exactly 👍🏾

Olshansk:
@eddyzags Our team is on holiday next week. If you'd like to get this in, please finish it by EOD tomorrow so we have time to review and merge it in by EOW.

Olshansk:
Going to close this out due to lack of response from @eddyzags but we can always reference the branch / PR when/if we pick up this effort again in the future.

docusaurus/docs/operate/configs/relayminer_config.md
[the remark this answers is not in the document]
eddyzags: Sorry, I don't know if my message was clear. But basically, 502 Gateway HTTP response occurs when a server acting as a gateway or proxy receives an invalid response from an upstream server.

pkg/relayer/proxy/proxy.go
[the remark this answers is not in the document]
eddyzags: We are now relying on `errors.Join` 👍🏾 


> I'm also wondering if we really need to log the error in addition to returning it. Could you please double-check whether and where the returned error is printed?

To clarify a bit, if an error occurs during scenario 1 (Ex: The `supplier[].backend_url` in the relayminer's configuration is not reachable at startup), an error is logged here https://github.com/pokt-network/poktroll/blob/7e059965c860cc85bd48027f0dde31be92dcf3fc/cmd/poktrolld/main.go#L16 


Then, if an error occurs during scenario 2 (Ex: The `supplier[].backend_url` in the relayminer's configuration is not reachable at runtime, and the relayminer is receiving a request to /ping endpoint), an error is logged here https://github.com/eddyzags/poktroll/blob/6129d78b58df0141bc5cb69fd4c3ab0db1932313/pkg/relayer/proxy/proxy.go#L229

In my opinion, both logs are relevant because they are two distinct workflows. wdyt?

pkg/relayer/config/types.go
eddyzags: https://github.com/eddyzags/poktroll/commit/d6165d9497cc1efacb0c6377fb81daef2156edf3

pkg/relayer/proxy/proxy.go
eddyzags: see comment here https://github.com/pokt-network/poktroll/pull/744#discussion_r1913831274

Furthermore, your suggested change will also log something in the output, right? I am not familiar with the zerolog package, but it seems that `Send()` is doing the same as `Msg("")` without printing a `message` field. ([link to documentation](https://pkg.go.dev/github.com/rs/zerolog#readme-contextual-logging))

Olshansk:
@eddyzags Given your recent followup on some of the comments, are you planning to pick this up again?

Tiltfile
eddyzags: Upon revisiting my initial approach, I realized it could be simplified by using the `--set` argument in Helm. Before diving into the solution, let me explain the reasoning behind this change.

The issue stemmed from the fact that the relayminer now ensures the backend URLs of the suppliers are reachable, so it became necessary to align the `poktroll/localnet_config.yaml` file with the relayminer configuration.

```
if the ollama containers is disabled on the `localnet_config.yaml`
    we MUST NOT specify any ollama suppliers in the relayminer's configuration
if the ollama containers is enabled on the `localnet_config.yaml`
    we MUST specify the the ollama suppliers in the relayminer's configuration
```
( This example above is also applicable to `anvil` and `rest` suppliers).
If this is not done, the relayminer will keep failing while launching the Localnet environment saying that one or more suppliers are not reachable.

My initialize approach was to multiple configuration, but I think we can rely on a single configuration per relayminer (1, 2 & 3), and dynamically assign the `config.suppliers` items. I've just pushed the modifications in my branch https://github.com/eddyzags/poktroll/blob/7d36e71e560e1dbcb4d7605046cee26063bd855d/Tiltfile#L278-L298.

I would love to have your thoughts on that 🙏🏾

pkg/relayer/relayminer_test.go
eddyzags: Hey @bryanchriswhite ! Thanks for sharing your thoughts on the test coverage for the Ping, and I agree.

I've just pushed all the test cases listed above in my branch. I think it was better to define these tests at the relay proxy level to have more flexibility to define the different relayminers 

https://github.com/eddyzags/poktroll/blob/7d36e71e560e1dbcb4d7605046cee26063bd855d/pkg/relayer/proxy/proxy_test.go#L560-L994

eddyzags:
Hey @Olshansk @bryanchriswhite @okdas ! Happy new year!

> are you planning to pick this up again?

I had time to pick this up for a few days now. And all the comments have been addressed. So, I am waiting for your valuable feedbacks!

In addition, I've taken some time to stabilize the flaky tests inside `poktroll/testproxy/relayproxy.go`. The reason was that sometimes the supplier's simple HTTP server didn't have the time to bind to a port and serve HTTP requests. So, I've separated the `Listen` and `Serve` processes to ensure that the HTTP server listens before creating a go routine, and then serves the HTTP requests inside the go-routine. I let you try on your end.

Should we re-open the PR?

Olshansk:
@eddyzags Thanks for following up!

> Should we re-open the PR?

Can you use the same branch and open up a new PR?

eddyzags:
Follow up available here https://github.com/pokt-network/poktroll/pull/1037 @Olshansk

PR #750, opened 2024-08-21
[Docs] Tokenomics Documentation

Olshansk:
## Summary

First iteration of tokenomics documentation.

[Token Logic Modules _ Poktroll.pdf](https://github.com/user-attachments/files/16697554/Token.Logic.Modules._.Poktroll.pdf)

@bryanchriswhite @red-0ne In addition to review this, please read through and understand it. Hopefully it serves as the foundation for how we tackle [this issue](https://www.notion.so/buildwithgrove/Off-chain-Application-Stake-Tracking-6a8bebb107db4f7f9dc62cbe7ba555f7?pvs=4) but 10x simpler.

## Issue

- #732

## Type of change

Select one or more:

- [ ] New feature, functionality or library
- [ ] Bug fix
- [ ] Code health or cleanup
- [x] Documentation
- [ ] Other (specify)

## Testing

- [x] **Documentation**: `make docusaurus_start`; only needed if you make doc changes

PR #740, opened 2024-08-16
[WIP][Tokenomics] Global Inflation Reimbursement Request

Olshansk:
## Summary

Add the `TLMGlobalMintReimbursementRequest` TLM which:

- Overcharge Applications settling sessions
- Uses the global inflation parameter
- Creates an event with the result of the overcharged applications
- Enables permissionless demand, where PNF is responsible for reimbursing gateways overpaying

## Issue

- #732

## Type of change

Select one or more:

- [x] New feature, functionality or library
- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

- [ ] `make docusaurus_start`; only needed if you make doc changes
- [ ] **Unit Tests**: `make go_develop_and_test`
- [ ] **LocalNet E2E Tests**: `make test_e2e`
- [ ] **DevNet E2E Tests**: Add the `devnet-test-e2e` label to the PR.

## Sanity Checklist

- [ ] I have tested my changes using the available tooling
- [ ] I have commented my code
- [ ] I have performed a self-review of my own code; both comments & source code
- [ ] I create and reference any new tickets, if applicable
- [ ] I have left TODOs throughout the codebase, if applicable

Issue #732, opened 2024-08-09
[Tokenomics] MVP of Inflation, Governance, Reimbursement & Docs


Olshansk:
## Objective

Introduce an inflation-based governance param as a follow-up to: #708

## Origin Document

- #708

## Goals

- Set up Shannon for a MainNet launch
- Allow launching with permissionless demand 
- Do not require the final version of what inflationary permissionless demand logic should look like

## Deliverables

- [ ] Add a new governance param called "Inflation"
- [ ] Inflation should act on a "claim" based basis
- [ ] The inflation MUST go to the DAO
- [ ] How the inflation is distributed between the supplier, gateway, application and service owner should be easily tracked on chain (though the funds still go to the DAO).

## Non-goals / Non-deliverables

- Enable free-market permissionless demand.

## General deliverables

- [ ] **Comments**: Add/update TODOs and comments alongside the source code so it is easier to follow.
- [ ] **Testing**: Add new tests (unit and/or E2E) to the test suite.
- [ ] **Makefile**: Add new targets to the Makefile to make the new functionality easier to use.
- [ ] **Documentation**: Update architectural or development READMEs; use [mermaid](https://mermaid-js.github.io/mermaid/) diagrams where appropriate.

---

**Creator**: @Olshansk

PR #745, opened 2024-08-19
[Relay Mining] Use per-service difficulty in the relayer

adshmh:
## Summary

- Relayer is updated to query the tokenmoics module to get the relay difficulty target hash specific to the service corresponding to the relay. This is used to determine if a relay is volume applicable.

- Relayer is updated to query the service module to get the compute units per relay specific to the service corresponding to the relay. This is used as the weight of the relay when updating the Relayer's SMT for the session corresponding to the relay.

- Claims are validated at the time of submission to ensure the total claimed compute units is equal to the number of relays multiplied by the compute units per relay for the corresponding service.

- All #705 references in the code have been addressed.

- Relay test fixture generator has been updated to use the difficulty threshold hash to be consistent with how the relayer now operates.

## Issue

- Relayer should use a service-specific difficulty target hash.
- Relayer should use the service-specific ComputeUnitsPerRelay when adding a relay to the SMT for claiming.

- #705

## UPNEXT 
- Remove the relay_difficulty_target_hash parameter
- Addressing any remaining instances of `TODO_FOLLOWUP(@olshansk, #690)`

These have been skipped to: 1) align on the PR before making further changes and 2) restrict the PR size and scope, as these were not strictly required for the above enhancements to function.

Depending on the review results, I can add commits to address the above in this PR as well.

## Type of change

Select one or more:

- [x] New feature, functionality or library
- [ ] Bug fix
- [x] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

**Documentation changes** (only if making doc changes)
- [ ] `make docusaurus_start`; only needed if you make doc changes

**Local Testing** (only if making code changes)
- [x] **Unit Tests**: `make go_develop_and_test`
- [ ] **LocalNet E2E Tests**: `make test_e2e`
- See [quickstart guide](https://dev.poktroll.com/developer_guide/quickstart) for instructions

**PR Testing** (only if making code changes)
- [ ] **DevNet E2E Tests**: Add the `devnet-test-e2e` label to the PR.
    - **THIS IS VERY EXPENSIVE**, so only do it after all the reviews are complete.
    - Optionally run `make trigger_ci` if you want to re-trigger tests without any code changes
    - If tests fail, try re-running failed tests only using the GitHub UI as shown [here](https://github.com/pokt-network/poktroll/assets/1892194/607984e9-0615-4569-9452-4c730190c1d2)

## Sanity Checklist

- [x] I have tested my changes using the available tooling
- [x] I have commented my code
[...]

PR #755, opened 2024-08-22
[Tokenomics] Preparation for Global Mint Reimbursement Request

Olshansk:
## Summary

Implement "max claimable amount" in preparation for "global mint reimbursement request".

Specific changes:
- Added lots of `TODO_BETA` in preparation for beta testnet
- Added a lot more comments and explanation in the TLM processors

## Issue

- #732
- Related to #750
- Superceed #740

## Type of change

Select one or more:

- [ ] Bug fix
- [ ] Code health or cleanup
- [x] Documentation
- [ ] Other (specify)

## Testing

## Sanity Checklist


x/tokenomics/keeper/token_logic_modules.go
```diff
@@ -275,8 +305,8 @@ func (k Keeper) TokenLogicModuleRelayBurnEqualsMint(
 	}
 	logger.Info(fmt.Sprintf("minted (%v) coins in the supplier module", settlementCoin))
 
-	amount := settlementCoin.Amount.Uint64()
-	if err = k.distributeSupplierRewardsToShareHolders(ctx, ownerAddr.String(), service.Id, amount); err != nil {
+	// Distribute the rewards to the supplier's shareholders based on the rev share percentage.
+	if err = k.distributeSupplierRewardsToShareHolders(ctx, ownerAddr.String(), service.Id, settlementCoin.Amount.Uint64()); err != nil {
```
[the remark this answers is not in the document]
bryanchriswhite: I feel like we should have a test which mitigates this regression; i.e., one which asserts that the correct rev share-holder address(es) are used.

x/tokenomics/keeper/token_logic_modules.go
```diff
@@ -319,79 +336,104 @@ func (k Keeper) TokenLogicModuleRelayBurnEqualsMint(
...
+	if newMintCoin.Amount.Int64() == 0 {
+		return tokenomicstypes.ErrTokenomicsMintAmountZero
+	}
 
 	// Mint new uPOKT to the tokenomics module account
-	if err := k.bankKeeper.MintCoins(ctx, tokenomictypes.ModuleName, newMintCoins); err != nil {
+	if err := k.bankKeeper.MintCoins(ctx, tokenomictypes.ModuleName, sdk.NewCoins(newMintCoin)); err != nil {
 		return tokenomicstypes.ErrTokenomicsModuleMintFailed.Wrapf(
-			"minting %s to the tokenomics module account: %v", newMintCoins, err)
+			"minting %s to the tokenomics module account: %v", newMintCoin, err)
```
bryanchriswhite, line 364:
```suggestion
			"minting (%s) to the tokenomics module account: %v", newMintCoin, err)
```

x/tokenomics/keeper/token_logic_modules.go
```diff
@@ -319,79 +336,104 @@ func (k Keeper) TokenLogicModuleRelayBurnEqualsMint(
...
 	}
-	logger.Info(fmt.Sprintf("minted (%v) coins in the tokenomics module", newMintCoins))
+	logger.Info(fmt.Sprintf("minted (%v) coins in the tokenomics module", newMintCoin))
```
bryanchriswhite, line 366:
```suggestion
	logger.Info(fmt.Sprintf("minted (%s) to the tokenomics module account", newMintCoin))
```

x/tokenomics/keeper/token_logic_modules.go
```diff
@@ -319,79 +336,104 @@ func (k Keeper) TokenLogicModuleRelayBurnEqualsMint(
...
 
 	// Send a portion of the rewards to the application
-	appCoin, err := k.sendRewardsToAccount(ctx, application.Address, newMintAmtFloat, MintAllocationApplication)
+	appCoin, err := k.sendRewardsToAccount(ctx, tokenomictypes.ModuleName, application.Address, &newMintAmtFloat, MintAllocationApplication)
```
bryanchriswhite, line 369:
```suggestion
	appCoin, err := k.sendRewardsToAccount(ctx, tokenomictypes.ModuleName, application.GetAddress(), &newMintAmtFloat, MintAllocationApplication)
```

x/tokenomics/keeper/token_logic_modules.go
```diff
@@ -319,79 +336,104 @@ func (k Keeper) TokenLogicModuleRelayBurnEqualsMint(
...
 
 	// Send a portion of the rewards to the supplier shareholders.
-	coinsToShareAmt := calculateGlobalMintAllocationFromSettlementAmount(newMintAmtFloat, MintAllocationSupplier)
-	if err = k.distributeSupplierRewardsToShareHolders(ctx, supplier.OperatorAddress, service.Id, uint64(coinsToShareAmt)); err != nil {
+	supplierCoinsToShareAmt := calculateAllocationAmount(&newMintAmtFloat, MintAllocationSupplier)
+	supplierCoin := cosmostypes.NewCoin(volatile.DenomuPOKT, math.NewInt(supplierCoinsToShareAmt))
+	// Send funds from the tokenomics module to the supplier module account
+	if err = k.bankKeeper.SendCoinsFromModuleToModule(ctx, tokenomicstypes.ModuleName, suppliertypes.ModuleName, sdk.NewCoins(supplierCoin)); err != nil {
+		return tokenomicstypes.ErrTokenomicsSupplierModuleSendFailed.Wrapf(
+			"sending %s from the tokenomics module to the supplier module account: %v",
```
bryanchriswhite, line 381:
```suggestion
			"transferring (%s) from the tokenomics module account to the supplier module account: %v",
```

x/tokenomics/keeper/token_logic_modules.go
```diff
@@ -533,22 +596,34 @@ func (k Keeper) distributeSupplierRewardsToShareHolders(
...
+	// TODO_MAINNET: Consider using fixed point arithmetic for deterministic results.
+	settlementAmtFloat := new(big.Float).SetUint64(settlementCoin.Amount.Uint64())
+	newMintAmtFloat := new(big.Float).Mul(settlementAmtFloat, big.NewFloat(MintPerClaimedTokenGlobalInflation))
+	newMintAmtInt, _ := newMintAmtFloat.Int64()
+	mintAmtCoin := cosmostypes.NewCoin(volatile.DenomuPOKT, math.NewInt(newMintAmtInt))
+	return mintAmtCoin, *newMintAmtFloat
+}
+
+// calculateAllocationAmount does big float arithmetic to determine the absolute
+// amount from amountFloat based on the allocation percentage provided.
```
bryanchriswhite, line 621: Is it worth mentioning and/or quantifying the precision loss here?

x/tokenomics/keeper/token_logic_modules.go
```diff
@@ -319,79 +336,104 @@ func (k Keeper) TokenLogicModuleRelayBurnEqualsMint(
...
+	// Send funds from the tokenomics module to the supplier module account
+	if err = k.bankKeeper.SendCoinsFromModuleToModule(ctx, tokenomicstypes.ModuleName, suppliertypes.ModuleName, sdk.NewCoins(supplierCoin)); err != nil {
+		return tokenomicstypes.ErrTokenomicsSupplierModuleSendFailed.Wrapf(
+			"sending %s from the tokenomics module to the supplier module account: %v",
+			supplierCoin,
+			err,
+		)
+	}
+	// Distribute the rewards from within the supplier's module account.
+	if err = k.distributeSupplierRewardsToShareHolders(ctx, supplier.OperatorAddress, service.Id, uint64(supplierCoinsToShareAmt)); err != nil {
```
bryanchriswhite, line 387:
```suggestion
	if err = k.distributeSupplierRewardsToShareHolders(ctx, supplier, service.Id, uint64(supplierCoinsToShareAmt)); err != nil {
```

x/tokenomics/keeper/token_logic_modules.go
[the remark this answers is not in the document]
bryanchriswhite: See feedback on `k.distributeSupplierRewardsToShareHolders()` regarding modifying the method signature.

```suggestion
	if err = k.distributeSupplierRewardsToShareHolders(ctx, supplier, service.Id, settlementCoin.Amount.Uint64()); err != nil {
```

x/tokenomics/keeper/token_logic_modules.go
```diff
@@ -245,23 +274,24 @@ func (k Keeper) ProcessTokenLogicModules(
...
 		return err
 	}
 
-	// NB: We are doing a mint & burn + transfer, instead of a simple transfer
-	// of funds from the supplier to the application in order to enable second
-	// order economic effects with more optionality. This could include funds
-	// going to pnf, delegators, enabling bonuses/rebates, etc...
-
 	// Mint new uPOKT to the supplier module account.
 	// These funds will be transferred to the supplier below.
```
bryanchriswhite, line 296:
```suggestion
	// Mint new uPOKT to the supplier module account.
	// These funds will be transferred to the supplier and revenue share-holders below.
```

bryanchriswhite: I'm not seeing where the supplier balance is modiefied; i.e., where the supplier is paid. :thinking:

It looks like we're only doing the following:
1. Minting to the supplier module account
2. Distributing newly minted tokens (from the supplier module account) to rev-share-holders

Am I missing something or do we lacking test coverage around this?

x/tokenomics/keeper/token_logic_modules.go
```diff
@@ -130,109 +150,118 @@ func (k Keeper) ProcessTokenLogicModules(
...
+	logger = logger.With("actual_settlement_upokt", actualSettlementCoin)
+
+	logger.Info(fmt.Sprintf("About to start processing TLMs for (%d) relays equaling to (%s) upokt claimed", numRelays, actualSettlementCoin))
 	// Execute all the token logic modules processors
 	for tlm, tlmProcessor := range tokenLogicModuleProcessorMap {
-		logger.Info(fmt.Sprintf("Starting to execute TLM %q", tlm))
-		if err := tlmProcessor(k, ctx, &service, &application, &supplier, settlementCoin, &relayMiningDifficulty); err != nil {
+		logger.Info(fmt.Sprintf("Starting TLM processing: %q", tlm))
+		if err := tlmProcessor(k, ctx, &service, claim.SessionHeader, &application, &supplier, actualSettlementCoin, &relayMiningDifficulty); err != nil {
 			return err
```
bryanchriswhite, line 251: I feel like there should be a sentinel error which wraps the TLM name and error here:

```suggestion
			return ErrTokenomicsTLMError.Wrapf("TLM %q: %v", tlm err)
```

x/tokenomics/keeper/token_logic_modules.go
```diff
@@ -319,79 +336,104 @@ func (k Keeper) TokenLogicModuleRelayBurnEqualsMint(
...
 	logger.Debug(fmt.Sprintf("sent (%v) newley minted coins from the tokenomics module to the proposer with address %q", proposerCoin, proposerAddr))
 
-	// TODO_MAINNET: Verify that the total distributed coins equals the settlement coins which could happen due to float rounding
-	totalDistributedCoins := appCoin.Add(supplierCoin).Add(*daoCoin).Add(*serviceCoin).Add(*proposerCoin)
-	if totalDistributedCoins.Amount.BigInt().Cmp(settlementCoins.Amount.BigInt()) != 0 {
-		logger.Error(fmt.Sprintf("TODO_MAINNET: The total distributed coins (%v) does not equal the settlement coins (%v)", totalDistributedCoins, settlementCoins.Amount.BigInt()))
-	}
-	logger.Info(fmt.Sprintf("distributed (%v) coins to the application, supplier, DAO, source owner, and proposer", totalDistributedCoins))
+	// Check and log the total amount of coins distributed
+	totalMintDistributedCoin := appCoin.Add(supplierCoin).Add(*daoCoin).Add(*serviceCoin).Add(*proposerCoin)
```
bryanchriswhite, line 420: Wdyt about `globalMintDistributedCoin` instead of `totalMintDistributedCoin` to reduce potential confusion as this only pertains to this TLM, whereas other TLMs may also mint tokens?

x/tokenomics/keeper/token_logic_modules.go
```diff
@@ -400,68 +442,88 @@ func (k Keeper) TokenLogicModuleGlobalMint(
...
+// ensureClaimAmountLimits checks if the application was overserviced and handles
+// the case if it was.
+// Per Algorithm #1 in the Relay Mining paper, the maximum amount that a single supplier
+// can claim in a session is AppStake/NumSuppliersPerSession.
+// If this is not the case, then the supplier essentially did "free work" and the
+// actual claim amount is less than what was claimed.
+// Ref: https://arxiv.org/pdf/2305.10672
+func (k Keeper) ensureClaimAmountLimits(
 	ctx context.Context,
+	methodLogger log.Logger,
```
bryanchriswhite, line 478: nit:
```suggestion
	logger log.Logger,
```

x/tokenomics/keeper/token_logic_modules.go
```diff
@@ -400,68 +442,88 @@ func (k Keeper) TokenLogicModuleGlobalMint(
...
-		application.Stake,
-	))
-
-	// TODO_MAINNET(@Olshansk, @RawthiL): The application was over-serviced in the last session so it basically
-	// goes "into debt". Need to design a way to handle this when we implement
-	// probabilistic proofs and add all the parameter logic. Do we touch the application balance?
-	// Do we just let it go into debt? Do we penalize the application? Do we unstake it? Etc...
-	// See this document from @red-0ne and @bryanchriswhite for more context: notion.so/buildwithgrove/Off-chain-Application-Stake-Tracking-6a8bebb107db4f7f9dc62cbe7ba555f7
-	expectedBurn := settlementCoins
+	logger := methodLogger.With("helper", "ensureClaimAmountLimits")
```
bryanchriswhite, line 486: nit:
```suggestion
	logger = logger.With("helper", "ensureClaimAmountLimits")
```

x/tokenomics/keeper/token_logic_modules.go
```diff
@@ -400,68 +442,88 @@ func (k Keeper) TokenLogicModuleGlobalMint(
...
+
+	// TODO_BETA_OR_MAINNET(@red-0ne): The application stake gets reduced with every claim
+	// settlement. Relay miners use the appStake at the beginning of a session to determine
+	// the maximum amount they can claim. We need to somehow access and propagate this
+	// value (via context?) so it is the same for all TLM processors for each claim.
+	// Note that this will also need to incorporate MintPerClaimGlobalInflation because
+	// applications are being overcharged by that amount in the meantime. Whatever the
+	// solution and implementation ends up being, make sure to KISS.
```
bryanchriswhite, line 494: :kiss:

x/tokenomics/keeper/token_logic_modules.go
```diff
@@ -400,68 +442,88 @@ func (k Keeper) TokenLogicModuleGlobalMint(
...
+	appStake := application.GetStake()
+
+	// Determine the max claimable amount for the supplier based on the application's stake
+	// in this session.
+	maxClaimableCoin := sdk.NewCoin(volatile.DenomuPOKT, appStake.Amount.Quo(math.NewInt(sessionkeeper.NumSupplierPerSession)))
```
bryanchriswhite, line 499: How do we ensure that applications/gateways know when they've reached the `maxClaimableCoin` for a given supplier such that they SHOULD/MUST send traffic to another supplier during a given session?

Is this solved and/or do we have an issue for it?

x/tokenomics/keeper/token_logic_modules.go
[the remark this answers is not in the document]
red-0ne: Good catch @bryanchriswhite ! `distributeSupplierRewardsToShareHolders` needs the operator address!

pkg/relayer/session/sessiontree.go
```diff
@@ -65,6 +69,8 @@ type sessionTree struct {
 // NewSessionTree creates a new sessionTree from a Session and a storePrefix. It also takes a function
 // removeFromRelayerSessions that removes the sessionTree from the RelayerSessionsManager.
 // It returns an error if the KVStore fails to be created.
+// TODO_BETA(@red-0ne): When starting a new session, check how many relays the app can handle.
```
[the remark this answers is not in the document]
Olshansk:
```go
// TODO_BETA(@red-0ne): When starting a new session, check what the MaxClaimableAmount
// (in uPOKT) by the Supplier as a function of
// (app_stake, compute_units_per_relay_for_service, global_compute_units_to_token_multiplier).
// Whether or not the RelayMiner stop handling requests when the max is reached should be
// configurable by the operator.
```

proto/poktroll/shared/service.proto
```diff
@@ -13,7 +13,7 @@ message Service {
   // For example, what if we want to request a session for a certain service but with some additional configs that identify it?
   string id = 1; // Unique identifier for the service
 
-  // TODO_MAINNET: Remove this.
+  // TODO_BETA: Either remove this or rename it to alias.
```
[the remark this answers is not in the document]
Olshansk:
```
// TODO_BETA(@bryanchriswhite): Only embed `ServiceId string ` instead of `Service Service` whenever the
// outer message only needs the ID of the service.
```

and

```
	// TODO_BETA(@bryanchriswhite): Either remove this or rename it to alias.
	// Ensure (i.e. double check) that a unique `Id` is enforced but remove the requirement
	// of a unique alias because that is controlled byt the service owner address.
```

x/tokenomics/keeper/token_logic_modules.go
```diff
@@ -319,79 +336,104 @@ func (k Keeper) TokenLogicModuleRelayBurnEqualsMint(
...
 	}
 	logger.Debug(fmt.Sprintf("sent (%v) newley minted coins from the tokenomics module to the proposer with address %q", proposerCoin, proposerAddr))
 
-	// TODO_MAINNET: Verify that the total distributed coins equals the settlement coins which could happen due to float rounding
-	totalDistributedCoins := appCoin.Add(supplierCoin).Add(*daoCoin).Add(*serviceCoin).Add(*proposerCoin)
-	if totalDistributedCoins.Amount.BigInt().Cmp(settlementCoins.Amount.BigInt()) != 0 {
-		logger.Error(fmt.Sprintf("TODO_MAINNET: The total distributed coins (%v) does not equal the settlement coins (%v)", totalDistributedCoins, settlementCoins.Amount.BigInt()))
-	}
-	logger.Info(fmt.Sprintf("distributed (%v) coins to the application, supplier, DAO, source owner, and proposer", totalDistributedCoins))
+	// Check and log the total amount of coins distributed
```
red-0ne, line 419: Could we have a dedicated function to this?

x/tokenomics/keeper/token_logic_modules.go
Olshansk:
```golang
// TODO_BETA(@red-0ne): When starting a new session, check what the MaxClaimableAmount
// (in uPOKT) by the Supplier as a function of
// (app_stake, compute_units_per_relay_for_service, global_compute_units_to_token_multiplier).
// TODO_CONFIG_NOTE: Whether or not the RelayMiner stop handling requests when the max is reached should be
// configurable by the operator.
// TODO_ERROR_NOTE: If overservicing is set to false, create a new error that the relay is rejected
// specifically because the supplier has reached the max claimable amount, so the caller should relay
// the request to another supplier.
```

See the `TODO_ERROR_NOTE`. No way to really enforce it but it's something!

red-0ne: In `AppGateServer` this is not an issue due to the round-robin endpoint selection. `Path` needs to account for it.

pkg/relayer/session/sessiontree.go
```diff
@@ -19,7 +19,11 @@ import (
 var _ relayer.SessionTree = (*sessionTree)(nil)
 
 // sessionTree is an implementation of the SessionTree interface.
-// TODO_TEST: Add tests to the sessionTree.
+// TODO_BETA(@red-0ne): Per the Relay Mining paper, we need to optimistically store
+// the number of requests that an application can pay for. This needs to be tracked
+// based on the app's stake in the beginning of a session and the number of nodes
+// per session. An operator should be able to specify "overservicing_okay" whereby
```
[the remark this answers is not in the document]
red-0ne: I'm not sure keeping track of the app's stake would be relevant.

This is because at the beginning of a new session, there may be one (or more) settlements from previous sessions that will occur after the new session starts (due to C&P window offsets) and that will alter the stake.

red-0ne (commented):
Glad to see the tokenomics module progressing. And there's a couple of great catches BTW!

I Still have to review the TLM tests.

image

x/tokenomics/keeper/update_relay_mining_difficulty.go
```diff
@@ -190,3 +182,26 @@ func computeEma(alpha *big.Float, prevEma, currValue uint64) uint64 {
 	newEma, _ := new(big.Float).Add(weightedCurrentContribution, weightedPreviousContribution).Uint64()
 	return newEma
 }
+
+// newDefaultRelayMiningDifficulty is a helper that creates a new RelayMiningDifficulty
+// structure if one is not available. It is often used to set the default when a service's
+// difficulty is being initialized for the first time.
```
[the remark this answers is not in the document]
Olshansk: TY!

pkg/relayer/session/sessiontree.go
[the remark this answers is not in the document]
Olshansk: I'm changing this to a `TODO_IN_THIS_PR_DISCUSS` until we are in alignment.

@red-0ne This is the foundation of optimistic rate limiting and the best we can do. The app's stake in escrow is a signal into rate limiting. 

Window-base centralized rate limiting solutions are better (because it's 100% centralized and controlled).

We need to be in alignment that this is the goal, or we likely have to go back to the drawing board prior to mainNet.

x/session/keeper/session_hydrator.go
```diff
@@ -20,8 +20,8 @@ import (
 
 var SHA3HashLen = crypto.SHA3_256.Size()
 
-// TODO_BLOCKER(@bryanchriswhite, #21): Make these configurable governance param
 const (
+	// TODO_BETA(@bryanchriswhite): Make this a governance parameter
```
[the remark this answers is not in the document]
Olshansk: Yes

x/tokenomics/keeper/token_logic_modules.go
```diff
@@ -61,21 +78,22 @@ func (tlm TokenLogicModule) EnumIndex() int {
...
 // TokenLogicModuleProcessor is the method signature that all token logic modules
 // are expected to implement.
-// IMPORTANT SIDE EFFECTS: Please note that TLMS may update the application and supplier
-// objects, which is why they are passed in as pointers. However, this IS NOT persisted.
-// The persistence to the keeper is currently done by ProcessTokenLogicModules only.
-// This may be an interim state of the implementation and may change in the future.
+// IMPORTANT_SIDE_EFFECTS: Please note that TLMs may update the application and supplier
+// objects, which is why they are passed in as pointers. NOTE THAT THIS IS NOT PERSISTED.
+// The persistence to the keeper is currently done by the TLM processor: ProcessTokenLogicModules.
+// This design and separation of concerns may change in the future.
```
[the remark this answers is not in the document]
Olshansk: You got it!

x/tokenomics/keeper/token_logic_modules.go
Olshansk: This is not global, it's specifically for this claim.

I'm using `total` because it is the sum of all the others.

I think we should leave it as is.

testutil/integration/app.go
```diff
@@ -509,9 +509,10 @@ func NewCompleteIntegrationApp(t *testing.T) *App {
 
 	// Prepare a new default service
 	defaultService := sharedtypes.Service{
-		Id:           "svc1",
-		Name:         "svcName1",
-		OwnerAddress: sample.AccAddress(),
+		Id:                   "svc1",
+		Name:                 "svcName1",
+		ComputeUnitsPerRelay: 1,
```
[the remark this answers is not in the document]
Olshansk: No need. @adshmh implemented the proper solution in #745!

x/tokenomics/keeper/token_logic_modules.go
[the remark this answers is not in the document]
Olshansk: If you look in `distributeSupplierRewardsToShareHolders`, the address was only being used for logging!

Regardless, this is a very good catch and improvement.

In the `token_logic_modules_test.go`, we have the following code snippet from @red-0ne :

```go
	// Assert that the supplier shareholders account balances have *increased* by
	// the appropriate amount w.r.t token distribution.
	shareAmounts := tokenomicskeeper.GetShareAmountMap(supplierRevShares, appBurn.Uint64())
	for shareHolderAddr, expectedShareAmount := range shareAmounts {
		shareHolderBalance := getBalance(t, ctx, keepers, shareHolderAddr)
		require.Equal(t, int64(expectedShareAmount), shareHolderBalance.Amount.Int64())
	}
```

x/tokenomics/keeper/token_logic_modules.go
Olshansk: This was a new feature implemented by @red-0ne: https://dev.poktroll.com/operate/configs/supplier_staking_config#default_rev_share_percent

THEORETICALLY, the `operator` and `owner` don't have to receive any of the rewards from claims & proofs.

In practice, they almost always are part of the list.

Olshansk:
@bryanchriswhite @red-0ne THAT WAS QUITE THE REVIEW! Thank you so much 🙏 

Aside from catching all the issues & improvements, it really feels like you guys have more visibility into what we're going into prod with.

In terms of next steps:
1. Make sure to do a diff since you last reviewed it to see all the changes
2. I left a handful of comments unresolved where I thought the comments were worth it, so please resolve them yourself if it's okay.
3. There is one `TODO_IN_THIS_PR` that we need to align on before merging.
4. I anticipate some merge conflicts as a couple other PRs get in before this but not expecting any issues and will handle it.

x/tokenomics/keeper/token_logic_modules_test.go
```diff
@@ -312,50 +249,175 @@ func TestProcessTokenLogicModules_AppStakeTooLow(t *testing.T) {
...
+		Stake:           &supplierStake,
+		Services:        []*sharedtypes.SupplierServiceConfig{{Service: service, RevShare: supplierRevShares}},
+	}
+	keepers.SetSupplier(ctx, supplier)
+
+	// Prepare the claim for which the supplier did work for the application
+	claim := prepareTestClaim(numRelays, service, &app, &supplier)
+
+	// Prepare addresses
+	daoAddr := authtypes.NewModuleAddress(govtypes.ModuleName)
```
red-0ne, line 365: NIT for naming consistency

```suggestion
	daoAddress := authtypes.NewModuleAddress(govtypes.ModuleName)
```

x/tokenomics/keeper/token_logic_modules_test.go
```diff
@@ -312,50 +249,175 @@ func TestProcessTokenLogicModules_AppStakeTooLow(t *testing.T) {
...
+	appAddress := app.Address
+	proposerAddress := sample.AccAddressFromConsAddress(validatorConsAddr)
+	// supplierOperatorAddr := supplier.OperatorAddress
+
+	// Determine balances before inflation
+	daoBalanceBefore := getBalance(t, ctx, keepers, daoAddr.String())
+	propBalanceBefore := getBalance(t, ctx, keepers, proposerAddress)
+	serviceOwnerBalanceBefore := getBalance(t, ctx, keepers, service.OwnerAddress)
```
red-0ne, line 373: Asserting on service owner balance is redundant, it's already in the shareholders.

x/tokenomics/keeper/token_logic_modules_test.go
```diff
@@ -312,50 +249,175 @@ func TestProcessTokenLogicModules_AppStakeTooLow(t *testing.T) {
...
+
+	// Compute mint per actor
+	numTokensMinted := numTokensClaimed * tokenomicskeeper.MintPerClaimedTokenGlobalInflation
+	daoMint := math.NewInt(int64(numTokensMinted * tokenomicskeeper.MintAllocationDAO))
+	propMint := math.NewInt(int64(numTokensMinted * tokenomicskeeper.MintAllocationProposer))
+	serviceOwnerMint := math.NewInt(int64(numTokensMinted * tokenomicskeeper.MintAllocationSourceOwner))
+	appMint := math.NewInt(int64(numTokensMinted * tokenomicskeeper.MintAllocationApplication))
+	supplierMint := float32(numTokensMinted * tokenomicskeeper.MintAllocationSupplier)
+
+	// Ensure the balance was increase be the appropriate amount
```
red-0ne, line 404:
```suggestion
	// Ensure the balance was increased be the appropriate amount
```

pkg/relayer/session/sessiontree.go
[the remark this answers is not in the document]
red-0ne: @Olshansk I agree and think that max claimable amount with stake check is the best coordination-free solution to over servicing.

My only concern is that the relay miners will rely on an outdated stake (at least by one update). I spent quite some time trying to workaround this but no solution not involving off-chain suppliers cooperation came to mind.

The best mitigation I can imagine is suppliers assuming that the application consumed max claimable amount for each non-yet-settled closed session (probably 1 or 2 sessions).

To conclude, I think we are aligned that we'll have to use max claimable amount and app stake in a more or less elaborated way to rate-limit the application. Future solutions will be added when appropriate.

red-0ne (approved):
This PR greatly clarifies what we're going to work on in the future. Thank you for setting all of this up. :railway_track: 

I believe we're in line on where our efforts will be now.

Preemptively approving as I only added a few NITs.

x/tokenomics/keeper/token_logic_modules.go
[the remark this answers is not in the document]
bryanchriswhite: > If you look in `distributeSupplierRewardsToShareHolders`, the address was only being used for logging!

At the time of review (4c335e7d), [token_logic_modules.go:557](https://github.com/pokt-network/poktroll/blob/4c335e7dd53aa8f766d63cd9b854aa2fb29f15e3/x/tokenomics/keeper/token_logic_modules.go#L557) was using that argument to retrieve the keeper from the store.

x/tokenomics/keeper/token_logic_modules_test.go
Olshansk: 1. It's still a different case (a service owner)
2. I looked through the code and it is a completely different address as far as I can tell

x/tokenomics/keeper/token_logic_modules.go
```diff
@@ -285,22 +316,9 @@ func (k Keeper) TokenLogicModuleRelayBurnEqualsMint(
...
-		}
-	}
-
 	// Burn uPOKT from the application module account which was held in escrow
 	// on behalf of the application account.
-	if err = k.bankKeeper.BurnCoins(
+	if err := k.bankKeeper.BurnCoins(
 		ctx, apptypes.ModuleName, sdk.NewCoins(settlementCoin),
 	); err != nil {
 		return tokenomicstypes.ErrTokenomicsApplicationModuleBurn.Wrapf("burning %s from the application module account: %v", settlementCoin, err)
```
bryanchriswhite, line 324: I'm noticing (in the lines below: 329-333) that by the time we determine whether or not the application has enough stake to decrement, we've already burnt their tokens from escrow (app module acct.). Since we return an error in the case where app stake is insufficient and leave the stake unchanged, the app will continue to have stake representing tokens in escrow which have already been burnt (and then some).

Wdyt about updating the application stake before burning from the app module acct.?

x/tokenomics/keeper/token_logic_modules.go
```diff
@@ -19,31 +16,51 @@ import (
...
+	// larger than what is distributed to pocket network participants
+	// TODO_MAINNET: Figure out if we can avoid this tolerance and use fixed point arithmetic.
+	MintDistributionAllowableTolerance = 0.02
 )
 
 type TokenLogicModule int
 
 const (
+	// TLMRelayBurnEqualsMint is the token logic module that burns the application's
+	// stake based on the amount of work done by the supplier.
```
bryanchriswhite, line 55:
```suggestion
	// stake balance based on the amount of work done by the supplier.
```

x/tokenomics/keeper/token_logic_modules.go
```diff
@@ -319,149 +337,216 @@ func (k Keeper) TokenLogicModuleRelayBurnEqualsMint(
...
-	// TODO_MAINNET(@Olshansk, @RawthiL): The application was over-serviced in the last session so it basically
-	// goes "into debt". Need to design a way to handle this when we implement
-	// probabilistic proofs and add all the parameter logic. Do we touch the application balance?
-	// Do we just let it go into debt? Do we penalize the application? Do we unstake it? Etc...
-	// See this document from @red-0ne and @bryanchriswhite for more context: notion.so/buildwithgrove/Off-chain-Application-Stake-Tracking-6a8bebb107db4f7f9dc62cbe7ba555f7
-	expectedBurn := settlementCoins
+	logger = logger.With("helper", "ensureClaimAmountLimits")
+
+	// TODO_BETA_OR_MAINNET(@red-0ne): The application stake gets reduced with every claim
+	// settlement. Relay miners use the appStake at the beginning of a session to determine
```
bryanchriswhite, line 513: TL;DR, I think we need to consider adding support for in-session stake increases (i.e. taking stake increases into account within the session they're committed).

I find myself thinking about this from the perspective of the application actor's UX, which seems to factor into how we handle overserviced applications (e.g. unstake them, etc.) and minimum stake balances:

I think this fact (`Relay miners use...`) means that an application operator who would top up their stake from an insufficient balance to a sufficient one in session N will have to wait for session N+1 before they can expect to be serviced by rational/pessimistic/non-altruistic suppliers?

It's been a minute since I looked at the session hydration code but I also find myself wondering about whether and how app stake balance is/should factor into session generation. It could be the case that an application is able to "get a session" via query but that suppliers may not service them until the next session.

bryanchriswhite (changes requested):
Took another pass but had to stop before I could get through the TLM test.

image

This is looking great! :rocket:

x/tokenomics/keeper/token_logic_modules.go
```diff
@@ -319,149 +337,216 @@ func (k Keeper) TokenLogicModuleRelayBurnEqualsMint(
...
 	application *apptypes.Application,
 	supplier *sharedtypes.Supplier,
-	settlementCoins cosmostypes.Coin,
+	settlementCoin cosmostypes.Coin,
 	relayMiningDifficulty *tokenomictypes.RelayMiningDifficulty,
 ) error {
 	logger := k.Logger().With("method", "TokenLogicModuleGlobalMint")
 
+	if MintPerClaimedTokenGlobalInflation == 0 {
+		logger.Warn("global inflation is set to zero. Skipping Global Mint TLM.")
```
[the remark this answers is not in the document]
Olshansk: Yup! That'll come in the next PR.

x/tokenomics/keeper/token_logic_modules.go
Olshansk:
```diff
type Session struct {
	Header              *SessionHeader      `protobuf:"bytes,1,opt,name=header,proto3" json:"header,omitempty"`
	SessionId           string              `protobuf:"bytes,2,opt,name=session_id,json=sessionId,proto3" json:"session_id,omitempty"`
	SessionNumber       int64               `protobuf:"varint,3,opt,name=session_number,json=sessionNumber,proto3" json:"session_number,omitempty"`
	NumBlocksPerSession int64               `protobuf:"varint,4,opt,name=num_blocks_per_session,json=numBlocksPerSession,proto3" json:"num_blocks_per_session,omitempty"`
	Application         *types1.Application `protobuf:"bytes,5,opt,name=application,proto3" json:"application,omitempty"`
	Suppliers           []*types.Supplier   `protobuf:"bytes,6,rep,name=suppliers,proto3" json:"suppliers,omitempty"`
+	// TODO_MAINNET(@bryanchriswhite): Add a RelayMining field that can be used for UX/DevX
+	// purposes capturing things like:
+	// - The probability of a proof being required for this session
+	// - The threshold of the probability that triggers a proof definitively
+	// - The maximum claimable pokt per supplier in this session
+	// See the discussion here: https://github.com/pokt-network/poktroll/pull/755#discussion_r1743286043
}
```

bryanchriswhite (approved):
Apologies for the delay, I made another pass this morning and didn't realize this review was still pending. :sweat_smile: 

Preemptively approving. Only one final ask regarding splitting token_logic_modules.go.

red-0ne (approved):

Issue #758, opened 2024-08-23
[Load Testing] Permissionless Demand Implementation Umbrella Ticket 


Olshansk:
## Objective

Ensure there will not be any scalability issues in Shannon.

## Origin Document

- [Probabilistic Proofs Document from Morse](https://github.com/pokt-network/pocket-core/blob/7f936ff7353249b161854e24435e4bc32d47aa3f/docs/proposals/probabilistic_proofs.md)
- [Permissionless Demand Load Testing Notion Doc](https://www.notion.so/buildwithgrove/Permissionless-Demand-Load-Testing-24feefb2f34f4399a941beb374bb6ea1?pvs=4)

- #711 

## Goals

- Understand missing gaps in the implementation of probabilistic proofs on both the supply (relayer) side & consensus (validator) side
- Implement everything necessary to make sure Shannon demand side scales scales
- Identify MUST have TODO for mainnet
- Come up with new ideas to improve the overall design while keeping things simple
- Update the documentation so all readers and ecosystem participants understand how things work
- Keep iterations and PRs small because the scope of this can get out of hand

## Deliverables

- [ ] Deliverable #1
- [ ] Deliverable #2

## Non-goals / Non-deliverables

- The following are requirements of #711 being built by @okdas 
	- Build the tooling for these load tests
	- Building observability tools for these tests
	- Gain an understanding into the block size
	- Get visibility 

## General deliverables

- [ ] **Comments**: Add/update TODOs and comments alongside the source code so it is easier to follow.
- [ ] **Testing**: Add new tests (unit and/or E2E) to the test suite.
- [ ] **Makefile**: Add new targets to the Makefile to make the new functionality easier to use.
- [ ] **Documentation**: Update architectural or development READMEs; use [mermaid](https://mermaid-js.github.io/mermaid/) diagrams where appropriate.

---

**Creator**: @Olshansk 
**Co-Owners**: @okdas @red-0ne

Olshansk:
Closing this out as an old ticket where A LOT of other effort has already taken place.

PR #765, opened 2024-08-27
[SessionManager] Implement off-chain proof params usage

red-0ne:
## Summary

This PR introduces the proof params to the `RelayMiner` so it does not submit proofs when not required.

The E2E and `SessionManager` tests have been updated to account for proof requirement.

## Issue

The `RelayMiner`'s `SessionManager` was systematically submitting proofs on-chain whether they were required or not.

This led to a state chain bloat since proofs are what takes most for the block space.

Given the permissionless nature of the protocol, this causing scalability issues.

- #758 

## Type of change

Select one or more:

- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

## Sanity Checklist


pkg/relayer/session/claim.go
```diff
@@ -231,3 +256,78 @@ func (rs *relayerSessionsManager) goCreateClaimRoots(
...
+	ctx context.Context,
+	sessionTree relayer.SessionTree,
+) (isProofRequired bool, err error) {
+	logger := rs.logger.With(
+		"session_id", sessionTree.GetSessionHeader().GetSessionId(),
+		"claim_root", fmt.Sprintf("%x", sessionTree.GetClaimRoot()),
+		"supplier_operator_address", sessionTree.GetSupplierOperatorAddress().String(),
+	)
+
+	// Create the claim object and use its methods to determine if a proof is required.
```
bryanchriswhite, line 272: :sunglasses:

pkg/relayer/session/claim.go
```diff
@@ -231,3 +256,78 @@ func (rs *relayerSessionsManager) goCreateClaimRoots(
...
+
+	proofParams, err := rs.proofQueryClient.GetParams(ctx)
+	if err != nil {
+		return false, err
+	}
+
+	logger = logger.With(
+		"num_claim_compute_units", numClaimComputeUnits,
+		"proof_requirement_threshold", proofParams.GetProofRequirementThreshold(),
+	)
```
bryanchriswhite, line 293: :ok_hand:

pkg/relayer/session/session_test.go
bryanchriswhite, line 1: :fire:

pkg/relayer/session/session.go
```diff
@@ -54,13 +54,17 @@ type relayerSessionsManager struct {
 
 	// sharedQueryClient is used to query shared module parameters.
 	sharedQueryClient client.SharedQueryClient
+	// proofQueryClient is used to query for the proof requirement threshold and
+	// requirement probability to determine whether a submitted claim requires a proof.
```
bryanchriswhite, line 58:
```suggestion
	// requirement probability governance parameters to determine whether a
	// submitted claim requires a proof.
```

bryanchriswhite (changes requested):
:raised_hands: Nice one @red-0ne!

e2e/tests/session.feature
```diff
@@ -2,12 +2,20 @@ Feature: Session Namespace
...
+    # previous relay count.
+    When the supplier "supplier1" has serviced a session with "21" relays for service "anvil" for application "app1"
     And the user should wait for the "proof" module "CreateClaim" Message to be submitted
     And the user should wait for the "proof" module "ClaimCreated" tx event to be broadcast
-    Then the claim created by supplier "supplier1" for service "svc1" for application "app1" should be persisted on-chain
     And the user should wait for the "proof" module "SubmitProof" Message to be submitted
     And the user should wait for the "proof" module "ProofSubmitted" tx event to be broadcast
+    # Observing the "ClaimSettled" block event also asserts that the claim was
+    # present at the moment of settlement. So there is no need to query the claim
+    # directly.
```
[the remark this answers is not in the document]
red-0ne: It was mainly to justify the removal of the flaky step. But I agree this should be a godoc of the step function.

e2e/tests/session.feature
```diff
@@ -2,12 +2,20 @@ Feature: Session Namespace
...
+    # The number of relays serviced is set to make the resulting compute units sum
+    # above the current ProofRequirementThreshold governance parameter so a proof
+    # is always required.
+    # TODO_TECHDEBT(#745): Once the SMST is updated with the proper weights,
+    # using the appropriate compute units per relay, we can then restore the
+    # previous relay count.
+    When the supplier "supplier1" has serviced a session with "21" relays for service "anvil" for application "app1"
     And the user should wait for the "proof" module "CreateClaim" Message to be submitted
     And the user should wait for the "proof" module "ClaimCreated" tx event to be broadcast
-    Then the claim created by supplier "supplier1" for service "svc1" for application "app1" should be persisted on-chain
```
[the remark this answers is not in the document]
red-0ne: I think this is coming from the assertion we make about the claims count
```
	require.Greater(s, len(allClaimsRes.Claims), len(preExistingClaims), "number of claims must have increased")
```

Which in some circumstances (tests repeated to quickly) fails because the previous claim gets removed.

I'll put back the step even if it's at the cost of some flakiness.

pkg/relayer/session/claim.go
```diff
@@ -180,9 +181,26 @@ func (rs *relayerSessionsManager) newMapClaimSessionsFn(
...
-		claimMsgs := make([]client.MsgCreateClaim, 0)
+		claimMsgs := make([]client.MsgCreateClaim, 0, len(sessionTrees))
+		// sessionTreesWithProofRequired will accumulate all the sessionTrees that
+		// will require a proof to be submitted.
+		sessionTreesWithProofRequired := make([]relayer.SessionTree, 0)
 		for _, sessionTree := range sessionTrees {
+			isProofRequired, err := rs.isProofRequired(ctx, sessionTree)
+
+			// If an error is encountered while determining if a proof is required,
+			// do not create the claim since the proof requirement is unknown.
```
Olshansk, line 192:
```suggestion
			// do not create the claim since the proof requirement is unknown.
			// Creating a claim and not submitting a proof (if necessary) could lead to a burn!!
```

pkg/relayer/session/claim.go
```diff
@@ -180,9 +181,26 @@ func (rs *relayerSessionsManager) newMapClaimSessionsFn(
...
+			if err != nil {
+				failedCreateClaimsSessionsPublishCh <- sessionTrees
+				rs.logger.Error().Err(err).Msg("failed to determine if proof is required")
```
Olshansk, line 195:
```suggestion
				rs.logger.Error().Err(err).Msg("failed to determine if proof is required so skipping claim creation")
```

pkg/relayer/session/claim.go
```diff
@@ -180,9 +181,26 @@ func (rs *relayerSessionsManager) newMapClaimSessionsFn(
...
-		// Map key is the supplier operator address.
-		claimMsgs := make([]client.MsgCreateClaim, 0)
+		claimMsgs := make([]client.MsgCreateClaim, 0, len(sessionTrees))
+		// sessionTreesWithProofRequired will accumulate all the sessionTrees that
+		// will require a proof to be submitted.
+		sessionTreesWithProofRequired := make([]relayer.SessionTree, 0)
 		for _, sessionTree := range sessionTrees {
+			isProofRequired, err := rs.isProofRequired(ctx, sessionTree)
+
+			// If an error is encountered while determining if a proof is required,
```
Olshansk, line 191: This is weird though, we ONLY know if a proof is required AFTER the claim was created. Otherwwise you could game the system. What am I missing?

pkg/relayer/session/claim.go
```diff
@@ -231,3 +256,83 @@ func (rs *relayerSessionsManager) goCreateClaimRoots(
...
+	ctx context.Context,
+	sessionTree relayer.SessionTree,
+) (isProofRequired bool, err error) {
+	logger := rs.logger.With(
+		"session_id", sessionTree.GetSessionHeader().GetSessionId(),
+		"claim_root", fmt.Sprintf("%x", sessionTree.GetClaimRoot()),
+		"supplier_operator_address", sessionTree.GetSupplierOperatorAddress().String(),
+	)
+
+	// Create the claim object and use its methods to determine if a proof is required.
```
Olshansk, line 272: Can you #PUC that the claim has already been created on-chain prior to this?

i.e. we are re-creating the claim object.

I want to ensure we are aligned on this.

pkg/relayer/session/claim.go
```diff
@@ -231,3 +256,83 @@ func (rs *relayerSessionsManager) goCreateClaimRoots(
...
+	if err != nil {
+		return false, err
+	}
+
+	// Sample a pseudo-random value between 0 and 1 to determine if a proof is required probabilistically.
+	var randFloat float32
+	randFloat, err = poktrand.SeededFloat32(claimHash[:])
+	if err != nil {
+		return false, err
+	}
```
Olshansk, line 310: **⚠️  CRITICAL ⚠️** 

When I saw this, it raised a critical issue that we overlooked.

If the `rand(seed(claimHash))` is all we need to determine the probability, anyone can grind different claims until you don't need a proof.

In the same way that we use the block hash AFTER the claim closes (see `GetPathForProof`), we HAVE to do the same thing when computing this probability as well.

Olshansk: Can you add a TODO_BETA in this PR but immediately tend to the correct business logic in another?

Olshansk (changes requested):

pkg/relayer/session/claim.go
red-0ne: _TL;DR; Seed should use more than claim hash, this has to be done both on/off-chain, will send a fix_

> re-creating the claim object

Actually no, we already have all we need to create a claim. There's nothing missing to do it exactly the same way on-chain does, see  [x/proof/keeper/msg_server_create_claim.go#L55](https://github.com/pokt-network/poktroll/blob/main/x/proof/keeper/msg_server_create_claim.go#L55)

>  we ONLY know if a proof is required AFTER

We also have all we need to know whether a poof is required or not, this flow reproduces how it's actually done on-chain, see: [main/x/tokenomics/keeper/settle_pending_claims.go#L282](https://github.com/pokt-network/poktroll/blob/main/x/tokenomics/keeper/settle_pending_claims.go#L282)

I agree that the actor can leverage the claim's root hash to get a seed that does not require a proof.

In order to fix that, that seed needs to include something that is not know as of the time of the claim submission.

If we add `ProofWindowOpenHeight` block hash to the random seed, we'll enforce that the probabilistic proof requirement is impossible to calculate before observing `ProofWindowOpenHeight`.

Which in turn will force the `RelayMiner` to push this logic to the proof logic.

If you agree with `ProofWindowOpenHeight` being the future value to use, I'll update both this logic and the on-chain part that enforces it.

pkg/relayer/session/claim.go
red-0ne: Grouping my reply in https://github.com/pokt-network/poktroll/pull/765#discussion_r1736436005

pkg/relayer/session/claim.go
red-0ne: Grouping my reply in https://github.com/pokt-network/poktroll/pull/765#discussion_r1736436005

pkg/relayer/session/claim.go
Olshansk: > TL;DR; Seed should use more than claim hash, this has to be done both on/off-chain, will send a fix

🤝  Agreed. Let's do it.

> There's nothing missing to do it...

 ❓  Can we create a shared helper?
 
 ---
 
 @red-0ne W.r.t to the rest, we're on the same page and aligned. WHat I was calling out is to following the exact same pattern as in `proof_validation.go`:
 
 ```go
 	// earliestSupplierProofCommitHeight - 1 is the block that will have its hash used as the
	// source of entropy for all the session trees in that batch, waiting for it to
	// be received before proceeding.
	proofPathSeedBlockHash := k.sessionKeeper.GetBlockHash(ctx, earliestSupplierProofCommitHeight-1)

	// TODO_BETA: Investigate "proof for the path provided does not match one expected by the on-chain protocol"
	// error that may occur due to block height differing from the off-chain part.
	k.logger.Info("E2E_DEBUG: height for block hash when verifying the proof", earliestSupplierProofCommitHeight, sessionHeader.GetSessionId())

	expectedProofPath := protocol.GetPathForProof(proofPathSeedBlockHash, sessionHeader.GetSessionId())
	if !bytes.Equal(proof.Path, expectedProofPath) {
		return types.ErrProofInvalidProof.Wrapf(
			"the path of the proof provided (%x) does not match one expected by the on-chain protocol (%x)",
			proof.Path,
			expectedProofPath,
		)
	}
```

File: pkg/relayer/session/session_test.go [hunk 11]
```diff
@@ -161,6 +169,43 @@ func TestRelayerSessionsManager_Start(t *testing.T) {
 	waitSimulateIO()
 }
 
+func TestRelayerSessionsManager_ProofThresholdRequired(t *testing.T) {
```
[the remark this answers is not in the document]
red-0ne: The `t.Run(t, callback)` doesn't seem to play well with the test setup :/

x/proof/types/params.go
```diff
@@ -22,6 +22,9 @@ var (
 	DefaultRelayDifficultyTargetHashHex = "ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff" // all relays are payable
 	DefaultRelayDifficultyTargetHash, _ = hex.DecodeString(DefaultRelayDifficultyTargetHashHex)
 
+	// TODO_BETA(@red-0ne): Iterate on the parameters below by adding unit suffixes,
+	// whether proof requirement threshold should be in function of the stake amount.
```
Olshansk, line 26: The first sentence LGTM, but I don't understand what you're intending to say with the second one here.

testutil/testclient/testsupplier/client.go
```diff
@@ -40,10 +40,13 @@ func NewLocalnetClient(
 	return supplierClient
 }
 
-func NewOneTimeClaimProofSupplierClientMap(
+// NewClaimProofSupplierClientMap creates and returns a map of supplier to supplier
+// mock client. Which expect to create one claim and a given number of proofs.
```
Olshansk, line 44:
```suggestion
// client mocks. Each supplier client is expected to submit exactly 1 claim and exactly proofCount proofs.
```

x/tokenomics/keeper/settle_pending_claims.go
```diff
@@ -286,9 +288,16 @@ func (k Keeper) proofRequirementForClaim(ctx sdk.Context, claim *prooftypes.Clai
 		return requirementReason, err
 	}
 
+	proofPathSeedBlockHash, err := k.getEarliestSupplierProofCommitBlockHash(ctx, claim)
```
Olshansk, line 291: s/proofPathSeedBlockHash/blockHashEarliestProofCommitHeight

x/tokenomics/keeper/settle_pending_claims.go
```diff
@@ -286,9 +288,16 @@ func (k Keeper) proofRequirementForClaim(ctx sdk.Context, claim *prooftypes.Clai
...
+	if err != nil {
+		return requirementReason, err
+	}
+
+	proofRequirementSeed := append(claimHash, proofPathSeedBlockHash...)
+
 	// Sample a pseudo-random value between 0 and 1 to determine if a proof is required probabilistically.
 	var randFloat float32
-	randFloat, err = poktrand.SeededFloat32(claimHash[:])
+	randFloat, err = poktrand.SeededFloat32(proofRequirementSeed)
```
Olshansk, line 300: Can you move lines 291-303 into a helper and rename `randFloat` to `proofRequiredProbability`

pkg/relayer/session/claim.go
```diff
@@ -180,8 +180,7 @@ func (rs *relayerSessionsManager) newMapClaimSessionsFn(
 			return either.Success(sessionTrees), false
 		}
 
-		// Map key is the supplier operator address.
-		claimMsgs := make([]client.MsgCreateClaim, 0)
+		claimMsgs := make([]client.MsgCreateClaim, 0, len(sessionTrees))
 		for _, sessionTree := range sessionTrees {
 			claimMsgs = append(claimMsgs, &prooftypes.MsgCreateClaim{
```
Olshansk, line 185:
```suggestion
		for idx, sessionTree := range sessionTrees {
			claimMsgs[idx] = &prooftypes.MsgCreateClaim{
```

Olshansk: You'll need to remove the `0` from above

pkg/relayer/session/proof.go
```diff
@@ -222,17 +224,38 @@ func (rs *relayerSessionsManager) newMapProveSessionsFn(
 func (rs *relayerSessionsManager) goProveClaims(
 	ctx context.Context,
 	sessionTrees []relayer.SessionTree,
-	sessionPathBlock client.Block,
+	proofPathSeedBlock client.Block,
```
Olshansk, line 227: Can you 
```suggestion
	proofPathSeedBlock client.Block, // the hash of this block is used to determine which branch the proof should be generated for
```

pkg/relayer/session/proof.go
```diff
@@ -258,6 +281,94 @@ func (rs *relayerSessionsManager) goProveClaims(
...
+	}
+
+	// Append the hash of the proofPathSeedBlock to the claim hash to seed the random number generator
+	// to ensure that the proof requirement probability is unknown until the proofPathSeedBlock is observed.
+	// The on-chain claim settlement routine will use the same seed to determine if a proof is required.
+	proofRequirementSeed := append(claimHash, proofPathSeedBlock.Hash()...)
+
+	// Sample a pseudo-random value between 0 and 1 to determine if a proof is required probabilistically.
+	var randFloat float32
+	randFloat, err = poktrand.SeededFloat32(proofRequirementSeed)
```
Olshansk, line 343: Don't like this `randFloat` naming. See my other comment and reuse the helper that I recommend creating from the onchain code.

pkg/relayer/session/proof.go
```diff
@@ -258,6 +281,94 @@ func (rs *relayerSessionsManager) goProveClaims(
...
 }
+
+// isProofRequired determines whether a proof is required for the given session's
+// claim based on the current proof module governance parameters.
+// TODO_TECHDEBT: Once the on/off-chain loggers are unified, move this logic to
+// a shared helper used by both off-chain and on-chain routines.
+func (rs *relayerSessionsManager) isProofRequired(
+	ctx context.Context,
+	sessionTree relayer.SessionTree,
+	proofPathSeedBlock client.Block,
```
Olshansk, line 295:
```suggestion
	proofRequirementSeedBlock client.Block,  // the hash of this block is used to determine whether the proof is required w.r.t the probabilistic features
```

pkg/relayer/session/proof.go
```diff
@@ -258,6 +281,94 @@ func (rs *relayerSessionsManager) goProveClaims(
...
+		"session_id", sessionTree.GetSessionHeader().GetSessionId(),
+		"claim_root", fmt.Sprintf("%x", sessionTree.GetClaimRoot()),
+		"supplier_operator_address", sessionTree.GetSupplierOperatorAddress().String(),
+	)
+
+	// Create the claim object and use its methods to determine if a proof is required.
+	claim := claimFromSessionTree(sessionTree)
+
+	// Get the number of compute units accumulated through the given session.
+	numClaimComputeUnits, err := claim.GetNumComputeUnits()
```
Olshansk, line 307: Can you please create a helper for `getComputeUnits` that gets

1. Compute units from tree
2. Num relays in tree * service.ComputeUnitsPerRelay
3. Errors if (1) & (2) are not the same
4. Returns a single value

pkg/relayer/session/proof.go
```diff
@@ -258,6 +281,94 @@ func (rs *relayerSessionsManager) goProveClaims(
...
+		return false, err
+	}
+
+	logger = logger.With(
+		"num_claim_compute_units", numClaimComputeUnits,
+		"proof_requirement_threshold", proofParams.GetProofRequirementThreshold(),
+	)
+
+	// Require a proof if the claim's compute units meets or exceeds the threshold.
+	if numClaimComputeUnits >= proofParams.GetProofRequirementThreshold() {
```
Olshansk, line 323:
```suggestion
	// TODO_MAINNET: This should be proportional to the supplier's stake as well.
	if numClaimComputeUnits >= proofParams.GetProofRequirementThreshold() {
```

pkg/relayer/session/proof.go
```diff
@@ -258,6 +281,94 @@ func (rs *relayerSessionsManager) goProveClaims(
...
+	}
+
+	// Get the hash of the claim to seed the random number generator.
+	var claimHash []byte
+	claimHash, err = claim.GetHash()
+	if err != nil {
+		return false, err
+	}
+
+	// Append the hash of the proofPathSeedBlock to the claim hash to seed the random number generator
```
Olshansk, line 336: Make sure to update comments as well as the var name

pkg/relayer/session/session_test.go
```diff
@@ -161,6 +169,43 @@ func TestRelayerSessionsManager_Start(t *testing.T) {
...
+}
+
+func TestRelayerSessionsManager_ProofProbabilityRequired(t *testing.T) {
+	proofParams := prooftypes.DefaultParams()
+
+	// Set proof requirement threshold to max uint64 to skip the threshold check.
+	proofParams.ProofRequirementThreshold = math.MaxUint64
+	// Set proof request probability to 1 so a proof is always requested.
+	proofParams.ProofRequestProbability = 1
+
```
Olshansk, line 190: #PUC why

pkg/relayer/session/session_test.go
```diff
@@ -161,6 +169,43 @@ func TestRelayerSessionsManager_Start(t *testing.T) {
...
+func TestRelayerSessionsManager_ProofNotRequired(t *testing.T) {
+	proofParams := prooftypes.DefaultParams()
+
+	// Set proof requirement threshold to max uint64 to skip the threshold check.
+	proofParams.ProofRequirementThreshold = math.MaxUint64
+	// Set proof request probability to 0 so a proof is never requested.
+	proofParams.ProofRequestProbability = 0
+
+	numExpectedProofs := 0
```
Olshansk, line 204: #PUC why

pkg/relayer/session/session_test.go
```diff
@@ -161,6 +169,43 @@ func TestRelayerSessionsManager_Start(t *testing.T) {
 	waitSimulateIO()
 }
 
+func TestRelayerSessionsManager_ProofThresholdRequired(t *testing.T) {
+	proofParams := prooftypes.DefaultParams()
+
+	// Set proof requirement threshold to a low enough value so a proof is always requested.
+	proofParams.ProofRequirementThreshold = 1
+
+	numExpectedProofs := 1
```
Olshansk, line 178: #PUC why

Olshansk (changes requested):
Great job. Let's get this in!

pkg/relayer/session/proof.go
red-0ne: Now that #745 is merged, It fixes [1] `smst.Update` to correctly update `numComputeUnits` for the tree, so it's safe to continue using `claim.GetNumComputeUnits`

[1] https://github.com/pokt-network/poktroll/blob/main/pkg/relayer/session/session.go#L436

pkg/relayer/session/proof.go
red-0ne: Renaming to `proofRequirementCheckValue` so it's distinguishable from the gov param. Here's how it'll show in logs:
```go
logger = logger.With(
	"proof_requirement_check_value", proofRequirementCheckValue,
	"proof_request_probability", proofParams.GetProofRequestProbability(),
)
```

pkg/relayer/session/proof.go
```diff
@@ -216,32 +197,45 @@ func (rs *relayerSessionsManager) newMapProveSessionsFn(
...
 	for _, sessionTree := range sessionTrees {
-		select {
-		case <-ctx.Done():
-			return
-		default:
+		isProofRequired, err := rs.isProofRequired(ctx, sessionTree, proofPathSeedBlock)
+
+		// If an error is encountered while determining if a proof is required,
+		// do not create the claim since the proof requirement is unknown.
+		// Creating a claim and not submitting a proof (if necessary) could lead to a stake burn!!
```
Olshansk, line 219:
```suggestion
		// WARNING: Creating a claim and not submitting a proof (if necessary) could lead to a stake burn!!
```

pkg/relayer/session/proof.go
```diff
@@ -258,6 +252,81 @@ func (rs *relayerSessionsManager) goProveClaims(
...
 	}
 
-	failSubmitProofsSessionsCh <- failedProofs
-	proofsGeneratedCh <- successProofs
+	return successProofs, failedProofs
+}
+
+// isProofRequired determines whether a proof is required for the given session's
+// claim based on the current proof module governance parameters.
+// TODO_TECHDEBT: Once the on/off-chain loggers are unified, move this logic to
```
Olshansk, line 260: Is it overkill to just pass a logger, unify it now, and add a TECHDEBT explaining why we pas sin a logger?

x/proof/types/params.go
```diff
@@ -22,6 +22,9 @@ var (
 	DefaultRelayDifficultyTargetHashHex = "ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff" // all relays are payable
 	DefaultRelayDifficultyTargetHash, _ = hex.DecodeString(DefaultRelayDifficultyTargetHashHex)
 
+	// TODO_BETA(@red-0ne): Iterate on the parameters below by adding unit suffixes and
+	// consider having the proof_requirement_threshold to be a function of the stake amount.
```
Olshansk, line 26:
```suggestion
	// consider having the proof_requirement_threshold to be a function of the supplier's stake amount.
```

x/tokenomics/keeper/settle_pending_claims.go
```diff
@@ -279,29 +280,25 @@ func (k Keeper) proofRequirementForClaim(ctx sdk.Context, claim *prooftypes.Clai
...
-	claimHash, err = claim.GetHash()
+	earliestProofCommitBlockHash, err := k.getEarliestSupplierProofCommitBlockHash(ctx, claim)
 	if err != nil {
 		return requirementReason, err
 	}
 
-	// Sample a pseudo-random value between 0 and 1 to determine if a proof is required probabilistically.
-	var randFloat float32
-	randFloat, err = poktrand.SeededFloat32(claimHash[:])
+	proofRequirementCheckValue, err := claim.GetProofRequirementCheckValue(earliestProofCommitBlockHash)
```
Olshansk, line 288: s/CheckValue/SampleValue

Let's use proper math terminology. Ditto everywhere

x/tokenomics/keeper/settle_pending_claims.go
```diff
@@ -311,8 +308,35 @@ func (k Keeper) proofRequirementForClaim(ctx sdk.Context, claim *prooftypes.Clai
...
+		return nil, err
+	}
+
+	sessionEndHeight := claim.GetSessionHeader().GetSessionEndBlockHeight()
+	supplierOperatorAddress := claim.GetSupplierOperatorAddress()
+
+	proofWindowOpenHeight := shared.GetProofWindowOpenHeight(sharedParams, sessionEndHeight)
+	proofWindowOpenBlockHash := k.sessionKeeper.GetBlockHash(ctx, proofWindowOpenHeight)
+
+	earliestSupplierProofCommitHeight := shared.GetEarliestSupplierProofCommitHeight(
```
Olshansk, line 334: Please add a TODO_TECHDEBT next to `GetEarliestSupplierProofCommitHeight` to take in just `sharedParams & claim`

x/proof/types/claim.go
```diff
@@ -27,3 +28,30 @@ func (claim *Claim) GetHash() ([]byte, error) {
 
 	return crypto.Sha256(claimBz), nil
 }
+
+// GetProofRequirementCheckValue returns a pseudo-random value between 0 and 1 to
```
Olshansk, line 32: s/Check/Sample

x/proof/types/claim.go
```diff
@@ -27,3 +28,30 @@ func (claim *Claim) GetHash() ([]byte, error) {
...
+// determine if a proof is required probabilistically.
+func (claim *Claim) GetProofRequirementCheckValue(
```
Olshansk, line 34: Great use of a reciever

Olshansk, line 34:
```suggestion
// IMPORTANT: It is assumed that the caller has ensured the hash of the block seed
// was committed AFTER the claim window has closed.
func (claim *Claim) GetProofRequirementCheckValue(
```

Olshansk (changes requested):

x/tokenomics/keeper/settle_pending_claims.go
red-0ne: If we only pass `sharedParams` and `claim`, this function would no longer be static, as it needs to fetch the a block hash (which is done differently on-chain and off-chain).

pkg/relayer/session/proof.go
red-0ne: The on-chain logger `cosmossdk.io/log` and off-chain one `github.com/pokt-network/poktroll/pkg/polylog` seem to be incompatible.

I'm maybe missing something but we can't pass in a logger in both on/off-chain.

Olshansk: Did you confirm that this doesn't work?

I believe this is the problem `polylog` was meant to solve. cc @bryanchriswhite 

Screenshot 2024-09-06 at 3 38 18 PM

x/tokenomics/keeper/settle_pending_claims.go
Olshansk: Please add a `TODO_TECHDEBT` next to `GetEarliestSupplierProofCommitHeight` that says:

```
TODO_TECHDEBT: Update the method header of this function to accept (sharedParams, Claim, BlockHash).
After doing so, please review all calling sites and simplify them accordingly.
```

Olshansk (approved):
@red-0ne Left a couple more comments. Please review and update TODOs accordingly. Feel free to merge in after. 

Approving pre-emtively.

pkg/relayer/session/proof.go
red-0ne: I tried, but had the following:
image

Will sync-up with @bryanchriswhite on this.

red-0ne: Updated the comments to include investigation.

bryanchriswhite: https://github.com/pokt-network/polylog/issues/3

PR #779, opened 2024-09-02
[Proof] Proof submission fee param

red-0ne:
## Summary

This PR adds `ProofSubmissionFee` governance param as a `Coin` type.

It is based on PR #778 

## Issue

Proof messages take up a lot of block space. We need to deter submitting unnecessary ones by making their submission payable by a governance adjusted parameter.

The current PR is only about adding the governance param. The feature itself will be submitted as a follow-up PR.

- #758 

## Type of change

Select one or more:

- [x] New feature, functionality or library
- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

- [ ] **Documentation**: `make docusaurus_start`; only needed if you make doc changes
- [x] **Unit Tests**: `make go_develop_and_test`
- [x] **LocalNet E2E Tests**: `make test_e2e`
- [ ] **DevNet E2E Tests**: Add the `devnet-test-e2e` label to the PR.

## Sanity Checklist

- [x] I have tested my changes using the available tooling
- [x] I have commented my code
- [x] I have performed a self-review of my own code; both comments & source code
- [ ] I create and reference any new tickets, if applicable
- [x] I have left TODOs throughout the codebase, if applicable

PR #780, opened 2024-09-02
Feat/payable proof submission

red-0ne:
## Summary

This PR implements payable proof submission given the proof submission governance param.

This is a follow-up feature PR to #779 which adds the proof submission governance param.

## Issue

Proof messages take up significant block space. We need to deter submitting non-required ones by making proof submission payable by a governance adjusted parameter.

- #758 

## Type of change

Select one or more:

- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

## Sanity Checklist


e2e/tests/session_steps_test.go
```diff
@@ -42,7 +42,21 @@ const (
 )
 
 func (s *suite) TheUserShouldWaitForTheModuleMessageToBeSubmitted(module, msgType string) {
-	s.waitForTxResultEvent(newEventMsgTypeMatchFn(module, msgType))
+	event := s.waitForTxResultEvent(newEventMsgTypeMatchFn(module, msgType))
```
Olshansk, line 45: Should this not be after the `msgType == ...`?

e2e/tests/session_steps_test.go
```diff
@@ -42,7 +42,21 @@ const (
...
+
+	// Currently the proof submission fee may be higher than the reward amount,
```
Olshansk, line 47: NIT: Please review and reword sentences 2-4. A bit hard to understand (see my following comment first though)

e2e/tests/session_steps_test.go
```diff
@@ -42,7 +42,21 @@ const (
...
+	// an adjustment the current balance to account for the proof submission fee
+	// is needed so validateAmountChange can assert on the reward change with the
+	// given condition.
+	if msgType == "SubmitProof" {
```
Olshansk, line 51: I see what you're doing here, but this is a "hidden side effect".

I think you should just either:
1. Add a step call `EnsureBalanceIsHighEnoughForProofSubmission`
OR
2. Update the default configs of LocalNet to give everyone high enough balances that this doesn't matter

My preference is (2).

e2e/tests/session_steps_test.go
```diff
@@ -336,3 +353,14 @@ func combineEventMatchFns(fns ...func(*abci.Event) bool) func(*abci.Event) bool
...
+// getMsgSubmitProofSenderAddress returns the sender address from the given event.
+func getMsgSubmitProofSenderAddress(event *abci.Event) string {
+	for _, attribute := range event.Attributes {
+		if attribute.Key == "sender" {
+			return attribute.Value
+		}
+	}
+
+	return ""
+}
```
Olshansk, line 366: Optional, you can do something like this:

```go
import "golang.org/x/exp/slices"

// getMsgSubmitProofSenderAddress returns the sender address from the given event.
func getMsgSubmitProofSenderAddress(event *abci.Event) *string {
	return slices.FindFunc(event.Attributes, func(attribute abci.EventAttribute) bool {
		return attribute.Key == "sender"
	})
```

testutil/integration/app.go
```diff
@@ -591,6 +594,19 @@ func NewCompleteIntegrationApp(t *testing.T) *App {
 	err = bankKeeper.MintCoins(integrationApp.sdkCtx, apptypes.ModuleName, moduleBaseMint)
 	require.NoError(t, err)
 
+	// Fund the supplier operator account to be able to submit proofs
+	err = bankKeeper.SendCoinsFromModuleToAccount(
```
Olshansk, line 598: Can you create a helper called "fundAccount" that mints new tokens in the banker module and sends it over?

x/proof/keeper/msg_server_submit_proof.go
```diff
@@ -149,3 +155,40 @@ func (k msgServer) SubmitProof(
...
+		return types.ErrProofNotEnoughFunds.Wrapf(
+			"account has %s, but the proof submission fee is %s",
+			accBalance, proofSubmissionFee,
+		)
+	}
+
+	// Deduct the proof submission fee from the supplier operator's balance.
+	proofSubmissionFeeCoins := cosmostypes.NewCoins(*proofSubmissionFee)
+	err = k.bankKeeper.SendCoinsFromAccountToModule(ctx, supplierOperatorAccAddress, types.ModuleName, proofSubmissionFeeCoins)
+	if err != nil {
```
Olshansk, line 186:
```suggestion
	if err = k.bankKeeper.SendCoinsFromAccountToModule(ctx, supplierOperatorAccAddress, types.ModuleName, proofSubmissionFeeCoins); err != nil {
```

Olshansk (approved):
Just a couple small changes but approving pre-emtively. Great work again!

e2e/tests/session_steps_test.go
red-0ne: This is actually not about having enough funds.

It is to make the step `Then the account balance of "supplier1" should be "420" uPOKT "more" than before` succeed without making it more complicated
* `fee > reward` -> `lower balance`
* `proof not required` -> `no fee`
* step needing to know about previous ones
* ...

There's actually an elegant solution without side effect.
Take a snapshot of the balance after the proof submission.

e2e/tests/session_steps_test.go
red-0ne: There seem to be no `FindFunc` in slices. I'll be using `slices.IndexFunc`

PR #743, opened 2024-08-16
[Application] feat: app stake transfer

bryanchriswhite:
## Summary

- Add `MsgTransferApplicationStake`.
- Implement application stake transfer to new "beneficiary" application (unstakes original application).

## Dependents

- #788
- #789

## Issue

- #657

## Type of change

Select one or more:

- [x] New feature, functionality or library
- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

**Documentation changes** (only if making doc changes)
- [ ] `make docusaurus_start`; only needed if you make doc changes

**Local Testing** (only if making code changes)
- [x] **Unit Tests**: `make go_develop_and_test`
- [x] **LocalNet E2E Tests**: `make test_e2e`
- See [quickstart guide](https://dev.poktroll.com/developer_guide/quickstart) for instructions

**PR Testing** (only if making code changes)
- [x] **DevNet E2E Tests**: Add the `devnet-test-e2e` label to the PR.
    - **THIS IS VERY EXPENSIVE**, so only do it after all the reviews are complete.
    - Optionally run `make trigger_ci` if you want to re-trigger tests without any code changes
    - If tests fail, try re-running failed tests only using the GitHub UI as shown [here](https://github.com/pokt-network/poktroll/assets/1892194/607984e9-0615-4569-9452-4c730190c1d2)

## Sanity Checklist

- [x] I have tested my changes using the available tooling
- [x] I have commented my code
- [x] I have performed a self-review of my own code; both comments & source code
- [ ] I create and reference any new tickets, if applicable
- [x] I have left TODOs throughout the codebase, if applicable

Issue #657, opened 2024-07-04
[Utility][Morse Parity] Staked App Transfer


Olshansk:
## Objective

Achieve Morse parity w.r.t app stake transfer.

## Origin Document

- [PIP-35: Introduce a secure way to transfer a staked app to a new account ](https://forum.pokt.network/t/pip-35-introduce-a-secure-way-to-transfer-a-staked-app-to-a-new-account/4806)
- [Allow an AppMsgStake tx to change the address of a staked app ](https://github.com/pokt-network/pocket-core/pull/1585)

## Goals

- Allow staked apps to transfer their stake without needing to unbond
- Get to Morse parity

## Deliverables

- [ ] Re-implemnt the feature in `pocket-core` in Morse

## Non-goals / Non-deliverables

- Introducing any new functionality above what exists in Morse already

## General deliverables

- [ ] **Comments**: Add/update TODOs and comments alongside the source code so it is easier to follow.
- [ ] **Testing**: Add new tests (unit and/or E2E) to the test suite.
- [ ] **Makefile**: Add new targets to the Makefile to make the new functionality easier to use.
- [ ] **Documentation**: Update architectural or development READMEs; use [mermaid](https://mermaid-js.github.io/mermaid/) diagrams where appropriate.

---

**Creator**: @Olshansk 
**Co-Owners**: @moatus

Olshansk:
@adshmh Let me know when you have cycles to pick this one up. I'll provide some more details, and can help with the design of the user-interface, but it'll involve a few things like:

1. Using ignite to scaffold new types + transactions -> committing those
2. Implementing the business logic, CLI component, etc...
3. Documentation for the CLI + docs
4. Unit + Integration + E2E tests
5. Ensuring parity with more

Let's pair when you get around to it.

PR #788, opened 2024-09-04
[Code Health] refactor: rename `ApplicationTransfer` msgs

bryanchriswhite:
## Summary

- Renaming the application module `TransferApplicationStake` messages to `TransferApplication`

## Depends on 

- #743

## Dependents

- #789

## Issue

- #657

## Type of change

Select one or more:

- [ ] Bug fix
- [x] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

## Sanity Checklist


Olshansk (changes requested):
@bryanchriswhite I like the change from `TransferAppStake` to `TransferApp` but I don't think we should conflate `ParamApplicationTransferAndUnbondingPeriodSessions`.

We need an unbonding period to avoid weak subjectivity, but transferring applications should be g2g at the beginning of the next session. 

Lmk if this doesn't align with what we discussed offline yesterday.

bryanchriswhite:
> We need an unbonding period to avoid weak subjectivity, but transferring applications should be g2g at the beginning of the next session.

You are correct. :+1: Thanks for calling this out and for adding more perspective! :pray:

Olshansk (changes requested):
@bryanchriswhite Per your [last comment](https://github.com/pokt-network/poktroll/pull/788#issuecomment-2330832035), can you update the PR title and description please. I'm not sure what to review atm.

red-0ne:
> We need an unbonding period to avoid weak subjectivity, but transferring applications should be g2g at the beginning of the next session.

@bryanchriswhite, @Olshansk. The reason why the unbonding period (or similar) needs to be applied here is that if the source application has requested work, then it should pay for it when settling pending claims.

Otherwise the application would no longer exist to have its stake burned. Unless we keep some kind of track in order to burn the destination app stake, which I believe is more complicated than enforcing an unbonding period.

The unbonding period logic has all the pre-requisites to avoid missing the last burn:
* Marks the application as inactive for the upcoming session (no work could be requested after) [x/session/keeper/session_hydrator.go#L149](https://github.com/pokt-network/poktroll/blob/main/x/session/keeper/session_hydrator.go#L149)
* Keeps the application hanging until the next `SettlePendingClaims`.

bryanchriswhite:
> Otherwise the application would no longer exist to have its stake burned. Unless we keep some kind of track in order to burn the destination app stake, which I believe is more complicated than enforcing an unbonding period.

I've updated the transfer period to include the blocks between session end and proof window close in #789. I.e., the destination is created and the source removed after settlement of the last "active" session. It's quite similar to how the unbonding period works; `Application#IsActive()` is sensitive to the transfer session end now as well.

Olshansk (approved):

red-0ne (approved):
Clean and focused changes :sparkles:

PR #789, opened 2024-09-04
[Application] chore: implement app transfer period

bryanchriswhite:
## Summary

- Reconcile the need for a transfer period with #743.
- Ensure the app transfer message handler returns grpc status errors consistently.
- Improve some shared E2E tests.
- Implement `ApplicationIntegrationSuite` test suite.
- Port app transfer E2E tests to integration tests.

## Depends on

- #827 
- #788 
- #743

## Issue

- #657

## Type of change

Select one or more:

- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

## Sanity Checklist


x/application/types/application.go
```diff
@@ -13,13 +13,22 @@ func (s *Application) IsUnbonding() bool {
...
 // given query height.
 // An application that has not submitted an unstake message is always active.
 // An application that has submitted an unstake message is active until the end of
 // the session containing the height at which unstake message was submitted.
 func (s *Application) IsActive(queryHeight int64) bool {
-	return !s.IsUnbonding() || uint64(queryHeight) <= s.UnstakeSessionEndHeight
+	return !s.IsUnbonding() || !s.HasPendingTransfer() ||
+		uint64(queryHeight) <= s.GetUnstakeSessionEndHeight() ||
+		uint64(queryHeight) <= s.GetPendingTransfer().GetSessionEndHeight()
 }
```
Olshansk, line 32:
```go
// IsActive returns whether the application is allowed to request services at the
// given query height.
// An application that has not submitted an unstake message is always active.
// An application that has submitted an unstake message is active until the end of
// the session containing the height at which unstake message was submitted.
// An application that has a pending transfer is active until the end of the session
// containing the height at which the transfer was initiated.
func (s *Application) IsActive(queryHeight int64) bool {
    isUnstakingButStillActive := !s.IsUnbonding() || uint64(queryHeight) <= s.GetUnstakeSessionEndHeight()
    isTransferingButStillActive := !s.HasPendingTransfer() || uint64(queryHeight) <= s.GetPendingTransfer().GetSessionEndHeight()

    return isUnstakingButStillActive || isTransferingButStillActive
}
```

x/application/module/abci.go
```diff
@@ -20,5 +20,9 @@ func EndBlocker(ctx sdk.Context, k keeper.Keeper) error {
 		return err
 	}
 
+	if err := k.EndBlockerTransferApplication(ctx); err != nil {
```
Olshansk, line 23: 1. Do we know whether Tokenomics end blocker runs first? We need to settle the session before transferring.
2. Need to find a way to enforce & document this (in the code)

x/application/keeper/transfer_applications.go
```diff
@@ -0,0 +1,132 @@
...
+	"fmt"
+
+	cosmostypes "github.com/cosmos/cosmos-sdk/types"
+
+	"github.com/pokt-network/poktroll/telemetry"
+	"github.com/pokt-network/poktroll/x/application/types"
+	"github.com/pokt-network/poktroll/x/shared"
+)
+
+func (k Keeper) EndBlockerTransferApplication(ctx context.Context) error {
```
Olshansk, line 14: godoc plz

x/application/keeper/transfer_applications.go
```diff
@@ -0,0 +1,132 @@
...
+		func() bool { return isSuccessful },
+	)
+
+	sdkCtx := cosmostypes.UnwrapSDKContext(ctx)
+	currentHeight := sdkCtx.BlockHeight()
+	sharedParams := k.sharedKeeper.GetParams(sdkCtx)
+	logger := k.Logger().With("method", "EndBlockerTransferApplication")
+
+	// Only process application transfers at the end of the session.
+	// TODO_CONSIDER: refactoring this logic into a shared function, `IsCurrentSessionEndHeight`.
```
Olshansk, line 28: Let's do it. It'll read nicely

x/application/keeper/transfer_applications.go
```diff
@@ -0,0 +1,132 @@
...
+	// TODO_IMPROVE: Use an index to iterate over the applications that have initiated
+	// the transfer action instead of iterating over all of them.
+	var transferredSrcApps []types.Application
+	for _, srcApp := range k.GetAllApplications(ctx) {
+		// Ignore applications that have not initiated the transfer action.
+		if !srcApp.HasPendingTransfer() {
+			continue
+		}
+
+		transferEndHeight := types.GetApplicationTransferEndHeight(&sharedParams, &srcApp)
```
Olshansk, line 44:
```suggestion
		// Ignore applications that have initiated a transfer but still active
		transferEndHeight := types.GetApplicationTransferEndHeight(&sharedParams, &srcApp)
```

x/application/keeper/transfer_applications.go
```diff
@@ -0,0 +1,132 @@
...
+	// Ensure destination application was not staked during transfer period.
+	_, isDstFound := k.GetApplication(ctx, srcApp.GetPendingTransfer().GetDestination())
+	if isDstFound {
+		return types.ErrAppDuplicateAddress.Wrapf(
+			"destination application (%s) was staked during transfer period of application (%s)",
+			srcApp.GetPendingTransfer().GetDestination(), srcApp.GetAddress(),
+		)
+	}
+
+	dstApp := srcApp
```
Olshansk, line 103:
```suggestion
	dstApp := srcApp // intentional copy
```

tools/scripts/params/shared_all.json
```diff
@@ -5,12 +5,14 @@
         "@type": "/poktroll.shared.MsgUpdateParams",
         "authority": "pokt10d07y265gmmuvt4z0w9aw880jnsr700j8yv32t",
         "params": {
-          "num_blocks_per_session": "4",
+          "num_blocks_per_session": "10",
```
Olshansk, line 8: This can have impact on E2E tests. Dangerous to just change in the middle of this large PR.

proto/poktroll/application/types.proto
```diff
@@ -31,10 +31,17 @@ message Application {
...
 
 // UndelegatingGatewayList is used as the Value of `pending_undelegations`.
 // It is required to store a repeated list of strings as a map value.
 message UndelegatingGatewayList {
   repeated string gateway_addresses = 2 [(cosmos_proto.scalar) = "cosmos.AddressString", (gogoproto.nullable) = false];
+}
+
+// PendingTransfer is used to store the details of a pending transfer.
+message PendingTransfer {
+  string destination = 2 [(cosmos_proto.scalar) = "cosmos.AddressString"];
```
Olshansk, line 45:
```suggestion
  string destination_address = 2 [(cosmos_proto.scalar) = "cosmos.AddressString"];
```

proto/poktroll/application/types.proto
```diff
@@ -31,10 +31,17 @@ message Application {
...
 }
 
 // UndelegatingGatewayList is used as the Value of `pending_undelegations`.
 // It is required to store a repeated list of strings as a map value.
 message UndelegatingGatewayList {
   repeated string gateway_addresses = 2 [(cosmos_proto.scalar) = "cosmos.AddressString", (gogoproto.nullable) = false];
+}
+
+// PendingTransfer is used to store the details of a pending transfer.
+message PendingTransfer {
```
Olshansk, line 44:
```suggestion
message PendingApplicationTransfer {
```

x/application/keeper/msg_server_transfer_application_stake.go
```diff
@@ -13,52 +17,69 @@ import (
...
 
-	// Remove the transferred app from the store
-	k.RemoveApplication(ctx, srcApp.GetAddress())
-	logger.Info(fmt.Sprintf("Successfully removed the application: %+v", srcApp))
+	if err := sdkCtx.EventManager().EmitTypedEvent(&types.EventTransferBegin{
+		SourceAddress:      srcApp.GetAddress(),
+		DestinationAddress: srcApp.GetPendingTransfer().GetDestination(),
+		SourceApplication:  &srcApp,
+	}); err != nil {
+
```
Olshansk, line 81: What then!?

e2e/tests/stake_app_transfer.feature
```diff
@@ -2,21 +2,37 @@ Feature: App Stake Transfer Namespace
...
         And the user verifies the "application" for account "app2" is not staked
         And the account balance of "app2" should be "0" uPOKT "more" than before
+        # Cleanup for other tests
+        # TODO(#): Until a network state reset API is implemented, we SHOULD
+        # manually unstake the applications to mitigate failures in other features
+        # which may otherwise make false assumptions about the starting state.
+        When the user successfully unstakes a "application" from the account "app3"
+
+#    TODO_TEST: Scenario: User cannot start an Application stake transfer from Application which has a pending transfer
+#    TODO_TEST: Scenario: Application stake transfer fails if the destination Application stakes before the transfer period elapses
```
Olshansk, line 36: TODO_TEST: If two applications start a transfer to the same destination, one of them fails and the other succeeds?

Would love to have this test done in this PR.

Olshansk (changes requested):

tools/scripts/params/shared_all.json
bryanchriswhite: This is not used anywhere except for a single makefile target (`params_update_shared_all`) [1]. 

My intention here was to updating it to match the (newer) default. These JSON files exist to support their respective makefile targets and to serve as a reference. They've historically contained the default values, which is convenient when debugging new governance params or their E2E tests (i.e. easy reset).

image
> [1] Search results for the regex `params/.+/\.json`

x/application/types/application.go
bryanchriswhite: Nice catch - thanks! :raised_hands:

x/application/keeper/msg_server_transfer_application_stake.go
bryanchriswhite: :sweat_smile: log!

e2e/tests/stake_app_transfer.feature
bryanchriswhite: :ok_hand: :100:

x/application/module/abci.go
bryanchriswhite: Great question! :thinking: 

_TL;DR, application module end blockers currently run first; we would need to change the order of the module names in [`endBlockers`](https://github.com/pokt-network/poktroll/blob/main/app/app_config.go#L178). Just need to make sure that doesn't cause any other issue(s)._

[Cosmos-sdk docs](https://docs.cosmos.network/main/build/building-modules/beginblock-endblock) say:
> It is possible for developers to define the order of execution between the BeginBlocker/EndBlocker functions of each of their application's modules via the module's manager SetOrderBeginBlocker/SetOrderEndBlocker methods.

We have an `endBlockers` var defined in [`app_config.go:178`](https://github.com/pokt-network/poktroll/blob/main/app/app_config.go#L178):
image

Which is passed into a `ModuleConfig` [below](https://github.com/pokt-network/poktroll/blob/main/app/app_config.go#L249) that is passed to `ModuleManager.SetOrderEndBlockers()` [in the cosmos-sdk runtime by `App#Load()`](https://github.com/cosmos/cosmos-sdk/blob/v0.50.7/runtime/app.go#L131)

We could enforce this with a script that runs in CI that knows about the constraints but it may be just as effective to leave comments in/around the `endBlockers` assignment articulating the constraints.

x/application/keeper/transfer_applications.go
```diff
@@ -0,0 +1,130 @@
...
+	// TODO_IMPROVE: Use an index to iterate over the applications that have initiated
+	// the transfer action instead of iterating over all of them.
+	var transferredSrcApps []types.Application
+	for _, srcApp := range k.GetAllApplications(ctx) {
+		// Ignore applications that have not initiated the transfer action.
+		if !srcApp.HasPendingTransfer() {
+			continue
+		}
+
+		transferEndHeight := srcApp.GetPendingTransfer().GetSessionEndHeight()
```
bryanchriswhite, line 42:
```suggestion
		// Ignore applications that have initiated a transfer but still active
		transferEndHeight := srcApp.GetPendingTransfer().GetSessionEndHeight()
```

tools/scripts/params/shared_all.json
```diff
@@ -5,12 +5,14 @@
...
           "grace_period_end_offset_blocks": "1",
           "claim_window_open_offset_blocks": "1",
           "claim_window_close_offset_blocks": "4",
           "proof_window_open_offset_blocks": "0",
-          "proof_window_close_offset_blocks": "4"
+          "proof_window_close_offset_blocks": "4",
+          "supplier_unbonding_period_sessions": "4",
+          "application_transfer_and_unbonding_period_sessions": "4"
```
bryanchriswhite, line 15:
```suggestion
          "application_unbonding_period_sessions": "4"
```

Olshansk (commented):
@red-0ne Do you mind taking a stab at this first given that you're closer to the problem?

e2e/tests/stake_app_transfer.feature
```diff
@@ -2,21 +2,87 @@ Feature: App Stake Transfer Namespace
...
+        # Stake with 1 uPOKT more than the current stake used in genesis to make
+        # the transaction succeed.
+        And the account "app2" has a balance greater than "1000070" uPOKT
         And the user successfully stakes a "application" with "1000070" uPOKT for "anvil" service from the account "app2"
+        # Begin transfer app1 --> app3
+        When the user transfers the "application" stake from account "app1" to account "app3"
+        Then the user should be able to see standard output containing "txhash:"
+        And the user should be able to see standard output containing "code: 0"
+        And the pocketd binary should exit without error
+        # Begin transfer app3 --> app3
```
red-0ne, line 54: Align comment with the step below.

e2e/tests/stake_app_transfer.feature
```diff
@@ -2,21 +2,87 @@ Feature: App Stake Transfer Namespace
...
+        # change after the transfer period
+        And the user verifies the "application" for account "app2" is not staked
+        And the account balance of "app2" should be "0" uPOKT "more" than before
+        # Cleanup for other tests
+        # TODO(#): Until a network state reset API is implemented, we SHOULD
+        # manually unstake the applications to mitigate failures in other features
+        # which may otherwise make false assumptions about the starting state.
+        And the user successfully unstakes a "application" from the account "app3"
+
+    Scenario: Only one Application transfer with a given destination address in the same session will succeed
```
red-0ne, line 40: This is the reason why I think we should allow transferring to an existing application.

Currently, having multiple transfers to the same destination address is non deterministic (depends on how the application list is iterated over in `transfer_applications.go`)

And it's being more problematic as the outcome is not known until the end of the transfer period.

If we agree on allowing multiple transfers/transfers to existing apps, let's just leave TODO.

red-0ne (commented):
The feature looks great! Worth the time spent on it :+1: 

Left some comments but mainly concerned about multiple transfers:
* Transfer of the same app to different dst addresses (we should only allow the first one)
* Transfer of multiple apps to the same address (should be deterministic)

x/application/keeper/transfer_applications.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+
+	dstApp := srcApp // intentional copy
+	dstApp.Address = srcApp.GetPendingTransfer().GetDestinationAddress()
+	dstApp.PendingTransfer = nil
+
+	sdkCtx := cosmostypes.UnwrapSDKContext(ctx)
+	sessionEndHeight := k.sharedKeeper.GetSessionEndHeight(ctx, sdkCtx.BlockHeight())
+
+	srcApp.PendingTransfer = &types.PendingApplicationTransfer{
```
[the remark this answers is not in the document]
bryanchriswhite: :thinking: probably copypasta. :sweat_smile:

bryanchriswhite:
> * Transfer of the same app to different dst addresses (we should only allow the first one)
**TL;DR, How strongly do we feel about the need for the requirements below? The app transfer UX could be improved and the design could be simplified if we support transferring to an existing application.**

As discussed elsewhere, I disagree. From a UX perspective, I think it's convenient for the last transfer message for a given source application address to "win". I'm imagining an application operator who copy/pastes the wrong address, for example. They will have to pay tx fees again but would not have to wait a whole session to correct the mistake.

> * Transfer of multiple apps to the same address (should be deterministic)

I agree with you on this one. :sweat_smile: Unfortunately (again, as discussed elsewhere), the current requirements/design can't support rejecting such transfers upon initiation. It seems to me that we would have to introduce a new map into the application module's state which is tracking pending transfers in order to do so.

---

### Requirements
1. Destination MUST NOT exist at transfer begin
1. Destination MUST NOT be a pending transfer destination

### Current Design
```mermaid
flowchart TD

%% subgraph r[Requirements]
%%     newdst[Destination MUST NOT exist]
%%     penddst[Destination MUST NOT be a pending transfer destination]
%% end

%% subgraph d[Design]
    subgraph appm["Application Module"]
        %% kv[KV Store]
        subgraph th[Transfer Application Logic]
            th_begin[MsgTransferApplication Handler]
            th_end[EndBlockerTransferApplication]
        end
        
    %% end

    subgraph app_map["Applications Map"]
        srcaddr["address (source)"]
        subgraph pt[pending_transfer]
            dstaddr[destination_address]
            seh[session_end_height]
        end
    end
end

%% kv --> app_map
th_begin --> |Add/update pending_transfer field| pt
th_begin -.-> |Ensure destination does not exist| srcaddr
th_end -.-> |Ensure transfer destination does not exist| app_map
th_end ---> |Remove transfer source app| app_map
th_end ---> |Create transfer destination app| app_map

%% class penddst unresolved
class th_begin,th_end,pt change
%% classDef unresolved fill:#d373;
classDef change fill:#353,color:#ccc;
```

### Alternative 1 (satisfy current requirements)
```mermaid
flowchart TD

%% subgraph r[Requirements]
%%     newdst[Destination MUST NOT exist]
%%     penddst[Destination MUST NOT be an existing pending transfer destination]
%% end

%% subgraph d[Design]
    subgraph appm["Application Module"]
        %% kv[KV Store]
        subgraph th[Transfer Application Logic]
            th_begin[MsgTransferApplication Handler]
            th_end[EndBlockerTransferApplication]
        end

    subgraph appp["Application (proto)"]
        srcaddr["address (source)"]
        subgraph pt[pending_transfer]
            srcaddr_[source_address]
            dstaddr[destination_address]
            seh[session_end_height]
        end
    end

        subgraph ptmap[Pending Transfers Map]
            %% ptm_dst_k[destination_address]
                        ptm_dst_idx(["destination_address index (optional)"])

            subgraph ptm_pt[pending_transfer]
                ptm_src[source_address]
                ptm_dst[destination_address]
                ptm_seh[session_end_height]
            end

            ptm_src_k["source_address (key/primary idx)"]
        end
    %% end


end

ptm_src_k --> ptm_pt

%% kv --> appp
%% th -.-> |Ensure destination is not alredy pending| ptm_dst


th_begin -.-> |Ensure destination is not already pending| ptm_dst_idx
th_begin -----> |Add/update pending_transfer| ptmap
th_begin -..-> |Ensure destination does not exist| srcaddr
th_begin --> |Add/update pending_transfer field| pt

th_end -------> |Remove pending_transfer| ptm_src_k
th_end ----> |Remove transfer source app| appp
th_end ----> |Create transfer destination app| appp

ptm_dst_idx -.-> ptm_dst

class th_begin,th_end,pt,ptmap change
classDef change fill:#353,color:#ccc;
```

### Alternative 2 (drop requirements)
```mermaid
flowchart TD

%% subgraph r[Requirements]
%%     newdst[Destination MUST NOT exist]
%%     penddst[Destination MUST NOT be an existing pending transfer destination]
%% end

%% subgraph d[Design]
    subgraph appm["Application Module"]
        %% kv[KV Store]
        subgraph th[Transfer Application Logic]
            th_begin[MsgTransferApplication Handler]
            th_end[EndBlockerTransferApplication]
        end
    %% end

    subgraph app_map["Applications Map"]
        srcaddr["address (source)"]
        subgraph pt[pending_transfer]
            dstaddr[destination_address]
            seh[session_end_height]
        end
    end

    if_dst_exists{if destination exists}
end

%% kv --> app_map
%% th_begin ---> |Get transfer source app| app_map
th_begin --> |Add/update pending_transfer field| pt

th_end --> if_dst_exists
if_dst_exists --> |YES:<br>Merge with destination app| app_map
if_dst_exists --> |NO:<br>Create destination app| app_map

%% th_end -.-> |Merge if destination exists| srcaddr
th_end --> |Remove source app| app_map
%% th_end --> |Create destination app| app_map

class th_begin,th_end,pt change
classDef change fill:#353,color:#ccc;
```

red-0ne:
You convinced me about overriding destination address on a transfer.

W.r.t. transfer determinism, I think these diagrams sum-up our alternatives.

To me Alt2 (dropping the non-existing dst requirement) is the way to go.

x/application/keeper/msg_server_transfer_application_stake.go
```diff
@@ -13,52 +17,65 @@ import (
...
-		return nil, types.ErrAppNotFound.Wrapf("source application %q not found", msg.GetSourceAddress())
+		return nil, status.Error(
+			codes.InvalidArgument,
+			types.ErrAppNotFound.Wrapf("source application (%s) not found", msg.GetSourceAddress()).Error(),
+		)
 	}
 
 	// Ensure source application is not already unbonding.
-	// TODO_TEST: Add E2E coverage to assert that an unbonding app cannot be transferred.
 	if srcApp.IsUnbonding() {
```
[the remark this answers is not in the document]
bryanchriswhite: > Although I believe it's safe to change the transfer destination (and update the pending transfer height), it's worth ensuring that no multiple transfer events will be emitted.

1. The `session_end_height` in the pending transfer data structure would not change (same session).
2. As discusses elsewhere w.r.t. transfer events, a different (e.g. `EventTransferUpdated`) event will be emitted (as opposed to `EventTransferBegin`) when updating an existing pending transfer.

Olshansk:
@bryanchriswhite 

Re `Alternative 1 (satisfy current requirements)` - I haven't done a deep dive into it (yet) but from a high-level, would like to avoid so many branches if possible.

Re `Transfer of the same app to different dst addresses (we should only allow the first one)` - Override makes sense to me. Just make sure to document this clearly in the docs & CLI.

Re `Transfer of multiple apps to the same address (should be deterministic)` - If it already exists, just fail the transfer. This can (realistically) only happen if the same operator anciently tried to do:

Tx1: Transfer A -> C
Tx2: Transfer B -> C

Assume Tx1 & Tx2 will be applied by the block proposer in some random order. During the actual transfer, the second one will fail because `C` is already an application.

----

I'm not sure what `Merge` means in your diagram so lmk if my responses are sufficient.

bryanchriswhite:
@Olshansk are you saying we should stick with the current design and that the suboptimal UX is fine for now? 

> Re `Transfer of the same app to different dst addresses (we should only allow the first one)` - Override makes sense to me. Just make sure to document this clearly in the docs & CLI.

👍

> Re `Transfer of multiple apps to the same address (should be deterministic)` - If it already exists, just fail the transfer. This can (realistically) only happen if the same operator anciently tried to do:
> 
> Tx1: Transfer A -> C Tx2: Transfer B -> C
> 
> Assume Tx1 & Tx2 will be applied by the block proposer in some random order. During the actual transfer, the second one will fail because `C` is already an application.

This seems accurate to me. For clarity, which transfer is the "second one" is not well defined and not known until transfer end (not necessarily the second transfer message which was broadcast). 

**The issue at hand is that "just fail[ing] the transfer" can only happen AFTER the transfer session has ended under the current design, as opposed to during the offending transfer begin. The shortest path to rectify this while adhering to the requirements that in questioning would be what I outlined at alternative 1, which nobody seems keen on.**

> I'm not sure what `Merge` means in your diagram so lmk if my responses are sufficient.

"Merging" applications here would mean to sum the stakes and take the union of the sets of distinct delegations.

Olshansk:
@bryanchriswhite I'm having a hard time providing a direct response to your question / comment, so am following up in a different way.

Lmk if the requirements + mermaid diagram below answers your question.

---

## Requirements

1. `Enable override` - The last successful pending transfer is the one to be attempted
2. `Avoid merging` - Avoid "merging app" logic to lower complexity (for the time being)
3. `Non-determinism is okay` - It is hard to account for all edge cases and that's fine (for the time being)

## Example Flow.

```mermaid
timeline
    section Session 1 <br> [1,10]
      Any block in [1,10]:
        Stake Application A:
        Stake Application B:
        Stake Application C:
        Stake Application D
    section Session 2 <br> [11,20]
      Block 11: Apps A, B, C & D staked to use onchain services
      Block 12:
        Transfer A -> E (pending):
        Transfer B -> D <br> (fails immediately because D is staked)
      Block 13:
        Transfer A -> F <br> (overrides A -> E <br> and pending):
        Transfer C -> G (pending):
        Transfer D -> G (pending)
    section Session 3 <br> [21,30]
        Block 21 :
            App A unstaked and <br> App F staked (successful transfer) :
            App B staked (unchanged):
            App G staked <br> (using either C or D):
            One of App C or D are unchanged (non-deterministic)
```

Olshansk:
Alt solution using merging apps

```mermaid
timeline
    section Session 1 <br> [1,10]
      Any block in [1,10]:
        Stake Application A:
        Stake Application B:
        Stake Application C:
        Stake Application D
    section Session 2 <br> [11,20]
      Block 11: Apps A, B, C & D staked to use onchain services
      Block 12:
        Transfer A -> E (pending):
        Transfer B -> D <br> (fails immediately because D is staked)
      Block 13:
        Transfer A -> F <br> (overrides A -> E <br> and pending):
        Transfer C -> G (pending):
        Transfer D -> G (pending)
    section Session 3 <br> [21,30]
        Block 21 :
            App A unstaked and <br> App F staked (successful transfer) :
            App B staked (unchanged):
            App C unstaked:
            App D unstaked:
            App G staked <br> (using sum C AND D)
```

Olshansk:
Open questions remain related to user experience:
1. **Option 1 is in the first image**: non-deterministic and forces waiting until the next session for failure
2. **Option 2 is in the second image**: summation of stakes
3. **Extension of option 2(not shown)**: let users specify amount of stake to transfer & implicitly transfer all stake (default to option 2) if not specified


Screenshot 2024-09-11 at 10 51 27 AM

Screenshot 2024-09-11 at 10 50 54 AM

fredteumer:
My vote here is for **Option 2** for the following reasons:
a. Option 1 - I don't like the idea of non-determinism for the transfer operation -- it _feels_ scary to me not having 100% certainty the outcome that will happen.

b. Option 3 - I don't love the idea of a partial app-stake transfer. It feels weird. IMO, you're either staked or you're not but I'm not 100% of all the instruments and restaking potential this does for the network if left enabled. It also doesn't encourage new app stakes necessarily -- if you make a big stake, you can break it down into N smaller chunks, as opposed to needing to create more apps to stake and theoretically lock up more tokens.

c. Option 2 seems to mimic the existing behavior: have concrete stakes, and allow for consolidation of stakes which I definitely can see as a valid use case. Also gives concrete outcomes for users so they know exactly what they're going to get.

Let me know if you need any other thoughts here.

Olshansk:
@fredteumer Thanks for the feedback!

@bryanchriswhite Let's go with option (2) like Fred called out. It keeps option (3) as a potential for the future.

For posterity, I'll say that my vote was for option (1) which is actually closer to how things work today (there are edge cases unacounted for), but the feedback here is super valuable.

e2e/tests/init_test.go
```diff
@@ -511,22 +532,51 @@ func (s *suite) TheUserWaitsForTheSupplierForAccountUnbondingPeriodToFinish(accN
...
+			return app.IsUnbonding()
+		}
+	case "transfer":
+		msgType = "TransferApplication"
+		isAppInState = func(application *apptypes.Application) bool {
+			return application.HasPendingTransfer()
+		}
+	}
+
+	s.waitForTxResultEvent(newEventMsgTypeMatchFn("application", msgType))
```
bryanchriswhite, line 556: Revert...

e2e/tests/init_test.go
```diff
@@ -409,10 +416,24 @@ func (s *suite) TheUserVerifiesTheForAccountIsNotStaked(actorType, accName strin
...
+	require.Truef(s, ok, "account %s of type %s SHOULD be staked", accName, actorType)
+	require.Greaterf(s, int64(stakeAmount), minActorStake, "account %s stake is not greater than minimum %d", accName, minActorStake)
 	s.scenarioState[accStakeKey(actorType, accName)] = stakeAmount // save the stakeAmount for later
 }
 
+// TODO_UP_NEXT(@bryanchriswhite, #612): replace with the actual minimum stake values.
+func (s *suite) getMinStake(actorType string) int64 {
+	return 0
+}
+
```
bryanchriswhite, line 436: Revert? Not being used in this commit or to test this feature but may be useful in the future for other scenarios.

e2e/tests/init_test.go
```diff
@@ -786,6 +836,28 @@ func (s *suite) getApplicationUnbondingHeight(accName string) int64 {
 	return unbondingHeight
 }
 
+// getApplicationTransferEndHeight returns the height at which the application will be transferred to the destination.
+func (s *suite) getApplicationTransferEndHeight(accName string) int64 {
```
bryanchriswhite, line 840: Revert...

tests/integration/application/application_transfer_test.go
```diff
@@ -0,0 +1,262 @@
...
+	require.Truef(s.T(), ok, "insufficient pre-generated accounts available")
+
+	//sendToAppMsg := &banktypes.MsgSend{
+	//	FromAddress: integration.FaucetAddrStr,
+	//	ToAddress:   appAccount.Address.String(),
+	//	// TODO_IN_THIS_PR: move amount to a constant.
+	//	Amount: cosmostypes.NewCoins(cosmostypes.NewInt64Coin(volatile.DenomuPOKT, 100000000)),
+	//}
+	//s.GetApp().RunMsg(s.T(), sendToAppMsg)
+
```
bryanchriswhite, line 200: Remove...

tools/scripts/params/shared_all.json
```diff
@@ -5,12 +5,14 @@
...
-          "num_blocks_per_session": "4",
+          "num_blocks_per_session": "10",
           "grace_period_end_offset_blocks": "1",
           "claim_window_open_offset_blocks": "1",
           "claim_window_close_offset_blocks": "4",
           "proof_window_open_offset_blocks": "0",
-          "proof_window_close_offset_blocks": "4"
+          "proof_window_close_offset_blocks": "4",
+          "supplier_unbonding_period_sessions": "4",
+          "application_unbonding_period_sessions": "4"
```
bryanchriswhite, line 15:
```suggestion
          "supplier_unbonding_period_sessions": "1",
          "application_unbonding_period_sessions": "1"
```

x/application/keeper/transfer_applications.go
```diff
@@ -0,0 +1,162 @@
...
+			continue
+		}
+
+		// Ignore applications that have initiated a transfer but still active
+		transferEndHeight := types.GetApplicationTransferHeight(&sharedParams, &srcApp)
+		if sdkCtx.BlockHeight() < transferEndHeight {
+			continue
+		}
+
+		// Transfer the stake of the source application to the destination application.
```
bryanchriswhite, line 51:
```suggestion
		// Transfer the stake of the source application to the destination application and
		// merge their gateway delegations and service configs.
```

x/application/keeper/transfer_applications.go
```diff
@@ -0,0 +1,163 @@
...
+	}
+
+	isSuccessful = true
+	return nil
+}
+
+// transferApplication transfers the fields of srcApp, except for address and pending_transfer,
+// to an application whose address is the destination address of the pending transfer of srcApp.
+// If the destination application does not exist, it is created. If it does exist, the stake of
+// the destination application is incremented by the stake of the source application, and the
```
bryanchriswhite, line 78:
```suggestion
// the destination application stake is incremented by the stake of the source application, and the
```

x/application/keeper/transfer_applications.go
```diff
@@ -0,0 +1,163 @@
...
+// delegatees and service configs of the destination application are set to the union of the
+// source application's delegatees and service configs and the destination application's
+// delegatees. It is intended to be called during the EndBlock ABCI method.
```
bryanchriswhite, line 81:
```suggestion
// source application's delegatees and service configs and the destination application's
// delegatees. It is intended to be called during the EndBlock ABCI method.
```
```suggestion
// source and destination applications' delegatees and service configs. It is intended
// to be called during the EndBlock ABCI method.
```

tests/integration/application/application_transfer_test.go
```diff
@@ -0,0 +1,434 @@
...
+	require.NoError(s.T(), err)
+	require.NotNil(s.T(), balanceRes)
+
+	require.EqualValues(s.T(),
+		cosmostypes.NewInt64Coin(volatile.DenomuPOKT, appFundAmount-stakeAmount),
+		*balanceRes.GetBalance(),
+	)
+}
+
+func (s *AppTransferSuite) TestMultipleSourceToSameNonexistentDestinationSucceedsForFirst() {
```
bryanchriswhite, line 152:
```suggestion
func (s *AppTransferSuite) TestMultipleSourceToSameNonexistentDestinationMergesSources() {
```

red-0ne (commented):
Left some comments but I love the shape this is taking!

x/application/keeper/msg_server_transfer_application_stake_test.go
```diff
@@ -44,93 +48,89 @@ func TestMsgServer_TransferApplication_Success(t *testing.T) {
...
 
 	srcApp.Address = ""
 	dstApp.Address = ""
 	require.EqualValues(t, srcApp, dstApp)
 
 	// Verify that the source app was unstaked.
-	srcApp, isSrcFound = k.GetApplication(ctx, srcAddr)
+	srcApp, isSrcFound = k.GetApplication(ctx, srcBech32)
 	require.False(t, isSrcFound)
 }
```
Olshansk, line 136: Great test!

testutil/integration/suites/application.go
```diff
@@ -0,0 +1,125 @@
...
+
+	txMsgRes, err := s.GetApp().RunMsg(t, stakeAppMsg)
+	require.NoError(t, err)
+
+	return txMsgRes.(*apptypes.MsgStakeApplicationResponse)
+}
+
+// Transfer sends a MsgApplicationTransfer to begin an application transfer
+// from srcBech32 to dstDst32.
+func (s *ApplicationModuleSuite) Transfer(
```
Olshansk, line 66:
```suggestion
func (s *ApplicationModuleSuite) TransferAppStake(
```

Olshansk (changes requested):
Reviewed everything except for `tests/integration/application/application_transfer_test.go`

x/application/types/application.go
```diff
@@ -13,13 +13,24 @@ func (s *Application) IsUnbonding() bool {
 	return s.UnstakeSessionEndHeight != ApplicationNotUnstaking
 }
 
+// HasPendingTransfer returns true if the application has begun but not completed
```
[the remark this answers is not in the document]
bryanchriswhite: "Stake transfer" is not accurate. The entire application state is being transferred (incl. delegations). This is a method defined on the `Application` type, I don't see how this could be confused with other kinds of transfers.

x/application/keeper/transfer_applications.go
```diff
@@ -0,0 +1,163 @@
...
+	"fmt"
+
+	cosmostypes "github.com/cosmos/cosmos-sdk/types"
+
+	"github.com/pokt-network/poktroll/telemetry"
+	"github.com/pokt-network/poktroll/x/application/types"
+	"github.com/pokt-network/poktroll/x/shared"
+)
+
+// EndBlockerTransferApplication completes pending application transfers at the
```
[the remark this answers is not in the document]
bryanchriswhite: see https://github.com/pokt-network/poktroll/pull/789#discussion_r1772991745

x/application/keeper/transfer_applications.go
```diff
@@ -0,0 +1,163 @@
...
+		// Transfer the stake of the source application to the destination application and
+		// merge their gateway delegations and service configs.
+		if transferErr := k.transferApplication(ctx, srcApp); transferErr != nil {
+			logger.Warn(transferErr.Error())
+
+			// Application transfer failed, removing the pending transfer from the source application.
+			srcApp.PendingTransfer = nil
+			k.SetApplication(ctx, srcApp)
+
+			if err := sdkCtx.EventManager().EmitTypedEvent(&types.EventTransferError{
```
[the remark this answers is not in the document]
bryanchriswhite: see: https://github.com/pokt-network/poktroll/pull/789#discussion_r1772991745

x/application/keeper/transfer_applications.go
```diff
@@ -0,0 +1,163 @@
+package keeper
```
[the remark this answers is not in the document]
bryanchriswhite: - #790

testutil/integration/suites/application.go
bryanchriswhite: see: https://github.com/pokt-network/poktroll/pull/789#discussion_r1772991745

x/application/keeper/transfer_applications.go
```diff
@@ -0,0 +1,163 @@
...
+		if transferErr := k.transferApplication(ctx, srcApp); transferErr != nil {
+			logger.Warn(transferErr.Error())
+
+			// Application transfer failed, removing the pending transfer from the source application.
+			srcApp.PendingTransfer = nil
+			k.SetApplication(ctx, srcApp)
+
+			if err := sdkCtx.EventManager().EmitTypedEvent(&types.EventTransferError{
+				SourceAddress:      srcApp.GetAddress(),
+				DestinationAddress: srcApp.GetPendingTransfer().GetDestinationAddress(),
```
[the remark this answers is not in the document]
bryanchriswhite: Nice catch - thanks! :sunglasses:

x/application/keeper/transfer_applications.go
```diff
@@ -0,0 +1,163 @@
...
+	if !isDstFound {
+		dstApp = srcApp //intentional copy
+		dstApp.Address = srcApp.GetPendingTransfer().GetDestinationAddress()
+		dstApp.PendingTransfer = nil
+	} else {
+		srcStakeSumCoin := dstApp.GetStake().Add(*dstApp.GetStake())
+		dstApp.Stake = &srcStakeSumCoin
+
+		mergeAppDelegatees(&srcApp, &dstApp)
+		mergeAppServiceConfigs(&srcApp, &dstApp)
```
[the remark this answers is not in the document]
bryanchriswhite: Yessir - thank you!

tests/integration/application/application_transfer_test.go
```diff
@@ -0,0 +1,434 @@
...
+	})
+
+	// Assert the on-chain state shows the application 3 as NOT staked.
+	_, queryErr := s.GetAppQueryClient().GetApplication(s.SdkCtx(), s.app3Bech32)
+	require.ErrorContains(s.T(), queryErr, "application not found")
+	require.ErrorContains(s.T(), queryErr, s.app3Bech32)
+}
+
+func (s *AppTransferSuite) TestSingleSourceToNonexistentDestinationSucceeds() {
+	sharedParams := sharedtypes.DefaultParams()
```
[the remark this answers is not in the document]
bryanchriswhite: Wdyt about this for now?

```go
	// TODO_TECHDEBT(#826): Once the ParamsSuite is available, embed it in the
	// AppTransferSuite to fetch the current integration app's params instead.
	sharedParams := sharedtypes.DefaultParams()
```

x/application/keeper/transfer_applications.go
```diff
@@ -0,0 +1,163 @@
...
+
+	"github.com/pokt-network/poktroll/telemetry"
+	"github.com/pokt-network/poktroll/x/application/types"
+	"github.com/pokt-network/poktroll/x/shared"
+)
+
+// EndBlockerTransferApplication completes pending application transfers at the
+// end of the session in which they began by copying the current state of the source
+// application onto a new destination application, unstaking (removing) the source,
+// and staking (storing) the destination.
```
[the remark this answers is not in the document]
bryanchriswhite: Much easier on the eyes, thank you! :sunglasses:

x/application/keeper/transfer_applications.go
```diff
@@ -0,0 +1,163 @@
...
+	// Iterate over all applications and transfer the ones that have finished the transfer period.
+	// TODO_IMPROVE: Use an index to iterate over the applications that have initiated
+	// the transfer action instead of iterating over all of them.
+	for _, srcApp := range k.GetAllApplications(ctx) {
+		// Ignore applications that have not initiated the transfer action.
+		if !srcApp.HasPendingTransfer() {
+			continue
+		}
+
+		// Ignore applications that have initiated a transfer but still active
```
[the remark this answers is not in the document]
bryanchriswhite:
```go
// Ignore applications that have initiated a transfer but still active.
// This spans the period from the end of the session in which the transfer
// began to the end of settlement for that session.
```

x/application/keeper/transfer_applications.go
```diff
@@ -0,0 +1,163 @@
...
+// If the destination application does not exist, it is created. If it does exist, the stake of
+// the destination application stake is incremented by the stake of the source application, and the
+// delegatees and service configs of the destination application are set to the union of the
+// source and destination applications' delegatees and service configs. It is intended
+// to be called during the EndBlock ABCI method.
+func (k Keeper) transferApplication(ctx context.Context, srcApp types.Application) error {
+	logger := k.Logger().With("method", "transferApplication")
+
+	// Double-check that the source application is not unbonding.
+	// NB: This SHOULD NOT be possible because applications SHOULD NOT be able
```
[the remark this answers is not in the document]
bryanchriswhite: It already is logged as a warning. In this case, we return an error from `transferApplication()` which is logged in `EndBlockTransferApplication()`. We also emit an `EventTransferError` event in this case.

x/application/types/application.go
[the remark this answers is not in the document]
bryanchriswhite: I discovered some files which I failed to rename in #788, I've completed the renaming here.

tests/integration/application/application_transfer_test.go
```diff
@@ -0,0 +1,496 @@
...
+var (
+	appFundAmount = int64(100000000)
+	stakeAmount   = int64(100)
+
+	service1Id = "svc1"
+	service2Id = "svc2"
+	service3Id = "svc3"
+)
+
+type AppTransferSuite struct {
```
bryanchriswhite, line 31: Unexport

testutil/integration/suites/application.go
```diff
@@ -0,0 +1,202 @@
...
+	)
+
+	txMsgRes, err := s.GetApp().RunMsg(t, stakeAppMsg)
+	require.NoError(t, err)
+
+	return txMsgRes.(*apptypes.MsgStakeApplicationResponse)
+}
+
+// Transfer sends a MsgApplicationTransfer to begin an application transfer
+// from srcBech32 to dstDst32.
```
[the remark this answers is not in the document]
bryanchriswhite: My intention was to disambiguate between `cosmostypes.AccAddress` and the string representation. I agree that it's not the prettiest and gets monotonous more quickly than `addr`. Do you think `bech` is to far removed to be clear or have a better idea?

tests/integration/application/application_transfer_test.go
```diff
@@ -0,0 +1,496 @@
...
+}
+
+// shouldObserveTransferBeginEvent asserts that the transfer begin event from
+// expectedSrcApp to expectedDstAppBech32 is observed.
+func (s *appTransferTestSuite) shouldObserveTransferBeginEvent(
+	expectedSrcApp *apptypes.Application,
+	expectedDstAppBech32 string,
+) {
+	eventTypeURL := cosmostypes.MsgTypeURL(&apptypes.EventTransferBegin{})
+	transferBeginEvents := s.FilterEvents(events.NewEventTypeMatchFn(eventTypeURL))
```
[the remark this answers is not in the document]
bryanchriswhite: Nope, thank you!

tests/integration/application/application_transfer_test.go
```diff
@@ -0,0 +1,496 @@
...
+
+		// Assert application pending transfer field updated in the msg response.
+		pendingTransfer := srcApp.GetPendingTransfer()
+		require.NotNil(s.T(), pendingTransfer)
+
+		nextExpectedPendingTransfer := &apptypes.PendingApplicationTransfer{
+			DestinationAddress: expectedDstBech32,
+			SessionEndHeight:   uint64(sessionEndHeight),
+		}
+		if expectedPendingTransfer != nil {
```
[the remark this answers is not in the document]
bryanchriswhite: I added a comment and simplified:

```go
		// Assert that the PendingApplicationTransfer fields in both transfer
		// responses match the expected PendingApplicationTransfer.
		if expectedPendingTransfer == nil {
			expectedPendingTransfer = &apptypes.PendingApplicationTransfer{
				DestinationAddress: expectedDstBech32,
				SessionEndHeight:   uint64(sessionEndHeight),
			}
		}
		require.EqualValues(s.T(), expectedPendingTransfer, pendingTransfer)
```

tests/integration/application/application_transfer_test.go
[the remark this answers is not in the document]
bryanchriswhite: Nice catch! :raised_hands: I've refactored it and made it consistent between the two:

```go
	eventTypeURL := cosmostypes.MsgTypeURL(&apptypes.EventTransferBegin{})
	isTransferBeginEventFn := events.NewEventTypeMatchFn(eventTypeURL)
	targetTransferBeginEvent := s.LatestMatchingEvent(func(event *cosmostypes.Event) bool {
		if !isTransferBeginEventFn(event) {
			return false
		}

		eventSrcAddr, hasSrcAddr := events.GetAttributeValue(event, "source_address")
		require.True(s.T(), hasSrcAddr)

		return eventSrcAddr == expectedSrcApp.GetAddress()
	})
	require.NotEmpty(s.T(), targetTransferBeginEvent)
```

x/application/keeper/transfer_applications.go
```diff
@@ -0,0 +1,276 @@
...
+		// Build a set of gateway addresses from which the destination application
+		// has pending undelegations.
+		dstGatewayAddrsAtHeight := make(map[string]struct{})
+		for _, dstGatewayBech32 := range dstUndelegatingGatewaysList.GatewayAddresses {
+			if _, ok := dstGatewayAddrsAtHeight[dstGatewayBech32]; !ok {
+				dstGatewayAddrsAtHeight[dstGatewayBech32] = struct{}{}
+			}
+
+			// Also add the gateway address to the set of all pending undelegations.
+			if _, ok := allPendingUndelegationGatewayAddrs[dstGatewayBech32]; !ok {
```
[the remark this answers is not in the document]
bryanchriswhite: I think I was just too focused on updating the destination application's pending undelegations in-place. :sweat_smile: I simplified, similar to as you've suggested:

```go
	// Build a map of all gateway addresses which have pending undelegations
	// to their respective undelegation session end heights. If the source and destination
	// applications both contain pending undelegations from the same gateway address, the
	// source pending undelegation is ignored. E.g., app1 has a pending undelegation at
	// height 10 from gateway1 and app2 has pending undelegation from gateway1 at height 20;
	// only one pending undelegation from gateway1 should be present in the merged result.
	undelegationsUnionAddrToHeightMap := make(map[string]uint64)

	for srcHeight, srcUndelegetingGatewaysList := range srcApp.PendingUndelegations {
		for _, srcGateway := range srcUndelegetingGatewaysList.GatewayAddresses {
			undelegationsUnionAddrToHeightMap[srcGateway] = srcHeight
		}
	}

	for dstHeight, dstUndelegatingGatewaysList := range dstApp.PendingUndelegations {
		for _, dstGateway := range dstUndelegatingGatewaysList.GatewayAddresses {
			undelegationsUnionAddrToHeightMap[dstGateway] = dstHeight
		}
	}

	// Reset the destination application's pending undelegations and rebuild it
	// from the undelegations union map.
	dstApp.PendingUndelegations = make(map[uint64]types.UndelegatingGatewayList)
	for gatewayAddr, height := range undelegationsUnionAddrToHeightMap {
		dstPendingUndelegationsAtHeight, ok := dstApp.PendingUndelegations[height]
		if !ok {
			dstApp.PendingUndelegations[height] = types.UndelegatingGatewayList{
				GatewayAddresses: []string{gatewayAddr},
			}
		} else {
			dstPendingUndelegationsAtHeight.GatewayAddresses = append(dstApp.PendingUndelegations[height].GatewayAddresses, gatewayAddr)
			dstApp.PendingUndelegations[height] = dstPendingUndelegationsAtHeight
		}
	}
```

Thanks for the thorough review! :raised_hands:

e2e/tests/init_test.go
```diff
@@ -511,22 +518,51 @@ func (s *suite) TheUserWaitsForTheSupplierForAccountUnbondingPeriodToFinish(accN
 	s.waitForBlockHeight(unbondingHeight + 1) // Add 1 to ensure the unbonding block has been committed
 }
 
-func (s *suite) TheApplicationForAccountIsUnbonding(appName string) {
+func (s *suite) TheApplicationForAccountIsInThePeriod(appName, periodName string) {
```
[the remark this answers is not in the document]
Olshansk: Yea, I prefer window too but realized that everything else is period.

Let's keep it as is. I think consistency is more valuable here.

x/application/keeper/transfer_applications.go
```diff
@@ -0,0 +1,241 @@
...
+	)
+
+	sdkCtx := cosmostypes.UnwrapSDKContext(ctx)
+	currentHeight := sdkCtx.BlockHeight()
+	sharedParams := k.sharedKeeper.GetParams(ctx)
+	logger := k.Logger().With("method", "EndBlockerTransferApplication")
+
+	// Only process application transfers at the end of the session in
+	// order to avoid inconsistent/unpredictable mid-session behavior.
+	if !shared.IsSessionEndHeight(&sharedParams, currentHeight) {
```
Olshansk, line 35: Add a debug log please that includes current height and the next session end height

x/application/keeper/transfer_applications.go
```diff
@@ -0,0 +1,241 @@
...
+		return nil
+	}
+
+	// Iterate over all applications and transfer the ones that have finished the transfer period.
+	// TODO_MAINNET: Use an index to iterate over the applications that have initiated
```
Olshansk, line 40: Can you add your name here?

Would love it if you can start looking into indexing on-chain!

x/application/keeper/transfer_applications.go
```diff
@@ -0,0 +1,241 @@
...
+				return err
+			}
+		}
+	}
+
+	isSuccessful = true
+	return nil
+}
+
+// transferApplication transfers the fields of srcApp, except for address and pending_transfer,
```
Olshansk, line 81: Can we remove 81-87?

x/application/keeper/msg_server_transfer_application.go
```diff
@@ -0,0 +1,85 @@
...
+
+	sdkCtx := sdk.UnwrapSDKContext(ctx)
+	sessionEndHeight := k.sharedKeeper.GetSessionEndHeight(ctx, sdkCtx.BlockHeight())
+
+	srcApp.PendingTransfer = &types.PendingApplicationTransfer{
+		DestinationAddress: msg.GetDestinationAddress(),
+		SessionEndHeight:   uint64(sessionEndHeight),
+	}
+
+	// Update the dstApp in the store
```
Olshansk, line 65: srcApp?

x/application/keeper/msg_server_unstake_application.go
```diff
@@ -27,17 +27,26 @@ func (k msgServer) UnstakeApplication(
...
 
 	// Check if the application has already initiated the unstaking process.
 	if foundApp.IsUnbonding() {
-		logger.Warn(fmt.Sprintf("Application %s is still unbonding from previous unstaking", msg.Address))
-		return nil, types.ErrAppIsUnstaking
+		logger.Warn(fmt.Sprintf("Application (%s) is still unbonding from previous unstaking", msg.GetAddress()))
+		return nil, types.ErrAppIsUnstaking.Wrapf("application (%s)", msg.GetAddress())
+	}
+
+	// Check if the application has already initiated a transfer process.
```
Olshansk, line 43:
```suggestion
	// Check if the application has already initiated a transfer process.
	// Transferring applications CANNOT unstake.
```

x/application/keeper/transfer_applications.go
```diff
@@ -0,0 +1,241 @@
...
+		} else {
+			dstPendingUndelegationsAtHeight.GatewayAddresses = append(dstApp.PendingUndelegations[height].GatewayAddresses, gatewayAddr)
+			dstApp.PendingUndelegations[height] = dstPendingUndelegationsAtHeight
+		}
+	}
+}
+
+// MergeAppServiceConfigs takes the union of the srcApp and dstApp's service configs
+// and sets the result in dstApp. It is exported for testing purposes.
+func MergeAppServiceConfigs(srcApp, dstApp *types.Application) {
```
Olshansk, line 227: Do you have a followup PR that updates the docs on this?

Would love a **simple** excalidraw/mermaid diagram for it.

x/application/keeper/transfer_applications.go
```diff
@@ -0,0 +1,241 @@
...
+
+		logger.Info(fmt.Sprintf(
+			"transferring application from %q to new application %q",
+			srcApp.GetAddress(), dstApp.GetAddress(),
+		))
+	} else {
+		srcStakeSumCoin := dstApp.GetStake().Add(*dstApp.GetStake())
+		dstApp.Stake = &srcStakeSumCoin
+
+		MergeAppDelegatees(&srcApp, &dstApp)
```
Olshansk, line 127: Confirming this needs to be publically exposed?

Olshansk (approved):
@bryanchriswhite Approving since I won't be available and overall it LGTM.

My biggest concerns / asks for the future:

1. Think of ways to make sure we use the test helpers you introduced
2. Add *SIMPLE* (< 7 elements per )mermaid diagrams in the docs showing what merging does and how it works

red-0ne (approved):
LGTM! :partying_face:

x/application/keeper/transfer_applications.go
bryanchriswhite: I will include it in #790. :+1:

x/application/keeper/transfer_applications.go
bryanchriswhite: I would like to push back on this. **This would produce a log for every block that's not a session end height.** [`EndBlockerUnbondApplications()`]() and [`EndBlockerUnbondSuppliers()`]() both have a simliar conditional branch and neither includes the log that you're suggesting here.

Can you elaborate on the rationale or concern?

bryanchriswhite: I've added the current and session end height to the logger output:

```go
	currentHeight := sdkCtx.BlockHeight()
	sharedParams := k.sharedKeeper.GetParams(ctx)
	sessionEndHeight := shared.GetSessionEndHeight(&sharedParams, currentHeight)
	logger := k.Logger().
		With("method", "EndBlockerTransferApplication").
		With("current_height", currentHeight).
		With("session_end_height", sessionEndHeight)
```

x/application/keeper/transfer_applications.go
bryanchriswhite: Not sure how that happened - thanks for calling it out! :raised_hands:

x/application/keeper/transfer_applications.go
bryanchriswhite: https://github.com/pokt-network/poktroll/issues/854

PR #801, opened 2024-09-07
[GATEWAY POC] A couple helpers to help deploy a PATH gateway

Olshansk:
Update the docs to also stake a gateway so we can use it with PATH.


okdas (approved):
I didn't run this but looks good. Excited to try this soon!

Olshansk:
Merged with a minor CI issue captured here: #803

PR #826, opened 2024-09-13
[Testing] refractor: param update E2E tests as integration tests

bryanchriswhite:
## Summary

Ports the `MsgUpdateParams` and `MsgUpdateParam` E2E tests to integration tests, improving execution speed and maintainability.

## Depends on

- #827 
- #820 
- #821
- #815

## Dependents

- #809 

## Issue

- #799

## Type of change

Select one or more from the following:

- [ ] New feature, functionality or library
- [ ] Consensus breaking; add the `consensus-breaking` label if so. See #791 for details
- [ ] Bug fix
- [x] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

- [x] **Documentation**: `make docusaurus_start`; only needed if you make doc changes
- [ ] **Unit Tests**: `make go_develop_and_test`
- [ ] **LocalNet E2E Tests**: `make test_e2e`
- [x] **DevNet E2E Tests**: Add the `devnet-test-e2e` label to the PR.

## Sanity Checklist

- [x] I have tested my changes using the available tooling
- [x] I have commented my code
- [x] I have performed a self-review of my own code; both comments & source code
- [x] I create and reference any new tickets, if applicable
- [x] I have left TODOs throughout the codebase, if applicable

Issue #799, opened 2024-09-06
[Testing] E2E Test Refactoring


Olshansk:
## Objective

Improve developer productivity to avoid spending time on flaky E2E tests.

## Origin Document

[E2E Test Mindmap](https://www.notion.so/buildwithgrove/E2E-testing-824a9045aaa044b4ae41b07b560a34db?pvs=4)

## Goals

- Reliable E2E Tests
- Reliable Integration Tests
- DIrection on when to use E2E Tests vs Integraiton Tests
- Save time for developers while focusing on what matters

## Deliverables

- [ ] Identify E2E tests that are flaky
	- [ ] If it does not involve off chain actors -> refactor into an integration test
	- [ ] If it does involve off-chain actors -> Make it 100% reliable in CI 
- [ ] Identify & design state management solution for 100% reliability.
	- Could be one of or combination of the following:
		 - Dynamic introduction of new acotrs
		 - Endpoint to reset the on-chain of existing
- [ ] Document (and make visible) for developers what should and shouldn't be an E2E test
- [ ] Debug CI

## Non-goals / Non-deliverables

- Refactoring all E2E Tests w/ 0 off-chain actors into integration tests

---

**Creator**: @Olshansk 
**Co-Owners**: @bryanchriswhite @red-0ne

PR #827, opened 2024-09-16
[Testing] refactor: integration app / feat: add `IntegrationAppSuite`

bryanchriswhite:
## Summary

- Refactor  integration app base app construction to eliminate duplicate DB construction.
- Set default params for all poktroll modules, some were omitted.
- Expose integration app's pre-generated account iterator for use by integration tests.
- Add faucet account to integration app to consolidate funding mechanics for integration tests.
- Add authz & bank modules to integration app in preparation for param update test porting/refactoring as integration tests (#826).
- Add IntegrationAppSuite & base implementation.

## Dependents

- #789
- #826

## Issue

- #799

## Type of change

- [ ] Bug fix
- [x] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

## Sanity Checklist


testutil/integration/suites/base.go
```diff
@@ -0,0 +1,93 @@
...
+// NewApp constructs a new integration app and sets it on the suite.
+func (s *BaseIntegrationSuite) NewApp(t *testing.T) *integration.App {
+	t.Helper()
+
+	s.app = integration.NewCompleteIntegrationApp(t)
+	return s.app
+}
+
+// SetApp sets the integration app on the suite.
+func (s *BaseIntegrationSuite) SetApp(app *integration.App) {
```
red-0ne, line 55: Do we need this given that it's already initialized in `NewApp`?

testutil/integration/app.go
```diff
@@ -659,6 +755,11 @@ func (app *App) RunMsg(t *testing.T, msg sdk.Msg, option ...RunOption) *codectyp
 	if cfg.AutomaticCommit {
 		defer func() {
 			_, err := app.Commit()
+			if cfg.ErrorAssertion != nil && err != nil {
+				cfg.ErrorAssertion(err)
```
red-0ne, line 759: Should we log something when there's an error but `cfg.ErrorAssertion` is nil?

testutil/integration/options.go
```diff
@@ -1,14 +1,30 @@
...
 }
 
 // RunOption is a function that can be used to configure the integration app.
 type RunOption func(*RunConfig)
 
+// RunOptions is a list of RunOption. It implements the Append method for convenience.
+type RunOptions []RunOption
+
+// Append returns a new RunOptions with the given RunOptions appended.
+func (runOpts RunOptions) Append(newRunOpts ...RunOption) RunOptions {
```
red-0ne, line 24: Will there be a use case for `RunOptions` and it `Append` method?

red-0ne (commented):
Great up-front work preparing for better/more integration tests :rocket:

Left a few questions but looks good overall. Waiting to see it in action :+1:

testutil/integration/suites/base.go
bryanchriswhite: I think it's prudent. I can imagine a scenario where the integration app is constructed elsewhere. In the case where we want a test suite to inherit from multiple other suites, we would require this method because only one suite can be embedded and the other(s) would still need a way to set their app fields to the embedded app (rather than constructing new one(s)).

testutil/integration/app.go
bryanchriswhite: I don't think it's necessary and I don't think we have the scope here to know whether it's an issue. If `cfg.ErrorAssertion` is set then it should be responsible for determining whether a nil error is acceptable.

testutil/integration/options.go
bryanchriswhite: No longer. I've removed `RunOption` and `RunConfig` altogether.  :sweat_smile:

testutil/integration/options.go
```diff
@@ -1,25 +1,61 @@
...
+	// InitChainer function.
+	InitChainerModuleFns []InitChainerModuleFn
 }
 
-// RunOption is a function that can be used to configure the integration app.
-type RunOption func(*RunConfig)
+// IntegrationAppOption is a function that receives and has the opportunity to
+// modify the IntegrationAppConfig. It is intended to be passed during integration
+// App construction to modify the behavior of the integration App.
+type IntegrationAppOption func(*IntegrationAppConfig)
```
red-0ne, line 22:
```suggestion
type IntegrationAppOptionFn func(*IntegrationAppConfig)
```

testutil/integration/suites/base_test.go
```diff
@@ -0,0 +1,214 @@
...
+	require.Nil(s.T(), s.app)
+
+	app := s.NewApp(s.T())
+	require.Same(s.T(), app, s.app)
+}
+
+func (s *BaseIntegrationSuiteTestSuite) TestGetApp_ReturnsApp() {
+	app := s.NewApp(s.T())
+	require.Same(s.T(), app, s.GetApp())
+}
```
red-0ne, line 59:
```suggestion
}

```

testutil/integration/suites/base_test.go
```diff
@@ -0,0 +1,214 @@
...
+	// Assert that the events are cleared on the next block.
+	s.GetApp().NextBlock(s.T())
+	require.Equal(s.T(), 0, len(s.SdkCtx().EventManager().Events()), "expected no events in the next block")
+}
+
+func (s *BaseIntegrationSuiteTestSuite) TestFilterLatestEventsWithNewEventTypeMatchFn() {
+	expectedNumEvents := 3
+
+	s.NewApp(s.T())
+	s.emitPoktrollGatewayUnstakedEvents(expectedNumEvents)
```
red-0ne, line 140: Assert that `len(s.SdkCtx().EventManager().Events())` is zero before emitting and non-zero after, in contrast with The latest assertion in that test.

testutil/integration/app.go
```diff
@@ -228,19 +281,31 @@ func NewCompleteIntegrationApp(t *testing.T) *App {
...
-	sdkCtx := sdk.NewContext(cms, cmtproto.Header{
+
+	// Prepare the database and multi-store.
+	db := dbm.NewMemDB()
+
+	// Prepare the base application.
+	bApp := baseapp.NewBaseApp(appName, logger, db, txCfg.TxDecoder(), baseapp.SetChainID(appName))
+
+	// Prepare the context
+	sdkCtx := bApp.NewUncachedContext(false, cmtproto.Header{
```
red-0ne, line 300: :+1:

red-0ne (commented):
Left some comments but this is a really great integration test suite :sparkles:.

Eager to see it used in our tests.

testutil/integration/suites/interface.go
```diff
@@ -0,0 +1,48 @@
...
+	// in the integration app.
+	GetCosmosModuleNames() []string
+	// SdkCtx returns the integration app's SDK context.
+	SdkCtx() *cosmostypes.Context
+
+	// FundAddress sends amtUpokt coins from the faucet to the given address.
+	FundAddress(t *testing.T, addr cosmostypes.AccAddress, amtUpokt int64)
+	// GetBankQueryClient constructs and returns a query client for the bank module
+	// of the integration app.
+	GetBankQueryClient() banktypes.QueryClient
```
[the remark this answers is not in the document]
bryanchriswhite: I don't see a way to do this using generics as this is an interface method. I could imagine a solution that uses reflection but I think that would require passing either a zero-value object or nil pointer thereto of the desired concrete query client type, which doesn't sound very pretty IMHO. I was anticipating having module-specific test suites (e.g. `ApplicationIntegrationSuite`) which would implement the necessary query client getters and would be embedded in the suites which need them. And/or we could have a suite which exposes all the clients via getters.

testutil/integration/app.go
```diff
@@ -633,65 +644,121 @@ func (app *App) GetAuthority() string {
...
 	t.Helper()
 
-	// set options
-	cfg := &RunConfig{}
-	for _, opt := range option {
-		opt(cfg)
+	// Commit the updated state after the message has been handled.
+	var finalizeBlockRes *abci.ResponseFinalizeBlock
+	defer func() {
+		if _, commitErr := app.Commit(); err != nil {
```
bryanchriswhite, line 685:
```suggestion
		if _, commitErr := app.Commit(); commitErr != nil {
```

testutil/integration/suites/base_test.go
```diff
@@ -0,0 +1,219 @@
...
+	"github.com/stretchr/testify/suite"
+
+	"github.com/pokt-network/poktroll/app/volatile"
+	"github.com/pokt-network/poktroll/testutil/events"
+	"github.com/pokt-network/poktroll/testutil/integration"
+	"github.com/pokt-network/poktroll/testutil/sample"
+	gatewaytypes "github.com/pokt-network/poktroll/x/gateway/types"
+)
+
+var gatewayStakeAmount = int64(1000)
```
bryanchriswhite, line 20:
```suggestion
```

testutil/integration/suites/interface.go
```diff
@@ -0,0 +1,44 @@
...
+
+	// FilterLatestEvents returns the most recent events in the event manager that
+	// match the given matchFn.
```
bryanchriswhite, line 39:
```suggestion
	// FilterEvents returns the events from the event manager which match the given
	// matchFn. Events are returned in reverse order, i.e. the most recent event is
	// first.
```

red-0ne (approved):
LGTM!

testutil/integration/app.go
```diff
@@ -672,58 +657,112 @@ func (app *App) QueryHelper() *baseapp.QueryServiceTestHelper {
...
-			app.nextBlockUpdateCtx()
-		}()
+	require.Equal(t, 1, len(txMsgRes), "expected exactly 1 tx msg response")
+	return txMsgRes[0], err
+}
+
+// RunMsgs provides the ability to process messages by packing them into a tx and
+// driving the ABCI through block finalization. It returns a slice of tx.MsgResponse
+// (any) whose elements correspond to the request message of the same index. These
+// responses can be type asserted to the expected response type.
```
Olshansk, line 679:
```suggestion
// (any) whose elements correspond to the request message of the same index.
// These responses can be type asserted to the expected response type.
```

testutil/integration/suites/interface.go
```diff
@@ -0,0 +1,45 @@
...
+// IntegrationSuite is an interface intended to be used within test suites which
+// exercise an integration.App.
+type IntegrationSuite interface {
+	// NewApp constructs a new integration app and sets it on the suite.
+	NewApp(*testing.T, ...integration.IntegrationAppOptionFn) *integration.App
+	// SetApp sets the integration app on the suite.
+	SetApp(*integration.App)
+	// GetApp returns the integration app from the suite.
+	GetApp() *integration.App
+	// GetPoktrollModuleNames returns the list of the names of all poktroll modules
```
Olshansk, line 23: I understand why you have `poktroll` in the name now, but would like to avoid it. Let's call it `All` and `Cosmos` is explicit

testutil/integration/suites/interface.go
```diff
@@ -0,0 +1,45 @@
...
+
+	cosmostypes "github.com/cosmos/cosmos-sdk/types"
+	banktypes "github.com/cosmos/cosmos-sdk/x/bank/types"
+
+	"github.com/pokt-network/poktroll/testutil/integration"
+)
+
+// IntegrationSuite is an interface intended to be used within test suites which
+// exercise an integration.App.
+type IntegrationSuite interface {
```
Olshansk, line 16: This is great!

Just note, you'll be responsible for reviewing tests now ;)

I have a general idea of what's available, but one only gets good at it once you do something hands on.

testutil/integration/suites/base_test.go
```diff
@@ -0,0 +1,217 @@
...
+	"github.com/stretchr/testify/suite"
+
+	"github.com/pokt-network/poktroll/app/volatile"
+	"github.com/pokt-network/poktroll/testutil/events"
+	"github.com/pokt-network/poktroll/testutil/integration"
+	"github.com/pokt-network/poktroll/testutil/sample"
+	gatewaytypes "github.com/pokt-network/poktroll/x/gateway/types"
+)
+
+type BaseIntegrationSuiteTestSuite struct {
```
Olshansk, line 20: Confirming that this (and othesr) need to be publically exposed?

Can you add a godoc for all of them please. It'll help when we try to get others to adopt this.

testutil/integration/app.go
```diff
@@ -780,44 +836,167 @@ func (app *App) nextBlockUpdateCtx() {
...
 
-// CreateMultiStore is a helper for setting up multiple stores for provided modules.
-func CreateMultiStore(keys map[string]*storetypes.KVStoreKey, logger log.Logger) storetypes.CommitMultiStore {
-	db := dbm.NewMemDB()
-	cms := store.NewCommitMultiStore(db, logger, metrics.NewNoOpMetrics())
+// setupDefaultActorsState uses the integration app keepers to stake "default"
+// on-chain actors for use in tests. In creates a service, and stakes a supplier
+// and application as well as funding the bank balance of the default supplier.
+//
+// TODO_IMPROVE: Eliminate usage of and remove this function in favor of
```
Olshansk, line 854: Could you mark this as `TODO_TECHDEBT(@bryanchriswhite)`. I think your're one of the few people who will be able to do a good job at this.

Olshansk (changes requested):
@bryanchriswhite Not one of my best reviews as reviewing this is quite tough. Hopefully my partial review helps.

If I have the concentration energy tomorrow, will go deeper.

testutil/sample/sample.go
```diff
@@ -30,12 +30,21 @@ func AccAddressAndPubKeyEd25519() (string, cryptotypes.PubKey) {
...
+func ValAddress() string {
+	_, pk := AccAddressAndPubKey()
+	validatorAddress := tmhash.SumTruncated(pk.Address())
+	valAddress := sdk.ValAddress(validatorAddress)
+	return valAddress.String()
+}
+
+// ConsAddress returns a sample consensus node address, which has the prefix
+// of consensus nodes when converted to bech32.
 func ConsAddress() string {
```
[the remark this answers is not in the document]
bryanchriswhite: It is, I checked.

tests/integration/tokenomics/relay_mining_integration_test.go
```diff
@@ -21,28 +24,52 @@ import (
...
 
-func TestComputeNewDifficultyHash_RewardsReflectWorkCompleted(t *testing.T) {
-	// Test params
-	globalComputeUnitsToTokensMultiplier := uint64(1) // keeping the math simple
-	// serviceComputeUnitsPerRelay := uint64(1)          // keeping the math simple
+var (
+	// Test params.
+	computeUnitsToTokensMultiplier = uint64(1) // keeping the math simple
+	proofRequirementThreshold      = sdk.NewInt64Coin(volatile.DenomuPOKT, 1e18)
+	//serviceComputeUnitsPerRelay = uint64(1)   // keeping the math simple
```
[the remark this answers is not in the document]
bryanchriswhite: No objection here, it was already present and I was reluctant to remove it without knowing the intention (assuming intentional) behind leaving it in.

testutil/integration/suites/interface.go
bryanchriswhite: Can you elaborate on your rationale? I feel strongly about avoiding ambiguity around "all".

tests/integration/tokenomics/relay_mining_integration_test.go
```diff
@@ -1,18 +1,21 @@
+//go:build integration
```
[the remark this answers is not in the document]
bryanchriswhite: The rationale for adding the tag here is that these tests will depend on the `ParamsSuite` which is currently using the `integration` build constraint. My thinking was that since the makefile has `test_integration`, running tests shouldn't be problematic. However, I see no reason why we MUST use an `integration` tag and agree that simpler is better. :smile: I will remove them.

I believe that rather than an `integration` tag, it would be more effective to have an `in-memory-network` as those tests are more useful to separate out as they tend to be flaky (leveldb closed, address already in use, etc.).

testutil/integration/suites/base_test.go
bryanchriswhite:
```go
// baseIntegrationSuiteTestSuite is a test suite which embeds BaseIntegrationSuite.
// **in order to test it**. It is intended to be embedded in other test suites.
type baseIntegrationSuiteTestSuite struct {
	BaseIntegrationSuite
}
```

bryanchriswhite: Are you saying you would like to see godoc comments on the test methods?

testutil/sample/sample.go
```diff
@@ -30,12 +30,21 @@ func AccAddressAndPubKeyEd25519() (string, cryptotypes.PubKey) {
...
+// ValAddress returns a sample validator address, which has the prefix
+// of validators when converted to bech32.
+func ValAddress() string {
+	_, pk := AccAddressAndPubKey()
+	validatorAddress := tmhash.SumTruncated(pk.Address())
+	valAddress := sdk.ValAddress(validatorAddress)
+	return valAddress.String()
+}
+
+// ConsAddress returns a sample consensus node address, which has the prefix
```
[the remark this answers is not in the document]
bryanchriswhite: I'm not sure honestly, all I know so far is that they have different prefixes.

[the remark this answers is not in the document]
bryanchriswhite: [Found it](https://docs.cosmos.network/main/learn/beginner/accounts#addresses) - will link in a comment.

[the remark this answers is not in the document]
bryanchriswhite:
```go
// ValAddress returns a sample validator address, which has the prefix
// of validators when converted to bech32. Validator addresses identify
// the validator operator on-chain account and are derived using secp256k1.
// See: https://docs.cosmos.network/main/learn/beginner/accounts#addresses
func ValAddress() string {
	_, pk := AccAddressAndPubKey()
	validatorAddress := tmhash.SumTruncated(pk.Address())
	valAddress := sdk.ValAddress(validatorAddress)
	return valAddress.String()
}

// ConsAddress returns a sample consensus node address, which has the prefix
// of consensus nodes when converted to bech32. Consensus addresses identify
// the validator node in the consensus engine and are derived using ed25519.
// See: https://docs.cosmos.network/main/learn/beginner/accounts#addresses
func ConsAddress() string {
	_, pk := AccAddressAndPubKeyEd25519()
	consensusAddress := tmhash.SumTruncated(pk.Address())
	valAddress := sdk.ConsAddress(consensusAddress)
	return valAddress.String()
}
```

testutil/integration/app.go
```diff
@@ -74,24 +81,34 @@ import (
 
 const appName = "poktroll-integration-app"
 
+var (
+	// FaucetAddrStr is a random address which is funded with FaucetAmountUpokt
+	// coins such that it can be used as a faucet for integration tests.
```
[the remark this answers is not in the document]
bryanchriswhite: I renamed and moved `FaucetAddrStr` to an unexported field of `App` as `faucetBech32` and unexported `faucetAmountUpokt`.

testutil/integration/suites/interface.go
Olshansk: Fair enough. We'll change if/when we need to.

testutil/integration/suites/base_test.go
Olshansk: 1. The struct yets.
2. The methods. Yes?

My rationale is that you built a lot of helpers but I'm afraid I won't know how/when to use them and am beginning to think that [1] will be the way to go.

[1] https://pkg.go.dev/github.com/pokt-network/poktroll

Olshansk (approved):

testutil/integration/suites/base_test.go
bryanchriswhite: I agree that the godocs are the way to go but I think we're talking past each other a bit here.

This is "test code", as in code that testing the thing that we would actually want to have godocs for. **Test code does is omitted from godoc output, so any godoc-style comments added here are only for the benefit of someone looking at this code** (with the exception of [Testable Examples](https://go.dev/blog/examples)).

I added the comment above to the struct and unexported it to avoid confusion. The object under test (`BaseIntegrationSuite`) has godoc comments for its struct and methods. Is that sufficient?

bryanchriswhite: UPDATE: I'm realizing there was a typo in the comment, I most unfortunately omitted the word "NOT". :sweat_smile:. Corrected.

PR #809, opened 2024-09-10
[Gateway] Add `min_stake` gateway module param

bryanchriswhite:
## Summary

Adds the `min_stake_gateway` param to the gateway.

## Depends on

- #808 
- #827
- #826

## Issue

- #612

## Type of change

Select one or more from the following:

- [x] New feature, functionality or library
- [x] Consensus breaking; add the `consensus-breaking` label if so. See #791 for details
- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

- [x] **Documentation**: `make docusaurus_start`; only needed if you make doc changes
- [ ] **Unit Tests**: `make go_develop_and_test`
- [ ] **LocalNet E2E Tests**: `make test_e2e`
- [ ] **DevNet E2E Tests**: Add the `devnet-test-e2e` label to the PR.

## Sanity Checklist

- [x] I have tested my changes using the available tooling
- [x] I have commented my code
- [x] I have performed a self-review of my own code; both comments & source code
- [ ] I create and reference any new tickets, if applicable
- [ ] I have left TODOs throughout the codebase, if applicable

Issue #612, opened 2024-06-17
[Utility] Minimum staking values for each actor


Olshansk:
## Objective

Introduce a minimum stake (MinStake) requirement for `Gateways`, `Suppliers`, and `Applications` in the network.

## Status

- #808 
- #809 
- #843
- #844
- #845
- #847
- #848 
- #849
- #850
- #857
- ~~_TODO: [Supplier] Enforce minimum stake when slashing_~~ (added in #795; see [x/tokenomics/keeper/settle_pending_claims.go](https://github.com/pokt-network/poktroll/pull/795/files#diff-162fd3e506083088104ff9c7fa152d635766e18a54bc633b6723a0c6e88e98e7R326))
- #862

### Related

- #839 
- #851 
- #852 
- #861

## Origin Document

In Morse, if you run:

```bash
	curl --location -X 'POST' -H 'Content-Type: application/json' "${POCKET_ENDPOINT}/v1/query/allParams" | jq
```

We find the following param under `node_params`

```bash
    {
      "param_key": "pos/StakeMinimum",
      "param_value": "15000000000"
    },
```

## Goals

- Establish a clear MinStake requirement for `Gateways`, `Suppliers` & `Applications`
- Enhance network security and reliability by having "something at stake"
- Create AT LEAST Morse feature parity

## Deliverables

- [ ] Add the `MinStakeGateway` governance parameter
- [ ] Add the `MinStakeApplication` governance parameter
- [ ] Add the `MinStakeSupplier` governance parameter
- [ ] Update docusouras documentation on these params
- [ ] Add at least one minimum E2E and unit test for each
- [ ] Communicate the changes to the network participants.

## Non-goals / Non-deliverables

- Implementing any staking-related penalties or rewards beyond the initial MinStake requirements.
- Making changes to the staking mechanisms unrelated to the MinStake values.
- Addressing any other network roles not specified in this issue (e.g., Validators).

## General deliverables

- [ ] **Comments**: Add/update TODOs and comments alongside the source code so it is easier to follow.
- [ ] **Testing**: Add new tests (unit and/or E2E) to the test suite to ensure MinStake enforcement.
- [ ] **Makefile**: Add new targets to the Makefile to facilitate testing and deployment of MinStake functionality.
- [ ] **Documentation**: Update architectural or development READMEs; use [mermaid](https://mermaid-js.github.io/mermaid/) diagrams where appropriate.

---

**Creator**: @Olshansk 
[...]

Issue #384, opened 2024-02-14
[DevExp] Developer Workflow Experience & Improvements


Olshansk:
## Objective

Improve the developer experience for blockchain developers.

## Origin Document

I encountered a handful of problems going through #359, ranging from bugs, issues and institutional knowledge that we have to capture and iterate on.

## Goals

- Reduce friction to onboard new developers
- Share knowledge on tips & tricks for day-to-day tooling
- Remove `hidden expected undefined` behaviour such as `I know this doesn't work` that others might not be aware of
- Build small tools or add small pieces of documentation that can greatly make everyone more effective

## Deliverables

- [ ] `[E2E Testing]` Make it easier to run a single E2E test. Formulating the following CLI command is non-trivial and there is an opportunity to have a `make` helper to just put in a fuzzy matched name

	```bash
	go test -v -tags=test,e2e -count=1 -run \
	TestFeatures/Tokenomics_Namespaces/Basic_tokenomics_validation_that_Supplier_mint_equals_Application_burn \
	./e2e/tests/...
	```
- [ ] `[Tooling`] Make it easier to run any `poktrolld` command w/ default flags during development. Formulating the following CLI command is non-trivial and there is an opportunity to have a `make` helper to just put in a fuzzy matched name.

   ```bash
   	poktrolld query application show-application pokt1mrqt5f7qh8uxs27cjm9t7v9e74a9vvdnq5jva4 --home=./localnet/poktrolld --node=tcp://127.0.0.1:36657
   	```	
- [x] `[Infra]` Make it easier to maintain state across LocalNet restarts. Currently, every time we click `Trigger Update`, the blockchain restarts from `height=1`
- [x] `[Infra]` Enable hot reloading. This is not in place today unless we click `Trigger Update` explicitly.
- [x] `[Infra]` Document the order of operations if the `Sequencer` is restarted. Specifically, `relayminers` and `appgateservers` need to have `Trigger Update` clicked **AFTER** the `sequencer` is back up for things to function normally
- [ ] `[Logging]` Document how `on-chain` logging should be done as a best practice. For example

	```
	logger = logger.With(
		"session_id", sessionHeader.GetSessionId(),
		"session_end_height", sessionHeader.GetSessionEndBlockHeight(),
		"supplier", supplierAddr)
	logger.Info("validated the submitProof message ")
	```
	
	Consider which logger should be used:
[...]

PR #843, opened 2024-09-27
[Gateway] Enforce minimum stake when staking

bryanchriswhite:
## Summary

1. Adds minimum stake validation to the gateway stake message handler (i.e. total stake must be >= minimum stake).
2. Updates error returns in the same handler to ensure all error paths return appropriate gRPC status errors.
3. Replaces some warn and error level logs with info level, which I believe is more appropriate (until we have a practical debug level, at which point they should become debug logs).

## Dependencies

- #809 

## Dependents

- #844 
- #845 
- #847 
- #848 
- #849
- #850
- #857

## Issue

Gateway staking min stake validation.

- #612

## Type of change

- [x] Consensus breaking; add the `consensus-breaking` label if so. See #791 for details
- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

## Sanity Checklist


x/gateway/keeper/msg_server_stake_gateway.go
```diff
@@ -24,46 +26,72 @@ func (k msgServer) StakeGateway(
...
+			return nil, status.Error(
+				codes.InvalidArgument,
+				types.ErrGatewayInvalidStake.Wrapf(
+					"stake (%s) must be higher than previous stake (%s)",
+					msg.Stake, currGatewayStake,
+				).Error(),
+			)
 		}
-		logger.Info(fmt.Sprintf("Gateway is going to escrow an additional %+v coins", coinsToEscrow))
+		totalStake = currGatewayStake.Add(*msg.Stake)
```
red-0ne, line 72: `msg.Stake` is the new total stake, we should not add it to the current stake.
```
newStake == msg.Stake
i.e.
newStake == oldStake + coinsToEscrow
```

x/gateway/keeper/msg_server_stake_gateway.go
```diff
@@ -24,46 +26,72 @@ func (k msgServer) StakeGateway(
...
 
-	// Retrieve the address of the gateway
-	gatewayAddress, err := sdk.AccAddressFromBech32(msg.Address)
-	if err != nil {
-		// TODO_TECHDEBT(#384): determine whether to continue using cosmos logger for debug level.
-		logger.Error(fmt.Sprintf("could not parse address %q", msg.Address))
-		return nil, err
+	// MUST ALWAYS have at least minimum stake.
+	minStake := k.GetParams(ctx).MinStake
+	if totalStake.Amount.LT(minStake.Amount) {
```
red-0ne, line 88: As per my previous comment, this should be:
```suggestion
	if msg.Stake.Amount.LT(minStake.Amount) {
```

x/gateway/keeper/msg_server_stake_gateway_test.go
```diff
@@ -60,33 +61,61 @@ func TestMsgServer_StakeGateway_FailLoweringStake(t *testing.T) {
...
+	k, ctx := keepertest.GatewayKeeper(t)
+	srv := keeper.NewMsgServerImpl(k)
+
+	addr := sample.AccAddress()
+	gatewayStake := cosmostypes.NewInt64Coin(volatile.DenomuPOKT, 100)
+	minStake := gatewayStake.AddAmount(math.NewInt(1))
+	expectedErr := gatewaytypes.ErrGatewayInvalidStake.Wrapf("gateway %q must stake at least %s", addr, minStake)
+
+	// Set the minimum stake to be greater than the gateway stake.
+	err := k.SetParams(ctx, gatewaytypes.Params{
```
red-0ne, line 105: I'd prefer having this part as:
```
params := k.GetParams()
params.Minstake = minStake
k.SetParams(params)
```
I've been hit many times with overriding previous param values with their zero values.

It won't be the case here since gateway has only one param but I believe it's a good practice.

red-0ne (changes requested):
Small but great addition. Looking forward the other min stakes :)

I left a comment about totalStake change which I think should remain as it was.

x/gateway/keeper/msg_server_stake_gateway.go
bryanchriswhite: :facepalm: Thank you for catching this.

x/gateway/keeper/msg_server_stake_gateway.go
```diff
@@ -24,46 +26,67 @@ func (k msgServer) StakeGateway(
...
-		return nil, err
+		return nil, status.Error(codes.InvalidArgument, err.Error())
+	}
+
+	// Retrieve the address of the gateway
+	gatewayAddress, err := sdk.AccAddressFromBech32(msg.Address)
+	// NB: This SHOULD NEVER happen because msg.ValidateBasic() validates the address as bech32.
+	if err != nil {
+		// TODO_TECHDEBT(#384): determine whether to continue using cosmos logger for debug level.
+		logger.Info(fmt.Sprintf("could not parse address %q", msg.Address))
```
Olshansk, line 40: This should either be an error or a warning. Please also log the error in there as well

x/gateway/keeper/msg_server_stake_gateway.go
```diff
@@ -24,46 +26,67 @@ func (k msgServer) StakeGateway(
...
-		logger.Info(fmt.Sprintf("Gateway is going to escrow an additional %+v coins", coinsToEscrow))
+		logger.Info(fmt.Sprintf("gateway is going to escrow an additional %+v coins", coinsToEscrow))
 	}
 
-	// Must always stake or upstake (> 0 delta)
+	// MUST ALWAYS stake or upstake (> 0 delta).
 	if coinsToEscrow.IsZero() {
-		logger.Warn(fmt.Sprintf("Gateway %q must escrow more than 0 additional coins", msg.Address))
-		return nil, types.ErrGatewayInvalidStake.Wrapf("gateway %q must escrow more than 0 additional coins", msg.Address)
+		errFmt := "gateway %q must escrow more than 0 additional coins"
```
Olshansk, line 73: You're doing the same fmt twice. Maybe just prepare the error string?

x/gateway/keeper/msg_server_stake_gateway.go
```diff
@@ -24,46 +26,67 @@ func (k msgServer) StakeGateway(
...
+		logger.Info(fmt.Sprintf(errFmt, msg.Address))
```
Olshansk, line 74: Again, I think we shold be logging error/warn for errors

x/gateway/keeper/msg_server_stake_gateway.go
```diff
@@ -24,46 +26,67 @@ func (k msgServer) StakeGateway(
...
-	gatewayAddress, err := sdk.AccAddressFromBech32(msg.Address)
-	if err != nil {
-		// TODO_TECHDEBT(#384): determine whether to continue using cosmos logger for debug level.
-		logger.Error(fmt.Sprintf("could not parse address %q", msg.Address))
-		return nil, err
+	// MUST ALWAYS have at least minimum stake.
+	minStake := k.GetParams(ctx).MinStake
+	if msg.Stake.Amount.LT(minStake.Amount) {
+		errFmt := "gateway %q must stake at least %s"
+		logger.Info(fmt.Sprintf(errFmt, msg.Address, minStake))
```
Olshansk, line 85: Ditto

x/gateway/keeper/msg_server_stake_gateway_test.go
```diff
@@ -60,33 +61,61 @@ func TestMsgServer_StakeGateway_FailLoweringStake(t *testing.T) {
...
+
+	// Prepare the gateway.
+	stakeMsg := &gatewaytypes.MsgStakeGateway{
+		Address: addr,
+		Stake:   &gatewayStake,
+	}
+
+	// Attempt to stake the gateway & verify that the gateway does NOT exist.
+	_, err = srv.StakeGateway(ctx, stakeMsg)
+	require.ErrorContains(t, err, expectedErr.Error())
```
Olshansk, line 118: Personally, I think that checking if `ErrorContains(atewaytypes.ErrGatewayInvalidStake)` is enough so you don't have to update the test if the error string message changes.

Olshansk (changes requested):

x/gateway/keeper/msg_server_stake_gateway.go
bryanchriswhite: I have a strong and principled opinion here. I'm confident that we've discussed this before but it's been a while so please let me reiterate my thinking on this:

1. In my opinion, error, warn, and even info logs should be reserved for logs which are relevant to the **service operator**; as opposed to the end user, for example. These logs should be important, actionable, or indicative of a status change of the service. 

    The rationale here is that in the absence of such prescriptive usage of the log levels, **logs which are relevant to the service operator are lost in the noise** of irrelevant logs.
3. Debug logs should be used for everything else. In all the cases where I've been switching log levels, these logs are **not intended for the validator operator but instead for protocol developers during development and testing**.
4. Since the cosmos-sdk debug level is practically useless (full of cosmos-sdk internal logs), we chose to use the info level for debug logs on-chain.
5. The long-term solution I was imagining was to add a cosmos-sdk logger backed implementation of the [polylog interface](https://github.com/pokt-network/poktroll/blob/main/pkg/polylog/interface.go#L26) that would shunt the existing debug level logs to a new level (e.g. `cosmos-debug` or something) and give us back a useful debug level.

x/gateway/keeper/msg_server_stake_gateway.go
bryanchriswhite: see https://github.com/pokt-network/poktroll/pull/843#discussion_r1782233435

x/gateway/keeper/msg_server_stake_gateway_test.go
bryanchriswhite: My personal preference is that the test be sensitive to the error message changing as this could be indicative of a couple things:

1. Business logic changed such that a different error is returned for a given scenario. In many cases, this could be the same error type (e.g. `ErrAppParamInvalid`) which is wrapping (`.Wrapf(...)`) a message from a different error path. 

    This could lead to false negatives (i.e. tests which pass but should not) appearing over time as the codebase evolves.
2. The error or wrapped message was changed. 

    In this case, I think we should want to have coverage over the error message such that we're confident that our expectation of how it looks is accurate. We're likely going to encounter these logs and error messages in bug reports, which need to be immediately useful.

x/gateway/keeper/msg_server_stake_gateway.go
bryanchriswhite: Note that the `[Logging]` section of #384 is outstanding. 

Regarding automating/improving alignment with this approach: I could imagine a CI task which collects all the non debug logs added (plus a few lines of context) in a PR and leaves them in a comment.

red-0ne (approved):
Let's merge it! :+1:

x/gateway/keeper/msg_server_stake_gateway.go
bryanchriswhite, line 40:
```suggestion
		logger.Info(fmt.Sprintf("ERROR: could not parse address %q", msg.Address))
```

x/gateway/keeper/msg_server_stake_gateway.go
```diff
@@ -24,46 +26,67 @@ func (k msgServer) StakeGateway(
...
+		return nil, status.Error(codes.InvalidArgument, err.Error())
 	}
 
 	// Check if the gateway already exists or not
-	var err error
 	var coinsToEscrow sdk.Coin
 	gateway, isGatewayFound := k.GetGateway(ctx, msg.Address)
 	if !isGatewayFound {
-		logger.Info(fmt.Sprintf("Gateway not found. Creating new gateway for address %q", msg.Address))
+		logger.Info(fmt.Sprintf("gateway not found. Creating new gateway for address %q", msg.Address))
```
bryanchriswhite, line 48:
```suggestion
		logger.Info(fmt.Sprintf("gateway not found; creating new gateway for address %q", msg.Address))
```

x/gateway/keeper/msg_server_stake_gateway.go
```diff
@@ -24,46 +26,67 @@ func (k msgServer) StakeGateway(
...
 		gateway = k.createGateway(ctx, msg)
 		coinsToEscrow = *msg.Stake
 	} else {
-		logger.Info(fmt.Sprintf("Gateway found. About to try and update gateway for address %q", msg.Address))
+		logger.Info(fmt.Sprintf("gateway found. About to try and update gateway for address %q", msg.Address))
```
bryanchriswhite, line 52:
```suggestion
		logger.Info(fmt.Sprintf("gateway found; about to try and update gateway for address %q", msg.Address))
```

x/gateway/keeper/msg_server_stake_gateway.go
```diff
@@ -24,46 +26,67 @@ func (k msgServer) StakeGateway(
...
+	// MUST ALWAYS stake or upstake (> 0 delta).
 	if coinsToEscrow.IsZero() {
-		logger.Warn(fmt.Sprintf("Gateway %q must escrow more than 0 additional coins", msg.Address))
-		return nil, types.ErrGatewayInvalidStake.Wrapf("gateway %q must escrow more than 0 additional coins", msg.Address)
+		errMsg := fmt.Sprintf("gateway %q must escrow more than 0 additional coins", msg.GetAddress())
+		logger.Info(errMsg)
+		return nil, status.Error(
+			codes.InvalidArgument,
+			types.ErrGatewayInvalidStake.Wrap(errMsg).Error(),
+		)
```
bryanchriswhite, line 78:
```suggestion
		err := types.ErrGatewayInvalidStake.Wrap("gateway %q must escrow more than 0 additional coins", msg.GetAddress())
		logger.Info(fmt.Sprintf("ERROR: %s", err)
		return nil, status.Error(codes.InvalidArgument, err.Error())
```

x/gateway/keeper/msg_server_stake_gateway.go
```diff
@@ -24,46 +26,67 @@ func (k msgServer) StakeGateway(
...
-		return nil, err
+	// MUST ALWAYS have at least minimum stake.
+	minStake := k.GetParams(ctx).MinStake
+	if msg.Stake.Amount.LT(minStake.Amount) {
+		errFmt := "gateway %q must stake at least %s"
+		logger.Info(fmt.Sprintf(errFmt, msg.Address, minStake))
+		return nil, status.Error(
+			codes.InvalidArgument,
+			types.ErrGatewayInvalidStake.Wrapf(errFmt, msg.Address, minStake).Error(),
+		)
```
bryanchriswhite, line 89:
```suggestion
		err := types.ErrGatewayInvalidStake.Wrapf("gateway %q must stake at least %s", msg.Address, minStake)
		logger.Info(fmt.Sprintf("ERROR: %s", err))
		return nil, status.Error(codes.InvalidArgument, err.Error())
```

x/gateway/keeper/msg_server_stake_gateway.go
bryanchriswhite: I've updated the logs that I changed to follow this pattern:

```go
err := examplemodtypes.ErrExamplemodSomeError(...)
logger.Info(fmt.Sprintf("ERROR: %s", err)
return nil, status.Error(codes.InvalidArgument, error.Error())
```

Olshansk (approved):
Small comment add but LGTM otherwise.

x/gateway/keeper/msg_server_stake_gateway.go
```diff
@@ -24,46 +26,61 @@ func (k msgServer) StakeGateway(
...
-		return nil, err
+		return nil, status.Error(codes.InvalidArgument, err.Error())
+	}
+
+	// Retrieve the address of the gateway
+	gatewayAddress, err := sdk.AccAddressFromBech32(msg.Address)
+	// NB: This SHOULD NEVER happen because msg.ValidateBasic() validates the address as bech32.
+	if err != nil {
+		// TODO_TECHDEBT(#384): determine whether to continue using cosmos logger for debug level.
+		logger.Info(fmt.Sprintf("ERROR: could not parse address %q", msg.Address))
```
Olshansk, line 40: Calling out to revisit the `logger.Info` vs `logger.Debug`

Olshansk (approved):

Issue #859, opened 2024-10-04
[On-Chain] Use `AsUint64` in `MsgUpdateParam` where appropriate


bryanchriswhite:
## Objective

Add support for a `uint64` `MsgUpdateParam#AsType` type for use with `uint64` typed params.

## Origin Document

![image](https://github.com/user-attachments/assets/638f7757-2506-4269-a021-dbcc8d658868)

## Goals

- Untangle `int64` and `uint64` param update logic across all modules.
- Improve consistency and readibility across param update logic.

## Deliverables

- [ ] A PR for each module with currently uses `AsInt64` for `uint64` param types which adds support for `AsUint64` for that module and replaces `AsInt64` usage where appropriate.

## Non-goals / Non-deliverables

- Adding params or `MsgUpdateParam` to any module(s).
- Any additional param or test refactoring beyond what's necessary to satisfy the goals.

## General deliverables

- [ ] **Comments**: Add/update TODOs and comments alongside the source code so it is easier to follow.
- [ ] **Testing**: Add new tests (unit and/or E2E) to the test suite.
- [ ] **Makefile**: Add new targets to the Makefile to make the new functionality easier to use.
- [ ] **Documentation**: Update architectural or development READMEs; use [mermaid](https://mermaid-js.github.io/mermaid/) diagrams where appropriate.

---

**Creator**: @bryanchriswhite 
**Co-Owners**: @Olshansk

PR #863, opened 2024-10-07
[On-Chain] Refactor `uint64` type individual param update logic

bryanchriswhite:
## Summary

In the context of individual param updates (`MsgUpdateParam`), `uint64` typed params were previously supported via an `MsgUpdateParam_AsInt64` type for obsolete reasons. This PR aligns all `uint64` type params with the corresopnding `MsgUpdateParam#AsType`s using a new `MsgUpdateParam_AsUint64` type. 

## Issue

- #859

## Type of change

- [ ] Bug fix
- [x] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

## Sanity Checklist


red-0ne (approved):
Great to see us moving to uint64.

I see `AsString` being removed. Is it because it just happens that we did not need it and that we can re-introduce it once we have string params? (same for int64 itself)

Approving.

bryanchriswhite:
> I see `AsString` being removed. Is it because it just happens that we did not need it and that we can re-introduce it once we have string params? (same for int64 itself)

Yes, string, int64, and uint64 were all not being used in proof module params so they were removed from the corresponding `MsgUpdateParam_As...` types.

I believe that no module actually uses the string type so it might have also been removed from additional cases.

bryanchriswhite:
E2E tests passing locally but stalled in CI:
image

PR #865, opened 2024-10-07
[Dependencies] bump cosmos-sdk to v0.50.10

okdas:
## Summary

Bumps `cosmos-sdk` to the latest version, along with its dependencies.

## Type of change

`go get github.com/cosmos/cosmos-sdk@v0.50.10`

- [ ] Bug fix
- [x] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

## Sanity Checklist


okdas:
After this is merged, protocol team members need to run to replace [deprecated mockgen](https://github.com/golang/mock?tab=readme-ov-file) with uber fork.

`go install "go.uber.org/mock/mockgen@v0.4.0" && mockgen --version`

okdas:
Unable to generate mocks, even after upgrading to uber's mockgen (the supported one).

Will need to come back to investigate further. It seems like a change in `v0.50.10`.

[50.9..50.10 changes here](https://github.com/cosmos/cosmos-sdk/compare/v0.50.9...v0.50.10).

The error I'm getting:
```
Reflection: can't yet turn interface { ProtoMessage(); Reset(); String() string } (interface) into a model.Type
Reflection: can't yet turn interface { ProtoMessage(); Reset(); String() string } (interface) into a model.Type
prog.go:15:2: no required module provides package github.com/pokt-network/poktroll/pkg/client: go.mod file not found in current directory or any parent directory; see 'go help modules'
prog.go:13:2: no required module provides package go.uber.org/mock/mockgen/model: go.mod file not found in current directory or any parent directory; see 'go help modules'
2024/10/07 16:58:06 Loading input failed: exit status 1
pkg/client/interface.go:4: running "mockgen": exit status 1
```

bryanchriswhite:
### Preliminary observations:

Given the output, `mockgen` seems to be struggling with something that appears to be the [`cosmostypes.Msg`](https://pkg.go.dev/github.com/cosmos/cosmos-sdk@v0.50.10/types#Msg) / [`proto.Message`](https://pkg.go.dev/github.com/cosmos/gogoproto/proto#Message) interface:

```
{ ProtoMessage(); Reset(); String() string }
```

I've eliminated all `mockgen` calls as being problematic except for these two in pkg/client/interfaces.go:

```go
//go:generate mockgen -destination=../../testutil/mockclient/tx_client_mock.go -package=mockclient . TxContext,TxClient
// ...
//go:generate mockgen -destination=../../testutil/mockclient/cosmos_tx_builder_mock.go -package=mockclient github.com/cosmos/cosmos-sdk/client TxBuilder
```

Specifically, it seems to be the `TxClient` and `TxBuilder` types, which makes sense given they both use `cosmostypes.Msg`:

```go
type TxClient interface {
	SignAndBroadcast(
		ctx context.Context,
		msgs ...cosmostypes.Msg,
	) either.AsyncError
}
```
> [pkg/client/interface.go:81](https://github.com/pokt-network/poktroll/blob/main/pkg/client/interface.go#L81)

```go
	TxBuilder interface {
		GetTx() signing.Tx

		SetMsgs(msgs ...sdk.Msg) error
		// ...
	}
```
> [godocs](https://pkg.go.dev/github.com/cosmos/cosmos-sdk@v0.50.10/client#TxBuilder)

Investigating the mockgen source seems to indicate that it [can only handle](https://github.com/uber-go/mock/blob/main/mockgen/model/model.go#L469) interface types which either have no methods or are an `error`. However, the original mockgen seems to have [the same code](https://github.com/golang/mock/blob/v1.6.0/mockgen/model/model.go#L431) in this area (even in v1.6.0, what we were using); more context is required to understand why things are working differently in the uber fork. :thinking:

bryanchriswhite:
### More observations

I've been experimenting with the `TxClient` interface:

- Switching between `cosmostypes.Msg` and `proto.Message` doesn't seem to make any difference.
- Defining a new interface that either embeds or implements `proto.Message` makes the error go away.

    While this might be a slightly ugly workaround for `TxClient`, this doesn't solve for `TxBuilder` as it's buried in cosmos-sdk source.
- At this point, my next thought would be to explore using source mode instead of reflect mode but again, this doesn't solve for `TxBuilder` unless we start cloning or otherwise pulling down cosmos-sdk source files for development and testing. :roll_eyes:

okdas:
@Olshansk I've noticed this PR is closed now. What do you think we should do next to unblock ourselves from `cosmos-sdk` upgrades? I suspect all future `cosmos-sdk` will have the same side effect for us.

I can open a new GitHub issue so we account for that work in the future.

PR #876, opened 2024-10-11
[Docs] Fix example of how to set rev share precentage

Olshansk:
The existing docs for setting `rev_share_percentage` resulted in this error: `"cannot unmarshal !!seq into map[string]float32"`


Olshansk:
See https://github.com/pokt-network/poktroll/pull/877

PR #877, opened 2024-10-11
[Docs] Fix example of how to set rev share precentage

Olshansk:
The existing docs for setting `rev_share_percentage` resulted in this error: `"cannot unmarshal !!seq into map[string]float32"`

### DIFF
--- a/localnet/poktrolld/config/supplier1_stake_config.yaml
+++ b/localnet/poktrolld/config/supplier1_stake_config.yaml
@@ -2,8 +2,8 @@ owner_address: pokt1mrqt5f7qh8uxs27cjm9t7v9e74a9vvdnq5jva4
 operator_address: pokt1mrqt5f7qh8uxs27cjm9t7v9e74a9vvdnq5jva4
 stake_amount: 1000069upokt
 default_rev_share_percent:
-  - pokt1mrqt5f7qh8uxs27cjm9t7v9e74a9vvdnq5jva4: 80.5
-  - pokt1eeeksh2tvkh7wzmfrljnhw4wrhs55lcuvmekkw: 19.5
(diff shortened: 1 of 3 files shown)


localnet/poktrolld/config/supplier1_stake_config.yaml
```diff
@@ -2,17 +2,17 @@ owner_address: pokt1mrqt5f7qh8uxs27cjm9t7v9e74a9vvdnq5jva4
...
+  pokt1eeeksh2tvkh7wzmfrljnhw4wrhs55lcuvmekkw: 19.5
 services:
   - service_id: anvil
     endpoints:
       - publicly_exposed_url: http://relayminer1:8545
         rpc_type: JSON_RPC
   - service_id: ollama
     rev_share_percent:
-      - pokt1mrqt5f7qh8uxs27cjm9t7v9e74a9vvdnq5jva4: 50
-      - pokt1eeeksh2tvkh7wzmfrljnhw4wrhs55lcuvmekkw: 50
```
red-0ne, line 15: Thanks for the fix. I wonder how this did work in the past :thinking:

red-0ne (approved):

localnet/poktrolld/config/supplier1_stake_config.yaml
red-0ne: It did not work :sweat_smile:

Olshansk: How did you test it?

red-0ne: These specific config files are not being used. E2E tests are generating their own.

PR #878, opened 2024-10-14
[Tokenomics] Implement Global Mint Reimbursement Request

red-0ne:
## Summary

This PR adds
* GMRR TLM
* Ensures that the application has enough funds to cover for it.
* Deducts the total global mint amount from the application's stake.
* Sends the global mint amount from the application module account to the tokenomics module account to the PNF account.
* The application still gets its global inflation share sent to its account.

## Issue

- #732 

## Type of change

## Testing

## Sanity Checklist


x/tokenomics/keeper/token_logic_modules.go
```diff
@@ -468,6 +477,95 @@ func (k Keeper) TokenLogicModuleGlobalMint(
...
+	}
+
+	// Determine how much new uPOKT to mint based on global inflation
+	newMintCoin, _ := calculateGlobalPerClaimMintInflationFromSettlementAmount(actualSettlementCoin)
+	if newMintCoin.Amount.Int64() == 0 {
+		return tokenomicstypes.ErrTokenomicsMintAmountZero
+	}
+
+	// Update the application's on-chain stake
+	newAppStake, err := application.Stake.SafeSub(newMintCoin)
```
bryanchriswhite, line 506: I'm curious to understand if/how this interacts with the application min. stake and/or `maxClaimableAmount`. :thinking: 

_TL;DR, it seems to me like min. stake SHOULD be collateral for app overservicing AND GMRR. Is this that unnecessary? Do we need to update `MaxClaimableAmount`?_

`MaxClaimableAmount` prevents `actualSettlementCoin` from exceeding `application stake - min. stake` (once we add the unsettled sessions coefficient :sweat_smile:) but it seems like with GMRR we would also need to also account for the potential magnitude of `actualSettlementCoin + newMintCoin`. Wouldn't min. stake actually need to cover `actualSettlementCoin + newMintCoin` for **each claim** in **each session** between the current and currently settling sessions?

Since `newMintCoin` is a function of `actualSettlementCoin`, it seems like we should be able to consider it in the same scope as `maxClaimableAmount` if that's the way to go here.

bryanchriswhite (commented):
image

tests/integration/application/min_stake_test.go
```diff
@@ -83,6 +85,12 @@ func (s *applicationMinStakeTestSuite) TestAppIsUnbondedIfBelowMinStakeWhenSettl
 	// Stake a supplier for service 1.
 	s.stakeSupplier()
 
+	proofParams := s.keepers.ProofKeeper.GetParams(s.ctx)
+	proofParams.ProofRequestProbability = 0
+	proofRequirementThreshold := cosmostypes.NewInt64Coin(volatile.DenomuPOKT, math.MaxInt64)
+	proofParams.ProofRequirementThreshold = &proofRequirementThreshold
+	s.keepers.ProofKeeper.SetParams(s.ctx, proofParams)
```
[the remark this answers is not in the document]
red-0ne: Brilliant! :D

testutil/keeper/tokenomics.go
```diff
@@ -508,3 +513,24 @@ func WithProposerAddr(addr string) TokenomicsModuleKeepersOpt {
...
+			proofParams.ProofRequestProbability = 1
+			proofRequirementThreshold := cosmostypes.NewInt64Coin(volatile.DenomuPOKT, 0)
+			proofParams.ProofRequirementThreshold = &proofRequirementThreshold
+		} else {
+			proofParams.ProofRequestProbability = 0
+			proofRequirementThreshold := cosmostypes.NewInt64Coin(volatile.DenomuPOKT, math.MaxInt64)
+			proofParams.ProofRequirementThreshold = &proofRequirementThreshold
+		}
+
+		keepers.ProofKeeper.SetParams(ctx, proofParams)
```
bryanchriswhite, line 532: I see two options here:

1. Update `TokenomicsModuleKeepersOpt` to receive a `*testing.T` such that you can call `require.NoError(t, err)` here.
2. `panic(err)`; if it's going to panic, it'll happen on setup.

x/tokenomics/keeper/token_logic_modules.go
```diff
@@ -468,6 +477,95 @@ func (k Keeper) TokenLogicModuleGlobalMint(
...
+	if err != nil {
```
bryanchriswhite, line 507: I think we should reduce the app stake to zero in this case. Though we return an error here, lower in the stack, we will be logging and continuing to avoid chain halts. In this context, I think we MUST escrow the "collateral" in this case; otherwise, I don't think we're sufficiently mitigating a colluding application's ability to skew inflation.

x/tokenomics/keeper/token_logic_modules.go
```diff
@@ -468,6 +477,95 @@ func (k Keeper) TokenLogicModuleGlobalMint(
...
+		SessionId:       sessionHeader.SessionId,
+		Amount:          &newMintCoin,
+	}
+
+	eventManger := cosmostypes.UnwrapSDKContext(ctx).EventManager()
+	if err := eventManger.EmitTypedEvent(reimbursementRequestEvent); err != nil {
+		return tokenomicstypes.ErrTokenomicsEmittingEventFailed.Wrapf(
+			"error emitting event %v",
+			reimbursementRequestEvent,
+		)
```
bryanchriswhite, line 563:
```suggestion
		return tokenomicstypes.ErrTokenomicsEmittingEventFailed.Wrapf(
			"(%+v): %s", reimbursementRequestEvent, err
		)
```

x/tokenomics/keeper/token_logic_modules.go
```diff
@@ -557,13 +655,45 @@ func (k Keeper) ensureClaimAmountLimits(
...
 	appStake := application.GetStake()
 
+	// TODO_BETA(red-0ne): Make sure that the relay miner logic also accounts for this
+	// when deciding to serve an application.
+
+	// The application should have enough stake to cover for the global mint reimbursement.
+	// This amount is deducted from the maximum claimable amount.
+	globalInflationCoin, _ := calculateGlobalPerClaimMintInflationFromSettlementAmount(claimSettlementCoin)
+
+	// TODO_BETA(@red-0ne): Introduce a session sliding window to account for potential consumption
```
bryanchriswhite, line 670: :eyes::ok_hand:

x/tokenomics/keeper/token_logic_modules.go
```diff
@@ -557,13 +655,45 @@ func (k Keeper) ensureClaimAmountLimits(
...
+	// - Application already requested work for session 3
+	// Problem:
+	// - If the application consumes its entire stake in settlement of session 2
+	// - Then over-servicing in session 3 (i.e. No stake left to consume)
+	// Solution:
+	// - By dividing the claimable stake by 2 (3 - 2 + 1), settling session 2 assumes that
+	//   the application will consume its maxClaimableAmt the current session (3).
+	// - Off-chain actors could use this formula during the servicing of session num 3
+	//   and assume maxClaimableAmt will be settled in session 2.
+	// - Garantee no over-servicing at the cost of higher application stake requirements.
```
bryanchriswhite, line 687:
```suggestion
	// - Guarantee no over-servicing at the cost of higher application stake requirements.
```

x/tokenomics/keeper/token_logic_modules.go
```diff
@@ -557,13 +655,45 @@ func (k Keeper) ensureClaimAmountLimits(
...
+	maxCalibmableAmt := appStake.Amount.Quo(math.NewInt(sessionkeeper.NumSupplierPerSession)).Sub(globalInflationCoin.Amount)
```
bryanchriswhite, line 688:
```suggestion
	maxClaimableAmt := appStake.Amount.Quo(math.NewInt(sessionkeeper.NumSupplierPerSession)).Sub(globalInflationCoin.Amount)
```

bryanchriswhite (changes requested):
Nice one! :raised_hands: 

I just have a few more questions/suggestions. :sweat_smile:

proto/poktroll/tokenomics/event.proto
```diff
@@ -79,4 +79,13 @@ message EventSupplierSlashed {
     // Amount slashed from the supplier's stake due to the expired claims.
     // This is a function of the number of expired claims and proof missing penalty.
     cosmos.base.v1beta1.Coin slashing_amount = 3;
+}
+
+// EventApplicationReimbursementRequest is emitted when an application requests
+// a reimbursement
+message EventApplicationReimbursementRequest {
```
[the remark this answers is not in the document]
Olshansk: +1 to this

x/tokenomics/keeper/token_logic_modules.go
```diff
@@ -105,8 +114,9 @@ type TokenLogicModuleProcessor func(
...
 // tokenLogicModuleProcessorMap is a map of TLMs to their respective independent processors.
 var tokenLogicModuleProcessorMap = map[TokenLogicModule]TokenLogicModuleProcessor{
-	TLMRelayBurnEqualsMint: Keeper.TokenLogicModuleRelayBurnEqualsMint,
-	TLMGlobalMint:          Keeper.TokenLogicModuleGlobalMint,
+	TLMRelayBurnEqualsMint:            Keeper.TokenLogicModuleRelayBurnEqualsMint,
+	TLMGlobalMint:                     Keeper.TokenLogicModuleGlobalMint,
+	TLMGlobalMintReimbursementRequest: Keeper.TokenLogicModuleGlobalMintReimbursementRequest,
 }
 
 func init() {
```
Olshansk, line 122: Please add a compile-time enforcement in `init` that requests BOTH `TLMGlobalMintReimbursementRequest` and `TLMGlobalMint` to exist.

x/tokenomics/keeper/token_logic_modules.go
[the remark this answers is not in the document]
Olshansk: > I think we should reduce the app stake to zero in this case.

This should simply never happen. Otherwise we get into weird edge cases w.r.t collusion, self-dealing, etc...

GlobalMint should cap at `min(appStake, globalMint)` and we should log/emit the event appropriately.

testutil/keeper/tokenomics.go
```diff
@@ -508,3 +513,24 @@ func WithProposerAddr(addr string) TokenomicsModuleKeepersOpt {
 		return sdkCtx
 	}
 }
+
+// WithProofRequirement is an option to set the proof requirement in the tokenomics module keepers.
+func WithProofRequirement(required bool) TokenomicsModuleKeepersOpt {
```
Olshansk, line 518:
```suggestion
func WithProofRequirement(proofRequired bool) TokenomicsModuleKeepersOpt {
```

x/tokenomics/keeper/token_logic_modules_test.go
```diff
@@ -399,14 +399,15 @@ func TestProcessTokenLogicModules_TLMGlobalMint_Valid_MintDistributionCorrect(t
```
Olshansk, line 399: Depending on the outcome of this: https://github.com/pokt-network/poktroll/pull/878/files#r1810783728

Can you add a dedicated (short) test for it?

x/tokenomics/keeper/token_logic_modules.go
```diff
@@ -468,6 +477,95 @@ func (k Keeper) TokenLogicModuleGlobalMint(
...
+			application.Address, amountDiffCoin,
+		)
+	}
+	application.Stake = &newAppStake
+	logger.Info(fmt.Sprintf("updated application %q stake to %s", application.Address, newAppStake))
+
+	// Send the global per claim mint inflation uPOKT from the application module
+	// account to the tokenomics module account.
+	if err = k.bankKeeper.SendCoinsFromModuleToModule(
+		ctx, apptypes.ModuleName, tokenomicstypes.ModuleName, sdk.NewCoins(actualSettlementCoin),
```
Olshansk, line 520: Why are we using `actualSettlementCoin` instead of `newMintCoin` here and below?

x/tokenomics/keeper/token_logic_modules.go
```diff
@@ -468,6 +477,95 @@ func (k Keeper) TokenLogicModuleGlobalMint(
...
+		amountDiffCoin := actualSettlementCoin.Amount.Sub(application.Stake.Amount)
+		return tokenomicstypes.ErrTokenomicsApplicationReimbursementRequestFailed.Wrapf(
+			"application %q stake cannot be reduced to a negative amount -%s",
+			application.Address, amountDiffCoin,
+		)
+	}
+	application.Stake = &newAppStake
+	logger.Info(fmt.Sprintf("updated application %q stake to %s", application.Address, newAppStake))
+
+	// Send the global per claim mint inflation uPOKT from the application module
```
Olshansk, line 517: The comment and the logic seem to differ.

x/tokenomics/keeper/token_logic_modules.go
```diff
@@ -468,6 +477,95 @@ func (k Keeper) TokenLogicModuleGlobalMint(
...
+	// Determine how much new uPOKT to mint based on global inflation
+	newMintCoin, _ := calculateGlobalPerClaimMintInflationFromSettlementAmount(actualSettlementCoin)
+	if newMintCoin.Amount.Int64() == 0 {
+		return tokenomicstypes.ErrTokenomicsMintAmountZero
+	}
+
+	// Update the application's on-chain stake
+	newAppStake, err := application.Stake.SafeSub(newMintCoin)
+	if err != nil {
+		amountDiffCoin := actualSettlementCoin.Amount.Sub(application.Stake.Amount)
```
Olshansk, line 508: After we correctly compute `newMintCoin`, don't we stop using `actualSettlementCOin` altogether?

x/tokenomics/keeper/token_logic_modules.go
Olshansk, line 670: Looks like something you'll do in MainNet?

Olshansk (changes requested):
@red-0ne Submitting a partial review to ask some questions, but will do a deeper one later.

If we need to align offline (call or other) lmk.

x/tokenomics/keeper/token_logic_modules.go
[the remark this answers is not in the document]
red-0ne: `ensureClaimAmountLimits` should make sure to adjust everything to get a positive new app stake.

If we error here this would mean that `ensureClaimAmountLimits` failed its calculation and I think we should error.

This is just a safety check that the "updated" application stake does not fall below 0.

Changes in 568ecb9b make `ensureClaimAmountLimits` enforce these checks.

I don't mind removing these app stake tests altogether.

x/tokenomics/keeper/token_logic_modules_test.go
red-0ne: `TestProcessTokenLogicModules_TLMBurnEqualsMint_Valid_SupplierExceedsMaxClaimableAmount` has been adjusted and describes exactly that even if the claim is over max claimable, an event is emitted for over-charging and max claimable is effective.

Will add a TODO_TEST for edge values though.

x/tokenomics/keeper/token_logic_modules.go
red-0ne: My bad!

x/tokenomics/keeper/token_logic_modules.go
red-0ne: This is important, it's in next week's tasks.

testutil/keeper/tokenomics.go
```diff
@@ -2,10 +2,11 @@ package keeper
 
 import (
 	"context"
+	"math"
 	"testing"
 
 	"cosmossdk.io/log"
-	"cosmossdk.io/math"
+	sdkmath "cosmossdk.io/math"
```
bryanchriswhite, line 9: nit: Any objection to `cosmosmath` to disambiguate between our sdk and cosmos-sdk?

x/tokenomics/keeper/token_logic_modules_test.go
```diff
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	math2 "math"
 	"testing"
 
 	"cosmossdk.io/math"
```
bryanchriswhite, line 10:
```suggestion
	"math"
	"testing"

	cosmosmath "cosmossdk.io/math"
```

x/tokenomics/keeper/token_logic_modules.go
```diff
@@ -115,8 +125,11 @@ func init() {
 		panic("mint allocation percentages do not add to 1.0")
 	}
 
-	// TODO_UPNEXT(@Olshansk): Ensure that if `TLMGlobalMint` is present in the map,
-	// then TLMGlobalMintReimbursementRequest will need to be there too.
+	_, hasGlobalMintTLM := tokenLogicModuleProcessorMap[TLMGlobalMint]
+	_, hasGlobalMintReimbursementRequestTLM := tokenLogicModuleProcessorMap[TLMGlobalMintReimbursementRequest]
+	if hasGlobalMintTLM != hasGlobalMintReimbursementRequestTLM {
+		panic("TLMGlobalMint and TLMGlobalMintReimbursementRequest must be activated together")
+	}
```
bryanchriswhite, line 132: Nice! :ok_hand: 

```suggestion
	_, hasGlobalMintTLM := tokenLogicModuleProcessorMap[TLMGlobalMint]
	_, hasGlobalMintReimbursementRequestTLM := tokenLogicModuleProcessorMap[TLMGlobalMintReimbursementRequest]
	if hasGlobalMintTLM != hasGlobalMintReimbursementRequestTLM {
		panic("TLMGlobalMint and TLMGlobalMintReimbursementRequest must be (de-)activated together")
	}
```

x/tokenomics/keeper/token_logic_modules.go
```diff
@@ -482,6 +495,93 @@ func (k Keeper) TokenLogicModuleGlobalMint(
...
+		)
+	}
+	logger.Info(fmt.Sprintf(
+		"sent (%s) from the application module account to the tokenomics module account",
+		newMintCoin,
+	))
+
+	// Send the global per claim mint inflation uPOKT from the tokenomics module
+	// account to PNF/DAO.
+	daoAccountAddr, err := cosmostypes.AccAddressFromBech32(k.GetAuthority())
```
bryanchriswhite, line 548: TL;DR, (It doesn't have to be in this PR but) we either need to align on an on-chain DAO address (see #889 for my proposal) OR align the dao_genesis_authorizations.json on the requirements of using the gov module account instead.

I saw something like this line somewhere else but have since lost track of it. This confuses me a bit because we've **opted to use authz to delegate authority over the (default - gov module, and configurable) "authority" account, instead of setting any particular address in the module config**. As such, it seems incorrect to me that we're sending the DAO's rewards to the governance module account.

This seems to me, at best, to be a temporary measure to be iterated on. Furthermore, we are currently missing an authz authorization in the dao_genesis_authorizations.json for the dao address to be able to do anything other than update parameters and manage upgrades on behalf of the gov module account. It stands to reason that if we're going to be holding DAO rewards in the gov module account, then we should probably add genesis authorizations for the DAO to use the `cosmos.bank.v1beta1.MsgSend` msg as well.

x/tokenomics/keeper/token_logic_modules.go
```diff
@@ -482,6 +495,93 @@ func (k Keeper) TokenLogicModuleGlobalMint(
...
+		SessionId:            sessionHeader.SessionId,
+		Amount:               &newMintCoin,
+	}
+
+	eventManger := cosmostypes.UnwrapSDKContext(ctx).EventManager()
+	if err := eventManger.EmitTypedEvent(reimbursementRequestEvent); err != nil {
+		return tokenomicstypes.ErrTokenomicsEmittingEventFailed.Wrapf(
+			"(%+v): %s",
+			reimbursementRequestEvent, err,
+		)
```
bryanchriswhite, line 579: I've been following this pattern but am open to feedback:
```suggestion
		err = tokenomicstypes.ErrTokenomicsEmittingEventFailed.Wrapf(
			"(%+v): %s",
			reimbursementRequestEvent, err,
		)
		logger.Error(err.Error())
		return err
```

x/tokenomics/keeper/settle_pending_claims.go
```diff
@@ -35,6 +36,26 @@ func (k Keeper) SettlePendingClaims(ctx sdk.Context) (
 		return settledResult, expiredResult, err
 	}
 
+	// Capture the applications initial stake which will be used to calculate the
+	// max share any claim could receive from the application stake.
+	// This ensures that each claim can calculate the maximum amount it can take
+	// from an application's stake.
```
bryanchriswhite, line 42: Would you you taking another pass at this? I'm not sure if my suggestion aligns with what you were hoping to communicate:

```suggestion
	// Capture the applications initial stake which will be used to calculate the
	// max share any claim could burn from the application stake.
	// This ensures that each supplier can calculate the maximum amount it can take
	// from an application's stake.
```

x/tokenomics/keeper/settle_pending_claims.go
```diff
@@ -35,6 +36,26 @@ func (k Keeper) SettlePendingClaims(ctx sdk.Context) (
...
+		}
+
+		app, isAppFound := k.applicationKeeper.GetApplication(ctx, appAddress)
+		if !isAppFound {
+			err = apptypes.ErrAppNotFound.Wrapf("application address: %q", appAddress)
+			return settledResult, expiredResult, err
+		}
+
+		applicationInitialStakeMap[appAddress] = *app.GetStake()
+	}
```
bryanchriswhite, line 57: Wdyt about extracting this to a a keeper method?

x/tokenomics/keeper/settle_pending_claims.go
```diff
@@ -176,8 +197,11 @@ func (k Keeper) SettlePendingClaims(ctx sdk.Context) (
 		// 1. The claim does not require a proof.
 		// 2. The claim requires a proof and a valid proof was found.
 
+		claimApplication := claim.SessionHeader.ApplicationAddress
```
bryanchriswhite, line 200: nit: in the context of settling a particular claim, there is only one "application", no? Is there a need to disambiguate this application from another? Did I miss something?

```suggestion
		app := claim.GetSessionHeader().GetApplicationAddress()
```

bryanchriswhite: Wdyt about adding a TODO here about subscribing to app stake events to increment balances inbetween session settlements?

bryanchriswhite (changes requested):
I could't quite make it to the end of the last file but I got through 90%+. :fire: 

image


Great work @red-0ne! :raised_hands:

x/tokenomics/keeper/token_logic_modules.go
```diff
@@ -482,6 +495,93 @@ func (k Keeper) TokenLogicModuleGlobalMint(
...
+
+	// Update the application's on-chain stake
+	// This should not fall below zero since `ensureClaimAmountLimits` should have
+	// already checked and adjusted the settlement amount so that the application
+	// stake covers the global inflation.
+	newAppStake := application.Stake.Sub(newMintCoin)
+	application.Stake = &newAppStake
+	logger.Info(fmt.Sprintf("updated application %q stake to %s", application.Address, newAppStake))
+
+	// Send the global per claim mint inflation uPOKT from the application module
```
Olshansk, line 531: Let's create a variable called `globalInflationMintedCoinsForClaim := sdk.NewCoins(newMintCoin)` and use it below.

It'll have a good semantic meaning.

x/tokenomics/keeper/token_logic_modules.go
```diff
@@ -482,6 +495,93 @@ func (k Keeper) TokenLogicModuleGlobalMint(
...
+			"sending %s from the application module account to the tokenomics module account: %v",
+			newMintCoin, err,
+		)
+	}
+	logger.Info(fmt.Sprintf(
+		"sent (%s) from the application module account to the tokenomics module account",
+		newMintCoin,
+	))
+
+	// Send the global per claim mint inflation uPOKT from the tokenomics module
```
Olshansk, line 546: Please move this code block above so all the sends are next to each (easier to read)

x/tokenomics/keeper/token_logic_modules.go
```diff
@@ -482,6 +495,93 @@ func (k Keeper) TokenLogicModuleGlobalMint(
...
+	// Update the application's on-chain stake
+	// This should not fall below zero since `ensureClaimAmountLimits` should have
+	// already checked and adjusted the settlement amount so that the application
+	// stake covers the global inflation.
+	newAppStake := application.Stake.Sub(newMintCoin)
+	application.Stake = &newAppStake
+	logger.Info(fmt.Sprintf("updated application %q stake to %s", application.Address, newAppStake))
+
+	// Send the global per claim mint inflation uPOKT from the application module
+	// account to the tokenomics module account.
```
Olshansk, line 532:
```suggestion
	// account to the tokenomics module account as an intermediary step.
```

x/tokenomics/keeper/token_logic_modules.go
```diff
@@ -482,6 +495,93 @@ func (k Keeper) TokenLogicModuleGlobalMint(
...
+	// Send the global per claim mint inflation uPOKT from the tokenomics module
+	// account to PNF/DAO.
+	daoAccountAddr, err := cosmostypes.AccAddressFromBech32(k.GetAuthority())
+	if err != nil {
+		return tokenomicstypes.ErrTokenomicsApplicationReimbursementRequestFailed.Wrapf(
+			"getting PNF/DAO address: %v",
+			err,
+		)
+	}
+	if err := k.bankKeeper.SendCoinsFromModuleToAccount(
```
Olshansk, line 555: #PUC why we're doing this with reference to https://discord.com/channels/824324475256438814/997192534168182905/1299372745632649408

x/tokenomics/keeper/token_logic_modules.go
```diff
@@ -482,6 +495,93 @@ func (k Keeper) TokenLogicModuleGlobalMint(
...
+		ctx, tokenomicstypes.ModuleName, daoAccountAddr, sdk.NewCoins(newMintCoin),
```
Olshansk, line 556: Use the new variable here

x/tokenomics/keeper/token_logic_modules.go
```diff
@@ -482,6 +495,93 @@ func (k Keeper) TokenLogicModuleGlobalMint(
...
+	); err != nil {
+		return tokenomicstypes.ErrTokenomicsApplicationReimbursementRequestFailed.Wrapf(
+			"sending %s from the tokenomics module account to the PNF/DAO account: %v",
+			newMintCoin, err,
+		)
+	}
+
+	// Prepare and emit the event for the application being overcharged.
```
Olshansk, line 564:
```suggestion
	// Prepare and emit the event for the application that'll require reimbursement.
	// Recall that it is being overcharged to compensate for global inflation while
	// preventing self-dealing.
```

x/tokenomics/keeper/token_logic_modules.go
```diff
@@ -555,47 +655,81 @@ func (k Keeper) sendRewardsToAccount(
 // If this is not the case, then the supplier essentially did "free work" and the
 // actual claim amount is less than what was claimed.
 // Ref: https://arxiv.org/pdf/2305.10672
+// TODO_TEST: Add more tests for edge cases that exercise all the code paths
```
Olshansk, line 658: You don't need to add the tests, but can you please create the test headers + comments on what the edge cases are.

NOW is the best time to document them.

x/tokenomics/keeper/token_logic_modules.go
Olshansk: I'm skipping this long comment but will take a look next week

x/tokenomics/keeper/token_logic_modules.go
```diff
@@ -555,47 +655,81 @@ func (k Keeper) sendRewardsToAccount(
...
+	// Note that this also incorporates MintPerClaimGlobalInflation since applications
+	// are being overcharged by that amount and the funds are sent to the DAO/PNF
+	// before being reimbursed to the application in the future.
+	appStake := initialApplicationStake
+
+	// The application should have enough stake to cover for the global mint reimbursement.
+	// This amount is deducted from the maximum claimable amount.
+	globalInflationCoin, _ := calculateGlobalPerClaimMintInflationFromSettlementAmount(claimSettlementCoin)
+	globalInflationAmt := globalInflationCoin.Amount
+	stakeRequirementAmt := claimSettlementCoin.Amount.Add(globalInflationAmt)
```
Olshansk, line 684: Is this basically `minRequiredAppStake`?

If so, can we call it that?

x/tokenomics/keeper/token_logic_modules.go
```diff
@@ -555,47 +655,81 @@ func (k Keeper) sendRewardsToAccount(
...
+	// Solution:
+	// - By dividing the claimable stake by 2 (3 - 2 + 1), settling session 2 assumes that
+	//   the application will consume its maxClaimableAmt the current session (3).
+	// - Off-chain actors could use this formula during the servicing of session num 3
+	//   and assume maxClaimableAmt will be settled in session 2.
+	// - Guarantee no over-servicing at the cost of higher application stake requirements.
+	maxClaimableAmt := appStake.Amount.Quo(math.NewInt(sessionkeeper.NumSupplierPerSession))
+	maxClaimSettlementAmt := stakeToMaxSettlementAmount(maxClaimableAmt)
+
+	// Check if the claimable amount is capped by the max claimable amount.
```
Olshansk, line 708: #PUC

x/tokenomics/keeper/token_logic_modules.go
```diff
@@ -681,6 +815,21 @@ func calculateGlobalPerClaimMintInflationFromSettlementAmount(settlementCoin sdk
 	return mintAmtCoin, *newMintAmtFloat
 }
 
+// stakeToMaxSettlementAmount calculates the max amount of uPOKT to that the supplier
```
Olshansk, line 818: I find this function and comment confusing. 

- We're ONLY dealing with `%` here, no `uPOKT`
- Should `stakeTo` be `stakeShare`?
- Instead of `Amount`, should it be `Multiplier` like `1.2`?


----

After looking at the caller, I see that it's `maxClaimableAmt` being passed in, so maybe that should be the name of the parameter here.

---

Overall, I trust yo're doing the right thing but please revisit the naming of the function, the params, etc, so it's clear to a reader.

Olshansk (changes requested):

x/tokenomics/keeper/token_logic_modules.go
red-0ne: Implemented in #895

x/tokenomics/keeper/settle_pending_claims.go
red-0ne: Implemented in #895

x/tokenomics/keeper/token_logic_modules_test.go
```diff
@@ -640,6 +649,18 @@ func TestProcessTokenLogicModules_InvalidClaim(t *testing.T) {
 	}
 }
 
+func TestProcessTokenLogicModules_AppStakeMissingGlobalInflationAmount(t *testing.T) {
+	t.Skip("TODO: Test application stake that does not cover the global inflation amount")
```
Olshansk, line 653:
```suggestion
	t.Skip("TODO_MAINNET(@red-0ne): Test application stake that is insufficient to cover the global inflation amount, for reimbursement, and ???????{fill in what happens isntead}")
```

x/tokenomics/keeper/token_logic_modules_test.go
```diff
@@ -640,6 +649,18 @@ func TestProcessTokenLogicModules_InvalidClaim(t *testing.T) {
 	}
 }
 
+func TestProcessTokenLogicModules_AppStakeMissingGlobalInflationAmount(t *testing.T) {
```
Olshansk, line 652:
```suggestion
func TestProcessTokenLogicModules_AppStakeInsufficientToCoverGlobalInflationAmount(t *testing.T) {
```

x/tokenomics/keeper/token_logic_modules_test.go
```diff
@@ -640,6 +649,18 @@ func TestProcessTokenLogicModules_InvalidClaim(t *testing.T) {
...
+	t.Skip("TODO: Test application stake that does not cover the global inflation amount")
+}
+
+func TestProcessTokenLogicModules_ZeroAppStakeShareDueToPrecisionLoss(t *testing.T) {
+	t.Skip("TODO: Test application stake that is low enough to have a zero share due to precision loss")
```
Olshansk, line 657: What do you mean by "zero share"? #PUC and test name

Olshansk: Also `TODO_MAINNET(@red-0ne)`

x/tokenomics/keeper/token_logic_modules_test.go
```diff
@@ -649,6 +649,18 @@ func TestProcessTokenLogicModules_InvalidClaim(t *testing.T) {
...
+}
+
+func TestProcessTokenLogicModules_AppStakeGoesBelowMinStake(t *testing.T) {
```
Olshansk, line 660:
```suggestion
func TestProcessTokenLogicModules_AppStakeDropsBelowMinStakeAfterSession(t *testing.T) {
```

x/tokenomics/keeper/token_logic_modules_test.go
```diff
@@ -649,6 +649,18 @@ func TestProcessTokenLogicModules_InvalidClaim(t *testing.T) {
...
+	t.Skip("TODO: Test application stake that goes min stake")
```
Olshansk, line 661:
```suggestion
	t.Skip("TODO_MAINNET(@red-0ne): Test that application stake being auto-unbonding after the stake drops below the required minimum when settling session accounting")
```

x/tokenomics/keeper/token_logic_modules.go
```diff
@@ -815,16 +832,16 @@ func calculateGlobalPerClaimMintInflationFromSettlementAmount(settlementCoin sdk
 	return mintAmtCoin, *newMintAmtFloat
 }
 
-// stakeToMaxSettlementAmount calculates the max amount of uPOKT to that the supplier
+// stakeShareToMaxSettlementAmount calculates the max amount of uPOKT to that the supplier
```
Olshansk, line 835:
```suggestion
// stakeShareToMaxSettlementAmount calculates the max amount of uPOKT the supplier
```

x/tokenomics/keeper/token_logic_modules.go
```diff
@@ -815,16 +832,16 @@ func calculateGlobalPerClaimMintInflationFromSettlementAmount(settlementCoin sdk
...
 // can claim based on the stake share and the global inflation allocation percentage.
```
Olshansk, line 836: I still don't understand what `stakeShare` is?

- Is it an amount? If so, why is it called a "share"
- Is it a percentage/fraction? If so, a fraction of "what"

Only blocking this PR because I'm unclear on this topic, so let's find a better name.

x/tokenomics/keeper/token_logic_modules.go
```diff
@@ -703,15 +716,18 @@ func (k Keeper) ensureClaimAmountLimits(
 	//   and assume maxClaimableAmt will be settled in session 2.
 	// - Guarantee no over-servicing at the cost of higher application stake requirements.
 	maxClaimableAmt := appStake.Amount.Quo(math.NewInt(sessionkeeper.NumSupplierPerSession))
-	maxClaimSettlementAmt := stakeToMaxSettlementAmount(maxClaimableAmt)
+	maxClaimSettlementAmt := stakeShareToMaxSettlementAmount(maxClaimableAmt)
 
 	// Check if the claimable amount is capped by the max claimable amount.
-	if stakeRequirementAmt.GT(maxClaimableAmt) {
+	// As per the Relay Mining paper, the Supplier claim MUST NO exceed the application's
```
Olshansk, line 722:
```suggestion
	// As per the Relay Mining paper, the Supplier claim MUST NOT exceed the application's
```

x/tokenomics/keeper/token_logic_modules.go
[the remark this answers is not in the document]
Olshansk: Add a `TODO_POST_MAINNET: Consider removing this since it should never happen just to simplify the code`

Olshansk (changes requested):

x/tokenomics/keeper/token_logic_modules.go
red-0ne: It's the supplier's application stake "share" to max settlement amount.

I renamed it appropriately.

Olshansk (approved):

Issue #838, opened 2024-09-25
[Gateway] Migrate from `AppGateServer` to `PATH`


Olshansk:
## Objective

Replace `AppGateServer` with [PATH](https://github.com/buildwithgrove/path) for local `poktroll` development.

## Origin Document

https://www.notion.so/buildwithgrove/Permissionless-Load-Testing-Sync-09-25-2024-10ca36edfff680cca07ac7ace75f258d?pvs=4

<img width="1285" alt="Screenshot 2024-09-25 at 12 45 30 PM" src="https://github.com/user-attachments/assets/bee1d870-91e0-46a0-ab91-279a49c57a40">

## Goals

- Deprecate `AppGateServer`
- Make `PATH` the priority for all future
- Enable a local end-to-end load testing flow

## Deliverables

Local:
- [ ] [@red-0ne] Add `PATH` to `tilt` and all related `LocalNet` environment tooling
- [ ] [@red-0ne] Identify missing gaps in `PATH`
- [ ] [@red-0ne] Ensure all our documentation is updated to use `PATH` instead of `AppGateServer`
- [ ] [@okdas] Ensure `PATH` provides all the same metrics as `AppGateServer`
Remote:
- [ ] [@okdas] Ensure `PATH` is deployed to our DevNet + Staging environments
Secondary:
- [ ] [@red-0ne + @okdas] Remove `AppGateServer` completely once all of the above are done

---

**Creator**: @Olshansk 
**Co-Owners**: @red-0ne @okdas

PR #879, opened 2024-10-14
[LocalNet] Replace `AppGateServer` with `Path` in `LocalNet`

red-0ne:
## Summary

Add PATH to Tilt so it could run on LocalNet for E2E tests and local load testing.

## Issue

In order to deprecate AppGate server and have better integration with the production gateway component, Path needs to run on LocalNet for both E2E test and load testing.

- #838 

## Type of change

## Testing

## Sanity Checklist


config.yml
```diff
@@ -171,9 +171,9 @@ genesis:
           denom: upokt
       applicationList:
         - address: pokt1mrqt5f7qh8uxs27cjm9t7v9e74a9vvdnq5jva4
-          delegatee_gateway_addresses: []
+          delegatee_gateway_addresses: [pokt15vzxjqklzjtlz7lahe8z2dfe9nm5vxwwmscne4]
           service_configs:
-            - service_id: anvil
+            - service_id: "0021"
```
Olshansk, line 176: tl;dr I'm changing this back

I understand why you did this, but:
1. This is legacy from Morse adopted into PATH
- I'd rather change PATH and move forward
2. If we make this change, you'll need to change a lot more (see the search result below)
- This requires a TODO_IN_THIS_PR

Screenshot 2024-10-16 at 12 20 31 PM

red-0ne: From my understanding, Path does not know about / have "anvil" as an `EVM` service.

Olshansk: I added it. PR incoming soon.

Olshansk (changes requested):
@red-0ne Didn't review this, but putting a "cod review" checkpoint to reference when this is ready.

Olshansk (changes requested):
@red-0ne I've started reviewing + testing this locally and am getting stuck on the first.

Nothing actionable as I'm looking into it but just FYI.

Screenshot 2024-12-04 at 2 59 36 PM

.github/workflows-helpers/run-e2e-test-job-template.yaml
```diff
@@ -49,10 +49,8 @@ spec:
               value: "false" # Flip to true to see the command and result of the execution
             - name: POKTROLLD_HOME
               value: /root/.poktroll
-            - name: APPGATE_SERVER_URL
-              value: http://${NAMESPACE}-appgate-server:80
-            - name: GATEWAY_URL
-              value: http://${NAMESPACE}-gateway:80
+            - name: PATH_HOST
+              value: ${NAMESPACE}-path:3000
```
okdas, line 53: This should work on devnets now

Olshansk: Awesome!

Olshansk (changes requested):
@red-0ne 

1. Please review my changes
2. Please make sure this works WITHOUT localnet
3. I will take another look tomorrow

e2e/tests/node.go
```diff
@@ -188,13 +197,13 @@ func (p *pocketdBin) runCurlCmd(rpcBaseURL, service, method, path, appAddr, data
...
 		path = "/" + path
 	}
-
 	rpcUrl.Path = rpcUrl.Path + path
 
 	base := []string{
 		"-v",                                   // verbose output
 		"-sS",                                  // silent with error
 		"-H", `Content-Type: application/json`, // HTTP headers
+		"-H", fmt.Sprintf("Host: %s", virtualHost), // Add virtual host header
```
red-0ne, line 213: :muscle:

.github/workflows-helpers/run-e2e-test-job-template.yaml
```diff
@@ -49,7 +49,11 @@ spec:
               value: "false" # Flip to true to see the command and result of the execution
             - name: POKTROLLD_HOME
               value: /root/.poktroll
-            - name: PATH_HOST
+            - name: PATH_URL
+              value: http://${NAMESPACE}-path:3000/v1
+              # As we can't expose many DNS names, but the test requires to send requests to different subdomains (e.g. anvil., rest., etc.),
```
Olshansk, line 54: Can you update this comment. I read it a few times still don't understand the root source of the issue.

e2e/tests/node.go
```diff
@@ -25,6 +26,8 @@ var (
 	defaultHome = os.Getenv("POKTROLLD_HOME")
 	// defaultPathURL used by curl commands to send relay requests
 	defaultPathURL = os.Getenv("PATH_URL")
+	// defaultPathHostOverride overrides the host in the URL used to send requests
```
Olshansk, line 29: #PUC why

e2e/tests/node.go
```diff
@@ -178,8 +181,21 @@ func (p *pocketdBin) runCurlCmd(rpcBaseURL, service, method, path, appAddr, data
 		return nil, err
 	}
 
+	// Store the virtual host we want to send in the Host header
+	virtualHost := rpcUrl.Host
```
Olshansk, line 185: #PUC why we need a virtual host

e2e/tests/node.go
```diff
@@ -178,8 +181,21 @@ func (p *pocketdBin) runCurlCmd(rpcBaseURL, service, method, path, appAddr, data
...
 	if len(service) > 0 {
-		rpcUrl.Host = fmt.Sprintf("%s.%s", service, rpcUrl.Host)
+		// Strip port if it exists and add service prefix
```
Olshansk, line 187: Can you create a helper called "getVirtualHostFromHost" w/ a longer explanation of what's going on here?

Olshansk (changes requested):

Olshansk:
@red-0ne Also, please make sure to update the PR description.

e2e/tests/node.go
```diff
@@ -27,6 +27,9 @@ var (
 	// defaultPathURL used by curl commands to send relay requests
 	defaultPathURL = os.Getenv("PATH_URL")
 	// defaultPathHostOverride overrides the host in the URL used to send requests
+	// Since the current DevNet infrastructure does not support arbitrary subdomains,
+	// this is used to specify the host to connect to and the full host (with the service as a subdomain)
+	// will be sent in the "Host" request header.
```
Olshansk, line 32: Note that comments like `Since the current...` are not "evergreen".

This makes sense to me and you today because we have context and are reading it in the scope of this PR. Will it make sense to a random person on the internet in 12 months? Will it even be relevant then?

Ways to improve this:
 - Reference the PR where this is added
 - Add a date when this i being added
 - Add a TODO for future readers
 - Etc...

e2e/tests/node.go
```diff
@@ -263,3 +261,24 @@ func formatURLString(serviceAlias, rpcUrl, path string) string {
...
+		host, _, err := net.SplitHostPort(rpcUrl.Host)
+		if err != nil {
+			// No port in the host, use as-is
+			host = rpcUrl.Host
+		}
+		virtualHost = fmt.Sprintf("%s.%s", service, host)
+	}
+
+	return virtualHost
+}
```
Olshansk, line 284:
```suggestion
// getVirtualHostFromUrlForService returns a virtual host taking into consideration
// the URL's host and the service if it's non-empty.
// Specifically, it:
// 	1. Extract's the host from the rpcURL
//      2. Prefixes the service as a subdomain to (1)the given rpcUrl host stripped of the port
// TODO_FUTURE: This is needed as of PR #879 because the DevNet infrastructure does
// not support arbitrary subdomains due to TLS certificate-related complexities.
func getVirtualHostFromUrlForService(rpcUrl *url.URL, service string) string {
	// Store the virtual host we want to send in the Host header
	virtualHost := rpcUrl.Host

	if len(service) > 0 {
		// Strip port if it exists and add service prefix
		host, _, err := net.SplitHostPort(rpcUrl.Host)
		if err != nil {
			// err is non-nil if rpcUrl.Host does not have a port.
			// Use the entire host as is
			host = rpcUrl.Host
		}
		virtualHost = fmt.Sprintf("%s.%s", service, host)
	}

	return virtualHost
}
```

e2e/tests/node.go
```diff
@@ -263,3 +261,24 @@ func formatURLString(serviceAlias, rpcUrl, path string) string {
...
+
+// getVirtualHostFromHost returns the given rpcUrl host stripped of the port
+// and with the service included as a subdomain to be used in the "Host" request header.
+// This is needed by the current DevNet infrastructure that does not support arbitrary
+// subdomains which the PATH Gateway needs to identify the requested service.
+func getVirtualHostFromHost(rpcUrl *url.URL, service string) string {
+	// Store the virtual host we want to send in the Host header
+	virtualHost := rpcUrl.Host
+
+	if len(service) > 0 {
```
Olshansk, line 273: Can you #PUC when this would be zero?

e2e/tests/node.go
```diff
@@ -181,19 +184,14 @@ func (p *pocketdBin) runCurlCmd(rpcBaseURL, service, method, path, appAddr, data
...
-	}
+	// Get the virtual host that will be sent in the "Host" request header
+	virtualHost := getVirtualHostFromHost(rpcUrl, service)
 
-	// Override the actual connection address if the environment requires it
+	// The current DevNet infrastructure does not support routing requests to arbitrary subdomains.
+	// As a workaround, defaultPathHostOverride (which contains no subdomain) is used
+	// as the gateway's host:port to connect to, and the virtual host (which includes
+	// the service as subdomain) is sent in the "Host" request header.
+	// Override the actual connection address if the environment requires it.
```
Olshansk, line 194:
```suggestion
	// TODO_FUTURE: As of PR #879, DevNet infrastructure does not support routine requests to
	// arbitrary subdomains due to TLS certificate-related complexities.
	// As a workaround,  defaultPathHostOverride (which contains no subdomain) is used as:
	//     1. The gateway's 'host:port' to connect to
        //     2. The virtual host send as part of the "Host" request header (which includes the service as subdomain)
```

Olshansk: @red-0ne I still feel like `As a workaround,  defaultPathHostOverride (which contains no subdomain) is used as:`  and `2. The virtual host send as part of the "Host" request header (which includes the service as subdomain)` are contradictory, but I'm not sure what the correct comment i.s

Olshansk (changes requested):
@red-0ne PTAL at my comments and suggestions and ping me as soon as it's tended. Want to push this over the finish line quickly.

e2e/tests/node.go
red-0ne: This should never be zero, and it's not the responsibility of this function to check it.

I'll move the check to where it's relevant.

Olshansk (approved):

PR #894, opened 2024-10-25
[Upgrade] Alpha TestNet v0.0.10 upgrade

okdas:
## Summary

Prep for `v0.0.10` release and upgrade.

## Type of change

- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

Went through upgrade testing locally.

## Sanity Checklist


okdas:
Merged in [78a7445](https://github.com/pokt-network/poktroll/commit/78a74453d23c5fee0d7f15df0dd12d5039d6fa14), seems like there are no changes that would affect the release so gtg.

okdas:
Merged in https://github.com/pokt-network/poktroll/commit/ff764308c0478a4323335aad5591372ffefd3053 and began testing the upgrade locally.

The upgrade failed initially. Here's a gist with a panic: https://gist.github.com/okdas/535bc3c0282483fef27454c70d889aa6

I believe the reason is we've changed the data type of the protobuf field [here](https://github.com/pokt-network/poktroll/commit/09fc4ce08a5cc02a88e3a3d02bd034a978dbdbd0#diff-e5d764e19c94556d799d2ae492ceb425ddcd7cb93d7c5dcf603a3600389de4b8). Meaning the new version is unable to map the existing `uint64` in kvstore into a `Coin` that it now expects.

After I manually created a new `prooftypes.Params` the upgrade went through.

Screenshot 2024-10-30 at 6 26 29 PM

We probably need to treat protobuf files as they used are production - meaning no type/field changes, any change that needs to be done should go through adding a new field, ensuring backwards-compatibility.

Olshansk:
> We probably need to treat protobuf files as they used are production - meaning no type/field changes, any change that needs to be done should go through adding a new field, ensuring backwards-compatibility.

Noted. We'll start doing this in Beta. We'll have one last cleanup opportunity prior to mainnet.

cc @red-0ne

okdas:
I'm going to create a release using [f0d0e7a](https://github.com/pokt-network/poktroll/pull/894/commits/f0d0e7ae96e365a927dbfe712e6e6f39d5332d98) as this is the commit that has been tested using this branch.

Looking at new changes in main branch, there are more state changes required in the upgrade. As discussed in the meeting yesterday, we will add them in another upgrade shortly after.

Let's practice upgrades!

okdas:
The release has been created: https://github.com/pokt-network/poktroll/releases/tag/v0.0.10

I did not use `main` branch specifically as there are more commits now that would require a bit more testing. We'll issue the next upgrade with the new changes. Probably next week.

@Olshansk let's get this in main to preserve the upgrade code! :)

app/upgrades/types.go
```diff
@@ -22,3 +29,28 @@ type Upgrade struct {
 	// StoreUpgrades adds, renames and deletes KVStores in the state to prepare for a protocol upgrade.
 	StoreUpgrades storetypes.StoreUpgrades
 }
+
+type grantAuthorization struct {
```
Olshansk, line 33: 1. Why do we need to create a struct for this rather than using an existing protobuf type?
2. #PUC if we can't reuse something

app/upgrades/types.go
```diff
@@ -22,3 +29,28 @@ type Upgrade struct {
...
+	grantee       sdk.AccAddress
+	granter       sdk.AccAddress
+	authorization authz.Authorization
+	expiration    *time.Time
+}
+
+func newTestNetGrantAuthorization(msg string) grantAuthorization {
```
Olshansk, line 40: #PUC on what the goal of this function is

app/upgrades/types.go
```diff
@@ -1,13 +1,20 @@
...
 	upgradetypes "cosmossdk.io/x/upgrade/types"
+	sdk "github.com/cosmos/cosmos-sdk/types"
 	"github.com/cosmos/cosmos-sdk/types/module"
+	"github.com/cosmos/cosmos-sdk/x/authz"
 
 	"github.com/pokt-network/poktroll/app/keepers"
 )
 
+var testNetPnfAddress = "pokt1r6ja6rz6rpae58njfrsgs5n5sp3r36r2q9j04h"
+var testNetAuthorityAddress = "pokt10d07y265gmmuvt4z0w9aw880jnsr700j8yv32t"
```
Olshansk, line 16:
```suggestion
var (
	// The default PNF/DAO address in the genesis file for local development
	testNetPnfAddress = "pokt1r6ja6rz6rpae58njfrsgs5n5sp3r36r2q9j04h"
	
	// A new authority address to which the PNF/DAO address above authorizes to act on its behalf
	testNetAuthorityAddress = "pokt10d07y265gmmuvt4z0w9aw880jnsr700j8yv32t"
)
```

But, I looked in the `Makefile` and found this:

```
GOV_ADDRESS = pokt10d07y265gmmuvt4z0w9aw880jnsr700j8yv32t
# PNF acts on behalf of the DAO and who AUTHZ must delegate to
PNF_ADDRESS = pokt1eeeksh2tvkh7wzmfrljnhw4wrhs55lcuvmekkw
```

Can you add comments so it's clear what/why we're doing here? I'm pretty confused.

tools/scripts/upgrades/local_test_v0.0.10.json
```diff
@@ -0,0 +1,15 @@
+{
+    "body": {
+        "messages": [
+            {
+                "@type": "/cosmos.upgrade.v1beta1.MsgSoftwareUpgrade",
+                "authority": "pokt10d07y265gmmuvt4z0w9aw880jnsr700j8yv32t",
+                "plan": {
+                    "name": "v0.0.10",
+                    "height": "20",
+                    "info": "NOT NEEDED FOR LOCAL TESTING"
```
Olshansk, line 10: #PUC why we have this file then

app/upgrades/v0.0.10.go
```diff
@@ -0,0 +1,127 @@
...
+	storetypes "cosmossdk.io/store/types"
+	upgradetypes "cosmossdk.io/x/upgrade/types"
+	cosmosTypes "github.com/cosmos/cosmos-sdk/types"
+	"github.com/cosmos/cosmos-sdk/types/module"
+	"github.com/pokt-network/poktroll/app/keepers"
+	prooftypes "github.com/pokt-network/poktroll/x/proof/types"
+)
+
+// Upgrade_0_0_10 is the upgrade handler for v0.0.10 Alpha TestNet upgrade
+// Before/after validations should be done using the correct version (e.g. before - v0.0.9, after - v0.0.10)
```
Olshansk, line 16:
```suggestion
// Before/after validations should be done using the correct version:
//     - Before: v0.0.9
//     - After: v0.0.10
```

app/upgrades/v0.0.10.go
```diff
@@ -0,0 +1,127 @@
...
+var Upgrade_0_0_10 = Upgrade{
+	PlanName: "v0.0.10",
+	CreateUpgradeHandler: func(mm *module.Manager,
+		keepers *keepers.Keepers,
+		configurator module.Configurator,
+	) upgradetypes.UpgradeHandler {
+		return func(ctx context.Context, plan upgradetypes.Plan, vm module.VersionMap) (module.VersionMap, error) {
+			//
+			// Add missing parameters and changes from `config.yml`
```
Olshansk, line 25: I feel like there's a lot more happening in this diff.

- Is the comment just outdated?
- Am I missing something?

app/upgrades/v0.0.10.go
```diff
@@ -0,0 +1,127 @@
...
+			// https://github.com/pokt-network/poktroll/compare/v0.0.9-3...ff76430
+			//
+
+			// Add application min stake
+			// Validate with: `poktrolld q application params --node=https://testnet-validated-validator-rpc.poktroll.com/`
+			appParams := keepers.ApplicationKeeper.GetParams(ctx)
+			newMinStakeApp := cosmosTypes.NewCoin("upokt", math.NewInt(100000000))
```
Olshansk, line 32:
```suggestion
			newDefaultMinStakeApp := cosmosTypes.NewCoin("upokt", math.NewInt(100000000))
```

app/upgrades/v0.0.10.go
```diff
@@ -0,0 +1,127 @@
...
+			gatewayParams.MinStake = &newMinStakeGW
+			err = keepers.GatewayKeeper.SetParams(ctx, gatewayParams)
+			if err != nil {
+				return vm, err
+			}
+
+			// Adjust proof module parameters
+			// Validate with: `poktrolld q proof params --node=https://testnet-validated-validator-rpc.poktroll.com/`
+			//
+			// !!!! THE UPGRADE CURRENTLY BREAKS HERE !!!!
```
Olshansk, line 62: Is this still the case?

app/upgrades/v0.0.10.go
```diff
@@ -0,0 +1,127 @@
...
+			// https://gist.github.com/okdas/535bc3c0282483fef27454c70d889aa6
+			// proofParams := keepers.ProofKeeper.GetParams(ctx)
```
Olshansk, line 64: Remove?

app/upgrades/v0.0.10.go
Olshansk: After reviewing the code further, my understanding is that:
- We use the same diff everywhere
- You grouped things based on comments

In that case: 
- we should put the diff in the function header
- Call out "section migrating new params"...

app/upgrades/v0.0.10.go
```diff
@@ -0,0 +1,127 @@
...
+				)
+				if err != nil {
+					return vm, err
+				}
+			}
+
+			// Seems like RelayMiningDifficulty have been moved from `tokenomics` to `services`.
+			// In the ideal scenario, we should have migrated the data before removing query/msg types in `tokenomics`
+			// module. It would be hard to do that now. We know that new RelayMiningDifficulty will be created,
+			// we can skip this step now.
```
Olshansk, line 120:
```suggestion
			// RelayMiningDifficulty have been moved from `tokenomics` to `services` in this diff.
			// Ideally (in prod), we should have migrated the data before removing query/msg types in `tokenomics` module.
			// In practice (development), we don't want to spend time on it.
			// Since we know that new RelayMiningDifficulty will be re-created and real tokens are not in danger,
			// we can skip this step now.
```

Olshansk (changes requested):

app/upgrades/types.go
okdas: Removed it.

app/upgrades/types.go
okdas: Removed it.

app/upgrades/types.go
okdas: Moved vars and extended comments so it is obvious what it is and why.

app/upgrades/v0.0.10.go
okdas: Correct. Changed comments so this is more evident.

okdas:
Thanks for the review @Olshansk.

At this point, with many changes here, I want to delete the v0.0.10 version we currently have a release for, remove the merges from the main branch (so the code only contains the commits we wrote an upgrade for), and test the upgrade again.

Will re-create release and apply an upgrade on the network after that.

okdas:
I've verified all new params and grants locally - we're good to upgrade! 
I'll create a new release and issue an upgrade on alpha network now.

okdas:
@Olshansk the upgrade has been applied on Alpha TestNet. We don't really have to merge this, but I still would to have an example for future upgrades.

app/upgrades/types.go
```diff
@@ -8,6 +8,18 @@ import (
 	"github.com/pokt-network/poktroll/app/keepers"
 )
 
+// TODO_CONSIDER: different networks should have the same gov module address, but might have different DAO addresses,
```
Olshansk, line 11:
```suggestion
// TODO_MAINNET_DISCUSSION(@Olshansk): different networks should have the same gov module address, but might have different DAO addresses,
```

app/upgrades/types.go
```diff
@@ -8,6 +8,18 @@ import (
...
+// unless we specifically write in these addresses in the genesis file.
+// Should we use the same address/wallet for DAO or find a way to detect the network the upgrade is being applied to,
+// to pick different addresses depending on the name of the network? (e.g chain-id)
+
+const (
```
Olshansk, line 16: Ty for the comments

Olshansk (approved):

PR #920, opened 2024-11-13
[quick-pr] change full-node.sh script links

okdas:
Use new links introduced in https://github.com/pokt-network/pocket-network-genesis/pull/7


okdas:
Merging this small change so all links are the same everywhere. Lmk if something is off so I can address @Olshansk.

PR #946, opened 2024-11-21
[Supplier] Allow Supplier to downstake

red-0ne:
## Summary

Allow `Supplier`s to lower their stake down to `MinStake`.

## Issue

- TODO_BETA

## Type of change

- [ ] Bug fix
- [ ] Code health or cleanup
- [x] Documentation
- [ ] Other (specify)

## Testing

## Sanity Checklist


x/supplier/keeper/msg_server_stake_supplier.go
```diff
@@ -140,15 +124,46 @@ func (k msgServer) StakeSupplier(ctx context.Context, msg *suppliertypes.MsgStak
...
 
-	// Send the coins from the message signer account to the staked supplier pool
 	supplierStakingFee := k.GetParams(ctx).StakingFee
-	stakeWithFee := sdk.NewCoins(coinsToEscrow.Add(*supplierStakingFee))
-	err = k.bankKeeper.SendCoinsFromAccountToModule(ctx, msgSignerAddress, suppliertypes.ModuleName, stakeWithFee)
-	if err != nil {
-		logger.Info(fmt.Sprintf("ERROR: could not send %v coins from %q to %q module account due to %v", coinsToEscrow, msgSignerAddress, suppliertypes.ModuleName, err))
-		return nil, status.Error(codes.InvalidArgument, err.Error())
+
+	// The Supplier is increasing its stake, so escrow the difference
```
Olshansk, line 129: OPTIONAL NIT: COnsider creating a function on the keeper called `escrowSupplierStakeDiff` to avoid this mega function from getting too large.

x/supplier/keeper/msg_server_stake_supplier.go
```diff
@@ -140,15 +124,46 @@ func (k msgServer) StakeSupplier(ctx context.Context, msg *suppliertypes.MsgStak
...
+		if coinsToUnescrow.Amount.LTE(supplierStakingFee.Amount) {
+			err = types.ErrSupplierInvalidStake.Wrapf(
+				"supplier with owner %q must unescrow more than supplier staking fee (%s)",
+				msg.GetOwnerAddress(), supplierStakingFee,
+			)
+			logger.Info(fmt.Sprintf("ERROR: %s", err))
+			return nil, status.Error(codes.InvalidArgument, err.Error())
+		}
+
+		coinsAfterFee := sdk.NewCoins(coinsToUnescrow.Sub(*supplierStakingFee))
```
Olshansk, line 157: Don't we need to add the fee?

Every single time you stake/unstake you pay again. It's like a transaction.

x/supplier/keeper/msg_server_stake_supplier.go
```diff
@@ -140,15 +124,46 @@ func (k msgServer) StakeSupplier(ctx context.Context, msg *suppliertypes.MsgStak
...
+
+		// Send the coins from the staked supplier pool to the message signer account
+		if err := k.bankKeeper.SendCoinsFromModuleToAccount(ctx, types.ModuleName, msgSignerAddress, coinsAfterFee); err != nil {
+			logger.Info(fmt.Sprintf("ERROR: could not send %s coins from %q module account to %q due to %v", coinsToUnescrow, types.ModuleName, msgSignerAddress, err))
+			return nil, status.Error(codes.InvalidArgument, err.Error())
+		}
+
+		logger.Info(fmt.Sprintf("Successfully unescrowed %s from %q module account to %q", coinsToUnescrow, types.ModuleName, msgSignerAddress))
```
Olshansk, line 165: Where is the MinStake check happening to make sure there are no problems?

x/supplier/keeper/msg_server_stake_supplier_test.go
```diff
@@ -135,16 +135,18 @@ func TestMsgServer_StakeSupplier_FailRestakingDueToInvalidServices(t *testing.T)
 	require.Equal(t, "http://localhost:8080", supplierFound.Services[0].Endpoints[0].Url)
```
Olshansk, line 135: Will review this after my other comments are addressed.

x/supplier/keeper/msg_server_stake_supplier.go
```diff
@@ -8,8 +8,10 @@ import (
 	"google.golang.org/grpc/codes"
```
Olshansk, line 8: I'm sure you and @bryanchriswhite have discussed & thought through this, but don't we have issues are slashing misbehaving suppliers if they do this mid session?

Lots of nuances, edges cases and details that I'm omitting above.

Olshansk (changes requested):

x/supplier/keeper/msg_server_stake_supplier.go
red-0ne: The staking fee was incorporated into the amount sent from the supplier's account to the module.

I agree, we should indeed always treat it as an individual transfer in all cases. This way we'll have a distinct transfer event representing the fee.

x/supplier/keeper/msg_server_stake_supplier.go
red-0ne: At this point in the code the new stake amount is already checked against min stake.

The check is done here: https://github.com/pokt-network/poktroll/pull/946/files#diff-a21ef4c8a0a17759b5b114f280982371695f9467e42fbbc237bf29bbac43e784R111

x/supplier/keeper/msg_server_stake_supplier.go
red-0ne: Good call!

`MinStake` should always be enough to pay for any slashing. Otherwise a malicious `Supplier` could attack the network by staking `MinStake`.

x/supplier/keeper/msg_server_stake_supplier.go
```diff
@@ -265,3 +248,31 @@ func (k msgServer) updateSupplier(
 
 	return nil
 }
+
+// transferSupplierStakeDiff transfers the difference between the current and new stake
+// amounts by either escrowing when the stake is increased or unescrowing otherwise.
+func (k msgServer) transferSupplierStakeDiff(
+	ctx context.Context,
+	msgSignerAccAddress sdk.AccAddress,
+	supplierCurrentStake sdk.Coin,
```
Olshansk, line 257: Rename to `currentStake` so we have `currStake` and `newStake` side by side

x/supplier/keeper/msg_server_stake_supplier.go
```diff
@@ -265,3 +248,31 @@ func (k msgServer) updateSupplier(
 
 	return nil
 }
+
+// transferSupplierStakeDiff transfers the difference between the current and new stake
+// amounts by either escrowing when the stake is increased or unescrowing otherwise.
+func (k msgServer) transferSupplierStakeDiff(
+	ctx context.Context,
+	msgSignerAccAddress sdk.AccAddress,
```
Olshansk, line 256: s/msgSignerAccAddress/signerAddr since it's a helper function

x/supplier/keeper/msg_server_stake_supplier.go
```diff
@@ -265,3 +248,31 @@ func (k msgServer) updateSupplier(
...
+) error {
+	// The Supplier is increasing its stake, so escrow the difference
+	if supplierCurrentStake.Amount.LT(newStake.Amount) {
+		coinsToEscrow := sdk.NewCoins(newStake.Sub(supplierCurrentStake))
+
+		// Send the coins from the message signer account to the staked supplier pool
+		return k.bankKeeper.SendCoinsFromAccountToModule(ctx, msgSignerAccAddress, suppliertypes.ModuleName, coinsToEscrow)
+	}
+
+	// The supplier is decreasing its stake, unescrow the difference but deduct the staking fee.
```
Olshansk, line 268: #PUC why we no newStake is greater than the minStake.

Optionally, consider being redundant with this logic. Doesn't hurt.

x/supplier/keeper/msg_server_stake_supplier.go
```diff
@@ -265,3 +248,31 @@ func (k msgServer) updateSupplier(
...
+	if supplierCurrentStake.Amount.GT(newStake.Amount) {
+		coinsToUnescrow := sdk.NewCoins(supplierCurrentStake.Sub(newStake))
+
+		// Send the coins from the staked supplier pool to the message signer account
+		return k.bankKeeper.SendCoinsFromModuleToAccount(ctx, types.ModuleName, msgSignerAccAddress, coinsToUnescrow)
+	}
+
+	// The supplier is not changing its stake
```
Olshansk, line 276: Please log this scenario and (optionally) callout if this should ever happen.

x/supplier/keeper/msg_server_stake_supplier.go
```diff
@@ -140,15 +125,20 @@ func (k msgServer) StakeSupplier(ctx context.Context, msg *suppliertypes.MsgStak
...
 
-	// Send the coins from the message signer account to the staked supplier pool
 	supplierStakingFee := k.GetParams(ctx).StakingFee
-	stakeWithFee := sdk.NewCoins(coinsToEscrow.Add(*supplierStakingFee))
-	err = k.bankKeeper.SendCoinsFromAccountToModule(ctx, msgSignerAddress, suppliertypes.ModuleName, stakeWithFee)
-	if err != nil {
-		logger.Info(fmt.Sprintf("ERROR: could not send %v coins from %q to %q module account due to %v", coinsToEscrow, msgSignerAddress, suppliertypes.ModuleName, err))
-		return nil, status.Error(codes.InvalidArgument, err.Error())
+
+	if err = k.transferSupplierStakeDiff(ctx, msgSignerAddress, supplierCurrentStake, *msg.Stake); err != nil {
```
Olshansk, line 130: s/transfer/handle

x/supplier/keeper/msg_server_stake_supplier.go
```diff
@@ -140,15 +125,20 @@ func (k msgServer) StakeSupplier(ctx context.Context, msg *suppliertypes.MsgStak
...
+		logger.Info(fmt.Sprintf("ERROR: could not transfer supplier stake difference due to %v", err))
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+
+	// Deduct the staking fee from the supplier's account balance.
+	// This is called after the stake difference is transferred give the supplier
```
Olshansk, line 136:
```suggestion
	// This is called after the stake difference is transferred to give the supplier
```

x/supplier/keeper/msg_server_stake_supplier_test.go
```diff
@@ -168,6 +170,42 @@ func TestMsgServer_StakeSupplier_FailLoweringStake(t *testing.T) {
 	require.Len(t, supplierFound.Services, 1)
 }
 
+func TestMsgServer_StakeSupplier_SuccessLoweringStakeAboveMinStake(t *testing.T) {
```
Olshansk, line 173: Do we have a test where we increase it?

Olshansk (changes requested):

x/supplier/keeper/msg_server_stake_supplier.go
bryanchriswhite: Thought on "reconcile"? Handle seems even more generic to me.

x/supplier/keeper/msg_server_stake_supplier.go
```diff
@@ -140,15 +125,20 @@ func (k msgServer) StakeSupplier(ctx context.Context, msg *suppliertypes.MsgStak
...
-	// Send the coins from the message signer account to the staked supplier pool
 	supplierStakingFee := k.GetParams(ctx).StakingFee
-	stakeWithFee := sdk.NewCoins(coinsToEscrow.Add(*supplierStakingFee))
-	err = k.bankKeeper.SendCoinsFromAccountToModule(ctx, msgSignerAddress, suppliertypes.ModuleName, stakeWithFee)
-	if err != nil {
-		logger.Info(fmt.Sprintf("ERROR: could not send %v coins from %q to %q module account due to %v", coinsToEscrow, msgSignerAddress, suppliertypes.ModuleName, err))
-		return nil, status.Error(codes.InvalidArgument, err.Error())
+
+	if err = k.transferSupplierStakeDiff(ctx, msgSignerAddress, supplierCurrentStake, *msg.Stake); err != nil {
+		logger.Info(fmt.Sprintf("ERROR: could not transfer supplier stake difference due to %v", err))
```
bryanchriswhite, line 131: 1. Internal errors should ALWAYS be logged at the error level.
2. We don't need to prefix the log with the level, at least when it matches the level used (i.e. we use `info` for client-side errors).
3. We should ALWAYS use `%s` to interpolate errors into strings unless there's a specific reason to do otherwise.  `%v` will result in some undesirable output which includes the path to the errors.go file, and `%+v` will include the stack trace.

```suggestion
		logger.Error(fmt.Sprintf("could not transfer supplier stake difference due to %s", err))
```

x/supplier/keeper/msg_server_stake_supplier.go
```diff
@@ -140,15 +125,20 @@ func (k msgServer) StakeSupplier(ctx context.Context, msg *suppliertypes.MsgStak
...
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+
+	// Deduct the staking fee from the supplier's account balance.
+	// This is called after the stake difference is transferred give the supplier
+	// the opportunity to have enough balance to pay the fee.
+	if err = k.bankKeeper.SendCoinsFromAccountToModule(ctx, msgSignerAddress, suppliertypes.ModuleName, sdk.NewCoins(*supplierStakingFee)); err != nil {
+		logger.Info(fmt.Sprintf("ERROR: signer %q could not pay for the staking fee %s due to %v", msgSignerAddress, supplierStakingFee, err))
```
bryanchriswhite, line 139:
```suggestion
		logger.Info(fmt.Sprintf("ERROR: signer %q could not pay for the staking fee %s: %s", msgSignerAddress, supplierStakingFee, err))
```

x/supplier/keeper/msg_server_stake_supplier.go
```diff
@@ -140,15 +125,20 @@ func (k msgServer) StakeSupplier(ctx context.Context, msg *suppliertypes.MsgStak
...
+		return nil, status.Error(codes.FailedPrecondition, err.Error())
```
bryanchriswhite, line 140: :ok_hand:

x/supplier/keeper/msg_server_stake_supplier.go
```diff
@@ -265,3 +248,31 @@ func (k msgServer) updateSupplier(
 
 	return nil
 }
+
+// transferSupplierStakeDiff transfers the difference between the current and new stake
+// amounts by either escrowing when the stake is increased or unescrowing otherwise.
```
bryanchriswhite, line 253:
```suggestion
// amounts by either escrowing, when the stake is increased, or unescrowing otherwise.
```

x/supplier/keeper/msg_server_stake_supplier_test.go
```diff
@@ -168,6 +170,42 @@ func TestMsgServer_StakeSupplier_FailLoweringStake(t *testing.T) {
...
+	stakeMsg, _ := newSupplierStakeMsg(ownerAddr, operatorAddr, minStake, "svcId")
+
+	// Stake the supplier & verify that the supplier exists
+	_, err := srv.StakeSupplier(ctx, stakeMsg)
+	require.NoError(t, err)
+
+	_, isSupplierFound := supplierModuleKeepers.GetSupplier(ctx, operatorAddr)
+	require.True(t, isSupplierFound)
+
+	// Prepare an update supplier msg with a lower stake which is below staking fee.
```
bryanchriswhite, line 193:
```suggestion
	// Prepare an update supplier msg with a lower stake which is below the minimum stake.
```

bryanchriswhite (approved):

I just had a couple of suggestions/comments - nothing blocking, preemptively approving. :ship:

x/supplier/keeper/msg_server_stake_supplier_test.go
```diff
@@ -190,7 +190,7 @@ func TestMsgServer_StakeSupplier_SuccessLoweringStakeAboveMinStake(t *testing.T)
 	_, isSupplierFound := supplierModuleKeepers.GetSupplier(ctx, operatorAddr)
 	require.True(t, isSupplierFound)
 
-	// Prepare an update supplier msg with a lower stake which is below staking fee.
+	// Prepare an update supplier msg with a lower stake which is below the minimum staking fee.
```
Olshansk, line 193:
```suggestion
	// Prepare an updated supplier msg with a lower stake which is below the minimum staking fee.
```

Olshansk (approved):

PR #888, opened 2024-10-22
[Gateway] Add application indexing by gateway delegation

red-0ne:
## Summary

This PR aims to implement `Application` retrieval by delegated `Gateway`.
* It adds a `delegating_appplication_addresses` string slice to the `gateway` proto file.
* Adds up `AddGateway` and `RemoveGateway` methods to update the delegation references
* Use them in `DelegateToGateway`, `UndelegateFromGateway`, `UnstakeApplication`, `MsgTransferApplication` and `UnstakeGateway`.
* Add unbonding to the gateway to work around cyclic dependencies and notify the EndBlocker to update the gateway->apps map which is not possible using events.

## Issue

`Gateways` need an easy way to fetch the list of `Application`s that delegated to them.
image

https://github.com/buildwithgrove/path/compare/feat/localnet-path?expand=1#diff-dbf18c3c62c4cc61ad09d7e4eb5292d2a6bec3c0000b34f0635b7dc446aed9c3R226

## Type of change

Select one or more from the following:

- [x] New feature, functionality or library
- [x] Consensus breaking; add the `consensus-breaking` label if so. See #791 for details
- [x] Bug fix

## Testing

- [ ] **Documentation**: `make docusaurus_start`; only needed if you make doc changes
- [ ] **Unit Tests**: `make go_develop_and_test`
- [ ] **LocalNet E2E Tests**: `make test_e2e`
- [ ] **DevNet E2E Tests**: Add the `devnet-test-e2e` label to the PR.

## Sanity Checklist

- [ ] I have tested my changes using the available tooling
- [ ] I have commented my code
- [ ] I have performed a self-review of my own code; both comments & source code
- [ ] I create and reference any new tickets, if applicable
- [ ] I have left TODOs throughout the codebase, if applicable

PR #957, opened 2024-11-24
[Application] Support filtering apps by delegatee gateway address

adshmh:
## Summary

Support querying applications based on the delegated gateway address:
- Add a new field to the `QueryAllApplicationsRequest` type.
- Update the application keeper to process the query request based on the updated struct.
- Add unit tests to cover the new functionality.

Note: #888 adds a similar functionality by allowing access to delegating apps directly from a gateway struct. This PR  allows querying apps directly for a specific gateway delegation. Likely both PRs are needed in the long term.

## Issue

This is needed ot support integration with PATH. More specifically, it allows PATH to get the list of all applications which delegate to a specific gateway address:

https://github.com/buildwithgrove/path/blob/8533ee8b6dae4dcc5d4615bdd1b58d6229ef85a7/protocol/shannon/gateway_mode.go#L25-L26

## Type of change

- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

## Sanity Checklist


x/application/types/query_validation.go
```diff
@@ -0,0 +1,20 @@
...
+// ValidateBasic performs basic (non-state-dependant) validation on a QueryAllApplicationsRequest.
+func (query *QueryAllApplicationsRequest) ValidateBasic() error {
+	delegateeGatewayAddr := query.GetDelegateeGatewayAddress()
+	if delegateeGatewayAddr == "" {
+		return nil
+	}
+
+	// Validate the delegation gateway address if the request specifies it as a constraint.
+	if _, err := sdk.AccAddressFromBech32(delegateeGatewayAddr); err != nil {
+		return ErrAppInvalidGatewayAddress.Wrapf("invalid gateway address for query: %q; (%v)", delegateeGatewayAddr, err)
```
red-0ne, line 16:
```suggestion
		return ErrAppInvalidGatewayAddress.Wrapf("querying for apps with delegatee gateway address: %q; (%v)", delegateeGatewayAddr, err)
```

x/application/keeper/application_test.go
```diff
@@ -18,18 +19,35 @@ import (
...
+		}
+
 		keeper.SetApplication(ctx, apps[i])
 	}
 	return apps
 }
 
+// testAppModifierDelegateeAddr adds the supplied gateway address to the application's delegatee list if the application's address matches
+// the supplied address list.
+func testAppModifierDelegateeAddr(delegateeGatewayAddr string, appsWithDelegationAddr []string) testAppModifier {
```
red-0ne, line 43: Wdyt about naming it `withAppDelegateeGatewayAddr`?

x/application/keeper/query_application_test.go
```diff
@@ -122,3 +124,44 @@ func TestApplicationQueryPaginated(t *testing.T) {
...
+		addrInvalid := "invalid-address"
+		_, err := keeper.AllApplications(ctx, requestBuilder(addrInvalid))
+		expectedErr := status.Error(
+			codes.InvalidArgument,
+			types.ErrAppInvalidGatewayAddress.Error(),
+		)
+
+		require.ErrorContains(t, err, expectedErr.Error())
+	})
+}
```
red-0ne, line 167: It's worth adding a test with an empty address and assert that all apps are returned.

red-0ne (approved):
Thanks for putting this together.

Left a few comments but nothing blocking.

Also, could you please add the `devnet-test-e2e` label before pushing new changes to trigger e2e tests?

proto/poktroll/application/query.proto
```diff
@@ -53,6 +53,8 @@ message QueryGetApplicationResponse {
 
 message QueryAllApplicationsRequest {
   cosmos.base.query.v1beta1.PageRequest pagination = 1;
+  // delegatee_gateway_address, if specified, filters the application list to only include those with delegation to the specified gateway address.
+  string delegatee_gateway_address = 2;
```
Olshansk, line 57: Thoughts on renaming this to `gateway_address_delegated_to`.

I realize that `delegatee_gateway_address` more dramatically correct (probably?) but it confuses me when I read it and forces me to think.

When I read `gateway_address_delegated_to`, I don't need to think.

x/application/keeper/query_application_test.go
Olshansk: +1

Olshansk (commented):
@adshmh Leaving a comment (not a request for changes, nor an approval) because I'm curious to hear what you think of my feedback.

proto/poktroll/application/query.proto
Olshansk: 100% agreed.

TODO_MAINNET instead of beta though ;)

adshmh: Done, thank you for the reviews.

x/application/keeper/application_test.go
adshmh: Done, thank you for the review.

Olshansk (approved):

PR #967, opened 2024-11-26
[Upgrade] Alpha TestNet v0.0.11

okdas:
## Summary

An upgrade for Alpha TestNet from v0.0.10 to v0.0.11.

## Issue

Beta TestNet has been launched using v0.0.11 (release candidate), and we need to upgrade alpha so both networks use the same version.

## Type of change

Chain upgrade.

## Testing

- [] An upgrade is tested by upgrading a network provisioned by the old version (v0.0.10) to the new version (v0.0.11)

## Sanity Checklist


app/upgrades/v0.0.11.go
```diff
@@ -0,0 +1,104 @@
...
+			sessionParams.NumSuppliersPerSession = uint64(15)
+			err = keepers.SessionKeeper.SetParams(ctx, sessionParams)
+			if err != nil {
+				return err
+			}
+
+			// Set tokenomics params. The values are based on default values for LocalNet.
+			// Validate with: `poktrolld q tokenomics params --node=https://testnet-validated-validator-rpc.poktroll.com/`
+			tokenomicsParams := keepers.TokenomicsKeeper.GetParams(ctx)
+			tokenomicsParams.MintAllocationPercentages = tokenomicstypes.MintAllocationPercentages{
```
okdas, line 43: These are the same values as Beta TestNet btw.

okdas:
Upgrade has been tested and validated locally:

```
5:39PM INF applying upgrade "v0.0.11" at height: 20 module=x/upgrade
5:39PM INF Starting v0.0.11 upgrade handler module=server
5:39PM INF Starting parameter updates for v0.0.11 module=server
5:39PM INF Current session params module=server params={"num_suppliers_per_session":0}
5:39PM INF Successfully updated session params module=server new_params={"num_suppliers_per_session":15}
5:39PM INF Current tokenomics params module=server params={"dao_reward_address":"","mint_allocation_proposer":{"application":0,"dao":0,"proposer":0,"source_owner":0,"supplier":0}}
5:39PM INF Successfully updated tokenomics params module=server new_params={"dao_reward_address":"pokt1r6ja6rz6rpae58njfrsgs5n5sp3r36r2q9j04h","mint_allocation_proposer":{"application":0,"dao":0.1,"proposer":0.05,"source_owner":0.15,"supplier":0.7}}
5:39PM INF Starting authorization updates for v0.0.11 module=server
5:39PM INF Granting authorization module=server
5:39PM INF Successfully granted authorization module=server
5:39PM INF Running module migrations module=server
```

okdas:
This is interesting.

The [diff between versions](https://github.com/pokt-network/poktroll/compare/v0.0.10...v0.0.11-rc) shows we need to add a new authorization, but we already have it on Alpha TestNet:
```bash
poktrolld q authz grants-by-granter pokt10d07y265gmmuvt4z0w9aw880jnsr700j8yv32t --node=https://shannon-testnet-grove-seed-rpc.alpha.poktroll.com
```


```yaml
- authorization:
    type: cosmos-sdk/GenericAuthorization
    value:
      msg: /poktroll.session.MsgUpdateParam
  expiration: "2500-01-01T00:00:00Z"
  grantee: pokt1r6ja6rz6rpae58njfrsgs5n5sp3r36r2q9j04h
  granter: pokt10d07y265gmmuvt4z0w9aw880jnsr700j8yv32t
```

okdas:
Tested again after authorization code removed:

```

5:58PM INF ABCI Handshake App Info hash=48BAFBE9874B2A7625D9834A983AB231C98523767859E5208A277F8518CF66C0 height=19 module=consensus protocol-version=0 software-version=0.0.11-rc-bc2a3c44
5:58PM INF ABCI Replay Blocks appHeight=19 module=consensus stateHeight=19 storeHeight=20
5:58PM INF Replay last block using real app module=consensus
5:58PM INF applying upgrade "v0.0.11" at height: 20 module=x/upgrade
5:58PM INF Starting v0.0.11 upgrade handler module=server
5:58PM INF Starting parameter updates for v0.0.11 module=server
5:58PM INF Current session params module=server params={"num_suppliers_per_session":0}
5:58PM INF Successfully updated session params module=server new_params={"num_suppliers_per_session":15}
5:58PM INF Current tokenomics params module=server params={"dao_reward_address":"","mint_allocation_proposer":{"application":0,"dao":0,"proposer":0,"source_owner":0,"supplier":0}}
5:58PM INF Successfully updated tokenomics params module=server new_params={"dao_reward_address":"pokt1r6ja6rz6rpae58njfrsgs5n5sp3r36r2q9j04h","mint_allocation_proposer":{"application":0,"dao":0.1,"proposer":0.05,"source_owner":0.15,"supplier":0.7}}
5:58PM INF Running module migrations module=server
5:58PM INF adding a new module: 06-solomachine module=server
5:58PM INF adding a new module: 07-tendermint module=server
```

I think we're good to go. I'll wait for a review, before I cut a new version.

app/upgrades/v0.0.11.go
```diff
@@ -0,0 +1,93 @@
...
+	PlanName: "v0.0.11",
+	CreateUpgradeHandler: func(mm *module.Manager,
+		keepers *keepers.Keepers,
+		configurator module.Configurator,
+	) upgradetypes.UpgradeHandler {
+
+		// Adds new parameters using ignite's config.yml as a reference. Assuming we don't need any other parameters.
+		// https://github.com/pokt-network/poktroll/compare/v0.0.10...v0.0.11-rc
+		applyNewParameters := func(ctx context.Context) (err error) {
+			logger := cosmosTypes.UnwrapSDKContext(ctx).Logger()
```
okdas, line 28: thanks @bryanchriswhite for the hint!

app/upgrades/v0.0.11.go
```diff
@@ -0,0 +1,93 @@
...
+//   - After: v0.0.11
+var Upgrade_0_0_11 = Upgrade{
+	PlanName: "v0.0.11",
+	CreateUpgradeHandler: func(mm *module.Manager,
+		keepers *keepers.Keepers,
+		configurator module.Configurator,
+	) upgradetypes.UpgradeHandler {
+
+		// Adds new parameters using ignite's config.yml as a reference. Assuming we don't need any other parameters.
+		// https://github.com/pokt-network/poktroll/compare/v0.0.10...v0.0.11-rc
```
Olshansk, line 26: should this be `-rc` or just `0.1`?

app/upgrades/v0.0.11.go
Olshansk: #PUC?

app/upgrades/v0.0.11.go
```diff
@@ -0,0 +1,93 @@
...
+			tokenomicsParams := keepers.TokenomicsKeeper.GetParams(ctx)
+			logger.Info("Current tokenomics params", "params", tokenomicsParams)
+			tokenomicsParams.MintAllocationPercentages = tokenomicstypes.MintAllocationPercentages{
+				Dao:         0.1,
+				Proposer:    0.05,
+				Supplier:    0.7,
+				SourceOwner: 0.15,
+				Application: 0.0,
+			}
+			tokenomicsParams.DaoRewardAddress = AlphaTestNetPnfAddress
```
Olshansk, line 54: Should this be `AlphaTestNetPnfAddress` or `BetaTestNetPnfAddress`?

app/upgrades/v0.0.11.go
```diff
@@ -0,0 +1,93 @@
...
+				logger.Error("Failed to set tokenomics params", "error", err)
+				return err
+			}
+			logger.Info("Successfully updated tokenomics params", "new_params", tokenomicsParams)
+
+			return
+		}
+
+		// The diff shows that the only new authz authorization is for the `poktroll.session.MsgUpdateParam` message.
+		// However, this message is already authorized for the `pokt10d07y265gmmuvt4z0w9aw880jnsr700j8yv32t` address.
```
Olshansk, line 66: Is this the alpha or beta PNF address

Olshansk (changes requested):

app/upgrades/v0.0.11.go
okdas: It should be an `-rc` here. We created RC release to bootstrap beta TestNet. The release must have this code (the upgrade handler) inside the binary. So we're just using the RC as a version to create a release on.

Olshansk (approved):

app/upgrades/v0.0.11.go
```diff
@@ -0,0 +1,93 @@
...
+			err = keepers.SessionKeeper.SetParams(ctx, sessionParams)
+			if err != nil {
+				logger.Error("Failed to set session params", "error", err)
+				return err
+			}
+			logger.Info("Successfully updated session params", "new_params", sessionParams)
+
+			// Set tokenomics params. The values are based on default values for LocalNet/Beta TestNet.
+			// Validate with: `poktrolld q tokenomics params --node=https://testnet-validated-validator-rpc.poktroll.com/`
+			tokenomicsParams := keepers.TokenomicsKeeper.GetParams(ctx)
```
okdas, line 45: Here's the line that caused a panic: https://gist.github.com/okdas/50f417e3a86230e9e1aea85da841e1de

I need to investigate why I did not have this issue during local upgrade from v0.0.10 to v0.0.11. Will try to replicate using ONLY binaries provided in releases.

Could it be the before state is different on Alpha TestNet? If so, going to document this so we check for that in the future.

Olshansk: > Here's the line that caused a panic: [gist.github.com/okdas/50f417e3a86230e9e1aea85da841e1de](https://gist.github.com/okdas/50f417e3a86230e9e1aea85da841e1de)

Why did this cause the panic?

okdas: @Olshansk My first hypothesis it has happened due to the protobuf type change, but that should have been caught on the local test. For some reason, it did not. I'm going to be looking into that today.

I expect we will change an upgrade procedure to migrate each module individually, which can preserve the types between upgrades. There are some pros and cons associated with that. So far, I wanted the upgrade process to look simple, but it seems like additional complexity might worth it to avoid such issues in the future. 

Basically, I need to spend more time to provide a complete answer.

Olshansk: > I need to spend more time to provide a complete answer.

This is REALLY REALLY REALLY REALLY important, so please dive as deep as necessary to the finest grain of detail.

okdas:
I'm going to remove the merge from main into this branch and do a force push.
This branch is not ready for code from main yet, as I'm still figuring out what happened on Alpha TestNet.

Olshansk:
@okdas Following up on an offline discussion.

> This branch is not ready for code from main yet, as I'm still figuring out what happened on Alpha TestNet.

If you can't resolve it by mid-day today, let's do the following:

1. Post your latest update + investigation + details
2. Please move over to load-testing
3. @bryanchriswhite will pick it up next week
4. If @bryanchriswhite doesn't figure it out by mid week next week, I'll just in and help
5. @okdas Please be ready (aside from working on (2)) to potentially help out with some questions

Olshansk (changes requested):
Undoing my approval per the latest discussion in this thread

okdas:
Here's my investigation so far: https://www.notion.so/buildwithgrove/Alpha-TestNet-v0-0-11-Postmortem-152a36edfff680d58411c8fa384f877c?pvs=4

TLDR: KVStore had a different representation of empty params for the tokenomics module (alpha testnet vs local node), which explains the difference in behavior. 

I've also included some action items for us (me mostly) to do to avoid this and issues that were uncovered during this upgrade / postmortem investigation.

@bryanchriswhite I would love you to look into this as well, and I'm happy to pair/help.

okdas:
OK, and now that I sent the previous message, it struck me what exactly happened here.

Alpha TestNet was provisioned using `v0.0.9`, which had a different protobuf for the `tokenomics` module params: https://github.com/pokt-network/poktroll/blob/v0.0.9/proto/poktroll/tokenomics/params.proto#L16

And it had a value, too: https://github.com/pokt-network/pocket-network-genesis/blob/master/shannon/testnet-alpha/genesis.json#L1018

That protobuf was then removed in `v0.0.10`, but the bytes were still preserved in the KVStore: `082A` - [see them decoded here](https://gchq.github.io/CyberChef/#recipe=From_Hex('Auto')Protobuf_Decode('',false,false)JSON_Minify()&input=MDgyQQ).

As I did not see any changes in the protobuf when comparing `v0.0.10` to `v0.0.11`, I deemed it safe to read the parameters from KVstore, while it was not.

If we did not remove the parameter or used a different field ID for the new fields, this wouldn't happen.

bryanchriswhite:
**TL;DR, we SHOULD use `reserved` for deleted fields AND we SHOULD be versioning our protobuf types (e.g. `proto/tokenomics/v1beta1`).**

> OK, and now that I sent the previous message, it struck me what exactly happened here.

:exploding_head: Nice sleuthing! :rocket: 

> If we did not remove the parameter or used a different field ID for the new fields, this wouldn't happen.

Point taken that going forward, we need to be much more careful about backwards compatibility. In this particular case, I don' think there's any way we could've avoided removing this field as it was necessary to resolve a dependency cycle. In cases like this, it seems like the correct thing to do, going forward, would be to remove the field but also add its field number (and possibly field name) as [`reserved`](https://protobuf.dev/programming-guides/proto3/#reserved-field-numbers).

FWIW, this case would also be a bit unusual, IMHO, to resolve with versioned types as we would end up having multiple branches in the effected module's `InitGenesis` function to handle the two different versions of genesis. I think this makes much less sense then the typical case, which would be where a message or query handler has conditional branches for each version. "Genesis backwards-compatibility" seems like a bit of an oxymoron. :joy: 

This (re-)raises the point of **versoined protobuf source**, which we currently are not setup for, neither from a go package nor a proto source package persective. I think we will need to do something like the following in order to pre-allocate sufficient degrees of freedom to ensure we can safely increment the protobuf types into the future, without causing caos from both the go module/pkg and buf.build perspectives:

### Current proto directory structure:
```
./proto
└── poktroll
    ├── application
    │   └── *.proto
    │   └── *.proto
    ├── gateway
    │   └── *.proto
    ├── proof
    │   └── *.proto
    ├── service
    │   └── *.proto
    ├── session
    │   └── *.proto
    ├── shared
    │   └── *.proto
    ├── supplier
    │   └── *.proto
    └── tokenomics
    │   └── *.proto
```

### Conventional (versioned):
```
./proto
└── poktroll
    ├── application
    │   └── v1beta1
    │       └── *.proto
    ├── gateway
    │   └── v1beta1
    │       └── *.proto
    ├── proof
    │   └── v1beta1
    │       └── *.proto
    ├── service
    │   └── v1beta1
    │       └── *.proto
    ├── session
    │   └── v1beta1
    │       └── *.proto
    ├── shared
    │   └── v1beta1
    │       └── *.proto
    ├── supplier
    │   └── v1beta1
    │       └── *.proto
    └── tokenomics
        └── v1beta1
            └── *.proto
```

Olshansk:
Firstly, great investigation @okdas!

@bryanchriswhite I want to avoid the `module/v1beta1/*.proto` refactor just now.

Let's continue with:
- Only incrementing IDs
- Using `reserved fields` for deprecated fields

Things we'll look out for:
- How do other cosmos chains do this?
- When (if) we hit major proto refactors, we'll revisit package versioning

okdas:
I'll have to force push to remove the merge.
Screenshot 2025-01-23 at 4 20 06 PM

We should only let commits until v0.0.11. Please don't update the branch.

okdas:
Now that Alpha TestNet is on `v0.0.11` - I merged main and this should be good to merge.

app/upgrades/v0.0.11.go
```diff
@@ -0,0 +1,99 @@
...
+// Beta TestNet was launched with v0.0.11, so this upgrade is exclusively for Alpha TestNet.
+//   - Before: v0.0.10
+//   - After: v0.0.11
+var Upgrade_0_0_11 = Upgrade{
+	PlanName: "v0.0.11",
+	CreateUpgradeHandler: func(mm *module.Manager,
+		keepers *keepers.Keepers,
+		configurator module.Configurator,
+	) upgradetypes.UpgradeHandler {
+
```
Olshansk, line 25:
```suggestion
```

app/upgrades/v0.0.11.go
```diff
@@ -0,0 +1,99 @@
...
+		// Adds new parameters using ignite's config.yml as a reference. Assuming we don't need any other parameters.
```
Olshansk, line 26: Is this comment still correct?

app/upgrades/v0.0.11.go
```diff
@@ -0,0 +1,99 @@
...
+		// https://github.com/pokt-network/poktroll/compare/v0.0.10...v0.0.11-rc
```
Olshansk, line 27: Is this still an rc?

app/upgrades/v0.0.11.go
```diff
@@ -0,0 +1,99 @@
...
+		applyNewParameters := func(ctx context.Context) (err error) {
+			logger := cosmosTypes.UnwrapSDKContext(ctx).Logger()
+			logger.Info("Starting parameter updates for v0.0.11")
+
+			// Set num_suppliers_per_session to 15
+			// Validate with: `poktrolld q session params --node=https://testnet-validated-validator-rpc.poktroll.com/`
```
Olshansk, line 33: Is this the correct URL and value has been validated?

Olshansk (approved):
Approving to unblock you, but just leaving a few comments for you to take into consideration.

app/upgrades/v0.0.11.go
okdas: Yes

app/upgrades/v0.0.11.go
okdas: Yes, and always will be. We needed the diff between v0.0.10 and v0.0.11-rc. This is a historical code. We don't even need it in the main branch. But I want to preserve to use as a reference.

app/upgrades/v0.0.11.go
okdas: Yes and yes.

Issue #543, opened 2024-05-21
[Off-Chain] ModuleParamsClient & Historical Params


bryanchriswhite:
## Objective

Enable support for off-chain processes to cache on-chain parameters and react to parameter updates asynchronously.

## Origin Document

1. Off-chain actors currently query the network much too frequently.

2. Observations made while working on #517. In order for off-chain processes to use the correct governance params in the appropriate places in the protocol, they MUST be aware of the most recent param values. In some cases, off-chain processes MUST also know the historical values of some params. 

### Assumptions
1. It doesn't make sense for changes to protocol parameters to take effect **during** a session.
2. We SHOULD limit the amount of historical data that is persisted to only what is necessary; let external indexers support non-essential indexing.

### Example
A param update message is committed in the middle of a session, modifying `num_blocks_per_session` & `min_relay_difficulty_bits` by the time a proof is submitted & validated on-chain. The `num_blocks_per_session` and `min_relay_difficulty_bits` at the time of the session to which the proof corresponds (i.e. will not be the current session) MUST be used. There may also be off-chain use cases (presently or in the future) which require previous values for params.

```mermaid
sequenceDiagram

participant SessionQueryClient as Supplier
participant Poktrolld
participant PNF

Note over SessionQueryClient,PNF: Session N Start

PNF->>Poktrolld: Update params

break
    SessionQueryClient-->Poktrolld: (wait for sesison end)
end

Note over SessionQueryClient,PNF: Session N End

break
    SessionQueryClient-->Poktrolld: (wait for grace period end)
end

Note over SessionQueryClient,PNF: Session N Grace Period End

SessionQueryClient ->> Poktrolld: Create claim

break
    SessionQueryClient-->Poktrolld: (wait for next grace period end)
end

Note over SessionQueryClient,PNF: Session N+1 Grace Period End

SessionQueryClient ->> Poktrolld: Submit proof

Poktrolld->>Poktrolld: Validate proof

```

## Goals

- To support off-chain reactivity to parameter updates (in-between sessions).
  - Implement `ModuleParamsClient` (see: [notion - `ModuleParamsClient`](https://www.notion.so/buildwithgrove/ModuleParamsClient-4af05dff9a0b4a22b157d2d92f2b262f))
- Support on-chain persistence (& pruning) of historical param values.

### On-chain persistence of historical params

When a param update message is committed:
1. The corresponding on-chain params value(s) are updated on commit.
[...]

PR #994, opened 2024-12-11
[Off-chain] feat: in-memory query cache(s)

bryanchriswhite:
## Summary

Adds the `QueryCache[T any]` and `HistoricalQueryCache[T any]` interfaces, `InMemoryCache[T any]` implementation, configurations, and options functions.

## Issue

- #543

## Type of change

- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

## Sanity Checklist

### DIFF
--- a/pkg/client/interface.go
+++ b/pkg/client/interface.go
@@ -360,3 +360,11 @@ type BankQueryClient interface {
 	// GetBalance queries the chain for the uPOKT balance of the account provided
 	GetBalance(ctx context.Context, address string) (*cosmostypes.Coin, error)
 }
+
+// QueryCache handles a single type of cached data
+type QueryCache[T any] interface {
+	Get(key string) (T, error)
+	Set(key string, value T) error
+	Delete(key string)
+	Clear()
+}
--- a/pkg/client/interface.go
+++ b/pkg/client/interface.go
@@ -368,3 +368,12 @@ type QueryCache[T any] interface {
 	Delete(key string)
 	Clear()
 }
+
+// HistoricalQueryCache extends QueryCache to support historical values at different heights
+type HistoricalQueryCache[T any] interface {
+	QueryCache[T]
+	// GetAtHeight retrieves the nearest value <= the specified height
+	GetAtHeight(key string, height int64) (T, error)
+	// SetAtHeight adds or updates a value at a specific height
+	SetAtHeight(key string, value T, height int64) error
+}
--- a/pkg/client/query/cache/config.go
+++ b/pkg/client/query/cache/config.go
@@ -1,0 +1,76 @@
+package cache
+
+import (
+	"time"
+)
+
+// EvictionPolicy determines how items are removed when cache is full.
+type EvictionPolicy int64
+
+const (
+	FirstInFirstOut = EvictionPolicy(iota)
+	LeastRecentlyUsed
+	LeastFrequentlyUsed
+)
+
+// CacheConfig is the configuration options for a cache.
+type CacheConfig struct {
+	// MaxKeys is the maximum number of items the cache can hold.
+	MaxKeys int64
+	// EvictionPolicy is how items should be removed when the cache is full.
+	EvictionPolicy EvictionPolicy
+	// TTL is how long items should remain in the cache
+	TTL time.Duration
+
+	// historical is whether the cache will cache a single value for each key
+	// (false) or whether it will cache a history of values for each key (true).
+	historical bool
+	// pruneOlderThan is the number of past blocks for which to keep historical
+	// values. If 0, no historical pruning is performed.
+	pruneOlderThan int64
+}
+
+// CacheOption defines a function that configures a CacheConfig
+type CacheOption func(*CacheConfig)
+
+// HistoricalCacheConfig extends the basic CacheConfig with historical settings.
+type HistoricalCacheConfig struct {
+	CacheConfig
+
+	// MaxHeightsPerKey is the maximum number of different heights to store per key
+	MaxHeightsPerKey int
+	// PruneOlderThan specifies how many blocks back to maintain in history
+	// If 0, no historical pruning is performed
+	PruneOlderThan int64
+}
+
+// WithHistoricalMode enables historical caching with the given pruneOlderThan
+// configuration, if 0 no historical pruning is performed.
+func WithHistoricalMode(pruneOlderThan int64) CacheOption {
+	return func(cfg *CacheConfig) {
+		cfg.historical = true
+		cfg.pruneOlderThan = pruneOlderThan
+	}
+}
+
+// WithMaxKeys sets the maximum number of distinct key/value pairs the cache will
+// hold before evicting according to the configured eviction policy.
+func WithMaxKeys(size int64) CacheOption {
+	return func(cfg *CacheConfig) {
+		cfg.MaxKeys = size
+	}
+}
+
+// WithEvictionPolicy sets the eviction policy
+func WithEvictionPolicy(policy EvictionPolicy) CacheOption {
+	return func(cfg *CacheConfig) {
+		cfg.EvictionPolicy = policy
+	}
+}
+
+// WithTTL sets the time-to-live for cache entries
+func WithTTL(ttl time.Duration) CacheOption {
+	return func(cfg *CacheConfig) {
+		cfg.TTL = ttl
+	}
+}
--- a/pkg/client/query/cache/memory.go
+++ b/pkg/client/query/cache/memory.go
@@ -1,0 +1,245 @@
+package cache
+
+import (
+	"sync"
+	"sync/atomic"
+	"time"
+
+	"github.com/pokt-network/poktroll/pkg/client"
+)
+
+var (
+	_ client.QueryCache[any]           = (*InMemoryCache[any])(nil)
+	_ client.HistoricalQueryCache[any] = (*InMemoryCache[any])(nil)
+)
+
+// InMemoryCache provides a concurrency-safe in-memory cache implementation with
+// optional historical value support.
+type InMemoryCache[T any] struct {
+	config       CacheConfig
+	latestHeight atomic.Int64
+
+	itemsMu sync.RWMutex
+	// items type depends on historical mode:
+	// | historical mode |  type                                   |
+	// | --------------- | --------------------------------------- |
+	// | false           | map[string]cacheItem[T]                 |
+	// | true            | map[string]map[int64]heightCacheItem[T] |
+	items map[string]any
+}
+
+// cacheItem wraps cached values with metadata
+type cacheItem[T any] struct {
+	value     T
+	timestamp time.Time
+}
+
+// heightCacheItem is used when the cache is in historical mode
+type heightCacheItem[T any] struct {
+	value     T
+	timestamp time.Time
+}
+
+// NewInMemoryCache creates a new cache with the given configuration
+func NewInMemoryCache[T any](opts ...CacheOption) *InMemoryCache[T] {
+	config := CacheConfig{
+		EvictionPolicy: FirstInFirstOut,
+	}
+
+	for _, opt := range opts {
+		opt(&config)
+	}
+
+	return &InMemoryCache[T]{
+		items:  make(map[string]interface{}),
+		config: config,
+	}
+}
+
+// Get retrieves an item from the cache
+func (c *InMemoryCache[T]) Get(key string) (T, error) {
+	if c.config.historical {
+		return c.GetAtHeight(key, c.latestHeight.Load())
+	}
+
+	c.itemsMu.RLock()
+	defer c.itemsMu.RUnlock()
+
+	var zero T
+
+	item, exists := c.items[key]
+	if !exists {
+		return zero, ErrCacheMiss
+	}
+
+	cItem := item.(cacheItem[T])
+	if c.config.TTL > 0 && time.Since(cItem.timestamp) > c.config.TTL {
+		// TODO_QUESTION: should we prune here?
+		return zero, ErrCacheMiss
+	}
+
+	return cItem.value, nil
+}
+
+// GetAtHeight retrieves an item from the cache at or before the specified height
+func (c *InMemoryCache[T]) GetAtHeight(key string, height int64) (T, error) {
+	var zero T
+
+	if !c.config.historical {
+		return zero, ErrHistoricalModeNotEnabled
+	}
+
+	c.itemsMu.RLock()
+	defer c.itemsMu.RUnlock()
+
+	heightMap, exists := c.items[key]
+	if !exists {
+		return zero, ErrCacheMiss
+	}
+
+	versions := heightMap.(map[int64]heightCacheItem[T])
+	var nearestHeight int64 = -1
+	for h := range versions {
+		if h <= height && h > nearestHeight {
+			nearestHeight = h
+		}
+	}
+
+	if nearestHeight == -1 {
+		return zero, ErrCacheMiss
+	}
+
+	item := versions[nearestHeight]
+	if c.config.TTL > 0 && time.Since(item.timestamp) > c.config.TTL {
+		return zero, ErrCacheMiss
+	}
+
+	return item.value, nil
+}
+
+// Set adds or updates an item in the cache
+func (c *InMemoryCache[T]) Set(key string, value T) error {
+	if c.config.historical {
+		return c.SetAtHeight(key, value, c.latestHeight.Load())
+	}
+
+	if c.config.MaxKeys > 0 && int64(len(c.items)) >= c.config.MaxKeys {
+		c.evict()
+	}
+
+	c.itemsMu.Lock()
+	defer c.itemsMu.Unlock()
+
+	c.items[key] = cacheItem[T]{
+		value:     value,
+		timestamp: time.Now(),
+	}
+
+	return nil
+}
+
+// SetAtHeight adds or updates an item in the cache at a specific height
+func (c *InMemoryCache[T]) SetAtHeight(key string, value T, height int64) error {
+	if !c.config.historical {
+		return ErrHistoricalModeNotEnabled
+	}
+
+	// Update latest height if this is newer
+	latestHeight := c.latestHeight.Load()
+	if height > latestHeight {
+		// NB: Only update if c.latestHeight hasn't changed since we loaded it above.
+		c.latestHeight.CompareAndSwap(latestHeight, height)
+	}
+
+	c.itemsMu.Lock()
+	defer c.itemsMu.Unlock()
+
+	var history map[int64]heightCacheItem[T]
+	if existing, exists := c.items[key]; exists {
+		history = existing.(map[int64]heightCacheItem[T])
+	} else {
+		history = make(map[int64]heightCacheItem[T])
+		c.items[key] = history
+	}
+
+	// Prune old heights if configured
+	if c.config.pruneOlderThan > 0 {
+		for h := range history {
+			if height-h > c.config.pruneOlderThan {
+				delete(history, h)
+			}
+		}
+	}
+
+	history[height] = heightCacheItem[T]{
+		value:     value,
+		timestamp: time.Now(),
+	}
+
+	return nil
+}
+
+// Delete removes an item from the cache.
+func (c *InMemoryCache[T]) Delete(key string) {
+	c.itemsMu.Lock()
+	defer c.itemsMu.Unlock()
+
+	delete(c.items, key)
+}
+
+// Clear removes all items from the cache
+func (c *InMemoryCache[T]) Clear() {
+	c.itemsMu.Lock()
+	defer c.itemsMu.Unlock()
+
+	c.items = make(map[string]interface{})
+	c.latestHeight.Store(0)
+}
+
+// evict removes one item according to the configured eviction policy
+func (c *InMemoryCache[T]) evict() {
+	switch c.config.EvictionPolicy {
+	case FirstInFirstOut:
+		var oldestKey string
+		var oldestTime time.Time
+		first := true
+
+		for key, item := range c.items {
+			var itemTime time.Time
+			if c.config.historical {
+				versions := item.(map[int64]heightCacheItem[T])
+				for _, v := range versions {
+					if itemTime.IsZero() || v.timestamp.Before(itemTime) {
+						itemTime = v.timestamp
+					}
+				}
+			} else {
+				itemTime = item.(cacheItem[T]).timestamp
+			}
+
+			if first || itemTime.Before(oldestTime) {
+				oldestKey = key
+				oldestTime = itemTime
+				first = false
+			}
+		}
+		delete(c.items, oldestKey)
+
+	case LeastRecentlyUsed:
+		// TODO: Implement LRU eviction
+		// This will require tracking access times
+		panic("LRU eviction not implemented")
+
+	case LeastFrequentlyUsed:
+		// TODO: Implement LFU eviction
+		// This will require tracking access times
+		panic("LFU eviction not implemented")
+
+	default:
+		// Default to FIFO if policy not recognized
+		for key := range c.items {
+			delete(c.items, key)
+			return
+		}
+	}
+}
--- a/pkg/client/query/cache/memory_test.go
+++ b/pkg/client/query/cache/memory_test.go
@@ -1,0 +1,343 @@
+package cache
+
+import (
+	"context"
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+)
+
+// TestInMemoryCache_NonHistorical tests the basic cache functionality without historical mode
+func TestInMemoryCache_NonHistorical(t *testing.T) {
+	t.Run("basic operations", func(t *testing.T) {
+		cache := NewInMemoryCache[string]()
+
+		// Test Set and Get
+		err := cache.Set("key1", "value1")
+		require.NoError(t, err)
+		val, err := cache.Get("key1")
+		require.NoError(t, err)
+		require.Equal(t, "value1", val)
+
+		// Test missing key
+		_, err = cache.Get("nonexistent")
+		require.ErrorIs(t, err, ErrCacheMiss)
+
+		// Test Delete
+		cache.Delete("key1")
+		_, err = cache.Get("key1")
+		require.ErrorIs(t, err, ErrCacheMiss)
+
+		// Test Clear
+		err = cache.Set("key2", "value2")
+		require.NoError(t, err)
+		cache.Clear()
+		_, err = cache.Get("key2")
+		require.ErrorIs(t, err, ErrCacheMiss)
+	})
+
+	t.Run("TTL expiration", func(t *testing.T) {
+		cache := NewInMemoryCache[string](
+			WithTTL(100 * time.Millisecond),
+		)
+
+		err := cache.Set("key", "value")
+		require.NoError(t, err)
+
+		// Value should be available immediately
+		val, err := cache.Get("key")
+		require.NoError(t, err)
+		require.Equal(t, "value", val)
+
+		// Wait for TTL to expire
+		time.Sleep(150 * time.Millisecond)
+
+		// Value should now be expired
+		_, err = cache.Get("key")
+		require.ErrorIs(t, err, ErrCacheMiss)
+	})
+
+	t.Run("max size eviction", func(t *testing.T) {
+		cache := NewInMemoryCache[string](
+			WithMaxKeys(2),
+			WithEvictionPolicy(FirstInFirstOut),
+		)
+
+		// Add items up to max size
+		err := cache.Set("key1", "value1")
+		require.NoError(t, err)
+		err = cache.Set("key2", "value2")
+		require.NoError(t, err)
+
+		// Add one more item, should trigger eviction
+		err = cache.Set("key3", "value3")
+		require.NoError(t, err)
+
+		// First item should be evicted
+		_, err = cache.Get("key1")
+		require.ErrorIs(t, err, ErrCacheMiss)
+
+		// Other items should still be present
+		val, err := cache.Get("key2")
+		require.NoError(t, err)
+		require.Equal(t, "value2", val)
+
+		val, err = cache.Get("key3")
+		require.NoError(t, err)
+		require.Equal(t, "value3", val)
+	})
+}
+
+// TestInMemoryCache_Historical tests the historical mode functionality
+func TestInMemoryCache_Historical(t *testing.T) {
+	t.Run("basic historical operations", func(t *testing.T) {
+		cache := NewInMemoryCache[string](
+			WithHistoricalMode(100),
+		)
+
+		// Test SetAtHeight and GetAtHeight
+		err := cache.SetAtHeight("key", "value1", 10)
+		require.NoError(t, err)
+		err = cache.SetAtHeight("key", "value2", 20)
+		require.NoError(t, err)
+
+		// Test getting exact heights
+		val, err := cache.GetAtHeight("key", 10)
+		require.NoError(t, err)
+		require.Equal(t, "value1", val)
+
+		val, err = cache.GetAtHeight("key", 20)
+		require.NoError(t, err)
+		require.Equal(t, "value2", val)
+
+		// Test getting intermediate height (should return nearest lower height)
+		val, err = cache.GetAtHeight("key", 15)
+		require.NoError(t, err)
+		require.Equal(t, "value1", val)
+
+		// Test getting height before first entry
+		_, err = cache.GetAtHeight("key", 5)
+		require.ErrorIs(t, err, ErrCacheMiss)
+
+		// Test getting height after last entry
+		val, err = cache.GetAtHeight("key", 25)
+		require.NoError(t, err)
+		require.Equal(t, "value2", val)
+	})
+
+	t.Run("historical TTL expiration", func(t *testing.T) {
+		cache := NewInMemoryCache[string](
+			WithHistoricalMode(100),
+			WithTTL(100*time.Millisecond),
+		)
+
+		err := cache.SetAtHeight("key", "value1", 10)
+		require.NoError(t, err)
+
+		// Value should be available immediately
+		val, err := cache.GetAtHeight("key", 10)
+		require.NoError(t, err)
+		require.Equal(t, "value1", val)
+
+		// Wait for TTL to expire
+		time.Sleep(150 * time.Millisecond)
+
+		// Value should now be expired
+		_, err = cache.GetAtHeight("key", 10)
+		require.ErrorIs(t, err, ErrCacheMiss)
+	})
+
+	t.Run("pruning old heights", func(t *testing.T) {
+		cache := NewInMemoryCache[string](
+			WithHistoricalMode(10), // Prune entries older than 10 blocks
+		)
+
+		// Add entries at different heights
+		err := cache.SetAtHeight("key", "value1", 10)
+		require.NoError(t, err)
+		err = cache.SetAtHeight("key", "value2", 20)
+		require.NoError(t, err)
+		err = cache.SetAtHeight("key", "value3", 30)
+		require.NoError(t, err)
+
+		// Add a new entry that should trigger pruning
+		err = cache.SetAtHeight("key", "value4", 40)
+		require.NoError(t, err)
+
+		// Entries more than 10 blocks old should be pruned
+		_, err = cache.GetAtHeight("key", 10)
+		require.ErrorIs(t, err, ErrCacheMiss)
+		_, err = cache.GetAtHeight("key", 20)
+		require.ErrorIs(t, err, ErrCacheMiss)
+
+		// Recent entries should still be available
+		val, err := cache.GetAtHeight("key", 30)
+		require.NoError(t, err)
+		require.Equal(t, "value3", val)
+
+		val, err = cache.GetAtHeight("key", 40)
+		require.NoError(t, err)
+		require.Equal(t, "value4", val)
+	})
+
+	t.Run("non-historical operations on historical cache", func(t *testing.T) {
+		cache := NewInMemoryCache[string](
+			WithHistoricalMode(100),
+		)
+
+		// Set some historical values
+		err := cache.SetAtHeight("key", "value1", 10)
+		require.NoError(t, err)
+		err = cache.SetAtHeight("key", "value2", 20)
+		require.NoError(t, err)
+
+		// Regular Set should work with latest height
+		err = cache.Set("key", "value3")
+		require.NoError(t, err)
+
+		// Regular Get should return the latest value
+		val, err := cache.Get("key")
+		require.NoError(t, err)
+		require.Equal(t, "value3", val)
+
+		// Delete should remove all historical values
+		cache.Delete("key")
+		_, err = cache.GetAtHeight("key", 10)
+		require.ErrorIs(t, err, ErrCacheMiss)
+		_, err = cache.GetAtHeight("key", 20)
+		require.ErrorIs(t, err, ErrCacheMiss)
+		_, err = cache.Get("key")
+		require.ErrorIs(t, err, ErrCacheMiss)
+	})
+}
+
+// TestInMemoryCache_ErrorCases tests various error conditions
+func TestInMemoryCache_ErrorCases(t *testing.T) {
+	t.Run("historical operations on non-historical cache", func(t *testing.T) {
+		cache := NewInMemoryCache[string]()
+
+		// Attempting historical operations should return error
+		err := cache.SetAtHeight("key", "value", 10)
+		require.ErrorIs(t, err, ErrHistoricalModeNotEnabled)
+
+		_, err = cache.GetAtHeight("key", 10)
+		require.ErrorIs(t, err, ErrHistoricalModeNotEnabled)
+	})
+
+	t.Run("zero values", func(t *testing.T) {
+		cache := NewInMemoryCache[string]()
+
+		// Test with empty key
+		err := cache.Set("", "value")
+		require.NoError(t, err)
+		val, err := cache.Get("")
+		require.NoError(t, err)
+		require.Equal(t, "value", val)
+
+		// Test with empty value
+		err = cache.Set("key", "")
+		require.NoError(t, err)
+		val, err = cache.Get("key")
+		require.NoError(t, err)
+		require.Equal(t, "", val)
+	})
+}
+
+// TestInMemoryCache_ConcurrentAccess tests thread safety of the cache
+func TestInMemoryCache_ConcurrentAccess(t *testing.T) {
+	t.Run("concurrent access non-historical", func(t *testing.T) {
+		cache := NewInMemoryCache[int]()
+		const numGoroutines = 10
+		const numOperations = 100
+
+		// Create a context with timeout
+		ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+		defer cancel()
+
+		var wg sync.WaitGroup
+		wg.Add(numGoroutines)
+
+		for i := 0; i < numGoroutines; i++ {
+			go func(routineID int) {
+				defer wg.Done()
+				for j := 0; j < numOperations; j++ {
+					// Check for timeout
+					select {
+					case <-ctx.Done():
+						t.Errorf("test timed out: %v", ctx.Err())
+						return
+					default:
+						key := "key"
+						err := cache.Set(key, j)
+						require.NoError(t, err)
+						_, _ = cache.Get(key)
+					}
+				}
+			}(i)
+		}
+
+		// Wait with timeout
+		done := make(chan struct{})
+		go func() {
+			wg.Wait()
+			close(done)
+		}()
+
+		select {
+		case <-ctx.Done():
+			t.Errorf("test timed out waiting for goroutines to complete: %v", ctx.Err())
+		case <-done:
+			// Test completed successfully
+		}
+	})
+
+	t.Run("concurrent access historical", func(t *testing.T) {
+		cache := NewInMemoryCache[int](
+			WithHistoricalMode(100),
+		)
+		const numGoroutines = 10
+		const numOpsPerGoRoutine = 100
+
+		// Create a context with timeout
+		ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+		t.Cleanup(cancel)
+
+		var wg sync.WaitGroup
+		wg.Add(numGoroutines)
+
+		for i := 0; i < numGoroutines; i++ {
+			go func(routineID int) {
+				defer wg.Done()
+				for j := 0; j < numOpsPerGoRoutine; j++ {
+					// Check for timeout
+					select {
+					case <-ctx.Done():
+						t.Errorf("test timed out: %v", ctx.Err())
+						return
+					default:
+						key := "key"
+						err := cache.SetAtHeight(key, j, int64(j))
+						require.NoError(t, err)
+						_, _ = cache.GetAtHeight(key, int64(j))
+					}
+				}
+			}(i)
+		}
+
+		// Wait with timeout
+		done := make(chan struct{})
+		go func() {
+			wg.Wait()
+			close(done)
+		}()
+
+		select {
+		case <-ctx.Done():
+			t.Errorf("test timed out waiting for goroutines to complete: %v", ctx.Err())
+		case <-done:
+			// Test completed successfully
+		}
+	})
+}
--- a/pkg/client/query/cache/errors.go
+++ b/pkg/client/query/cache/errors.go
@@ -1,0 +1,10 @@
+package cache
+
+import "cosmossdk.io/errors"
+
+const codesace = "client/query/cache"
+
+var (
+	ErrCacheMiss                = errors.Register(codesace, 1, "cache miss")
+	ErrHistoricalModeNotEnabled = errors.Register(codesace, 2, "historical mode not enabled")
+)


pkg/client/query/cache/memory.go
```diff
@@ -0,0 +1,245 @@
...
+	defer c.itemsMu.RUnlock()
+
+	heightMap, exists := c.items[key]
+	if !exists {
+		return zero, ErrCacheMiss
+	}
+
+	versions := heightMap.(map[int64]heightCacheItem[T])
+	var nearestHeight int64 = -1
+	for h := range versions {
```
bryanchriswhite, line 102: Prune and/or ensure a value older than `pruneOlderThan` isn't returned.

pkg/client/query/cache/config.go
```diff
@@ -0,0 +1,69 @@
...
+// WithHistoricalMode enables historical caching with the given pruneOlderThan
+// configuration; if 0, no historical pruning is performed.
+func WithHistoricalMode(pruneOlderThan int64) QueryCacheOptionFn {
+	return func(cfg *queryCacheConfig) {
+		cfg.historical = true
+		cfg.pruneOlderThan = pruneOlderThan
+	}
+}
+
+// WithMaxKeys sets the maximum number of distinct key/value pairs the cache will
```
Olshansk, line 48: TOL: MaxEntries sounds pretty bad but that's what this is, no? Just curious what you think.

pkg/client/query/cache/config.go
```diff
@@ -0,0 +1,69 @@
...
+}
+
+// WithEvictionPolicy sets the eviction policy.
+func WithEvictionPolicy(policy EvictionPolicy) QueryCacheOptionFn {
+	return func(cfg *queryCacheConfig) {
+		cfg.EvictionPolicy = policy
+	}
+}
+
+// WithTTL sets the time-to-live for cached items. Items older than the TTL
```
Olshansk, line 63: ` Items older than the TTL MAY not be evicted but SHOULD not be considered as cache hits.`

Given that pruning is actually implemented, the MAY and SHOULD doesn't make as much sense, but I may be misinterpting. 

Also, let's capitalize NOT as well.

pkg/client/query/cache/memory.go
```diff
@@ -0,0 +1,304 @@
...
+	"github.com/pokt-network/poktroll/pkg/client"
+)
+
+var (
+	_ client.QueryCache[any]           = (*InMemoryCache[any])(nil)
+	_ client.HistoricalQueryCache[any] = (*InMemoryCache[any])(nil)
+
+	DefaultQueryCacheConfig = queryCacheConfig{
+		EvictionPolicy: FirstInFirstOut,
+		// TODO_MAINNET(@bryanchriswhite): Consider how we can "guarantee" good
```
Olshansk, line 18: 👍

pkg/client/query/cache/memory.go
```diff
@@ -0,0 +1,304 @@
...
+		// alignment between the TTL and the block production rate.
```
Olshansk, line 19:
```suggestion
		// alignment between the TTL and the block production rate,
		// by accessing onchain block times directly.
```

pkg/client/query/cache/memory.go
```diff
@@ -0,0 +1,304 @@
...
+		TTL: time.Minute,
+	}
+)
+
+// InMemoryCache provides a concurrency-safe in-memory cache implementation with
```
Olshansk, line 24: Was there really no OSS library for this?

pkg/client/query/cache/memory.go
```diff
@@ -0,0 +1,304 @@
...
+// optional historical value support.
+type InMemoryCache[T any] struct {
```
Olshansk, line 26: Does this need to be publically exposed

pkg/client/query/cache/memory.go
```diff
@@ -0,0 +1,304 @@
...
+}
+
+// cacheItem wraps cached values with a timestamp for later comparison against
+// the configured TTL.
+type cacheItem[T any] struct {
+	value     T
+	timestamp time.Time
+}
+
+// cacheItemHistory stores cachedItems by height and maintains a sorted list of
```
Olshansk, line 46: Awesome.

Beginning to remind me of our interview question a bit 😅

pkg/client/query/cache/memory.go
```diff
@@ -0,0 +1,304 @@
...
+	return &InMemoryCache[T]{
+		items:  make(map[string]interface{}),
+		config: config,
+	}
+}
+
+// Get retrieves the value from the cache with the given key. If the cache is
+// configured for historical mode, it will return the value at the latest **known**
+// height, which is only updated on calls to SetAtHeight, and therefore is not
+// guaranteed to be the current height.
```
Olshansk, line 74:
```suggestion
// guaranteed to be the current height w.r.t the blockchain.
```

pkg/client/query/cache/memory.go
```diff
@@ -0,0 +1,304 @@
...
+	var zero T
+
+	item, exists := c.items[key]
+	if !exists {
+		return zero, ErrCacheMiss.Wrapf("key: %s", key)
+	}
+
+	cItem := item.(cacheItem[T])
+	if c.config.TTL > 0 && time.Since(cItem.timestamp) > c.config.TTL {
+		// DEV_NOTE: Intentionally not pruning here to improve concurrent speed;
```
Olshansk, line 92: So is there ever cleanup? Is it something we could defer or do lazyl elsewhere?

pkg/client/query/cache/memory.go
```diff
@@ -0,0 +1,304 @@
...
+		// otherwise, the read lock would be insufficient. The value will be
+		// overwritten by the next call to Set().
+		return zero, ErrCacheMiss.Wrapf("key: %s", key)
+	}
+
+	return cItem.value, nil
+}
+
+// GetAtHeight retrieves the value from the cache with the given key, at the given
+// height. If a value is not found for that height, the value at the nearest previous
```
Olshansk, line 102: > If a value is not found for that height, the value at the nearest previous height is returned.

Thoughts on adding a `nearestOk` bool to make this functionality ultra explicit?

pkg/client/query/cache/memory.go
```diff
@@ -0,0 +1,304 @@
...
+	defer c.itemsMu.RUnlock()
+
+	var zero T
+
+	item, exists := c.items[key]
+	if !exists {
+		return zero, ErrCacheMiss.Wrapf("key: %s", key)
+	}
+
+	cItem := item.(cacheItem[T])
```
Olshansk, line 90: Consider checking if `ok` because this is a type assertion and not just a type cast

https://stackoverflow.com/questions/20494229/what-is-the-difference-between-type-conversion-and-type-assertion

pkg/client/query/cache/memory.go
```diff
@@ -0,0 +1,304 @@
...
+	if c.config.TTL > 0 && time.Since(cItem.timestamp) > c.config.TTL {
```
Olshansk, line 91:
```suggestion
	isTTLEnabled := c.config.TTL > 0
	isCacheItemExpired := time.Since(cItem.timestamp) > c.config.TTL
	if isTTLEnabled && isCacheItemExpired {
```

pkg/client/query/cache/memory.go
```diff
@@ -0,0 +1,304 @@
...
+
+	c.itemsMu.RLock()
+	defer c.itemsMu.RUnlock()
+
+	itemHistoryAny, exists := c.items[key]
+	if !exists {
+		return zero, ErrCacheMiss.Wrapf("key: %s, height: %d", key, getHeight)
+	}
+
+	itemHistory := itemHistoryAny.(cacheItemHistory[T])
```
Olshansk, line 120: Ditto regarding assertion & casting.

pkg/client/query/cache/memory.go
```diff
@@ -0,0 +1,304 @@
...
+			break
+		}
+	}
+
+	if nearestCachedHeight == -1 {
+		return zero, ErrCacheMiss.Wrapf("key: %s, height: %d", key, getHeight)
+	}
+
+	item := itemHistory.itemsByHeight[nearestCachedHeight]
+	if c.config.TTL > 0 && time.Since(item.timestamp) > c.config.TTL {
```
Olshansk, line 138: Consider using the same language here as above

pkg/client/query/cache/memory.go
```diff
@@ -0,0 +1,304 @@
...
+			// getHeight.
+			break
+		}
+	}
+
+	if nearestCachedHeight == -1 {
+		return zero, ErrCacheMiss.Wrapf("key: %s, height: %d", key, getHeight)
+	}
+
+	item := itemHistory.itemsByHeight[nearestCachedHeight]
```
Olshansk, line 137: Should we check if the item exists?

Olshansk, line 137: Consider consolidating `item`/`value` language.

pkg/client/query/cache/memory.go
```diff
@@ -0,0 +1,304 @@
...
+// Set adds or updates the value in the cache for the given key. If the cache is
+// configured for historical mode, it will store the value at the latest **known**
+// height, which is only updated on calls to SetAtHeight, and therefore is not
+// guaranteed to be the current height.
+func (c *InMemoryCache[T]) Set(key string, value T) error {
+	if c.config.historical {
+		return c.SetAtHeight(key, value, c.latestHeight.Load())
+	}
+
+	if c.config.MaxKeys > 0 && int64(len(c.items)) >= c.config.MaxKeys {
```
Olshansk, line 158:
```suggestion
	maxKeysEnabled := c.config.MaxKeys > 0
	maxKeysReached := int64(len(c.items)) >= c.config.MaxKeys
	if maxKeysEnabled &&  maxKeysReached {
```

pkg/client/query/cache/memory.go
```diff
@@ -0,0 +1,304 @@
...
+	config       queryCacheConfig
+	latestHeight atomic.Int64
+
+	itemsMu sync.RWMutex
+	// items type depends on historical mode:
+	// | historical mode |  type                          |
+	// | --------------- | ------------------------------ |
+	// | false           | map[string]cacheItem[T]        |
+	// | true            | map[string]cacheItemHistory[T] |
+	items map[string]any
```
Olshansk, line 36: Why items instead of values?

pkg/client/query/cache/memory_test.go
```diff
@@ -0,0 +1,339 @@
+package cache
```
Olshansk, line 1: @bryanchriswhite I'm not reviewing this but lmk if there's a particular section in the tests you'd want me to take a look at.

pkg/client/query/cache/memory.go
```diff
@@ -0,0 +1,304 @@
...
+
+	return nil
+}
+
+// Delete removes an item from the cache.
+func (c *InMemoryCache[T]) Delete(key string) {
+	c.itemsMu.Lock()
+	defer c.itemsMu.Unlock()
+
+	delete(c.items, key)
```
Olshansk, line 245: What if we delete the (key,value) pair for `latestHeight`?

pkg/client/query/cache/memory_test.go
```diff
@@ -0,0 +1,339 @@
...
+
+		// Wait for TTL to expire
+		time.Sleep(150 * time.Millisecond)
+
+		// Value should now be expired
+		_, err = cache.Get("key")
+		require.ErrorIs(t, err, ErrCacheMiss)
+	})
+
+	t.Run("max size eviction", func(t *testing.T) {
```
Olshansk, line 62:
```suggestion
	t.Run("max key count eviction", func(t *testing.T) {
```

Olshansk (changes requested):
@bryanchriswhite I did a superficial review but did not dive into the validation of the business logic line-by-line.

Is there any section where you'd want another pair of 👀 ?

pkg/client/query/cache/memory.go
bryanchriswhite: This work (and #995) is more about interface discovery and laying a solid foundation than implementing an in-memory cache, which isn't the end goal. In fact, we MUST iterate towards a persistent cache as it's necessary in order for off-chain binaries to tolerate restarts; e.g., a relayminer restarting in-between claim and proof windows.

My intuition regarding the persistent cache is that it should be backed by badger (or whatever we're already using for off-chain persistence), making the cache implementation itself, a thin wrapper around the badger API. Additionally, I think we should consider designing an off-chain "multi-store", similar in princple to the on-chain multi-store. The rationale is that we should prefer consolidating off-chain persistence into a single database over maintaining multiple, feature-specific databases. I don't see this as a huge lift given the solid foundational architecture/conventions we've got going (e.g. `context.Context`, `depinject`, etc.).

All things considered, I determined that it didn't make sense to use an external dependency here. Was there a particular library that had caught your attention?

I'd be happy to continue the discussion here or elsewhere, and/or open an issue.

pkg/client/query/cache/memory.go
bryanchriswhite: Thanks for the explanation, I'm familiar with the distinction. :blush:

pkg/client/query/cache/memory.go
bryanchriswhite: Will #PUC.

pkg/client/query/cache/memory.go
bryanchriswhite: Could you elaborate, I'm not quite following.

pkg/client/query/cache/memory.go
bryanchriswhite: Yes, will #PUC.

pkg/client/query/cache/config.go
```diff
@@ -0,0 +1,69 @@
...
+	LeastRecentlyUsed
+	LeastFrequentlyUsed
+)
+
+// queryCacheConfig is the configuration for query caches. It is intended to be
+// configured via QueryCacheOptionFn functions.
+type queryCacheConfig struct {
+	// MaxKeys is the maximum number of items (key/value pairs) the cache can
+	// hold before it starts evicting.
+	MaxKeys        int64
```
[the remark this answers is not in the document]
bryanchriswhite: Good catch - thanks!

pkg/client/query/cache/config.go
```diff
@@ -0,0 +1,69 @@
...
+	EvictionPolicy EvictionPolicy
+	// TTL is how long items should remain in the cache. Items older than the TTL
+	// MAY not be evicted but SHOULD not be considered as cache hits.
```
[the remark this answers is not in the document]
bryanchriswhite: > What determines if it will or will not be evicted? MAY NOT is a bit confusing.

The implementation determines that, this config doesn't know anything about the implementation.
In the case of `InMemoryCache`, eviction only happens when on a per-key/value basis, when a key/value is updated.

> Should we have a WILL NOT be considered cache hits? I can't tell what the behaviour is with SHOULD NOT

:+1: Agreed.

pkg/client/query/cache/config.go
bryanchriswhite: I considered `MaxEntries` but didn't feel like it was as clear as `MaxKeys`. It would be more idiomatic if we were in a different language (e.g. javascript, python, ruby, C#, etc.). :sweat_smile:

pkg/client/query/cache/memory.go
bryanchriswhite: Nope - good catch! :raised_hands:

pkg/client/query/cache/memory.go
bryanchriswhite: I restructured things a bit and did some renaming:

```go
// inMemoryCache provides a concurrency-safe in-memory cache implementation with
// optional historical value support.
type inMemoryCache[T any] struct {
	config       queryCacheConfig
	latestHeight atomic.Int64

	// valuesMu is used to protect values AND valueHistories from concurrent access.
	valuesMu sync.RWMutex
	// values holds the cached values in non-historical mode.
	values map[string]cacheValue[T]
	// valueHistories holds the cached historical values in historical mode.
	valueHistories map[string]cacheValueHistory[T]
}

// cacheValue wraps cached values with a timestamp for later comparison against
// the configured TTL.
type cacheValue[T any] struct {
	value     T
	timestamp time.Time
}

// cacheValueHistory stores cachedItems by height and maintains a sorted list of
// heights for which cached items exist. This list is sorted in descending order
// to improve performance characteristics by positively correlating index with age.
type cacheValueHistory[T any] struct {
	// sortedDescHeights is a list of the heights for which values are cached.
	// It is sorted in descending order.
	sortedDescHeights []int64
	heightMap         map[int64]cacheValue[T]
}
```

pkg/client/query/cache/memory.go
bryanchriswhite:
```go
	if isTTLEnabled && isCacheItemExpired {
		// DEV_NOTE: Intentionally not pruning here to improve concurrent speed;
		// otherwise, the read lock would be insufficient. The value will be
		// overwritten by the next call to Set(). If usage is such that values
		// aren't being subsequently set, maxKeys (if configured) will eventually
		// cause the pruning of values with expired TTLs.
		return zero, ErrCacheMiss.Wrapf("key: %s", key)
	}
```

pkg/client/query/cache/memory.go
```diff
@@ -10,116 +10,119 @@ import (
...
+// cacheValueHistory stores cachedItems by height and maintains a sorted list of
 // heights for which cached items exist. This list is sorted in descending order
 // to improve performance characteristics by positively correlating index with age.
-type cacheItemHistory[T any] struct {
+type cacheValueHistory[T any] struct {
 	// sortedDescHeights is a list of the heights for which values are cached.
 	// It is sorted in descending order.
 	sortedDescHeights []int64
-	itemsByHeight     map[int64]cacheItem[T]
+	heightMap         map[int64]cacheValue[T]
```
Olshansk, line 53: I was considering `heightCacheValueMap`, but `heightMap` just doesn't feel like enough.
```suggestion
	heightValueMap         map[int64]cacheValue[T]
```

pkg/client/query/cache/memory.go
```diff
@@ -10,116 +10,119 @@ import (
...
-	value     T
-	timestamp time.Time
+type cacheValue[T any] struct {
+	value    T
+	cachedAt time.Time
 }
 
-// cacheItemHistory stores cachedItems by height and maintains a sorted list of
+// cacheValueHistory stores cachedItems by height and maintains a sorted list of
 // heights for which cached items exist. This list is sorted in descending order
```
Olshansk, line 47:
```suggestion
// heights for which cached values exist. This list is sorted in descending order
```

pkg/client/query/cache/memory.go
Olshansk: When I see a function like `GetAtHeight`, my immediate thought is that it'll return the value at the height provided (if available) or nil otherwise.

For `height=5`, it's the difference between `select * from table where height=5` and `select * from table where height <= 5`. 

There could be business logic that depends on state transitions AT A SPECIFIC HEIGHT, so I'm just afraid that it's not explicit enough.

pkg/client/query/cache/memory.go
```diff
@@ -134,37 +137,47 @@ func (c *InMemoryCache[T]) GetAtHeight(key string, getHeight int64) (T, error) {
...
 	}
 
-	item := itemHistory.itemsByHeight[nearestCachedHeight]
-	if c.config.TTL > 0 && time.Since(item.timestamp) > c.config.TTL {
+	value, exists := valueHistory.heightMap[nearestCachedHeight]
+	if !exists {
+		return zero, ErrCacheInternal.Wrapf("failed to load historical value for key: %s, height: %d", key, getHeight)
+	}
+
+	if c.config.ttl > 0 && time.Since(value.cachedAt) > c.config.ttl {
```
Olshansk, line 145: Looks like `isTTLEnabled` dissapeared: https://github.com/pokt-network/poktroll/pull/994#discussion_r1886643319

pkg/client/query/cache/memory.go
Olshansk: This is great, ty!

pkg/client/query/cache/memory.go
Olshansk: I think the `if isTTLEnabled` part was lost as part of another commit.

pkg/client/query/cache/memory.go
Olshansk: Yea, this comment doesn't make sense if you don't know where I was looking or what I was thinking...

If we delete the key for the latest height, do we need to also make an appropriate call to `c.latestHeight.Store(??)`?

If this is indeed the case, please add a unit test for it.

If not, would just appreciate a quick (no need for anything big) explanation.

pkg/client/query/cache/memory.go
Olshansk: Haha - too many late nights on python bindings ;)

pkg/client/query/cache/memory.go
Olshansk: > All things considered, I determined that it didn't make sense to use an external dependency here. Was there a particular library that had caught your attention?

Nope, just wanted to get your thought process!

> I'd be happy to continue the discussion here or elsewhere, and/or open an issue.

I'll request one better.

Why not transfer the explanation you have here into a `pkg/client/query/cache/README.md` with ` TODO_FUTURE` assigned to your name?

pkg/client/query/cache/memory.go
```diff
@@ -0,0 +1,362 @@
...
+	// Update sortedDescHeights and ensure the list is sorted in descending order.
+	if _, setHeightExists := valueHistory.heightMap[setHeight]; !setHeightExists {
+		valueHistory.sortedDescHeights = append(valueHistory.sortedDescHeights, setHeight)
+		sort.Slice(valueHistory.sortedDescHeights, func(i, j int) bool {
+			return valueHistory.sortedDescHeights[i] > valueHistory.sortedDescHeights[j]
+		})
+	}
+
+	// Prune historical values for this key, where the setHeight
+	// is oder than the configured numHistoricalValues.
```
red-0ne, line 222:
```suggestion
	// is older than the configured numHistoricalValues.
```

pkg/client/query/cache/memory.go
```diff
@@ -0,0 +1,362 @@
...
+	if c.config.numHistoricalValues > 0 {
+		lenCachedHeights := int64(len(valueHistory.sortedDescHeights))
+		for heightIdx := lenCachedHeights - 1; heightIdx >= 0; heightIdx-- {
+			cachedHeight := valueHistory.sortedDescHeights[heightIdx]
+
+			// DEV_NOTE: Since the list is sorted, and we're iterating from lowest
+			// (oldest) to highest (youngest) height, once we encounter a cachedHeight
+			// that is younger than the configured numHistoricalValues, ALL subsequent
+			// heights SHOULD also be younger than the configured numHistoricalValues.
+			if setHeight-cachedHeight <= c.config.numHistoricalValues {
```
red-0ne, line 232: I struggle to make sense of this.

As of my understanding `config.numHistoricalValues` would be the "size" of the historical values retained. If so, the condition to delete an entry would be:
```
if
  len(valueHistory.sortedDescHeights) > config.numHistoricalValues &&
  setHeight.isYoungerThan(oldestCachedHeight)
```

red-0ne (commented):
Great caching design.

Left some comments but did not tend to `pkg/client/query/cache/memory_test.go` yet. Which I'll do right after this one.

pkg/client/query/cache/memory.go
bryanchriswhite: Gotcha, good question. I don't think we need to update `c.latestHeight`, regardless of which key is being deleted.  For `Get`s, this should be completely transparent to the caller; for `Set`s, it means that if you call `Delete`, followed by `Set` (as opposed to `SetAtHeight`), then the height you're setting will be the last `c.latestHeight`. I think this behavior is correct. 

> If not, would just appreciate a quick (no need for anything big) explanation.
I did see :point_up_2: but had to think it through for myself anyways:

```mermaid
flowchart LR

subgraph hc2[Historical Cache]
    lh2["latestHeight: 2"]
    subgraph k2_2h["key2 history"]
        k2_2h1["height: 1; value 100"]
    end
    subgraph k2_1h["key1 history"]
        k2_1h1["height: 1; value 10"]
        k2_1h2["height: 2; value 20"]
    end
end

subgraph hc3[Historical Cache]
    lh3["latestHeight: 2"]
    subgraph k3_2h["key2 history"]
        k3_2h1["height: 1; value 100"]
    end
end


subgraph hc4[Historical Cache]
    lh4["latestHeight: 2"]
    subgraph k4_1h["key1 history"]
        k4_1h2["height: 2; value 30"]
    end
end

hc2 --"Delete(key1)"--> hc3
hc3 --"Set(key1, 30)"--> hc4

lh2 --> k2_1h2
lh4 --> k4_1h2
```

pkg/client/query/cache/memory.go
bryanchriswhite: Great catch - thank you! :raised_hands:

pkg/client/query/cache/memory.go
bryanchriswhite: I take your point, and agree that his is a valid concern. However, I think we're getting hung up on an implementation detail that is actually an optimization. If the implementation were completely honest, it would be populating the history at **each** height, where the value would be the same as the previous height so long as it wasn't changed (or deleted; we have few of these objects: claims/proofs).

The fact that the heights are omitted here is an optimization which takes advantage of the assumption that this pattern holds due to the blockchain context that we're in. Even in the case of the few onchain objects which do violate this assumption , I don't think it becomes problematic. I also don't think we have a good reason to consider caching claims or proofs yet.

A concrete example would be something like:
1. A claim was created at height 10
2. The claim was queried and cached at height 12
3. The claim was settled and deleted at height 20
4. The claim was queried at height 30, for height 22
5. The claim would be returned even though it was deleted two blocks prior to the queried height

pkg/client/query/cache/config.go
```diff
@@ -13,57 +13,77 @@ const (
...
+		return ErrQueryCacheConfigValidation.Wrapf("eviction policy %d not imlemented", cfg.evictionPolicy)
+	}
+
+	return nil
+}
+
+// WithHistoricalMode enables historical caching with the given numHistoricalValues
 // configuration; if 0, no historical pruning is performed.
-func WithHistoricalMode(pruneOlderThan int64) QueryCacheOptionFn {
+func WithHistoricalMode(numHistoricalBlocks int64) QueryCacheOptionFn {
```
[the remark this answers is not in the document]
bryanchriswhite: I intentionally chose `numHistoricalBlocks` over `numHistoricalValues` to disambiguate "key/value" from "historical value" and to imply the frequency. I suppose it's not quite appropriate to imply the frequency at this level, as it's out of the scope this component's concerns. After a chat w/ @red-0ne I would rename it to `numRetainedVersions`. Wdyt?

pkg/client/query/cache/memory.go
```diff
@@ -10,116 +10,119 @@ import (
...
-type cacheItem[T any] struct {
-	value     T
-	timestamp time.Time
+type cacheValue[T any] struct {
+	value    T
+	cachedAt time.Time
 }
 
-// cacheItemHistory stores cachedItems by height and maintains a sorted list of
+// cacheValueHistory stores cachedItems by height and maintains a sorted list of
```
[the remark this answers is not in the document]
bryanchriswhite: That was the last one. :sweat_smile:

pkg/client/query/cache/memory.go
bryanchriswhite: Continuing the direction of https://github.com/pokt-network/poktroll/pull/994#discussion_r1890025275, I'm going with `heightVersionsMap` here.

pkg/client/query/cache/memory.go
bryanchriswhite:
```go
// HistoricalQueryCache extends QueryCache to support getting and setting values
// at multiple heights for a given key.
type HistoricalQueryCache[T any] interface {
	QueryCache[T]
	// GetAsOfHeight retrieves the nearest value <= the specified height
	GetAsOfHeight(key string, height int64) (T, error)
	// SetAsOfHeight adds or updates a value at a specific height
	SetAsOfHeight(key string, value T, height int64) error
}
```

pkg/client/query/cache/config.go
```diff
@@ -0,0 +1,89 @@
...
+// Validate ensures that the queryCacheConfig isn't configured with incompatible options.
+func (cfg *queryCacheConfig) Validate() error {
+	switch cfg.evictionPolicy {
+	case FirstInFirstOut:
+	// TODO_IMPROVE: support LeastRecentlyUsed and LeastFrequentlyUsed policies.
+	default:
+		return ErrQueryCacheConfigValidation.Wrapf("eviction policy %d not imlemented", cfg.evictionPolicy)
+	}
+
+	return nil
```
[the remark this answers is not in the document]
bryanchriswhite:
```go
	if cfg.numRetainedVersions > 0 && cfg.historical == false {
		return ErrQueryCacheConfigValidation.Wrap("numRetainedVersions > 0 requires historical mode to be enabled")
	}

	if cfg.historical && cfg.numRetainedVersions < 0 {
		return ErrQueryCacheConfigValidation.Wrapf("numRetainedVersions MUST be >= 0, got: %d", cfg.numRetainedVersions)
	}
```

pkg/client/query/cache/memory.go
```diff
@@ -0,0 +1,362 @@
...
+	var zero T
+
+	if !c.config.historical {
+		return zero, ErrHistoricalModeNotEnabled
+	}
+
+	c.valuesMu.RLock()
+	defer c.valuesMu.RUnlock()
+
+	valueHistory := c.valueHistories[key]
```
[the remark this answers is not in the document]
bryanchriswhite: :pray:

pkg/client/query/cache/memory.go
```diff
@@ -0,0 +1,362 @@
...
+// configured for historical mode, it will store the value at the latest **known**
+// height, which is only updated on calls to SetAtHeight, and therefore is not
+// guaranteed to be the current height.
+func (c *inMemoryCache[T]) Set(key string, value T) error {
+	if c.config.historical {
+		return c.SetAtHeight(key, value, c.latestHeight.Load())
+	}
+
+	isMaxKeysConfigured := c.config.maxKeys > 0
+	cacheHasMaxKeys := int64(len(c.values)) >= c.config.maxKeys
```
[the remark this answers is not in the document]
bryanchriswhite: Great catch - thank you! :raised_hands:

pkg/client/query/cache/memory.go
```diff
@@ -0,0 +1,362 @@
...
+	return value.value, nil
+}
+
+// Set adds or updates the value in the cache for the given key. If the cache is
+// configured for historical mode, it will store the value at the latest **known**
+// height, which is only updated on calls to SetAtHeight, and therefore is not
+// guaranteed to be the current height.
+func (c *inMemoryCache[T]) Set(key string, value T) error {
+	if c.config.historical {
+		return c.SetAtHeight(key, value, c.latestHeight.Load())
```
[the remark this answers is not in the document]
bryanchriswhite: I think this is more about the specific behavior of `Set()` in historical mode. This is indeed an edge case, and I don't think we should find ourselves doing this unless for a very specific reason. In such a case, I think it's more accurate to say that instead of `fetchValue()` updating the latest version (height), it would be an intermediate call to `SetAsOfVersion()` where the version is newer than the current latest version. I think this is the scenario I outlined above: https://github.com/pokt-network/poktroll/pull/994#discussion_r1889842645.

pkg/client/query/cache/memory.go
bryanchriswhite: :hankey:  :joy:

pkg/client/query/cache/memory.go
bryanchriswhite: > Why not transfer the explanation you have here into a pkg/client/query/cache/README.md with  TODO_FUTURE assigned to your name?

I was compelled to open an issue instead. :sweat_smile: 

- https://github.com/pokt-network/poktroll/issues/1014

pkg/client/query/cache/memory.go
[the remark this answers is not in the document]
bryanchriswhite: As discussed, I've disabled `#Set()` in historical mode.

pkg/client/query/cache/memory.go
```diff
@@ -0,0 +1,378 @@
...
+	if !c.config.historical {
+		return ErrHistoricalModeNotEnabled
+	}
+
+	// Update c.latestVersion if the given version is newer (higher).
+	latestVersion := c.latestVersion.Load()
+	if version > latestVersion {
+		// NB: Only update if c.latestVersion hasn't changed since we loaded it above.
+		if c.latestVersion.CompareAndSwap(latestVersion, version) {
+			latestVersion = version
```
red-0ne, line 209:
```suggestion
			latestVersion = version
			c.latestVersion.Store(version)
```

pkg/client/query/cache/memory.go
```diff
@@ -0,0 +1,362 @@
...
+		config:         config,
+	}, nil
+}
+
+// Get retrieves the value from the cache with the given key. If the cache is
+// configured for historical mode, it will return the value at the latest **known**
+// height, which is only updated on calls to SetAtHeight, and therefore is not
+// guaranteed to be the current height w.r.t the blockchain.
+func (c *inMemoryCache[T]) Get(key string) (T, error) {
+	if c.config.historical {
```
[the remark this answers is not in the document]
red-0ne: Looking at this function it was not clear why this conditional block is excluded from the `valuesMu` lock.

Now I know that it `GetAsOfVersion` is locking too, I could no longer justify placing #PUCing it :sweat_smile:

red-0ne (commented):
Asking for a small change, but otherwise LGTM.

Not approving yet as I want review the tests too.

pkg/client/query/cache/memory.go
bryanchriswhite: Can you elaborate on the thinking here? In this case, the `latestVersion` was updated to an even "newer/later" concurrently. The point of using `#CompareAndSwap()` here was to avoid the latest version from regressing.

red-0ne: My bad. I overlooked the fact that `CompareAndSwap` do swap too :sweat_smile:

pkg/client/query/cache/memory.go
Olshansk: _tl;dr I still disagree but will leave the thread here :)_

There are "implicit assumptions"  in this conversation/thread.

Whenever that happens, I always think it's something that requires context or institutional knowledge. Further, it's something I can see us forgetting as well.

In either case, you are the main implementor and designer here. If this becomes an issue, I'm sure it'll resurface.

pkg/client/query/cache/memory.go
Olshansk: Given the explanation and diagram, I don't fully understand the need for `latestHeight` at all. In fact, I think we should delete it now if the behaviour you described is expected.

pkg/client/interface.go
```diff
@@ -375,8 +375,8 @@ type QueryCache[T any] interface {
 // at multiple heights for a given key.
 type HistoricalQueryCache[T any] interface {
 	QueryCache[T]
-	// GetAtHeight retrieves the nearest value <= the specified height
-	GetAtHeight(key string, height int64) (T, error)
-	// SetAtHeight adds or updates a value at a specific height
-	SetAtHeight(key string, value T, height int64) error
+	// GetAsOfVersion retrieves the nearest value <= the specified version number.
+	GetAsOfVersion(key string, version int64) (T, error)
```
Olshansk, line 379: Can we go with one of the following instead:

1. GetVersioned
2. GetAtVersion
3. GetHistorical

Don't care which one but `AsOf` in a function name just doesn't feel right.

pkg/client/query/cache/config.go
```diff
@@ -13,57 +13,86 @@ const (
...
+	// TODO_CONSIDERATION:
+	//
+	// maxValueSize is the maximum cumulative size of all values in the cache.
+	// maxValueSize int64
+	// maxCacheSize is the maximum cumulative size of all keys AND values in the cache.
+	// maxCacheSize int64
+
+	// evictionPolicy determines which values are removed when number of keys in the cache reaches maxKeys.
+	evictionPolicy EvictionPolicy
+	// ttl is how long values should remain valid in the cache. Items older than the
```
Olshansk, line 32: This updated comment is great

pkg/client/query/cache/config.go
```diff
@@ -13,57 +13,86 @@ const (
...
+	// ttl MAY NOT be evicted immediately, but are NEVER considered as cache hits.
+	ttl time.Duration
+	// historical determines whether each key will point to a single values (false)
+	// or a history (i.e. reverse chronological list) of values (true).
 	historical bool
-	// pruneOlderThan is the number of past blocks for which to keep historical
-	// values. If 0, no historical pruning is performed. It only applies when
-	// historical is true.
-	pruneOlderThan int64
+	// maxVersionAge is the max difference between the latest known version and
```
Olshansk, line 38: this sentence is hard to grok.

Any chance you could improve it with an example?

pkg/client/query/cache/memory.go
```diff
@@ -10,295 +10,369 @@ import (
...
-	itemsMu sync.RWMutex
-	// items type depends on historical mode:
-	// | historical mode |  type                          |
-	// | --------------- | ------------------------------ |
-	// | false           | map[string]cacheItem[T]        |
-	// | true            | map[string]cacheItemHistory[T] |
-	items map[string]any
+type inMemoryCache[T any] struct {
+	config        queryCacheConfig
+	latestVersion atomic.Int64
```
Olshansk, line 29: Per my other comment, can you #PUC why we need this and how its used?

Olshansk (changes requested):

pkg/client/interface.go
bryanchriswhite: In that case, `GetVersion` would be my preference. Any objections @red-0ne @Olshansk?

pkg/client/query/cache/memory.go
bryanchriswhite: :saluting_face: :sweat_smile:

pkg/client/query/cache/memory.go
bryanchriswhite: I neglected to add a comment to this thread but after discussing with @red-0ne, we decided that it should be an error to call `Set()` on a `HistoricalQueryCache`. I will update the interface and `inMemoryCache#Set()` to reflect this. In this context, `latestVersion` is no longer necessary and can be removed.

bryanchriswhite:
```
// QueryCache is a key/value store style interface for a cache of a single type.
// It is intended to be used to cache query responses (or derivatives thereof),
// where each key uniquely indexes the most recent query response.
type QueryCache[T any] interface {
	Get(key string) (T, error)
	Set(key string, value T) error
	Delete(key string)
	Clear()
}

// HistoricalQueryCache extends QueryCache to support getting and setting values
// at multiple heights for a given key.
type HistoricalQueryCache[T any] interface {
	// GetLatestVersion retrieves the historical value with the highest version number.
	GetLatestVersion(key string) (T, error)
	// GetAsOfVersion retrieves the nearest value <= the specified version number.
	GetAsOfVersion(key string, version int64) (T, error)
	// SetAsOfVersion adds or updates a value at a specific version number.
	SetAsOfVersion(key string, value T, version int64) error
}
```
(class diagrams also updated)

pkg/client/interface.go
bryanchriswhite: Also adding `GetLatestVersion`.

Olshansk: @bryanchriswhite Can we add `GetLtestVersino`

Olshansk (changes requested):

Olshansk:
@bryanchriswhite Could we get this over the finish line ASAP so it can be used in #1050?

red-0ne (commented):
Left few comments but it feels like the current implementation is calling for having an interface and implementation for each type of cache (historical, non-historical).

It will remove a lot of conditional branching and runtime errors.

pkg/client/query/cache/memory.go
```diff
@@ -0,0 +1,412 @@
...
+	if c.config.historical {
+		return c.GetLatestVersion(key)
+	}
+
+	c.valuesMu.RLock()
+	defer c.valuesMu.RUnlock()
+
+	cachedValue, exists := c.values[key]
+	if !exists {
+		return zero, ErrCacheMiss.Wrapf("key: %s", key)
```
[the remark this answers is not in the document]
bryanchriswhite: We can't literally return `cachedValue` because it is not of type `T`, it is a `cacheValue[T]`, the `#value` field of which would have our `T` type; however, since `cachedValue` is `nil`, we cannot access it.

pkg/client/query/cache/memory.go
```diff
@@ -0,0 +1,412 @@
...
+
+// evictHistorical removes one value (and all its versions) from the cache,
+// to make space for a new one, according to the configured eviction policy.
+func (c *inMemoryKVCache[T]) evictHistorical() error {
+	switch c.config.evictionPolicy {
+	case FirstInFirstOut:
+		var oldestKey string
+		var oldestTime time.Time
+		for key, valueHistory := range c.valueHistories {
+			mostRecentVersion := valueHistory.sortedDescVersions[0]
```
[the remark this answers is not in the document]
bryanchriswhite: I think we only disagree about the word "recent". I was seeing it as interchangeable with "latest", where both would refer to the "highest version number".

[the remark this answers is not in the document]
bryanchriswhite: I think we only disagree about the word "recent". I was seeing it as interchangeable with "latest", where both would refer to the "highest version number".

pkg/cache/memory/historical_kvcache_test.go
```diff
@@ -0,0 +1,179 @@
...
+		val, isCached = cache.GetVersion("key", 25)
+		require.True(t, isCached)
+		require.Equal(t, "value2", val)
+
+		// Test getting a version for a key that isn't cached
+		_, isCached = cache.GetVersion("key2", 20)
+		require.False(t, isCached)
+	})
+
+	t.Run("historical cache ignores TTL expiration", func(t *testing.T) {
```
red-0ne, line 74: Historical cache is still subject to `maxKeys` FIFO eviction policy.

I believe it's worth test covering it.

Olshansk (changes requested):
Just realized I never published my review from yesterday :(

pkg/cache/interface.go
```diff
@@ -0,0 +1,22 @@
...
+// It is intended to be used to cache query responses (or derivatives thereof),
+// where each key uniquely indexes the most recent query response.
+type KeyValueCache[T any] interface {
+	Get(key string) (T, bool)
+	Set(key string, value T) error
+	Delete(key string)
+	Clear()
+}
+
+// HistoricalKeyValueCache extends KeyValueCache to support getting and setting values
```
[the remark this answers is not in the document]
bryanchriswhite:
```go
// HistoricalKeyValueCache is a key/value store style interface for a cache of a single type.
// It is intended to be used to cache arbitrary data, where each key uniquely indexes
// a mapping of version numbers to values corresponding to the historical values of the data
// associated that key.
```

pkg/cache/errors.go
```diff
@@ -0,0 +1,10 @@
+package cache
+
+import "cosmossdk.io/errors"
+
+const codesace = "client/query/cache"
```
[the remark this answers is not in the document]
bryanchriswhite: Nice catch! :raised_hands: 

Given that it's no longer in the `client` pkg, I'm going with:
```go
const codespace = "cache"
```

pkg/cache/memory/kvcache.go
```diff
@@ -0,0 +1,150 @@
...
+			if first || value.cachedAt.Before(oldestTime) {
+				oldestKey = key
+				oldestTime = value.cachedAt
+			}
+			first = false
+		}
+		delete(c.values, oldestKey)
+		return
+
+	case LeastRecentlyUsed:
```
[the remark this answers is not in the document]
bryanchriswhite: I moved the comment above this line and rephrased slightly:
```go
	// DEV_NOTE: The following cases SHOULD NEVER happen, KeyValueCacheConfig#Validate, SHOULD prevent it.
	case LeastRecentlyUsed:
```

pkg/cache/memory/historical_kvcache.go
```diff
@@ -0,0 +1,233 @@
...
+		valueHistories: make(map[string]cacheValueHistory[T]),
+		config:         config,
+	}, nil
+}
+
+// GetVersion retrieves the value from the cache with the given key, as of the
+// given version. If a value is not found for that version, the value at the nearest
+// previous version is returned. If the cache is not configured for historical mode,
+// it returns an error.
+func (c *historicalKeyValueCache[T]) GetVersion(key string, version int64) (T, bool) {
```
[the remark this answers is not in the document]
bryanchriswhite: Thank you! :heart:

pkg/cache/memory/historical_kvcache.go
```diff
@@ -0,0 +1,233 @@
...
+	sortedDescVersions []int64
+	// versionToValueMap is a map from a version number to the cached value at
+	// that version number, if present.
+	versionToValueMap map[int64]cacheValue[T]
+}
+
+// NewHistoricalKeyValueCache creates a new historicalKeyValueCache with the
+// configuration generated by the given option functions.
+func NewHistoricalKeyValueCache[T any](opts ...KeyValueCacheOptionFn) (*historicalKeyValueCache[T], error) {
+	config := DefaultHistoricialKeyValueCacheConfig
```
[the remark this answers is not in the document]
bryanchriswhite: :pray:

pkg/cache/memory/historical_kvcache.go
```diff
@@ -0,0 +1,233 @@
...
+		return zero, false
+	}
+
+	value, exists := valueHistory.versionToValueMap[nearestCachedVersion]
+	if !exists {
+		// DEV_NOTE: This SHOULD NEVER happen. If it does, it means that the cache has been corrupted.
+		return zero, false
+	}
+
+	return value.value, true
```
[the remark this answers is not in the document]
bryanchriswhite: I went with `cachedValue` for consistency.

pkg/cache/memory/config.go
```diff
@@ -0,0 +1,163 @@
...
+	case FirstInFirstOut:
+	// TODO_IMPROVE: support LeastRecentlyUsed and LeastFrequentlyUsed policies.
+	default:
+		return cache.ErrKeyValueCacheConfigValidation.Wrapf("eviction policy %d not imlemented", cfg.evictionPolicy)
+	}
+	return nil
+}
+
+// Validate ensures that the historicalKeyValueCacheConfig isn't configured with incompatible options.
+func (cfg *historicalKeyValueCacheConfig) Validate() error {
```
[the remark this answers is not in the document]
bryanchriswhite:
```go
// Validate ensures that the historicalKeyValueCacheConfig isn't configured with incompatible options.
func (cfg *historicalKeyValueCacheConfig) Validate() error {
	if err := cfg.keyValueCacheConfig.Validate(); err != nil {
		return err
	}
...
```

pkg/cache/memory/historical_kvcache.go
```diff
@@ -0,0 +1,233 @@
...
+
+	return c.GetVersion(key, version)
+}
+
+// SetVersion adds or updates the historical value in the cache for the given key,
+// and at the version number. If the cache is not configured for historical mode, it
+// returns an error.
+func (c *historicalKeyValueCache[T]) SetVersion(key string, value T, version int64) error {
+	// DEV_NOTE: MUST call getLatestVersionNumber() before locking valuesMu.
+	latestVersion := c.getLatestVersionNumber(key)
```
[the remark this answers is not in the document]
bryanchriswhite: Agreed.

pkg/client/query/cache/config.go
```diff
@@ -0,0 +1,100 @@
...
+	evictionPolicy EvictionPolicy
+	// ttl is how long values should remain valid in the cache. Items older than the
+	// ttl MAY NOT be evicted immediately, but are NEVER considered as cache hits.
+	ttl time.Duration
+	// historical determines whether each key will point to a single values (false)
+	// or a history (i.e. reverse chronological list) of values (true).
+	historical bool
+	// maxVersionAge is the max difference between the latest known version and
+	// any other version, below which value versions are retained, and above which
+	// value versions are pruned.
```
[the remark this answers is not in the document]
bryanchriswhite: All of the above ("pruned" == "evicted" == "deleted"). As it states, they ARE pruned. The word "evict" has only been used so far in a slightly different context, related to the `MaxKeys` config field.

Do you have a specific suggestion on how to improve clarity here?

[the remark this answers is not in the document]
Olshansk:
```suggestion
// value versions are pruned (i.e. evicted).
```

Olshansk (commented):
@bryanchriswhite I did a "show changes since your last review" and it seems like all the files have changed?

Either way, lmk when this is ready for a re-review.

bryanchriswhite:
Based on prior feedback and the continued iteration here, I've refactored `GetVersion()` to ONLY return the specific version requested, and added a new `GetVersionLTE()` method which implements the previous behavior:
```
	// GetVersionLTE retrieves the value at the nearest version <= maxVersion.
	// It returns true if a value meeting the version criteria for the given
	// key is cached, false otherwise.
	GetVersionLTE(key string, maxVersion int64) (T, bool)
```

red-0ne (approved):
Love the shape it took :fire:

Thanks for your patience iterating over it.

Olshansk (approved):
Love the final shape of this!

bryanchriswhite:
PLEASE DO NOT DELETE THIS BRANCH UNTIL DOWNSTREAM PRS HAVE BEEN REBASED

PR #1043, opened 2025-01-24
[Upgrades] v0.0.12 upgrade

okdas:
## Summary

Adding an upgrade handler for `v0.0.12`

## Type of change

- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [x] Other (specify)

## Sanity Checklist


app/upgrades/v0.0.12.go
```diff
@@ -0,0 +1,130 @@
...
+			// GetParams() to avoid potential protobuf issues and all networks have the same value (no need to read).
+			// Validate with: `poktrolld q supplier params --node=https://testnet-validated-validator-rpc.poktroll.com/`
+			supplierParams := suppliertypes.Params{
+				MinStake: &cosmosTypes.Coin{
+					Denom:  "upokt",
+					Amount: math.NewInt(1000000),
+				},
+				StakingFee: &cosmosTypes.Coin{
+					Denom: "upokt",
+					// TODO_IN_THIS_PR: 100upokt a good value?
```
red-0ne, line 45: I think `100upokt` or `1000upokt` is a good value for now. We have the transaction fees that also protect from sibyl.

app/upgrades/v0.0.12.go
```diff
@@ -0,0 +1,130 @@
...
+			// Since we changed the type of RevSharePercent from float32 to uint64, we need to update all on-chain data.
+			// The easiest way to do this is to iterate over all suppliers and services and set the revshare to 100 by default.
+			suppliers := keepers.SupplierKeeper.GetAllSuppliers(ctx)
+			logger.Info("Updating all suppliers to have a 100% revshare to the supplier", "num_suppliers", len(suppliers))
+			for _, supplier := range suppliers {
+				for _, service := range supplier.Services {
+					// Force all services to have a 100% revshare to the supplier.
+					// Not something we would do on a real mainnet, but it's a quick way to resolve the issue.
+					// Currently, we don't break any existing suppliers (as all of them have a 100% revshare to the supplier).
+					service.RevShare = []*sharedtypes.ServiceRevenueShare{
```
red-0ne, line 106: Ideally we would be able to loop over the old revshare map by unmashalling to the old proto type then create the new `Supplier` proto with the new values.

But since we don't have proto versioning and we marked the old (float32) value as reserved, this is the next best option offered to us.

red-0ne (approved):
LGTM! I feel we are starting to improve our migration plans :clap:

app/upgrades/v0.0.12.go
```diff
@@ -0,0 +1,130 @@
...
+					Amount: math.NewInt(1000000),
+				},
+				StakingFee: &cosmosTypes.Coin{
+					Denom: "upokt",
+					// TODO_IN_THIS_PR: 100upokt a good value?
+					Amount: math.NewInt(100),
+				},
+			}
+
+			// ALL parameters must be present when setting params.
```
Olshansk, line 50: This sort of thing makes me question the avoidance of not using `GetParams` then?

Should we update the comments and docs to:
1. Show how to do a `GetParams`
2. Copy-paste the response
3. Tell the user to update only what they care about?

I know I'm being pedantic, but I'm going for the "stupid & lazy people (i.e. me) filter here"

app/upgrades/v0.0.12.go
Olshansk: It'll all be revisited prior to mainnet. 100upokt LGTM

app/upgrades/v0.0.12.go
```diff
@@ -0,0 +1,130 @@
...
+			err = keepers.SupplierKeeper.SetParams(ctx, supplierParams)
+			if err != nil {
+				logger.Error("Failed to set supplier params", "error", err)
+				return err
+			}
+			logger.Info("Successfully updated supplier params", "new_params", supplierParams)
+
+			// Add service module `target_num_relays` parameter per `config.yml`.
+			// We don't use `GetParams()` to avoid potential protobuf issues and all networks have the same value (no need to read).
```
Olshansk, line 59: See my edit above about adding details in "all networks".

1. When I read it, I didn't know i you meant all blockchain networks, all cosmos networks, all versions of shannon (alpha, beta, etc...).

2. Note that this is the case today but it's VERY likely to change soon. Given that we're not in a rush (and nothing is breaking / blocking), now is the time to do it right.

3. These examples will act as copy-pasta references for other people. I'd like to either:
	- Use GetParams if possible
	- Show copy-pasta instructions (docs, comments, etc) on how we ensure that all vars are accounted for.

app/upgrades/v0.0.12.go
```diff
@@ -0,0 +1,130 @@
...
+			serviceParams := servicetypes.Params{
```
Olshansk, line 60: NIT: What are your thoughts on small helpers functions (in this file) for each set of params for readability?

Might be helpful as these migrations become more complex.

app/upgrades/v0.0.12.go
```diff
@@ -0,0 +1,130 @@
...
+				TargetNumRelays: 100,
+			}
+			err = keepers.ServiceKeeper.SetParams(ctx, serviceParams)
+			if err != nil {
+				logger.Error("Failed to set service params", "error", err)
+				return err
+			}
+			logger.Info("Successfully updated service params", "new_params", serviceParams)
+
+			// Add tokenomics module `global_inflation_per_claim` parameter per `config.yml`.
```
Olshansk, line 74: When we say `per `config.yml`.`, it confuses me a bit.

To we mean "in line with the hard-coded config.yml in the repo"?

If so, let's update the comments to avoid confusion.

app/upgrades/v0.0.12.go
```diff
@@ -0,0 +1,130 @@
...
+			// We use GetParams() as `DaoRewardAddress` is different between networks and we don't want to hardcode it.
```
Olshansk, line 75: Ditto my other comments about using GetParams everywhere.

app/upgrades/v0.0.12.go
```diff
@@ -0,0 +1,130 @@
...
+			tokenomicsParams := keepers.TokenomicsKeeper.GetParams(ctx)
+			tokenomicsParams.GlobalInflationPerClaim = 0.1
```
Olshansk, line 77: Wdyt of of making local constants for all of the magic numbers in this file?

That way someone can look at the top of the file and quickly see the new values we're changing/setting, and all the "ugly business logic" is hidden in the implementation.

app/upgrades/v0.0.12.go
```diff
@@ -0,0 +1,130 @@
...
+		}
+
+		// Returns the upgrade handler for v0.0.12
+		return func(ctx context.Context, plan upgradetypes.Plan, vm module.VersionMap) (module.VersionMap, error) {
+			logger := cosmosTypes.UnwrapSDKContext(ctx).Logger()
+			logger.Info("Starting v0.0.12 upgrade handler")
+
+			err := applyNewParameters(ctx)
+			if err != nil {
+				logger.Error("Failed to apply new parameters", "error", err)
```
Olshansk, line 93: Thoughts on (here and elsewhere) adding the version upgrade a a (key,val) as well?

You can either have a local `v0.0.12` const, or update the logger context, or both!

app/upgrades/v0.0.12.go
```diff
@@ -0,0 +1,130 @@
...
+				return vm, err
+			}
+
+			// Since we changed the type of RevSharePercent from float32 to uint64, we need to update all on-chain data.
```
Olshansk, line 97: Can we moved this into a local helper to scope / separate the changes from the "boilerplate"?

app/upgrades/v0.0.12.go
```diff
@@ -0,0 +1,130 @@
...
+			}
+
+			logger.Info("Running module migrations")
+			vm, err = mm.RunMigrations(ctx, configurator, vm)
+			if err != nil {
+				logger.Error("Failed to run migrations", "error", err)
+				return vm, err
+			}
+
+			logger.Info("Successfully completed v0.0.12 upgrade handler")
```
Olshansk, line 124: Let's make `v0.0.12` a local constant and do some string formatting here.

It'll avoid copy-pasta errors in the future.

app/upgrades/v0.0.12.go
```diff
@@ -0,0 +1,130 @@
...
+				return vm, err
+			}
+
+			// Since we changed the type of RevSharePercent from float32 to uint64, we need to update all on-chain data.
+			// The easiest way to do this is to iterate over all suppliers and services and set the revshare to 100 by default.
+			suppliers := keepers.SupplierKeeper.GetAllSuppliers(ctx)
+			logger.Info("Updating all suppliers to have a 100% revshare to the supplier", "num_suppliers", len(suppliers))
+			for _, supplier := range suppliers {
+				for _, service := range supplier.Services {
+					// Force all services to have a 100% revshare to the supplier.
```
Olshansk, line 103: Can you change this to read the prior state and use it for the new state/

I realize it's not a big deal here, but I want us to take this as an opportunity to do it right, and set a right example/reference in the future.

Olshansk (changes requested):
@okdas Looks good!

Overall, I do want us to take this as an opportunity to create a good reference + example.

- keep it clean
- do things right
- iterate on process / template

Things aren't "on fire" and it's a "simple migration", so now is the time!

app/upgrades.go
```diff
@@ -11,9 +11,7 @@ import (
 // The chain upgrade can be scheduled AFTER the new version (with upgrade strategy implemented) is released,
```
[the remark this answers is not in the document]
okdas: At the moment of release, we don't know when the upgrade will happen. My thinking, this should be a separate PR. But! We can prepare some changes in advance - that's a good idea.

app/upgrades/types.go
```diff
@@ -16,8 +16,14 @@ import (
 const (
 	// The default PNF/DAO address in the genesis file for Alpha TestNet. Used to create new authz authorizations.
 	AlphaTestNetPnfAddress = "pokt1r6ja6rz6rpae58njfrsgs5n5sp3r36r2q9j04h"
+
 	// Authority address. Defaults to gov module address. Used to create new authz authorizations.
+	// DEV_NOTE: Use `keepers.UpgradeKeeper.Authority(ctx, &upgradetypes.QueryAuthorityRequest{})` to query the authority
+	// address for the current network. Keeping this variable for historical upgrades.
 	AlphaTestNetAuthorityAddress = "pokt10d07y265gmmuvt4z0w9aw880jnsr700j8yv32t"
+
+	// The default PNF/DAO address in the genesis file for Beta TestNet. Used to create new authz authorizations.
```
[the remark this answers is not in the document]
okdas: Correct. These addresses were generated when we created a genesis. We don't have to have them hardcoded, and we can get them from the on-chain data instead.

I thought it's a good idea to have them referenced in code as well - just in case.

app/upgrades/v0.0.12.go
okdas: @Olshansk, we implement `GetParams()` in this upgrade for another module (few lines below).

I skipped `GetParams()` in this place because it's just more room for error. We ran into a protobuf issue during v0.0.10 upgrade. Although I don’t anticipate any problems with this upgrade, I decided to hardcode all parameters for now since it didn’t cause any harm.

I don't mind switching to `GetParams()` - let me know.

app/upgrades/v0.0.12.go
okdas: Noted. I'll switch to `GetParams()` everywhere.

app/upgrades/v0.0.12.go
okdas: Looking at comments below - switching to `GetParams()` everywhere.

app/upgrades/v0.0.12.go
okdas: Can you please elaborate? `GetAllSuppliers` gives us the current state from what I understand.

I do want to add a warning if there's a supplier with more than 1 revshare, which would signal us if we broke somebody's setup.

localnet/poktrolld/config/relayminer_config_localnet_vscode.yaml
```diff
@@ -23,7 +23,7 @@ metrics:
   addr: :9070
 pocket_node:
   query_node_rpc_url: tcp://localhost:26657
-  query_node_grpc_url: tcp://localhost:36658
```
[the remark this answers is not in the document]
okdas: Oh - I wanted to test if relayminer works when connected to the node running on localhost.

app/upgrades/v0.0.12.go
```diff
@@ -0,0 +1,154 @@
...
+	upgradetypes "cosmossdk.io/x/upgrade/types"
+	cosmosTypes "github.com/cosmos/cosmos-sdk/types"
+	"github.com/cosmos/cosmos-sdk/types/module"
+	"github.com/pokt-network/poktroll/app/keepers"
+	sharedtypes "github.com/pokt-network/poktroll/x/shared/types"
+)
+
+const Upgrade_0_0_12_PlanName = "v0.0.12"
+
+// Upgrade_0_0_12 handles the upgrade to release `v0.0.12`. To be issued on both Alpha and Beta TestNets.
```
Olshansk, line 17:
```suggestion
// Upgrade_0_0_12 handles the upgrade to release `v0.0.12`.
// This is planned to be issued on both Pocket Network's Shannon Alpha & Beta TestNets.
```

app/upgrades/v0.0.12.go
```diff
@@ -0,0 +1,154 @@
...
+var Upgrade_0_0_12 = Upgrade{
+	PlanName: Upgrade_0_0_12_PlanName,
+	CreateUpgradeHandler: func(mm *module.Manager,
+		keepers *keepers.Keepers,
+		configurator module.Configurator,
+	) upgradetypes.UpgradeHandler {
+		// Parameter configurations aligned with repository config.yml specifications
```
Olshansk, line 24:
```suggestion
		// Parameter configurations aligned with repository config.yml specifications.
```

app/upgrades/v0.0.12.go
```diff
@@ -0,0 +1,154 @@
...
+		// These values reflect the delta between v0.0.11 and current main branch
```
Olshansk, line 25:
```suggestion
		// These values reflect the delta between v0.0.11 and the main branch as of #1043.
```

app/upgrades/v0.0.12.go
Olshansk: @okdas Seems like you missed/skipped this one.

Going to request changes to make sure we get your perspective here.

app/upgrades/v0.0.12.go
```diff
@@ -0,0 +1,154 @@
...
+		// Reference:
+		// - Comparison: https://github.com/pokt-network/poktroll/compare/v0.0.11..main
```
Olshansk, line 27: Should we be using `main` or a specific sha (that equals to main) when you're preparing this PR?

That way this will be "evergreen"

app/upgrades/v0.0.12.go
```diff
@@ -0,0 +1,154 @@
...
+		// - Direct diff: `git diff v0.0.11..main -- config.yml`
+		//
+		// Note: Parameter updates are derived from config.yml modifications, which serves
+		// as the source of truth for all parameter changes.
```
Olshansk, line 31:
```suggestion
		// DEV_NOTE: These parameter updates are derived from config.yml in the root directory 
		// of this repository, which serves as the source of truth for all parameter changes.
```

app/upgrades/v0.0.12.go
```diff
@@ -0,0 +1,154 @@
...
+		const (
+			supplierStakingFee                = 1000000
+			serviceTargetNumRelays            = 100
+			tokenomicsGlobalInflationPerClaim = 0.1
```
Olshansk, line 35:
```suggestion
			supplierStakingFee                = 1000000 # uPOKT
			serviceTargetNumRelays            = 100 # num relays
			tokenomicsGlobalInflationPerClaim = 0.1 # % of the claim amount
```

app/upgrades/v0.0.12.go
```diff
@@ -0,0 +1,154 @@
...
+			if err != nil {
+				logger.Error("Failed to set tokenomics params", "error", err)
+				return err
+			}
+			logger.Info("Successfully updated tokenomics params", "new_params", tokenomicsParams)
+			return nil
+		}
+
+		// Helper function to update all suppliers' RevShare to 100%. This is necessary to ensure that
+		// we have that value populated before suppliers are connected.
```
Olshansk, line 84:
```suggestion
		// Helper function to update all suppliers' RevShare to 100%.
		// This is necessary to ensure that we have that value populated before suppliers are connected.
```

app/upgrades/v0.0.12.go
Olshansk: @red-0ne Really appreciate this feedback & input.

@okdas Can you #PUC what the "best" approach would be an "why" - essentially paragraphsing @red-0ne's comments. 

In particular, explaining why we're not doing it would help. I'm thinking:
1. Requires multiple copies of the protobuf
2. Requiers multiple steps in the migration (overkill given that we're not in prod)
3. Etc...

app/upgrades/v0.0.12.go
```diff
@@ -0,0 +1,154 @@
...
+		updateSuppliersRevShare := func(ctx context.Context) error {
+			logger := cosmosTypes.UnwrapSDKContext(ctx).Logger()
+			suppliers := keepers.SupplierKeeper.GetAllSuppliers(ctx)
+			logger.Info("Updating all suppliers to have a 100% revshare to the supplier",
```
Olshansk, line 88:
```suggestion
			logger.Info("Updating (overriding) all suppliers to delegate 100% revenue share to the supplier's operator address",
```

app/upgrades/v0.0.12.go
```diff
@@ -0,0 +1,154 @@
...
+				"num_suppliers", len(suppliers))
+
+			for _, supplier := range suppliers {
+				for _, service := range supplier.Services {
+					// Add warning log if we're overwriting existing revshare settings. We can't get the previous
+					// revshare percentage due to a protobuf change, but we can at least log the number of previous revshares.
```
Olshansk, line 94:
```suggestion
					// Log a warning log if we're overwriting existing revshare settings.
					// We chose not to retrieve previous revshare settings due to the complexities of a 
					// managing protobuf changes, but we can at least log potential surprises.
```

app/upgrades/v0.0.12.go
```diff
@@ -0,0 +1,154 @@
...
+					if len(service.RevShare) > 1 {
+						logger.Warn(
+							"Overwriting existing revenue share configuration",
+							"supplier", supplier.OperatorAddress,
```
Olshansk, line 98:
```suggestion
							"supplier_operator", supplier.OperatorAddress,
```

app/upgrades/v0.0.12.go
```diff
@@ -0,0 +1,154 @@
...
+			logger.Info("Updating all suppliers to have a 100% revshare to the supplier",
+				"num_suppliers", len(suppliers))
+
+			for _, supplier := range suppliers {
+				for _, service := range supplier.Services {
+					// Add warning log if we're overwriting existing revshare settings. We can't get the previous
+					// revshare percentage due to a protobuf change, but we can at least log the number of previous revshares.
+					if len(service.RevShare) > 1 {
+						logger.Warn(
+							"Overwriting existing revenue share configuration",
```
Olshansk, line 97: Can you also log the supplier owner?

app/upgrades/v0.0.12.go
```diff
@@ -0,0 +1,154 @@
...
+			logger := cosmosTypes.UnwrapSDKContext(ctx).Logger()
+			suppliers := keepers.SupplierKeeper.GetAllSuppliers(ctx)
+			logger.Info("Updating all suppliers to have a 100% revshare to the supplier",
+				"num_suppliers", len(suppliers))
+
+			for _, supplier := range suppliers {
+				for _, service := range supplier.Services {
+					// Add warning log if we're overwriting existing revshare settings. We can't get the previous
+					// revshare percentage due to a protobuf change, but we can at least log the number of previous revshares.
+					if len(service.RevShare) > 1 {
```
Olshansk, line 95: You also need to check if the address is different

app/upgrades/v0.0.12.go
```diff
@@ -0,0 +1,154 @@
...
+						logger.Warn(
+							"Overwriting existing revenue share configuration",
+							"supplier", supplier.OperatorAddress,
+							"service", service.ServiceId,
+							"previous_revshare_count", len(service.RevShare),
```
Olshansk, line 100: Can we logs the entire previous rev share, not just the count?

app/upgrades/v0.0.12.go
```diff
@@ -0,0 +1,154 @@
...
+					service.RevShare = []*sharedtypes.ServiceRevenueShare{
+						{
+							Address:            supplier.OperatorAddress,
+							RevSharePercentage: uint64(100),
+						},
+					}
+				}
+				keepers.SupplierKeeper.SetSupplier(ctx, supplier)
+				logger.Info("Updated supplier",
+					"supplier", supplier.OperatorAddress)
```
Olshansk, line 112: Owner & operator please!

app/upgrades/v0.0.12.go
```diff
@@ -0,0 +1,154 @@
...
+			}
+			return nil
+		}
+
+		// Returns the upgrade handler for v0.0.12
+		return func(ctx context.Context, plan upgradetypes.Plan, vm module.VersionMap) (module.VersionMap, error) {
+			logger := cosmosTypes.UnwrapSDKContext(ctx).Logger()
+			logger.Info("Starting upgrade handler", "upgrade_plan_name", Upgrade_0_0_12_PlanName)
+
+			err := applyNewParameters(ctx)
```
Olshansk, line 122:
```suggestion
			// Update all governance parameter changes.
			// This includes adding params, removing params and changing values of existing params.
			err := applyNewParameters(ctx)
```

app/upgrades/v0.0.12.go
```diff
@@ -0,0 +1,154 @@
...
+			if err != nil {
+				logger.Error("Failed to apply new parameters",
+					"upgrade_plan_name", Upgrade_0_0_12_PlanName,
+					"error", err)
+				return vm, err
+			}
+
+			// Update all suppliers' RevShare
```
Olshansk, line 130:
```suggestion
			// Override all suppliers' RevShare to be 100% delegate to the supplier's operator address
```

app/upgrades/v0.0.12.go
```diff
@@ -0,0 +1,154 @@
...
+			err = updateSuppliersRevShare(ctx)
+			if err != nil {
+				logger.Error("Failed to update suppliers RevShare",
+					"upgrade_plan_name", Upgrade_0_0_12_PlanName,
+					"error", err)
+				return vm, err
+			}
+
+			logger.Info("Running module migrations", "upgrade_plan_name", Upgrade_0_0_12_PlanName)
```
Olshansk, line 139: Can we add a log like this before every section / update?

Olshansk (changes requested):
Looking great @okdas!

Just a handful of more comments & NITs but we should be g2g after the next set of changes.

app/upgrades/v0.0.12.go
okdas: Added a list of addresses.

app/upgrades/v0.0.12.go
okdas: Rephrased the whole block.

app/upgrades/v0.0.12.go
okdas: Changed the logic a bit.

If more than 1 supplier: log and overwrite,
if 1 supplier: just update the data keeping existing revshare address (assuming 100% goes to that one address)

localnet/poktrolld/config/relayminer_config_localnet_vscode.yaml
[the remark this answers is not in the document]
okdas: `9090` is the default gRPC port in `cosmos-sdk`. If you start the node without LocalNet, that'll be the port the gRPC endpoint is available on. This particular file is only used to debug in VSCode, probably just by me. I haven't seen anybody else using this setup.

app/upgrades/historical.go
```diff
@@ -40,6 +40,9 @@ func defaultUpgradeHandler(
 // For example, even if `ConsensusVersion` is not modified for any modules, it still might be beneficial to create
 // an upgrade so node runners are signaled to start utilizing `Cosmovisor` for new binaries.
 var UpgradeExample = Upgrade{
+	// PlanName can be any string.
+	// This code is executed when the upgrade with this plan name is submitted to the network.
+	// This does not necessarly need to be a version, but it's usually the case with consensus-breaking changes.
```
Olshansk, line 45:
```suggestion
	// This does not necessarily need to be a version, but it's usually the case with consensus-breaking changes.
```

app/upgrades/v0.0.12.go
Olshansk: Moving conversation to discord to get this over the finish line.

https://discord.com/channels/824324475256438814/1332065342746525778/1336434350660784239

app/upgrades/v0.0.12.go
```diff
@@ -80,36 +81,76 @@ var Upgrade_0_0_12 = Upgrade{
...
-			logger.Info("Updating all suppliers to have a 100% revshare to the supplier",
+			logger.Info("Updating (overriding) all suppliers to delegate 100% revenue share to the supplier's operator address",
 				"num_suppliers", len(suppliers))
 
 			for _, supplier := range suppliers {
 				for _, service := range supplier.Services {
-					// Add warning log if we're overwriting existing revshare settings. We can't get the previous
-					// revshare percentage due to a protobuf change, but we can at least log the number of previous revshares.
 					if len(service.RevShare) > 1 {
+						// Warning: Overwriting existing revshare settings without preserving history.
```
Olshansk, line 96:
```suggestion
						// WARNING: Overwriting existing revshare settings without preserving history.
```

app/upgrades/v0.0.12.go
```diff
@@ -80,36 +81,76 @@ var Upgrade_0_0_12 = Upgrade{
...
+						// Note: While the canonical approach would be using Module Upgrade (docs.cosmos.network/v0.46/building-modules/upgrade)
```
Olshansk, line 97:
```suggestion
						// NOTE: While the canonical approach would be using Module Upgrade (docs.cosmos.network/v0.46/building-modules/upgrade)
```

app/upgrades/v0.0.12.go
```diff
@@ -80,36 +81,76 @@ var Upgrade_0_0_12 = Upgrade{
...
+
+						// Create a string representation of just the revenue share addresses
+						revShareAddresses := "["
+						for i, rs := range service.RevShare {
+							if i > 0 {
+								revShareAddresses += ","
+							}
+							revShareAddresses += rs.Address
+						}
+						revShareAddresses += "]"
```
Olshansk, line 111: Thought on this?

```suggestion
addresses := make([]string, len(service.RevShare))
for i, rs := range service.RevShare {
    addresses[i] = rs.Address
}

revShareAddressesStr := "[" + strings.Join(addresses, ",") + "]"
```

Olshansk (approved):
A few NITs.

Preliminary approval so we can start working on the upgrade while tending to the discussion / comments in parallel.

Feel free to merge at your pace.

okdas:
Had a successful upgrade locally (using [this guide](https://dev.poktroll.com/protocol/upgrades/upgrade_procedure)).
Will create a release and plan an upgrade on Alpha.

PR #1050, opened 2025-01-31
[RelayMiner] Implement relayminer query caching

red-0ne:
## Summary
Implements caching layer for query clients to reduce network calls and improve performance

### Primary Changes:
- Added generic `KeyValueCache` and `ParamsCache` interfaces with thread-safe implementations
- Integrated caching across all query clients (`Account`, `Application`, `Bank`, `Service`, etc.)
- Added cache clearing on new blocks via `WithNewBlockCacheClearing` option

### Secondary Changes:
- Replaced manual `sync.Mutex` implementation in `accQuerier` with new cache interface
- Added cache configuration to integration tests
- Updated relayer dependencies to include cache initialization

## Issue

The RelayMiner RPC queries are not cached, which puts excessive load on the configured full node, degrading the performance of both off-chain and on-chain components.

## Type of change

- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Sanity Checklist

- [x] I added TODOs where applicable


pkg/client/query/interface.go
```diff
@@ -0,0 +1,16 @@
+package query
+
+// ParamsCache is an interface for a simple in-memory cache implementation for query parameters.
+type ParamsCache[T any] interface {
```
Olshansk, line 4: A `Param` can be represented as a `KeyValue`.

Any reason we can't consolidate the two?

pkg/relayer/session/session_test.go
```diff
@@ -73,9 +73,10 @@ func requireProofCountEqualsExpectedValueFromProofParams(t *testing.T, proofPara
 	}
```
Olshansk, line 73: I feel like we're very light with the tests.

Did you test this in any way?
- Unit?
- Integration?
- Manual?

pkg/relayer/session/session_test.go
```diff
@@ -73,9 +73,10 @@ func requireProofCountEqualsExpectedValueFromProofParams(t *testing.T, proofPara
...
 	// Set the supplier operator balance to be able to submit the expected number of proofs.
-	claimAndFeeGasCost := session.ClamAndProofGasCost.Amount.Int64()
-	numExpectedProofs := int64(2)
-	supplierOperatorBalance := claimAndFeeGasCost * numExpectedProofs
+	feePerProof := prooftypes.DefaultParams().ProofSubmissionFee.Amount.Int64()
+	gasCost := session.ClamAndProofGasCost.Amount.Int64()
+	proofCost := feePerProof + gasCost
+	supplierOperatorBalance := proofCost
 	supplierClientMap := testsupplier.NewClaimProofSupplierClientMap(ctx, t, supplierOperatorAddress, proofCount)
 	blockPublishCh, minedRelaysPublishCh := setupDependencies(t, ctx, supplierClientMap, emptyBlockHash, proofParams, supplierOperatorBalance)
```
Olshansk, line 81: Can you do a comparison of the performance before and update adding a cache? Similar to what you did with the SMT.

E.g. I don't know how to (easily) to a load test on DevNet so anything like:
- Makefile
- E2E test
- Docs / instructions 

Would be really nice!

pkg/relayer/cmd/cmd.go
```diff
@@ -198,7 +205,23 @@ func setupRelayerDependencies(
 		config.NewSupplyQueryClientContextFn(queryNodeGRPCUrl),            // leaf
 		config.NewSupplyTxClientContextFn(queryNodeGRPCUrl, txNodeRPCUrl), // leaf
 		config.NewSupplyDelegationClientFn(),                              // leaf
-		config.NewSupplySharedQueryClientFn(),                             // leaf
+
+		// Setup the params caches and configure them to clear on new blocks.
+		config.NewSupplyParamsCacheFn[sharedtypes.Params](cache.WithNewBlockCacheClearing),
```
Olshansk, line 210: Why aren't we doing this for all params?

Screenshot 2025-02-04 at 3 22 42 PM

pkg/relayer/cmd/cmd.go
```diff
@@ -198,7 +205,23 @@ func setupRelayerDependencies(
...
+		// Setup the key-value caches and configure them to clear on new blocks.
+		config.NewSupplyKeyValueCacheFn[sharedtypes.Service](cache.WithNewBlockCacheClearing),
+		config.NewSupplyKeyValueCacheFn[servicetypes.RelayMiningDifficulty](cache.WithNewBlockCacheClearing),
+		config.NewSupplyKeyValueCacheFn[apptypes.Application](cache.WithNewBlockCacheClearing),
+		config.NewSupplyKeyValueCacheFn[cosmostypes.AccountI](cache.WithNewBlockCacheClearing),
+		config.NewSupplyKeyValueCacheFn[sharedtypes.Supplier](cache.WithNewBlockCacheClearing),
+		config.NewSupplyKeyValueCacheFn[*sessiontypes.Session](cache.WithNewBlockCacheClearing),
+		config.NewSupplyKeyValueCacheFn[*cosmostypes.Coin](cache.WithNewBlockCacheClearing),
+
+		config.NewSupplySharedQueryClientFn(), // leaf
```
Olshansk, line 224: What is the leaf comment?

pkg/relayer/cmd/cmd.go
```diff
@@ -198,7 +205,23 @@ func setupRelayerDependencies(
...
+
+		// Setup the params caches and configure them to clear on new blocks.
+		config.NewSupplyParamsCacheFn[sharedtypes.Params](cache.WithNewBlockCacheClearing),
+		config.NewSupplyParamsCacheFn[apptypes.Params](cache.WithNewBlockCacheClearing),
+		config.NewSupplyParamsCacheFn[sessiontypes.Params](cache.WithNewBlockCacheClearing),
+		config.NewSupplyParamsCacheFn[prooftypes.Params](cache.WithNewBlockCacheClearing),
+
+		// Setup the key-value caches and configure them to clear on new blocks.
+		config.NewSupplyKeyValueCacheFn[sharedtypes.Service](cache.WithNewBlockCacheClearing),
+		config.NewSupplyKeyValueCacheFn[servicetypes.RelayMiningDifficulty](cache.WithNewBlockCacheClearing),
```
Olshansk, line 217: Can you group & separated poktroll types and cosmos types?
Screenshot 2025-02-04 at 3 24 50 PM

pkg/relayer/cmd/cmd.go
```diff
@@ -198,7 +205,23 @@ func setupRelayerDependencies(
...
+		config.NewSupplyKeyValueCacheFn[apptypes.Application](cache.WithNewBlockCacheClearing),
+		config.NewSupplyKeyValueCacheFn[cosmostypes.AccountI](cache.WithNewBlockCacheClearing),
+		config.NewSupplyKeyValueCacheFn[sharedtypes.Supplier](cache.WithNewBlockCacheClearing),
+		config.NewSupplyKeyValueCacheFn[*sessiontypes.Session](cache.WithNewBlockCacheClearing),
```
Olshansk, line 221: Why is this a pointer but the others are not?

pkg/client/query/interface.go
```diff
@@ -0,0 +1,16 @@
...
+	Get() (T, bool)
+	Set(T)
+	Clear()
+}
+
+// KeyValueCache is an interface for a simple in-memory key-value cache implementation.
+type KeyValueCache[V any] interface {
```
Olshansk, line 11: @bryanchriswhite and I spent a lot of time designing a more general purpose interface in.

I think we should merge that one in and leverage it here.

[1] https://github.com/pokt-network/poktroll/pull/994/files

pkg/client/query/sessionquerier.go
```diff
@@ -16,8 +18,11 @@ var _ client.SessionQueryClient = (*sessionQuerier)(nil)
 // querying of onchain session information through a single exposed method
 // which returns an sessiontypes.Session struct
 type sessionQuerier struct {
-	clientConn     grpc.ClientConn
-	sessionQuerier sessiontypes.QueryClient
+	clientConn        grpc.ClientConn
+	sessionQuerier    sessiontypes.QueryClient
+	sharedQueryClient client.SharedQueryClient
+	sessionsCache     KeyValueCache[*sessiontypes.Session]
```
Olshansk, line 24: Why is this the only pointer?

pkg/client/query/sessionquerier.go
```diff
@@ -48,6 +56,24 @@ func (sessq *sessionQuerier) GetSession(
...
 	blockHeight int64,
 ) (*sessiontypes.Session, error) {
+	// Get the shared parameters to calculate the session start height.
+	// Use the session start height as the canonical height to be used in the cache key.
+	sharedParams, err := sessq.sharedQueryClient.GetParams(ctx)
+	if err != nil {
+		return nil, err
+	}
+	sessionStartHeight := sharedtypes.GetSessionStartHeight(sharedParams, blockHeight)
+	// Construct the cache key using the appAddress, serviceId and session start height.
```
Olshansk, line 66: Move this into a helper that returns `sessionCacheKey`

pkg/client/query/sharedquerier.go
```diff
@@ -49,11 +53,19 @@ func NewSharedQuerier(deps depinject.Config) (client.SharedQueryClient, error) {
 // Once `ModuleParamsClient` is implemented, use its replay observable's `#Last()` method
 // to get the most recently (asynchronously) observed (and cached) value.
 func (sq *sharedQuerier) GetParams(ctx context.Context) (*sharedtypes.Params, error) {
+	// Get the params from the cache if they exist.
+	if params, found := sq.paramsCache.Get(); found {
```
Olshansk, line 57: I'm concerned (and don't fully understand) the lack of a "height" param when retrieving things from the cache.

pkg/client/query/sharedquerier.go
```diff
@@ -127,13 +139,20 @@ func (sq *sharedQuerier) GetEarliestSupplierClaimCommitHeight(ctx context.Contex
...
 	claimWindowOpenHeight := sharedtypes.GetClaimWindowOpenHeight(sharedParams, queryHeight)
-	claimWindowOpenBlock, err := sq.blockQuerier.Block(ctx, &claimWindowOpenHeight)
-	if err != nil {
-		return 0, err
-	}
 
-	// NB: Byte slice representation of block hashes don't need to be normalized.
-	claimWindowOpenBlockHash := claimWindowOpenBlock.BlockID.Hash.Bytes()
+	// Check if the block hash is already in the cache.
+	claimWindowOpenHeightStr := strconv.FormatInt(claimWindowOpenHeight, 10)
```
Olshansk, line 144: This is "implicit magic logic" of how the key is prepared. Move into a helper function.

pkg/client/query/sharedquerier.go
```diff
@@ -127,13 +139,20 @@ func (sq *sharedQuerier) GetEarliestSupplierClaimCommitHeight(ctx context.Contex
...
+	claimWindowOpenBlockHash, found := sq.blockHashCache.Get(claimWindowOpenHeightStr)
+	if !found {
+		claimWindowOpenBlock, err := sq.blockQuerier.Block(ctx, &claimWindowOpenHeight)
+		if err != nil {
+			return 0, err
+		}
+
+		// Cache the block hash for future use.
```
Olshansk, line 152: Where are we setting/updating the cache?

pkg/client/query/sharedquerier.go
```diff
@@ -157,18 +176,25 @@ func (sq *sharedQuerier) GetEarliestSupplierProofCommitHeight(ctx context.Contex
...
+	if !found {
+		// Fetch the block at the proof window open height. Its hash is used as part
+		// of the seed to the pseudo-random number generator.
+		proofWindowOpenHeight := sharedtypes.GetProofWindowOpenHeight(sharedParams, queryHeight)
+		proofWindowOpenBlock, err := sq.blockQuerier.Block(ctx, &proofWindowOpenHeight)
+		if err != nil {
+			return 0, err
+		}
+
+		// Cache the block hash for future use.
```
Olshansk, line 190: Where are we actually setting/updating the cache?

Olshansk (changes requested):
@red-0ne I did a first partial review but have a lot of comments & questions.

Here's a high-level summary but PTAL at the actual comments as well:

1. Need to understand if/how this can build on top of #994 w/ @bryanchriswhite 
2. See a few comments (logs + comments) that need to be addressed in multiple places
3. I’m a bit concerned (and don’t understand) how we’re not using “height” to retrieve things from the cache, especially when values are always changing
4. When does the cache ever get cleared?
5. Light on tests
6. I’d be interested to see numbers of performance improvement

pkg/client/query/accquerier.go
```diff
@@ -21,11 +20,7 @@ var _ client.AccountQueryClient = (*accQuerier)(nil)
 type accQuerier struct {
 	clientConn     grpc.ClientConn
 	accountQuerier accounttypes.QueryClient
-
-	// accountCache is a cache of accounts that have already been queried.
-	// TODO_TECHDEBT: Add a size limit to the cache and consider an LRU cache.
-	accountCache   map[string]types.AccountI
-	accountCacheMu sync.Mutex
```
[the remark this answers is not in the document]
red-0ne: This was a cache internal to the account querier, which is now replaced by the `KeyValueCache` that has its own mutex.

pkg/client/query/accquerier.go
```diff
@@ -21,11 +20,7 @@ var _ client.AccountQueryClient = (*accQuerier)(nil)
 type accQuerier struct {
 	clientConn     grpc.ClientConn
 	accountQuerier accounttypes.QueryClient
-
-	// accountCache is a cache of accounts that have already been queried.
-	// TODO_TECHDEBT: Add a size limit to the cache and consider an LRU cache.
```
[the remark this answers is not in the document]
red-0ne: Size limit would be implemented in the `KeyValueCache` to which the account querier should have no knowledge of.

pkg/client/query/interface.go
red-0ne: If we consolidate the `ParamsCache` and `KeyValueCache`, we'll end up with
```
ParamsCache[P]#Get(key)
ParamsCache[P]#Set(key, value)
```

Where `key` would be a made up key that we force the developer to provide.

This would leave us with the following options:
* Empty string key `paramsCache.Set("", param)`
* Any key `paramsCache.Set("anyKeyWouldWork`, param)`
* An enforced constant `paramsCache.Set(SingleValueConst, param)`

Those options were considered but opted for a keyless params cache for clarity.

pkg/relayer/session/session_test.go
red-0ne: You're right.

I'll add tests.

pkg/relayer/cmd/cmd.go
red-0ne: Some of the params (`tokenomics` and `gateway`) are not used by the `RelayMiner` and don't need to have a cache implemented.

pkg/relayer/cmd/cmd.go
red-0ne: Those have been around for a while now. They mean that whatever is supplied is a `leaf` in the dependency tree (i.e. has no dependency)

pkg/relayer/cmd/cmd.go
red-0ne: This is because `sessionQuerier.GetSession` return a pointer to `Session` while the others do not.

The cache should align with whatever the querier is returning.

pkg/client/query/sessionquerier.go
red-0ne: The reason is that the `sessionQueryClient` returns a `QueryGetSessionResponse` which has a `Session` pointer

```go
type QueryGetSessionResponse struct {
	Session *Session `protobuf:"bytes,1,opt,name=session,proto3" json:"session,omitempty"`
}
```

I tried to align the cache type with whatever the querier returns.

pkg/client/query/sharedquerier.go
red-0ne: This cache implementation does not add any new functionality besides caching whatever has been queried.

It does not alter the `RelayMiner`s current behavior
*  `RelayMiner` cold start
* React to `Param`s change

For those reasons, it does not leverage historical data that justifies the usage of `height` for cache querying.

pkg/client/query/interface.go
bryanchriswhite: This interface is consistent with what I was calling `QueryCache[T any]` in #994. It should be quite straightforward to refactor #994 to use this instead. I see `KeyValueCache[V any]` as a subsequent iteration of `QueryCache[T any]` which includes generalizing the name.

pkg/client/query/interface.go
bryanchriswhite: #PUC with a summary of this distinction. I also was initially questioning the need for distinct interfaces, and like your solution. :+1:

pkg/client/query/cache/paramscache.go
```diff
@@ -0,0 +1,54 @@
...
+	c.value = value
+}
+
+// Clear empties the cache.
+func (c *paramsCache[T]) Clear() {
+	c.cacheMu.Lock()
+	defer c.cacheMu.Unlock()
+
+	c.found = false
+	c.value = zeroValue[T]()
```
bryanchriswhite, line 48: Any reason to prefer this static function over either?:
```
var zero T
c.value = zero
```
Or
```
c.value = *new(T)
```

pkg/client/query/cache/paramscache.go
```diff
@@ -0,0 +1,54 @@
...
+import (
+	"sync"
+
+	"github.com/pokt-network/poktroll/pkg/client/query"
+)
+
+var _ query.ParamsCache[any] = (*paramsCache[any])(nil)
+
+// paramsCache is a simple in-memory cache implementation for query parameters.
+// It does not involve key-value pairs, but only stores a single value.
```
bryanchriswhite, line 12: Clever to avoid the temptation to compose `KeyValueCache` here! :sunglasses:

pkg/client/query/sessionquerier.go
bryanchriswhite: :+1:

pkg/client/query/sharedquerier.go
bryanchriswhite: My understanding is that the cache implementations here are NOT historical; i.e. ONLY the most recently observed value is cached for each `ParamsCache` instance (or key, in the case of `KeyValueCache`).

While #994 does include historical caching as well (via the `HistoricalQueryCache` interface, that's an additional and distinct feature.

This shouldn't be necessary here because we're clearing the cache on every new block. The end result being, somewhat sub-optimal, but significant caching. This reformulates the number off-chain queries from being a function of API usage, to no more than one per block, per cache.

pkg/deps/config/suppliers.go
```diff
@@ -507,3 +508,45 @@ func newSupplyTxClientsFn(
 
 	return depinject.Configs(deps, depinject.Supply(txClient)), nil
 }
+
+// NewSupplyTxClientFn returns a function which constructs a KeyValueCache of type T.
+// It take a list of cache options that can be used to configure the cache.
+func NewSupplyKeyValueCacheFn[T any](opts ...cache.CacheOption[query.KeyValueCache[T]]) SupplierFn {
```
bryanchriswhite, line 514: Comment doesn't match function name.

bryanchriswhite (commented):

Thanks for doing this! :heart:

pkg/client/query/interface.go
red-0ne: This PR and PR #994 have very similar interfaces and will ultimately be merged.

I considered building on top of #994 but decided not to, as this PR does not aim to provide:

* Granular (per-key) cache invalidation
* Time-based cache invalidation (it only clears the whole cache when a new block is observed)
* Access to historical values, which is out of scope and would imply adding new functionality (or not using the implementation)

I can put in the effort to rebase it on top of #994 if you and @bryanchriswhite think it would be worthwhile.

pkg/relayer/session/session_test.go
red-0ne:
```
// TODO_TECHDEBT: Consider a flag to change client queriers caching behavior.
// This would allow to easily switch between caching and non-caching queriers
// for benchmarking purposes.
```

testutil/testclient/testqueryclients/grpcserver.go
```diff
@@ -0,0 +1,168 @@
...
+
+func (c *callCounter) CallCount() int {
+	return c.callCount
+}
+
+func (c *callCounter) Increment() {
+	c.callCount++
+}
+
+// MockServiceQueryServer is a mock implementation of the servicetypes.QueryServer interface
```
Olshansk, line 33: Up until now we've always been using `gomock` to generate this sort of thing, which has support for call counters. 

I feel strongly that we should not be changing patterns now.

pkg/relayer/cmd/cmd.go
Olshansk: We have some `//leaf` comments before the new code and some (this one) after the new code.

What's the idea behind this code organization?

pkg/relayer/cmd/cmd.go
Olshansk: #PUC

pkg/relayer/cmd/cmd.go
Olshansk: #PUC

pkg/relayer/cmd/cmd.go
```diff
@@ -198,7 +206,29 @@ func setupRelayerDependencies(
 		config.NewSupplyQueryClientContextFn(queryNodeGRPCUrl),            // leaf
 		config.NewSupplyTxClientContextFn(queryNodeGRPCUrl, txNodeRPCUrl), // leaf
 		config.NewSupplyDelegationClientFn(),                              // leaf
-		config.NewSupplySharedQueryClientFn(),                             // leaf
+
+		// Setup the params caches and configure them to clear on new blocks.
+		// TODO_TECHDEBT: Consider a flag to change client queriers caching behavior.
```
Olshansk, line 211: I'm going to push on the fact that this is a TODO_IN_THIS_PR.

It's not that hard and I want to understand the benefit (if any) of this cache.

It feels like we've built a car but not actually checking if it works.

pkg/client/query/interface.go
Olshansk: I left a comment on discord, but I feel strongly that we should finish #994 and rebase on top of it.

Now is the time to do this right.

Screenshot 2025-02-11 at 4 31 05 PM

pkg/client/query/types/balance.go
```diff
@@ -0,0 +1,12 @@
...
+
+import (
+	cosmostypes "github.com/cosmos/cosmos-sdk/types"
+)
+
+// Balance represents a pointer to a Cosmos SDK Coin, specifically used for bank balance queries.
+// It is deliberately defined as a distinct type (not a type alias) to ensure clear dependency
+// injection and to differentiate it from other coin caches in the system. This type helps
+// maintain separation of concerns between different types of coin-related data in the caching
+// layer.
```
Olshansk, line 11:
```suggestion
// Balance represents a pointer to a Cosmos SDK Coin used for bank balance queries.
// It is defined as a distinct type (not an alias) to:
// - Ensure clear dependency injection
// - Differentiate from other coin caches in the system
// - Maintain separation of concerns between coin-related data in the caching layer
type Balance *sdk.Coin
```

@red-0ne Have you used the code-cleaner Claude project yet?

pkg/client/query/types/blockhash.go
```diff
@@ -0,0 +1,8 @@
+package types
+
+// BlockHash represents a byte slice, specifically used for bank balance query caches.
+// It is deliberately defined as a distinct type (not a type alias) to ensure clear
+// dependency injection and to differentiate it from other byte slice caches in the system.
+// This type helps maintain separation of concerns between different types of
+// byte slice data in the caching layer.
```
Olshansk, line 7:
```suggestion
// BlockHash represents a byte slice used for bank balance query caches.  
// It is defined as a distinct type (not an alias) to:
// - Ensure clear dependency injection 
// - Differentiate from other byte slice caches
// - Maintain separation of concerns between byte slice data in caching layer
```

pkg/client/query/types/blockhash.go
```diff
@@ -0,0 +1,8 @@
+package types
```
Olshansk, line 1: Having separate files for this feels like overkill. 

Can we just have a `types.go`?

pkg/client/query/supplierquerier.go
```diff
@@ -44,6 +51,16 @@ func (supq *supplierQuerier) GetSupplier(
...
 ) (sharedtypes.Supplier, error) {
+	logger := supq.logger.With("query_client", "supplier", "method", "GetSupplier")
+
+	// Check if the supplier is present in the cache.
+	if supplier, found := supq.suppliersCache.Get(operatorAddress); found {
+		logger.Debug().Msgf("cache hit for key: %s", operatorAddress)
+		return supplier, nil
+	}
+
+	logger.Debug().Msgf("cache miss for key: %s", operatorAddress)
```
Olshansk, line 62: Leaving one comment but please update everywhere. If these logs ever becomes the source for debugging, you want it to be ULTRA obvious.

```suggestion
	logger.Debug().Msgf("cache miss for operator address key: %s", operatorAddress)
```

Olshansk (changes requested):
Leaving a partial review.

1. @bryanchriswhite Can you please prioritize getting #994 in? It's the most mature/versatile cache, and I'd like us to just build on top of 1 thing.

2. @red-0ne See some of my nits/edits, but in particular around using gomock for proper mocks.

Will do a full review after (1) & (2) are done. 

Few notes:
1. If you think we should take a different direction, let's jump on a call.
2. We have other (large) parallel efforts going on, so there shouldn't be any blockers
3. I strongly believe we should benchmark in this PR. Seems like something an LLM can help get done in a couple of hours.

pkg/deps/config/suppliers.go
```diff
@@ -507,3 +513,75 @@ func newSupplyTxClientsFn(
...
+		deps depinject.Config,
+		cmd *cobra.Command,
+	) (depinject.Config, error) {
+		// Check if query caching is enabled
+		queryCachingEnabled, err := cmd.Flags().GetBool(FlagQueryCaching)
+		if err != nil {
+			return nil, err
+		}
+
+		// Use a NoOpKeyValueCache if query caching is disabled
```
Olshansk, line 531: NIT: Do you mind adding a log before & after this if structure just stating "Query Caching is enabled/disabled".

If it ever becomes an issue in debugging, we'll have this data point right away.

pkg/deps/config/suppliers.go
```diff
@@ -507,3 +513,75 @@ func newSupplyTxClientsFn(
...
+		cmd *cobra.Command,
+	) (depinject.Config, error) {
+		// Check if params caching is enabled
+		queryCachingEnabled, err := cmd.Flags().GetBool(FlagQueryCaching)
+		if err != nil {
+			return nil, err
+		}
+
+		// Use a NoOpParamsCache if query caching is disabled
+		if !queryCachingEnabled {
```
Olshansk, line 568: Ditto

pkg/relayer/cmd/cmd.go
```diff
@@ -78,6 +86,7 @@ for such operations.`,
 	cmd.Flags().StringVar(&flagLogLevel, cosmosflags.FlagLogLevel, "debug", "The logging level (debug|info|warn|error)")
 	cmd.Flags().Float64(cosmosflags.FlagGasAdjustment, 1.5, "The adjustment factor to be multiplied by the gas estimate returned by the tx simulation")
 	cmd.Flags().String(cosmosflags.FlagGasPrices, "1upokt", "Set the gas unit price in upokt")
+	cmd.Flags().Bool(config.FlagQueryCaching, true, "Enable or disable query caching")
```
Olshansk, line 89:
```suggestion
	cmd.Flags().Bool(config.FlagQueryCaching, true, "Enable or disable onchain query caching")
```

pkg/client/interface.go
```diff
@@ -370,3 +371,11 @@ type BankQueryClient interface {
 	// GetBalance queries the chain for the uPOKT balance of the account provided
 	GetBalance(ctx context.Context, address string) (*cosmostypes.Coin, error)
 }
+
+// ParamsCache is an interface for a simple in-memory cache implementation for query parameters.
```
Olshansk, line 375:
```suggestion
// ParamsCache is an interface for a simple in-memory cache implementation for onchain module parameter queries.
```

pkg/client/query/cache/noop.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/pokt-network/poktroll/pkg/client"
+)
+
+var _ client.ParamsCache[any] = (*noOpParamsCache[any])(nil)
+var _ cache.KeyValueCache[any] = (*noOpKeyValueCache[any])(nil)
+
+// noOpParamsCache is a no-op implementation of a ParamsCache.
+// It does not store any values.
+type noOpParamsCache[T any] struct {
+}
```
Olshansk, line 14:
```suggestion
type noOpParamsCache[T any] struct {}
```

Olshansk: Ditto below

pkg/client/query/servicequerier.go
```diff
@@ -67,6 +91,16 @@ func (servq *serviceQuerier) GetServiceRelayDifficulty(
...
 ) (servicetypes.RelayMiningDifficulty, error) {
+	logger := servq.logger.With("query_client", "service", "method", "GetServiceRelayDifficulty")
+
+	// Check if the relay mining difficulty is present in the cache.
+	if relayMiningDifficulty, found := servq.relayMiningDifficultyCache.Get(serviceId); found {
+		logger.Debug().Msgf("cache hit for service id key: %s", serviceId)
+		return relayMiningDifficulty, nil
+	}
+
+	logger.Debug().Msgf("cache miss for service id key: %s", serviceId)
```
Olshansk, line 102:
```suggestion
	logger.Debug().Msgf("relay mining difficulty cache miss for service id key: %s", serviceId)
```

pkg/client/query/servicequerier.go
```diff
@@ -47,6 +58,16 @@ func (servq *serviceQuerier) GetService(
...
 ) (sharedtypes.Service, error) {
+	logger := servq.logger.With("query_client", "service", "method", "GetService")
+
+	// Check if the service is present in the cache.
+	if service, found := servq.servicesCache.Get(serviceId); found {
+		logger.Debug().Msgf("cache hit for service id key: %s", serviceId)
+		return service, nil
+	}
+
+	logger.Debug().Msgf("cache miss for service id key: %s", serviceId)
```
Olshansk, line 69:
```suggestion
	logger.Debug().Msgf("service cache miss for service id key: %s", serviceId)
```

pkg/client/query/servicequerier.go
```diff
@@ -19,6 +21,12 @@ var _ client.ServiceQueryClient = (*serviceQuerier)(nil)
 type serviceQuerier struct {
 	clientConn     grpc.ClientConn
 	serviceQuerier servicetypes.QueryClient
+	logger         polylog.Logger
+
+	// servicesCache caches serviceQueryClient.Service requests
```
Olshansk, line 26:
```suggestion
	// servicesCache caches serviceQueryClient.Service query requests
```

pkg/client/query/servicequerier.go
```diff
@@ -19,6 +21,12 @@ var _ client.ServiceQueryClient = (*serviceQuerier)(nil)
...
+	servicesCache cache.KeyValueCache[sharedtypes.Service]
+	// relayMiningDifficultyCache caches serviceQueryClient.RelayMiningDifficulty requests
```
Olshansk, line 28:
```suggestion
	// relayMiningDifficultyCache caches serviceQueryClient.RelayMiningDifficulty query requests
```

pkg/client/query/sessionquerier.go
```diff
@@ -60,15 +93,45 @@ func (sessq *sessionQuerier) GetSession(
...
 	if err != nil {
 		return nil, ErrQuerySessionParams.Wrapf("[%v]", err)
 	}
+
+	// Cache the params for future queries.
+	sessq.paramsCache.Set(res.Params)
 	return &res.Params, nil
 }
+
+// getSessionCacheKey constructs the cache key for a session.
```
Olshansk, line 125:
```suggestion
// getSessionCacheKey constructs the cache key for a session in the form of: appAddress/serviceId/sessionStartHeight
```

pkg/client/query/sharedquerier.go
```diff
@@ -187,3 +242,8 @@ func (sq *sharedQuerier) GetComputeUnitsToTokensMultiplier(ctx context.Context)
 	}
 	return sharedParams.GetComputeUnitsToTokensMultiplier(), nil
 }
+
+// getBlockHashKacheKey constructs the cache key for a block hash.
```
Olshansk, line 246: #PUC how its constructed

Olshansk (approved):
@red-0ne Preliminary approval with a few NITS throughout the PR AND the comment below.

The ONLY thing I'd really like to see is us adding an E2E test with the cache enabled and the cache disabled.

I think one of the tokenomics ones (proofs, claims, blocks, etc...) would be a great use-case. You'd just need to:
1. Copy-paste an existing E2E test
2. Update the E2E framework to start the relayminer with a cache3. 
3. Profit

I don't think this is super trivial, so my ask is to add a TODO_MAINNET in that part of the code before merging this in and tending to it when you have some spare cycles this/next week.

pkg/deps/config/suppliers.go
red-0ne: Added in https://github.com/pokt-network/poktroll/blob/b1ef0add07118296d3fc84faa6a4043b52ba3bad/pkg/relayer/cmd/cmd.go#L150-L154

PR #1041, opened 2025-01-23
[Workflows] Update GitHub actions from v3 to v4

Olshansk:
## Summary

Upgrade github ations from `v3` to `v4`.

### Origin Document

https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/
Screenshot 2025-01-23 at 2 32 22 PM

### Example in action

Screenshot 2025-01-23 at 2 32 40 PM

https://github.com/buildwithgrove/path/actions/runs/12935368023/job/36078630477

PR #1056, opened 2025-02-04
[CI][Quick change] downgrade docker/login-action version

okdas:
We accidentally bumped the action version in #1041. Need to reverse to build artifacts for the release.


okdas:
Tested this using `act` - to the point of pushing the image, which I didn't populate credentials for. This should be good.

Issue #1107, opened 2025-03-05
[RelayMiner] Handle restarts gracefully


okdas:
## Objective

We want RelayMiner to handle restarts gracefully.

## Origin Document

Currently RelayMiner doesn't pick up the existing on-going smt sessions from the disk. RelayMiner, however, should discover them on startup.

## Goals

- Avoid slashing of suppliers when relayminer is restarted;
- Ensure we account for relays even when relayminer is restarted;

## Deliverables

- [ ] Pick up SMT for the on-going sessions;
- [ ] Clean up SMT for the old sessions that are no longer relevant and proofs can't be submitted;

## General deliverables

- [ ] **Comments**: Add/update TODOs and comments alongside the source code so it is easier to follow.
- [ ] **Testing**: Add new tests (unit and/or E2E) to the test suite.
- [ ] **Makefile**: Add new targets to the Makefile to make the new functionality easier to use.
- [ ] **Documentation**: Update architectural or development READMEs; use [mermaid](https://mermaid-js.github.io/mermaid/) diagrams where appropriate.

---

**Creator**: @okdas 
**Co-Owners**: @red-0ne

PR #1140, opened 2025-03-20
[RelayMiner] Session Persistence for RelayerSessionsManager

red-0ne:
## Summary

Add persistent storage for relayer sessions to survive restarts and improve claim/proof lifecycle management.

### Primary Changes:
- Add session persistence to disk for relayer sessions, enabling recovery after restarts
- Implement automatic claim/proof lifecycle management for persisted sessions  
- Test the RelayerSessionsManager for session persistence

### Secondary changes:
- Export `CometBlockResult` type for external use
- Add `GetClaim` method to `ProofQueryClient` interface with caching
- Add `GetSMSTRoot` and `Stop` methods to `SessionTree` interface

## Issue

- Issue: #1107 

## Type of change

- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Sanity Checklist


pkg/client/interface.go
```diff
@@ -73,8 +73,8 @@ type SupplierClient interface {
 		ctx context.Context,
 		sessionProofs ...MsgSubmitProof,
 	) error
-	// Address returns the operator address of the SupplierClient that will be submitting proofs & claims.
-	OperatorAddress() *cosmostypes.AccAddress
+	// OperatorAddress returns the bech32 string representation of the supplier operator address.
```
bryanchriswhite, line 76: :ok_hand:

pkg/relayer/interface.go
```diff
@@ -176,6 +183,9 @@ type SessionTree interface {
 
 	// GetTrieSpec returns the trie spec of the SMST.
 	GetTrieSpec() smt.TrieSpec
+
+	// Stop stops the session tree and closes the KVStore.
```
bryanchriswhite, line 187: I feel like it would be helpful to articulate how this is similar to and/or different from the `Flush()` method (i.e. do they both free the same resources?). Perhaps consider updating the `Flush()` comment as well if that seems appropriate.

pkg/relayer/session/options.go
```diff
@@ -11,3 +11,11 @@ func WithStoresDirectory(storesDirectory string) relayer.RelayerSessionsManagerO
 		relSessionMgr.(*relayerSessionsManager).storesDirectory = storesDirectory
 	}
 }
+
+// WithSessionTreesInspector exposes the session trees map to an so it can be inspected.
```
bryanchriswhite, line 15:
```suggestion
// WithSessionTreesInspector exposes the session trees map so that it can be inspected.
```

pkg/relayer/session/options.go
```diff
@@ -11,3 +11,11 @@ func WithStoresDirectory(storesDirectory string) relayer.RelayerSessionsManagerO
...
+// This is useful for testing purposes, as it allows inspecting the session trees present in memory.
+func WithSessionTreesInspector(sessionTreeMap *SessionsTreesMap) relayer.RelayerSessionsManagerOption {
```
bryanchriswhite, line 17: :sunglasses: :ok_hand:

pkg/relayer/session/persistence.go
```diff
@@ -0,0 +1,200 @@
...
+	prooftypes "github.com/pokt-network/poktroll/x/proof/types"
+	sharedtypes "github.com/pokt-network/poktroll/x/shared/types"
+)
+
+// populateSessionTreeMap reconstructs the in-memory session tree map from previously persisted sessions.
+// This method:
+// 1. Retrieves all persisted session metadata from storage
+// 2. Validates each session against current blockchain height to determine its lifecycle state
+// 3. Deletes expired sessions that are past their proof submission window
+// 4. Deletes sessions past their claim window that don't have an associated onchain claim
```
bryanchriswhite, line 18: supernit: I feel like this would read slightly better if the ordering of the clauses of one was reversed to match the other:

```suggestion
// 3. Deletes expired sessions that are past their proof submission window
// 4. Deletes sessions that don't have an associated onchain claim which are past their claim window 
```

Alternatively:
```suggestion
// 3. Deletes expired sessions that are past their proof submission window
// 4. Deletes unclaimed sessions that are past their claim creation window
```
OR
```suggestion
// 3. Deletes expired sessions whose proof submission window has elapsed
// 4. Deletes unclaimed sessions whose claim creation window has elapsed
```

pkg/relayer/session/persistence.go
```diff
@@ -0,0 +1,200 @@
...
+// 5. Retains and loads valid sessions into memory for continued processing
```
bryanchriswhite, line 19: 1. It seems like the "load"ing part already happened in 1. 
2. The word "retains" here seems either out of order and/or redundant to me. I would suggest that it should come after "loads", if at all.
3. It seems to me like all that's left to describe at this step would be the fact that we're ignoring corrupt/non-deseralizable sessions. Would it make sense to rephrase this step in those terms?

pkg/relayer/session/persistence.go
```diff
@@ -0,0 +1,200 @@
...
+			"supplier_operator_address", supplierOperatorAddress,
+			"session_end_height", sessionEndHeight,
+		)
+
+		// Session lifecycle states and corresponding actions:
+		// 1. Active and accepting relays: Retain session for processing new relays
+		// 2. Within claim window: Retain session as it's in the process of creating a claim
+		// 3. Past claim window with onchain claim: Retain for proof submission
+		// 4. Past claim window without onchain claim: Delete as it's too late for settlement
+		// 5. Past proof window: Delete as the session is completely expired
```
bryanchriswhite, line 60: :fire: :fire: :fire:

pkg/relayer/session/persistence.go
```diff
@@ -0,0 +1,200 @@
...
+			"session_id", sessionId,
+			"supplier_operator_address", supplierOperatorAddress,
+			"session_end_height", sessionEndHeight,
+		)
+
+		// Session lifecycle states and corresponding actions:
+		// 1. Active and accepting relays: Retain session for processing new relays
+		// 2. Within claim window: Retain session as it's in the process of creating a claim
+		// 3. Past claim window with onchain claim: Retain for proof submission
+		// 4. Past claim window without onchain claim: Delete as it's too late for settlement
```
bryanchriswhite, line 59:
```suggestion
		// 4. Past claim window without onchain claim: Delete as it's too late to claim
```

pkg/relayer/session/persistence.go
bryanchriswhite, line 60: Perhaps there's an opportunity here to log whether the (and which) supplier should be expecting to be slashed in relevant cases (claimed + proof window closed + proof required). Even if it's just a TODO for now, wdyt?

pkg/relayer/session/persistence.go
```diff
@@ -0,0 +1,200 @@
...
+
+		// The session has still a chance to reach settlement.
+		// Add it to the sessionTrees map of blockHeight->sessionId->supplierOperatorAddress->sessionTree
+		sessionTreesWithEndHeight, ok := rs.sessionsTrees[sessionEndHeight]
+
+		// If there is no map for sessions at the sessionEndHeight, create one.
+		if !ok {
+			sessionTreesWithEndHeight = make(map[string]map[string]relayer.SessionTree)
+			rs.sessionsTrees[sessionEndHeight] = sessionTreesWithEndHeight
+		}
```
bryanchriswhite, line 106: I feel like this should be extracted to a method. Do you think there are other places where this may be duplicated?

pkg/relayer/session/persistence.go
```diff
@@ -0,0 +1,200 @@
...
+
+		// Get the sessionTreeWithSessionId for the given session.
+		sessionTreeWithSessionId, ok := sessionTreesWithEndHeight[sessionId]
+
+		// If there is no map for session trees with the session id, create one.
+		if !ok {
+			sessionTreeWithSessionId = make(map[string]relayer.SessionTree)
+			sessionTreesWithEndHeight[sessionId] = sessionTreeWithSessionId
+		}
```
bryanchriswhite, line 115: same as above

pkg/relayer/session/persistence.go
```diff
@@ -0,0 +1,200 @@
...
+
+		_, ok = sessionTreeWithSessionId[supplierOperatorAddress]
+		if ok {
+			sessionLogger.Error().Msg("the session tree already exists, skipping")
```
bryanchriswhite, line 119: Is there any possibility that the in-memory and persisted session trees have diverged? Does it make sense to compare/reconcile them in this case?

pkg/relayer/session/persistence.go
```diff
@@ -0,0 +1,200 @@
...
+			continue
+		}
+
+		sessionTree, treeErr := ImportSessionTree(persistedSMT, claim, rs.storesDirectory, rs.logger)
+		if treeErr != nil {
+			sessionLogger.Error().Err(treeErr).Msg("failed to import session tree")
+			continue
+		}
+
+		sessionTreeWithSessionId[supplierOperatorAddress] = sessionTree
```
bryanchriswhite, line 129: #PUC (if I'm understanding correctly) that this is where each loaded persisted session tree is actually included in the set of session trees which are processed by the observable pipeline.

pkg/relayer/session/persistence.go
```diff
@@ -0,0 +1,200 @@
...
+	}
+
+	return nil
+}
+
+// deletePersistedSessionTree deletes the persisted session tree and its metadata.
```
bryanchriswhite, line 135:
```suggestion
// deletePersistedSessionTree deletes the persisted session tree key/value store and its metadata.
```

pkg/relayer/session/persistence.go
```diff
@@ -0,0 +1,200 @@
...
+		return err
+	}
+
+	delete(rs.sessionsTrees[sessionEndHeight][supplierOperatorAddress], sessionId)
+
+	return nil
+}
+
+// persistSessionMetadata persists the session metadata to the store.
+// It is used to persist the session metadata after a session has been created.
```
bryanchriswhite, line 170: #PUC it may be worth noting that the session ID is the key.

bryanchriswhite (dismissed):
Nice one @red-0ne! :sunglasses: 

I'm stopping here for this morning. Will resume later.

image

pkg/client/interface.go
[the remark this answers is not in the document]
red-0ne: It was not needed, we systematically called `String()` when accessing it.

pkg/relayer/session/persistence.go
red-0ne: `populateSessionTreeMap` is called and blocking in `relayerSessionsManager.Start` before mined relays start updating trees.

pkg/relayer/session/persistence.go
```diff
@@ -0,0 +1,261 @@
...
+
+	// If there is no map for sessions at the sessionEndHeight, create one.
+	if !ok {
+		sessionTreesWithEndHeight = make(map[string]relayer.SessionTree)
+		supplierSessionTrees[sessionEndHeight] = sessionTreesWithEndHeight
+	}
+
+	// Get the sessionTree for the given session id.
+	if _, ok = sessionTreesWithEndHeight[sessionId]; ok {
+		return false
```
Olshansk, line 260: WHy don't we allow overrides?

Note that I've updated the godoc comment, but still feel like this should be okay behaviour.

pkg/relayer/session/persistence.go
```diff
@@ -0,0 +1,261 @@
...
+			continue
+		}
+
+		// Insert the session tree into the in-memory map to be processed by the observable pipeline.
+		if ok := rs.insertSessionTree(sessionTree); !ok {
+			sessionLogger.Error().Msg("the session tree already exists, skipping")
+		}
+	}
+
+	rs.proveClaimedSessions(ctx)
```
Olshansk, line 153: Can you #PUC why we're doing this here and why the normal proof pipeline is not enough?

pkg/relayer/session/sessiontree.go
```diff
@@ -113,6 +114,67 @@ func NewSessionTree(
...
+	// Initialize the basic session tree structure with metadata
+	sessionTree := &sessionTree{
+		logger:                  logger,
+		sessionHeader:           persistedSMT.SessionHeader,
+		storePath:               storePath,
+		sessionMu:               &sync.Mutex{},
+		supplierOperatorAddress: supplierOperatorAddress,
+	}
+
+	// SCENARIO 1: Session has been claimed
```
Olshansk, line 149: These comments are great. Do you mind adding logging for it too?

pkg/relayer/session/sessiontree.go
```diff
@@ -113,6 +114,67 @@ func NewSessionTree(
...
+	// When a claim exists, the session is in a finalized state with an immutable root hash.
```
Olshansk, line 150: Not completely true. 

If the claim window is still open, you can recreated it.

Can you:
1. Confirm we're aligned.
2. If yes, #PUC
3. If not, let's discuss

pkg/relayer/session/session.go
```diff
@@ -104,7 +120,13 @@ func NewRelayerSessions(
 		opt(rs)
 	}
 
-	if err := rs.validateConfig(); err != nil {
+	if err = rs.validateConfig(); err != nil {
+		return nil, err
+	}
+
+	// Initialize the session metadata store.
+	sessionRegistryDir := path.Join(rs.storesDirectory, "sessions_metadata")
```
Olshansk, line 128: Align naming once we align on something. First time I'm seeing `sessionRegistryDir`.

pkg/relayer/session/session.go
```diff
@@ -134,16 +168,59 @@ func (rs *relayerSessionsManager) Start(ctx context.Context) {
...
+						"failed to stop session tree store for sessionId %q",
+						sessionTree.GetSessionHeader().GetSessionId(),
+					)
+				}
+			}
+		}
+	}
+
+	// Close the metadata store that tracks all sessions and release its resources.
+	// Then clear the in-memory sessions map for testing purposes.
```
Olshansk, line 232: Why testing purposes?

pkg/relayer/session/session.go
```diff
@@ -134,16 +168,59 @@ func (rs *relayerSessionsManager) Start(ctx context.Context) {
...
+	if err := rs.sessionsMetadataStore.Stop(); err != nil {
+		rs.logger.Error().Err(err).Msg("failed to stop sessions metadata store")
+	}
+	clear(rs.sessionsTrees)
```
Olshansk, line 237: Should we log the number of session trees we're clearing?

Olshansk (changes requested):
@red-0ne AMAZING WORK HERE!

1. PTAL at the NITs, edits, comments, etc I made here: [`5659d87` (#1140)](https://github.com/pokt-network/poktroll/pull/1140/commits/5659d876b4ad73819379e048cf8eda156c9eb3d1)
2. I reviewed everything except for pkg/relayer/session/session_persistence_test.go
3. Lmk if we need to discuss some naming stuff sync
4. I'll test this myself when I'm updating RelayMiner + Supplier docs
5. Since you're taking over load testing, we'll see this in action!

pkg/relayer/session/persistence.go
```diff
@@ -0,0 +1,261 @@
...
+			"session_end_height", sessionEndHeight,
+			"claim_window_close", claimWindowCloseHeight,
+			"proof_window_close", proofWindowCloseHeight,
+			"current_height", height,
+		)
+
+		var claim *prooftypes.Claim
+		if height >= claimWindowOpenHeight {
+			foundClaim, err := rs.proofQueryClient.GetClaim(ctx, supplierOperatorAddress, sessionId)
+			if err != nil && status.Convert(err).Code() != codes.NotFound {
```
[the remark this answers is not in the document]
red-0ne: We should not fail when the error IS a `NotFound` error.

Having an `else` block is plain wrong though. Fixing.

pkg/relayer/session/persistence.go
```diff
@@ -0,0 +1,261 @@
...
+		"session_end_height", sessionEndHeight,
+	)
+
+	metadataBz, err := persistedSMT.Marshal()
+	if err != nil {
+		logger.Error().Err(err).Msg("failed to marshal metadata")
+		return err
+	}
+
+	if err := rs.sessionsMetadataStore.Set([]byte(sessionId), metadataBz); err != nil {
```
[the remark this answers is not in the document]
red-0ne: Great suggestion! Will do

pkg/relayer/session/persistence.go
red-0ne: Note that this map is `supplierOperatorAddress->blockHeight->sessionId->sessionTree`

When loading session trees from the store, there should be only one corresponding to a given supplier+sessionId.

Updating this case with:
```go
	// Get the sessionTree for the given session id.
	// An already existing session tree means that the the sessionTreeStore has
	// duplicate session trees for the same supplier operator address and session id,
	// which should not happen.
	if _, ok = sessionTreesWithEndHeight[sessionId]; ok {
		rs.logger.Warn().
			Str("session_id", sessionId).
			Str("supplier_operator_address", supplierOperatorAddress).
			Msg("session tree already exists, skipping")
		return false
	}
```

pkg/relayer/session/persistence.go
red-0ne: Done. I also added this to the claim steps of the pipeline
```
		// TODO_POST_MAINNET(red-0ne): Support skipping sessionTrees that have already their
		// claim created and are ready to be proven.
```

pkg/relayer/session/sessiontree.go
red-0ne: You are right, the claim could also be updated if the claim window is still open.

pkg/relayer/session/session.go
red-0ne: I only needed it on tests, In a real world scenario, the process shut-down should take care of this.

Will remove that comment anyway, as it's misleading to think that it's ONLY here for testing purposes.

pkg/relayer/session/persistence.go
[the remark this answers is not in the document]
Olshansk: Awesome, makes a lot more sense now!

Olshansk (approved):
Great iteration, let's get this in!

Issue #1130, opened 2025-03-18
[RelayMiner] SubmitProof Tx Retry Mechanism


Olshansk:
## Objective

Add support for RelayMiners retrying the `SubmitProof` transaction if it failed.

### Background on Issue

In the [most recent load test](https://discord.com/channels/824324475256438814/1350962330086674502), we discovered that Suppliers are getting slashed because they're not submitting a `Proof` after submitting a `Claim`.

The following screenshot shows @red-0ne's [investigation](https://discord.com/channels/824324475256438814/1350962330086674502/1351275367989772418) based on the [logs in grafana here](https://grafana.poktroll.com/explore?schemaVersion=1&panes=%7B%22zy7%22%3A%7B%22datasource%22%3A%22ae7iqr8r0phc0b%22%2C%22queries%22%3A%5B%7B%22refId%22%3A%22A%22%2C%22datasource%22%3A%7B%22type%22%3A%22victoriametrics-logs-datasource%22%2C%22uid%22%3A%22ae7iqr8r0phc0b%22%7D%2C%22editorMode%22%3A%22builder%22%2C%22expr%22%3A%22kubernetes.container_name%3A+%5C%22relayminer%5C%22+%22%2C%22queryType%22%3A%22instant%22%7D%5D%2C%22range%22%3A%7B%22from%22%3A%221742000974000%22%2C%22to%22%3A%221742000976000%22%7D%7D%7D&orgId=1):

![Image](https://github.com/user-attachments/assets/d2424ad7-6f55-4951-a455-3cecde30cc89)

### Background on Proposed Solution

**To mitigate (not solve) the issue, the solution is to retry more than once.**

Network retry mechanisms are a very common paradigm, so building our own would necessitate a very strong need, but the cosmos SDK does not have it built into the [tx builder](https://github.com/cosmos/cosmos-sdk/blob/main/x/auth/tx/builder.go#L60)

**🙌  Asking in the [cosmos-tech slack channel](https://buildwithgrove.slack.com/archives/C047X1T6ERK/p1742251639380469) led to the [cosmos/relayer](https://github.com/cosmos/relayer/blob/99ab82d4ef2e3f52b30aee25811207f0e116e3d4/relayer/chains/cosmos/tx.go#L129) which uses [avast/retry-go](https://pkg.go.dev/github.com/avast/retry-go).**

[...]

PR #1143, opened 2025-03-22
[RelayMiner] Implement RPC calls with retry mechanisms

red-0ne:
## Summary

Add retry capabilities to RPC calls to improve relayer resilience against network issues.

### Primary Changes:
* Add generic `retry.Call<T>()` function that handles retrying operations that return values
* Add specialized `retry.GetParams<T>()` function for parameter queries with retry support
 * Wrap critical RPC calls throughout the relayer with retry logic

### Secondary Changes:
* Change DefaultConnRetryLimit from 10 to math.MaxInt for event replay client
* Add a simple backof strategy and a default config
 
## Issue

- Issue: #1130

## Type of change

- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Sanity Checklist


pkg/relayer/session/session_retry_test.go
```diff
@@ -0,0 +1,402 @@
+package session_test
```
red-0ne, line 1: Note to reviewer: Testing the `RelayerSessionsManager` retry integration is not finished yet.

pkg/retry/retry.go
```diff
@@ -77,3 +81,78 @@ func OnError(
...
+) (T, error) {
+	// Wrap the GetParams call in a function that matches the signature expected by Call
+	// This adapts the paramsGetter.GetParams method to be used with our retry mechanism
+	return Call(
+		func() (T, error) { return paramsGetter.GetParams(ctx) },
+		retryStrategy...,
+	)
+}
+
+// WithExponentialBackoffFn creates a retry strategy with exponential backoff.
```
bryanchriswhite, line 141: #PUC with an explanation for the case where `retryCount >= maxRetryCount`; i.e. continue retrying with max delay.

pkg/retry/retry.go
```diff
@@ -77,3 +81,78 @@ func OnError(
...
+// indicates that no more retries should be attempted.
+//
+// If no retry strategy is provided, it defaults to an exponential backoff strategy
+// with 5 max retries, 1000ms initial delay, and 16000ms max delay.
+//
+// Returns the result from the work function and any error that occurred if retries are exhausted.
+func Call[T any](
+	work func() (T, error),
+	retryStrategy ...RetryStrategyFunc,
+) (T, error) {
```
bryanchriswhite, line 95: :fire: :fire: :fire:

bryanchriswhite (commented):
I'm noticing that it seems like most (if not all) of the time we're using the default exponential retry strategy. Is there a particular case that you had in mind that wouldn't follow this pattern?

It seems to me like there's an opportunity to encapsulate the `retry.Call` usage within the clients. I'm imagining this would require adding a retry strategy option to the clients as well. I think the only edge case this doesn't cover is one-off client methods that need a different retry strategy. I feel like there would still be options to address that edge case as well if we needed to. Wdyt?

pkg/client/tx/client.go
```diff
@@ -290,7 +292,15 @@ func (txnClient *txClient) SignAndBroadcast(
 		return either.SyncErr(err)
 	}
 
-	txResponse, err := txnClient.txCtx.BroadcastTx(txBz)
+	txResponse, err := retry.Call(func() (*cosmostypes.TxResponse, error) {
+		response, txErr := txnClient.txCtx.BroadcastTx(txBz)
+		// Wrap timeout height error to make it non-retryable.
```
bryanchriswhite, line 297: I'm wondering whether this may behave any differently/unexpectedly when using offline mode. :thinking: 

Perhaps not, but I just wanted to ask that you think about it as well for a moment and #PUC if you think of anything that's worth noting in this context.

pkg/relayer/session/session.go
```diff
@@ -72,7 +72,10 @@ type relayerSessionsManager struct {
 // Required dependencies:
 //   - client.BlockClient
 //   - client.SupplierClientMap
+//   - client.SharedQueryClient
+//   - client.ServiceQueryClient
 //   - client.ProofQueryClient
+//   - client.BankQueryClient
```
bryanchriswhite, line 78: :pray:

pkg/retry/errors.go
```diff
@@ -0,0 +1,11 @@
+package retry
+
+import (
+	sdkerrors "cosmossdk.io/errors"
+)
+
+var (
+	codespace = "retry"
+	// ErrRetryable allows the retry strategy to stop retrying due to a non-retryable error.
```
bryanchriswhite, line 9:
```suggestion
	// ErrNonRetryable signals the retry strategy to stop retrying due to a non-retryable error.
```

pkg/retry/retry.go
```diff
@@ -77,3 +85,80 @@ func OnError(
 		}
 	}
 }
+
+// Call executes a function repeatedly until it succeeds or the retry strategy
+// indicates that no more retries should be attempted.
```
bryanchriswhite, line 90: #PUC it seems to me that the "retry strategy" doesn't determine when to stop, that's the responsibility of the "retry function". :thinking: 

```suggestion
// Call executes a function repeatedly, according to the retry strategy until
// it succeeds or returns an `ErrNonRetryable`, which indicates that no more
// retries should be attempted.
```

pkg/retry/retry.go
```diff
@@ -77,3 +85,80 @@ func OnError(
...
+//
+// If no retry strategy is provided, it defaults to an exponential backoff strategy
+// with 5 max retries, 1000ms initial delay, and 16000ms max delay.
```
bryanchriswhite, line 93: #PUC shouldn't these match the args passed on line 15?

pkg/retry/retry.go
```diff
@@ -77,3 +85,80 @@ func OnError(
...
+		if err == nil {
+			return result, err
+		}
+
+		// Stop retrying and return the result if the error is non-retryable
+		if ErrNonRetryable.Is(err) {
+			return result, err
+		}
+
+		// Stop retrying and return if the error is a gRPC error
```
bryanchriswhite, line 121: :ok_hand: Nice :sunglasses: 

I think we should make an exception for ["unavailable"](https://pkg.go.dev/google.golang.org/grpc/codes@v1.67.1#Unavailable) and perhaps ["resource exhausted"](https://pkg.go.dev/google.golang.org/grpc/codes@v1.67.1#ResourceExhausted) codes though because they both represent potentially transient errors.

pkg/retry/retry.go
bryanchriswhite: This seems unresolved. :thinking:

testutil/integration/suites/base.go
```diff
@@ -177,7 +177,7 @@ func newInitChainerCollectModuleNamesFn(suite *BaseIntegrationSuite) integration
 		}
 
 		modType := reflect.TypeOf(mod)
-		if strings.Contains(modType.PkgPath(), "pocket") {
+		if strings.Contains(modType.PkgPath(), "poktroll") {
```
bryanchriswhite, line 180:
```suggestion
		// TODO_INCOMPLETE: replace "poktroll" with "pocket" once the repo rename is complete.
		// This is correlated to the go package path which will only change when the go module is
		// renamed (i.e. go.mod line 1).
		if strings.Contains(modType.PkgPath(), "poktroll") {
```

pkg/client/tx/client_test.go
```diff
@@ -446,7 +450,153 @@ func TestTxClient_SignAndBroadcast_Timeout(t *testing.T) {
...
+
+	appStake := types.NewCoin(volatile.DenomuPOKT, math.NewInt(1000000))
+	appStakeMsg := &apptypes.MsgStakeApplication{
+		// Providing address to avoid panic from #GetSigners().
+		Address:  signingAddr.String(),
+		Stake:    &appStake,
+		Services: client.NewTestApplicationServiceConfig(testServiceIdPrefix, 1),
+	}
+
+	// Sing and broadcast the message.
```
bryanchriswhite, line 526: :notes: 
```suggestion
	// Sign and broadcast the message.
```

pkg/client/tx/client_test.go
```diff
@@ -446,7 +450,153 @@ func TestTxClient_SignAndBroadcast_Timeout(t *testing.T) {
...
+	// The error count should remain the same.
+	require.Equal(t, 4, callStatus.errorCount)
+	// There should be one successful attempt.
+	require.Equal(t, 1, callStatus.successCount)
+
+	// Instruct the tx client to return a non-retryable error when submitting the transaction.
+	// This will cause the transaction client to stop retrying and return the error.
+	callStatus.errorToReturn = sdkerrors.ErrTxTimeoutHeight.Wrap(fmt.Errorf("test error").Error())
+
+	// Sing and broadcast the message.
```
bryanchriswhite, line 551: (same as above)

bryanchriswhite (dismissed):
400 LoC challenge completed - very nice @red-0ne! :fire: :raised_hands: :heart_eyes: 

This is fantastic to see, especially in this form! :muscle:

pkg/client/tx/client.go
red-0ne: Can't think of any misbehavior w.r.t. offline mode.

I refactored the `SignAndBroadcast` to have a timeout height arg though. The caller knows best when the tx should land (e.g. `CreateClaim` should timeout when `ClaimWindowCloseHeight` is reached)

pkg/retry/retry.go
```diff
@@ -77,3 +114,86 @@ func OnError(
...
+			return result, err
+		}
+
+		// Stop retrying and return if the error is a non-transient gRPC error
+		status, isGRPCError := status.FromError(err)
+		if isGRPCError && !slices.Contains(transientGRPCErrorCodes, status.Code()) {
+			return result, err
+		}
+
+		if !retryStrategy[0](ctx, retryCount) {
```
Olshansk, line 173: #PUC on the approach/logic here.

- What's with the [0]?
- What is it doing?
- Is there an associated TODO?

pkg/retry/retry.go
```diff
@@ -77,3 +114,86 @@ func OnError(
...
+	ctx context.Context,
+	work func() (T, error),
+	retryStrategy ...RetryStrategyFunc,
+) (T, error) {
+	var (
+		result T
+		err    error
+	)
+	if retryStrategy == nil {
+		retryStrategy = []RetryStrategyFunc{DefaultExponentialDelay}
```
Olshansk, line 148: Why not use `GetStrategy(ctx)` here?

pkg/retry/retry.go
```diff
@@ -77,3 +114,86 @@ func OnError(
...
+	maxRetryCount int,
+	initialDelayMs int,
+	maxDelayMs int,
+) RetryStrategyFunc {
+	return func(ctx context.Context, retryCount int) bool {
+		if retryCount >= maxRetryCount {
+			return false
+		}
+
+		backoffDelay := initialDelayMs * int(math.Pow(2, float64(retryCount)))
```
Olshansk, line 217: Consider parameterizing `2` or just making it a local contstant

pkg/retry/retry_test.go
```diff
@@ -439,3 +440,48 @@ func TestOnError_NegativeRetryLimit(t *testing.T) {
...
+		result, err := retry.Call(
+			ctx,
+			func() (string, error) {
+				attempts++
+				if attempts < maxAttempts {
+					return "", errors.New("not yet")
+				}
+				return "success", nil
+			},
+			retry.WithExponentialBackoffFn(5, 1, 10), // Very short delays for testing
```
Olshansk, line 460: Make these local vars for the test. 

Makes it easier to understand what we're dealing with / testing.

pkg/client/tx/client_test.go
```diff
@@ -446,7 +451,153 @@ func TestTxClient_SignAndBroadcast_Timeout(t *testing.T) {
...
+	}
+
+	// Sign and broadcast the message.
+	go txClient.SignAndBroadcast(ctx, 0, appStakeMsg)
+
+	// Wait for 5 seconds to allow the retry strategy to perform 4 failing retries.
+	time.Sleep(5 * time.Second)
+
+	// All attempts should have failed.
+	require.Equal(t, 4, callStatus.errorCount)
```
Olshansk, line 538: I'm a little worried about potential CI flakinss here.

Have you thought and/or looked into it?

pkg/client/tx/client_test.go
```diff
@@ -446,7 +451,153 @@ func TestTxClient_SignAndBroadcast_Timeout(t *testing.T) {
...
+	require.Equal(t, 5, callStatus.errorCount)
+	require.Equal(t, 1, callStatus.successCount)
+}
+
 // TODO_TECHDEBT: add coverage for sending multiple messages simultaneously
 func TestTxClient_SignAndBroadcast_MultipleMsgs(t *testing.T) {
 	t.SkipNow()
 }
+
+func newTxContext(
```
Olshansk, line 572: #PUC for this and `callStatus` and what it is and how its used.

Olshansk (changes requested):
@red-0ne I pushed some NITs & edits in this commit: [`5e5896f` (#1143)](https://github.com/pokt-network/poktroll/pull/1143/commits/5e5896f116514ad6e18c23c3f6bb16313d5c2702)

Requesting a few minor edits & changes and also clarification on a few changes.

Don't feel like you have to refactor it (because we have to push this through), but I want to make sure we capture TECHDEBT if its applicable.

Other than that, this is great. Ty!

pkg/client/tx/client.go
```diff
@@ -69,15 +65,10 @@ type CometTxEvent struct {
 // regarding their status.
 // It also depends on the BlockClient as a timer, synchronized to block height,
 // to facilitate transaction timeout logic. If a transaction doesn't appear to
-// have been committed by commitTimeoutHeightOffset number of blocks have elapsed,
-// it is considered as timed out. Upon timeout, the client queries the network for
-// the last status of the transaction, which is used to derive the asynchronous
-// error that's populated in the either.AsyncError.
+// have been committed by timeoutHeight, it is considered as timed out.
```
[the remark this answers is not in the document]
red-0ne: The `txClient` does handle tx timeouts through its `txTimeoutPool`. But no longer don't provide the `timeoutHeight`.

Your comment refactor is correct and should be kept :+1:

pkg/client/interface.go
```diff
@@ -64,13 +64,15 @@ type SupplierClient interface {
 	// session's mined relays.
```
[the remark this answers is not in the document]
red-0ne: Passing the `timeoutHeight` to `SignAndBoradcast` is the way to go.
* Internally, the function has no knowledge of when a tx is no longer valid to be submitted.
* The caller, in our cases `CreateClaim` and `SubmitProof` are more in knowledge about when the tx should timeout (`ClaimWindowCloseHeight` and `ProofWindowCloseHeight` respectively) this is why it should not be a static/configurable value.
* We can use `context.WithValue` but I'd rather prefer not making it opaque.

Will #PUC

pkg/retry/retry.go
red-0ne: Was planning on having `retryStrategy` optional to `Call` but I agree that it does not provide better DX compared to using `GetStrategy`.

pkg/client/tx/client_test.go
red-0ne: Yes I did. The subject under test is by nature very time related but it showed no signs of flakiness.

The exponential retry params and sleep time is in the order of `seconds` so no signs of flakiness should be observed. I'll keep an eye on it though.

Olshansk (approved):
Pushed a last set of changes at [`7dcbf64` (#1143)](https://github.com/pokt-network/poktroll/pull/1143/commits/7dcbf648207a0dca2a0fc8fe23da1282ce60695f).

Let's 🚢 it!

PR #1147, opened 2025-03-23
[Relayminer] add forward capabilities to send request to suppliers

eddyzags:
## Summary

This PR aims to add capabilities to forward requests to HTTP and a WebSocket server in the relayminer. 

It sets up a configurable authenticated endpoint (/services/{service_id}/forward) whose responsibility would be to forward raw requests from the Relay Server to the service ID (more precisely, service node or data node).

## Issue

- Issue: https://github.com/pokt-network/poktroll/issues/447

## Type of change

- [ ] Bug fix
- [ ] Code health or cleanup
- [x] Documentation
- [ ] Other (specify)

## Sanity Checklist

### DIFF
--- a/docusaurus/docs/operate/configs/relayminer_config.md
+++ b/docusaurus/docs/operate/configs/relayminer_config.md
@@ -215,7 +215,7 @@ forward:
   token: 8e8cff9152db92e960b00b4159b23f82084192f9d6c589337caab9705b3e0693
 ```
 
-Example query using `relayminer1` with `ollama` as supplier in Localnet environment:
+Example query to foward a request using the `relayminer1` to `ollama` service in Localnet environment:
 
 ```shell
 ?> cat data.json
@@ -224,7 +224,13 @@ Example query using `relayminer1` with `ollama` as supplier in Localnet environm
     "path": "/api/generate",
     "data": "{\"model\": \"qwen:0.5b\", \"prompt\": \"hello world.\", \"stream\": false}"
 }
-?> curl -X POST -H "token: 8e8cff9152db92e960b00b4159b23f82084192f9d6c589337caab9705b3e0693" -H "service-id: ollama" localhost:10001/services/ollama/forward --data-binary "@data.json"
+?> curl -X POST -H "token: 8e8cff9152db92e960b00b4159b23f82084192f9d6c589337caab9705b3e0693" localhost:10001/services/ollama/forward --data-binary "@data.json"
+```
+
+Example query to foward a websocket connection using the  `relayminer1` to `anvilws` service in Localnet environment:
+
+```shell
+?>  websocat -H="token: 8e8cff9152db92e960b00b4159b23f82084192f9d6c589337caab9705b3e0693" ws://localhost:10001/services/anvilws/forward
 ```
 
 Helpers to generate token in Makefile:
--- a/docusaurus/docs/operate/configs/relayminer_config.md
+++ b/docusaurus/docs/operate/configs/relayminer_config.md
@@ -200,8 +200,8 @@ Configures a `forward` server to send request to a supplier for a given
 service name. It is intended for operational use only (e.g., a relayminer
 operator may need to send requests to suppliers to verify the correctness
 of their responses going through the relayminer proxy).
-Requests forwarded via this endpoint bypass the miner and session
-manager mechanisms.
+Requests forwarded via this endpoint bypass the miner, meter and session
+mechanisms.
 To prevent misuse by unauthorized users, this endpoint requires
 authentication. A token can be defined using a 32 bytes hexadecimal
 `token`.
--- a/docusaurus/docs/operate/configs/relayminer_config.md
+++ b/docusaurus/docs/operate/configs/relayminer_config.md
@@ -24,6 +24,7 @@ You can find a fully featured example configuration at [relayminer_config_full_e
   - [`metrics`](#metrics)
   - [`pprof`](#pprof)
   - [`ping`](#ping)
+  - [`forward`](#forward)
 - [Pocket node connectivity](#pocket-node-connectivity)
   - [`query_node_rpc_url`](#query_node_rpc_url)
   - [`query_node_grpc_url`](#query_node_grpc_url)
@@ -189,6 +190,47 @@ Example configuration:
 ping:
   enabled: true
   addr: localhost:8081
+```
+
+### `forward`
+
+_`Optional`_
+
+Configures a `forward` server to send request to a supplier for a given
+service name. It is intended for operational use only (e.g., a relayminer
+operator may need to send requests to suppliers to verify the correctness
+of their responses going through the relayminer proxy).
+Requests forwarded via this endpoint bypass the miner and session
+manager mechanisms.
+To prevent misuse by unauthorized users, this endpoint requires
+authentication. A token can be defined using a 32 bytes hexadecimal
+`token`.
+
+Example configuration:
+
+```yaml
+forward:
+  enabled: true
+  addr: localhost:8082
+  token: 8e8cff9152db92e960b00b4159b23f82084192f9d6c589337caab9705b3e0693
+```
+
+Example query using `relayminer1` with `ollama` as supplier in Localnet environment:
+
+```shell
+?> cat data.json
+{
+    "method": "POST",
+    "path": "/api/generate",
+    "data": "{\"model\": \"qwen:0.5b\", \"prompt\": \"hello world.\", \"stream\": false}"
+}
+?> curl -X POST -H "token: 8e8cff9152db92e960b00b4159b23f82084192f9d6c589337caab9705b3e0693" -H "service-id: ollama" localhost:10001/services/ollama/forward --data-binary "@data.json"
+```
+
+Helpers to generate token in Makefile:
+
+```shell
+?> make relayminer_forward_token_gen
 ```
 
 ## Pocket node connectivity
--- a/localnet/poktrolld/config/relayminer_config.yaml
+++ b/localnet/poktrolld/config/relayminer_config.yaml
@@ -26,3 +26,7 @@ pprof:
 ping:
   enabled: false
   addr: localhost:8082
+forward:
+  enabled: false
+  addr: localhost:10000
+  token: 8cc09793290cd64d8a9bc80eaae4fbeef5f7cf797b0c70e078d2a5b81d74f12c
--- a/makefiles/relay.mk
+++ b/makefiles/relay.mk
@@ -31,3 +31,7 @@ send_relay_path_WEBSOCKET: test_e2e_env ## Send a WEBSOCKET relay through PATH t
 
 # TODO_MAINNET(@olshansk): Add all the permissionless/delegated/centralized variations once
 # the following documentation is ready: https://www.notion.so/buildwithgrove/Different-Modes-of-Operation-PATH-LocalNet-Discussions-122a36edfff6805e9090c9a14f72f3b5?pvs=4#151a36edfff680d681a2dd7f4e5fee55
+
+.PHONY: relayminer_forward_token_gen
+relayminer_forward_token_gen:
+	@openssl rand -hex 32 | tr -d "\n"
--- a/pkg/relayer/cmd/cmd.go
+++ b/pkg/relayer/cmd/cmd.go
@@ -167,6 +167,12 @@ func runRelayer(cmd *cobra.Command, _ []string) error {
 		}
 	}
 
+	if relayMinerConfig.Forward.Enabled {
+		if err := relayMiner.ServeForward(ctx, "tcp", relayMinerConfig.Forward.Addr, relayMinerConfig.Forward.Token); err != nil {
+			return fmt.Errorf("serve forward: %w", err)
+		}
+	}
+
 	// Start the relay miner
 	logger.Info().Msg("Starting relay miner...")
 	if err := relayMiner.Start(ctx); err != nil && !errors.Is(err, http.ErrServerClosed) {
--- a/pkg/relayer/config/errors.go
+++ b/pkg/relayer/config/errors.go
@@ -11,4 +11,5 @@ var (
 	ErrRelayMinerConfigEmpty                 = sdkerrors.Register(codespace, 2104, "empty RelayMiner config")
 	ErrRelayMinerConfigInvalidSupplier       = sdkerrors.Register(codespace, 2105, "invalid supplier in RelayMiner config")
 	ErrRelayMinerConfigInvalidServer         = sdkerrors.Register(codespace, 2106, "invalid server in RelayMiner config")
+	ErrRelayerMinerWrongForwardToken         = sdkerrors.Register(codespace, 2107, "wrong or empty forward.token in configuration file. (you can use 'make relayminer_forward_token_gen' command to generate a token)")
 )
--- a/pkg/relayer/config/types.go
+++ b/pkg/relayer/config/types.go
@@ -25,6 +25,7 @@ type YAMLRelayMinerConfig struct {
 	SmtStorePath           string                         `yaml:"smt_store_path"`
 	Suppliers              []YAMLRelayMinerSupplierConfig `yaml:"suppliers"`
 	Ping                   YAMLRelayMinerPingConfig       `yaml:"ping"`
+	Forward                YAMLRelayMinerForwardConfig    `yaml:"forward"`
 }
 
 // YAMLRelayMinerPingConfig represents the configuration to expose a ping server.
@@ -32,6 +33,18 @@ type YAMLRelayMinerPingConfig struct {
 	Enabled bool `yaml:"enabled"`
 	// Addr is the address to bind to (format: 'hostname:port') where 'hostname' can be a DNS name or an IP
 	Addr string `yaml:"addr"`
+}
+
+// YAMLRelayMinerForwardConfig represents the configuration to expose a forward
+// request server.
+type YAMLRelayMinerForwardConfig struct {
+	Enabled bool `yaml:"enabled"`
+
+	// Addr is the address to bind to (format: 'hostname:port') where 'hostname' can be a DNS name or an IP.
+	Addr string `yaml:"addr"`
+
+	// Token must represent an 32 bytes hexadecimal string. (mandatory flag when activating forwarding capabilities)
+	Token string `yaml:"token"`
 }
 
 // YAMLRelayMinerPocketNodeConfig is the structure used to unmarshal the pocket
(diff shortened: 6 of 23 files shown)


eddyzags:
Here is the new PR @Olshansk ! 
I'm not sure what label I am supposed to set for myself and the reviewers. Is there any documentation about it please?

Olshansk:
@eddyzags I've added the `relayminer` label. Please hold off on adding `devnet-e2e` until we've gone through a few rounds of review. I want to avoid wasting infrastructure resources.

docusaurus/docs/operate/configs/relayminer_config.md
```diff
@@ -191,6 +192,53 @@ ping:
...
+
+Configures a `forward` server to send request to a supplier for a given
+service name. It is intended for operational use only (e.g., a relayminer
+operator may need to send requests to suppliers to verify the correctness
+of their responses going through the relayminer proxy).
+Requests forwarded via this endpoint bypass the miner, meter and session
+mechanisms.
+To prevent misuse by unauthorized users, this endpoint requires
+authentication. A token can be defined using a 32 bytes hexadecimal
+`token`.
```
Olshansk, line 207:
```suggestion
Configures a `forward` server to send requests to a particular supplier (i.e. operator)
for a given service name.

This endpoint:
- Is intended for operational use only (e.g., a relayminer operator may need to send 
  requests to an suppliers to verify the correctness of their responses going 
  through the relayminer proxy)
- Bypasses the onchain miner, meter and session mechanisms

To prevent misuse by unauthorized users:
- This endpoint requires authentication
- A token can be defined using a 32 bytes hexadecimal `token`
```

makefiles/relay.mk
```diff
@@ -30,4 +30,8 @@ send_relay_path_WEBSOCKET: test_e2e_env ## Send a WEBSOCKET relay through PATH t
 		-H "Target-Service-Id: anvilws"
 
 # TODO_MAINNET(@olshansk): Add all the permissionless/delegated/centralized variations once
-# the following documentation is ready: https://www.notion.so/buildwithgrove/Different-Modes-of-Operation-PATH-LocalNet-Discussions-122a36edfff6805e9090c9a14f72f3b5?pvs=4#151a36edfff680d681a2dd7f4e5fee55
\ No newline at end of file
+# the following documentation is ready: https://www.notion.so/buildwithgrove/Different-Modes-of-Operation-PATH-LocalNet-Discussions-122a36edfff6805e9090c9a14f72f3b5?pvs=4#151a36edfff680d681a2dd7f4e5fee55
+
+.PHONY: relayminer_forward_token_gen
+relayminer_forward_token_gen:
```
Olshansk, line 36: Please add a comment (like the others) so it shows up via `make help`

pkg/relayer/cmd/cmd.go
```diff
@@ -167,6 +167,12 @@ func runRelayer(cmd *cobra.Command, _ []string) error {
 		}
 	}
 
+	if relayMinerConfig.Forward.Enabled {
```
Olshansk, line 170: Please add a comment above of why a user would enable it

pkg/relayer/proxy/proxy.go
```diff
@@ -201,3 +202,26 @@ func (rp *relayerProxy) PingAll(ctx context.Context) error {
...
+	for _, srv := range rp.servers {
+		if err := srv.Forward(ctx, serviceID, w, req); err != nil {
+			if errors.Is(err, ErrRelayerProxyServiceIDNotFound) {
+				continue
+			}
+			return err
+		} else {
+			found = true
+			break
+		}
```
Olshansk, line 219:
```suggestion
		if err := srv.Forward(ctx, serviceID, w, req); err != nil {
			if errors.Is(err, ErrRelayerProxyServiceIDNotFound) {
				continue
			}
			return err
		}
		found = true
		break
```

pkg/relayer/proxy/proxy.go
```diff
@@ -201,3 +202,26 @@ func (rp *relayerProxy) PingAll(ctx context.Context) error {
 
 	return nil
 }
+
+// Forward iterates through all the relay servers, then tries to forward the given request.
+func (rp *relayerProxy) Forward(ctx context.Context, serviceID string, w http.ResponseWriter, req *http.Request) error {
+	found := false
```
Olshansk, line 208:
```suggestion
	successfullyForwardedRequest := false
```

pkg/relayer/config/errors.go
```diff
@@ -11,4 +11,5 @@ var (
 	ErrRelayMinerConfigEmpty                 = sdkerrors.Register(codespace, 2104, "empty RelayMiner config")
 	ErrRelayMinerConfigInvalidSupplier       = sdkerrors.Register(codespace, 2105, "invalid supplier in RelayMiner config")
 	ErrRelayMinerConfigInvalidServer         = sdkerrors.Register(codespace, 2106, "invalid server in RelayMiner config")
+	ErrRelayerMinerWrongForwardToken         = sdkerrors.Register(codespace, 2107, "wrong or empty forward.token in configuration file. (you can use 'make relayminer_forward_token_gen' command to generate a token)")
```
Olshansk, line 14: Can you add some documentation on the token type, it's expectations, etc...?

pkg/relayer/relayminer.go
```diff
@@ -182,3 +184,47 @@ func (rel *relayMiner) newPinghandlerFn(ctx context.Context) http.Handler {
 		w.WriteHeader(http.StatusNoContent)
 	})
 }
+
+// ServeForward exposes a forward HTTP server for administrators to send request to
```
Olshansk, line 188: Can you move this into a new file? `pkg/relayer/forward_server.go`

pkg/relayer/proxy/websockets/connection.go
```diff
@@ -81,9 +81,9 @@ type connection struct {
 	isClosed atomic.Bool
 }
 
-// connectServiceBackend establishes a websocket connection established by the
+// ConnectServiceBackend establishes a websocket connection established by the
 // relay miner client to the service backend endpoint.
-func connectServiceBackend(serviceBackendUrl *url.URL, header http.Header) (*websocket.Conn, error) {
+func ConnectServiceBackend(serviceBackendUrl *url.URL, header http.Header) (*websocket.Conn, error) {
```
Olshansk, line 86: Why did you make this public?

pkg/relayer/proxy/sync.go
```diff
@@ -227,3 +234,107 @@ func (server *relayMinerHTTPServer) sendRelayResponse(
 	_, err = writer.Write(relayResponseBz)
 	return err
 }
+
+// forwardPayload represents the HTTP request body format to forward a request
+// to the supplier.
+type forwardPayload struct {
```
Olshansk, line 240: can we move this into `pkg/relayer/proxy/types.go` along with the helpers below?

pkg/relayer/proxy/async.go
```diff
@@ -96,3 +98,62 @@ func (server *relayMinerHTTPServer) handleAsyncConnection(
 
 	return nil
 }
+
+// forwardAsyncConnection instantiates two websocket connections that:
```
Olshansk, line 102: Can you #PUC why we need two websockeet connections and one is not enough?

pkg/relayer/config/types.go
```diff
@@ -34,6 +38,18 @@ type YAMLRelayMinerPingConfig struct {
...
 
+// YAMLRelayMinerForwardConfig represents the configuration to expose a forward
+// request server.
+type YAMLRelayMinerForwardConfig struct {
+	Enabled bool `yaml:"enabled"`
+
+	// Addr is the address to bind to (format: 'hostname:port') where 'hostname' can be a DNS name or an IP.
+	Addr string `yaml:"addr"`
+
+	// Token must represent an 32 bytes hexadecimal string. (mandatory flag when activating forwarding capabilities)
```
Olshansk, line 49: #PUC how it can be generated using the make helper

pkg/relayer/config/types.go
```diff
@@ -104,6 +121,17 @@ type RelayMinerPingConfig struct {
...
 }
 
+// RelayMinerForwardConfig is the structure resulting from parsing the
+// forward server configuration.
+type RelayMinerForwardConfig struct {
+	Enabled bool
+	// Addr is the address to bind to (format: 'hostname:port') where 'hostname' can be a DNS name or an IP.
+	Addr string
+
+	// Token must represent an 32 bytes hexadecimal string. (mandatory flag when activating forwarding capabilities).
```
Olshansk, line 131: Ditto

pkg/relayer/config/types.go
```diff
@@ -104,6 +121,17 @@ type RelayMinerPingConfig struct {
 	Addr string
 }
 
+// RelayMinerForwardConfig is the structure resulting from parsing the
+// forward server configuration.
+type RelayMinerForwardConfig struct {
+	Enabled bool
+	// Addr is the address to bind to (format: 'hostname:port') where 'hostname' can be a DNS name or an IP.
```
Olshansk, line 128:
```suggestion

	// Addr is the address to bind to (format: 'hostname:port') where 'hostname' can be a DNS name or an IP.
```

Olshansk (changes requested):
@eddyzags Thanks for the change!

1. A few minor nits/comments/edits
2. Please merge with main
3. In case you missed it, do check out `make send_relay_path_WEBSOCKET`: it's websockets leveraging the PATH gateway
4. Can you provide some more detail/rationale on the:
- Token
- Use case
- Maybe an E2E mermaid diagram of how this works

I'm still struggling to understand what use case this serves that was possible before, but am pretty sure it's just an oversight on my part.

jorgecuesta:
@Olshansk @eddyzags is this a kind of `--simRelays` of more? if this is that, it should be on next iter, non-blocking release because you already have `ping` feature which will verify the communication to the backend, right?

Olshansk:
Delegating to @eddyzags to answer this question.

eddyzags:
> is this a kind of --simRelays of more?

Yes. As stated in this issue #447 , the objective is to:
_add a functionality to RelayMiner: an ability to send relay through a supplier without providing meta information (session, signature, etc.) so operators can verify their RelayMiner configuration._

> if this is that, it should be on next iter, non-blocking release because you already have ping feature which will verify the communication to the backend, right?

I don't have any opinions on whether this should be in the next iteration or a non-blocking release. But, the `/ping` and `/forward` endpoints serve two purposes (or assert two different things).

From an OSI standpoint, the `/ping` endpoint ensures that the Physical, Data Link, and network layers are working correctly in the intercommunication between the client (relayminer) and the server (data node).
On the other hand, the `/forward` endpoint asserts a bit more. It also gives operators confidence about the Session, Presentation, and Application layers.

I think we should give both capabilities to the Relayminer's users.

cc @jorgecuesta @Olshansk

Olshansk:
@eddyzags Your explanation above makes sense to me!

Love the OSI analogy you put together as well.

For reference, I'm working on a `pocketd relayminer relay` utility here: https://github.com/pokt-network/poktroll/pull/1298

Please re-request a review once this is ready.

pkg/relayer/proxy/websockets/connection.go
eddyzags: I made this function public so I can reuse it in the `relayMinerHTTPServer.forwardAsyncConnection()` method (`pkg/relayer/proxy/async.go`). I wanted the forward server to behave the same way as the relay server, so I reused the same WebSocket dialer configuration.

pkg/relayer/proxy/async.go
eddyzags: It is written in the comment ([103 & 104](https://github.com/pokt-network/poktroll/blob/647cf9d53a761ca0fda0779f12800ed95f89d16c/pkg/relayer/proxy/async.go#L103-L104)).

I need two connections to proxy the websocket traffic from the client to the backend websocket service. One connection (clientConn) between your proxy and the client, and one connection (serviceConn) between your proxy and the backend service.

eddyzags:
👋🏾 @Olshansk 

> Can you provide some more detail/rationale on the:
  Token

This is an authentication token for the forward server. I want to give this access exclusively to the operators because this is one way of bypassing the miner, meter, session mechanism. Maybe this should be name `AuthToken`? wdyt?

> Use case

Would you like more details on the use case of the whole feature? Or only the AuthToken?


> Maybe an E2E mermaid diagram of how this works

Same question. The whole feature or the Auth Token?

Olshansk:
@eddyzags Can you merge this with main and ensure it works?

@bryanchriswhite Can you take a look at this and review it before I take a second look?

pkg/relayer/forward_server.go
```diff
@@ -0,0 +1,54 @@
...
+// ServeForward exposes a forward HTTP server for administrators to send request to
+// specific service.
+func (rel *relayMiner) ServeForward(ctx context.Context, network, addr, token string) error {
+	ln, err := net.Listen(network, addr)
+	if err != nil {
+		return fmt.Errorf("net listen: %w", err)
+	}
+
+	muxRouter := chi.NewRouter()
+	muxRouter.HandleFunc("/services/{service_id}/forward", rel.newForwardHandlerFn(ctx, token))
```
bryanchriswhite, line 21: Neat! :eyes:

bryanchriswhite (changes requested):
image

Taking a break here. :sweat_smile: 
Looking great so far - thanks @eddyzags! :raised_hands:

docusaurus/docs/1_operate/3_configs/4_relayminer_config.md
```diff
@@ -190,6 +191,54 @@ ping:
...
+Configures a `forward` server to send requests to a particular supplier (i.e. operator)
+for a given service name.
+This endpoint:
+- Is intended for operational use only (e.g., a relayminer operator may need to send 
+  requests to an suppliers to verify the correctness of their responses going 
+  through the relayminer proxy)
+- Bypasses the onchain miner, meter and session mechanisms
+To prevent misuse by unauthorized users:
+- This endpoint requires authentication
+- A token can be defined using a 32 bytes hexadecimal `token`
```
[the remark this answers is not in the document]
eddyzags: I agree with both points. I would even challenge my approach. Right now, the forward server is exposing one endpoint, `/services/{service_id}/forward`. The handler will look into the HTTP header for a specific key and value to authenticate incoming requests.

```
Token: <token>
```

But why not use the Bearer Authentication scheme, a well-known authentication scheme. This approach would mean that:
1. There is no strict format required by the HTTP authorization header for Bearer tokens, but we strongly recommend using a 32-byte hexadecimal auth token
2. It would make sense to call this configuration field `auth_token`.

```
Authentication: Bearer <token>
```

wdyt?

makefiles/relay.mk
```diff
@@ -49,4 +48,8 @@ test_relay_util_100: check_path_up check_relay_util  ## Test anvil PATH behind G
 		-H "App-Address: pokt1mrqt5f7qh8uxs27cjm9t7v9e74a9vvdnq5jva4" \
 		-d '{"jsonrpc":"2.0","method":"eth_blockNumber","id":1}' \
 		-x 100 \
-		-b
\ No newline at end of file
+		-b
+
+.PHONY: relayminer_forward_token_gen
+relayminer_forward_token_gen: ## Generate 32 bytes hexadecimal token for relayminer forward configuration.
+	@openssl rand -hex 32 | tr -d "\n"
```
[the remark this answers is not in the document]
eddyzags: I agree. Done here https://github.com/pokt-network/poktroll/pull/1147/commits/62ed98e460237ec07d516ebad9c53cf827eeff7e

localnet/poktrolld/config/relayminer_config.yaml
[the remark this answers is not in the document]
eddyzags: 🧹 https://github.com/pokt-network/poktroll/pull/1147/commits/d274000ed2b645e77c17e3c369886bdae666414f

pkg/relayer/config/errors.go
```diff
@@ -11,4 +11,5 @@ var (
 	ErrRelayMinerConfigEmpty                 = sdkerrors.Register(codespace, 2104, "empty RelayMiner config")
 	ErrRelayMinerConfigInvalidSupplier       = sdkerrors.Register(codespace, 2105, "invalid supplier in RelayMiner config")
 	ErrRelayMinerConfigInvalidServer         = sdkerrors.Register(codespace, 2106, "invalid server in RelayMiner config")
+	ErrRelayerMinerWrongForwardToken         = sdkerrors.Register(codespace, 2107, "wrong or empty forward.token in configuration file. Expect a 32 bytes hexadecimal string (you can use 'make relayminer_forward_token_gen' command to generate a token)")
```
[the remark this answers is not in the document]
eddyzags: I think you got confused here. This error isn't used for unauthorized request when requesting a resource (`/services/{service_id}/forward`), but this error is used when the Relayminer's operator defines a wrong authentication token in the Relayminer's configuration file ([see code here](https://github.com/pokt-network/poktroll/blob/2be7c894d988597723c42d4a73ccc9bd70851980/pkg/relayer/config/relayminer_configs_reader.go#L55)). In other words, this error is used when the Relayminer field validation isn't what we are expecting.

That being said, this error no longer makes sense and must be deleted based on this discussion https://github.com/pokt-network/poktroll/pull/1147#discussion_r2113568196

pkg/relayer/config/types.go
```diff
@@ -104,6 +122,18 @@ type RelayMinerPingConfig struct {
...
+	Addr string
+
+	// Token is a 32 bytes hexadecimal string to authenticate request to the forward server. (mandatory flag when enabling forwarding capabilities)
+	// You can use 'make relayminer_forward_token_gen' command to print a well-formatted token.
+	Token string
```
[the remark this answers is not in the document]
eddyzags: ea0136afcab85bdf592a81eaeb290e64f749e3b2

pkg/relayer/proxy/http_server.go
```diff
@@ -159,6 +159,22 @@ func (server *relayMinerHTTPServer) Ping(ctx context.Context) error {
 	return nil
 }
 
+// Forward sends request to the appropriate service.
+// - It checks if the service id is managed by the relayminer.
```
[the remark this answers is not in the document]
eddyzags: Yes exactly 👍🏾

pkg/relayer/interface.go
```diff
@@ -94,6 +98,9 @@ type RelayServer interface {
 
 	// Ping tests the connection between the relay server and its backend URL.
 	Ping(ctx context.Context) error
+
+	// Forward sends a request to the supplier service.
```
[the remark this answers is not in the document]
eddyzags: e8dc4bd61fd591f6690e4b01503d2017b70a1884

Olshansk:
@eddyzags Do you mind resolving the conflicts with main so we can re-review and get this in?

eddyzags:
@Olshansk, there were quite a few changes to this part in the main branch. I'll have to understand it first to resolve the conflicts while keeping the code working. I am working on it. I try to push the changes as soon as I can.

Issue #1149, opened 2025-03-24
[Morse -> Shannon Migration] Claim Tx Fees


bryanchriswhite:
## Objective

Enable fee-less Morse account claim transactions.

## Origin Document

![Image](https://github.com/user-attachments/assets/22741a0b-5543-4b92-8a8e-907ad38d5110)

## Goals

- Remove blockers to Morse account migration

## Deliverables

- [ ] A PR which implements an antehandler that allows for a fee-less Morse claim message according to the flowchart preseted in the origin document section.

## Non-goals / Non-deliverables

- Comprehensive DoS mitigation to cover the attack vector exposed as a result - this SHOULD be addressed in a future issue (#xxx).

## General deliverables

- [ ] **Comments**: Add/update TODOs and comments alongside the source code so it is easier to follow.
- [ ] **Testing**: Add new tests (unit and/or E2E) to the test suite.
- [ ] **Makefile**: Add new targets to the Makefile to make the new functionality easier to use.
- [ ] **Documentation**: Update architectural or development READMEs; use [mermaid](https://mermaid-js.github.io/mermaid/) diagrams where appropriate.

---

**Creator**: @bryanchriswhite 
**Co-Owners**: @Olshansk @red-0ne @okdas

Olshansk:
@bryanchriswhite Thanks for creating the issue.

I know it's late in the iteration and will be spilled over, but I went ahead and moved it into this one.

@okdas and @red-0ne will also be involved in this effort. In particular, I'll need red0ne's help on reviews.

![Image](https://github.com/user-attachments/assets/28c21bd1-ca81-4163-8441-e4b1244019f5)

PR #1160, opened 2025-03-28
[Morse -> Shannon Migration] Waive fees for Morse claim txs

bryanchriswhite:
## Summary

Waive fees for txs containing ONLY Morse claim messages.

_NOTE: This PR **DOES NOT** make any client-side gas/fee related changes, which **WILL** be necessary once we turn fees on. See #1165._

### Testing these changes

Due to the need for some client-side and test helper changes, I've isolated those changes to a different (sloppier) branch: #1165.

Changes:

- Set a custom ante handler which does the following if the tx contains ONLY Morse claim messages:
  - uses a `SignatureVerificationGasConsumer` that waives fees..
  - uses a `TxFeeChecker` that bypasses any minimum gas/fee requirement..

## Issue

- Issue: #1149

## Type of change

- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Sanity Checklist


x/migration/keeper/msg_server_claim_morse_supplier.go
```diff
@@ -60,6 +82,7 @@ func (k msgServer) ClaimMorseSupplier(
 
 	// Ensure that the given MorseClaimableAccount has not already been claimed.
 	if morseClaimableAccount.IsClaimed() {
+		isAlreadyClaimed = true
```
Olshansk, line 85: What's this for?

x/migration/keeper/msg_server_claim_morse_supplier.go
```diff
@@ -29,6 +29,28 @@ func (k msgServer) ClaimMorseSupplier(
 	sdkCtx := cosmostypes.UnwrapSDKContext(ctx)
 	logger := k.Logger().With("method", "ClaimMorseSupplier")
 
+	// Ensure that gas fees are NOT waived if the claim is invalid.
```
Olshansk, line 32: Please start using the code cleaner comment claude project.

This is much easier to read:

```
// Ensure that gas fees are NOT waived if the claim is invalid.
// This restores the disincentive for Shannon account holders to spam invalid Morse claim txs:
//   • BUT ONLY if the spanner is using a funded account
//   • In cases where the tx contains ONLY one or more Morse claim messages:
//     - The validators which consider the tx WILL do slightly more work than the typical message
//     - Which would've been rejected prior to the message handler during CheckTx
```

Olshansk: Take this from someone who reads lots of documents, PRs, etc across backend, BD, infra, protocol, etc on a daily basis.

x/migration/keeper/msg_server_claim_morse_supplier.go
```diff
@@ -29,6 +29,28 @@ func (k msgServer) ClaimMorseSupplier(
...
+	// message, which would've been rejected prior to the message handler during CheckTx.
+	var (
+		morseClaimableAccount              migrationtypes.MorseClaimableAccount
+		isFound, isValid, isAlreadyClaimed bool
+	)
+	defer func() {
+		if !isFound || !isValid || isAlreadyClaimed {
+			// Attempt to charge the waived gas fee for invalid claims.
+			sdkCtx.GasMeter()
+			// DEV_NOTE: Assuming that the tx containing this message was signed
```
Olshansk, line 46: The assumption is fine. 

Any way we can check it just to make sure `x/` code errors rather than panic in the offchance.

app/ante.go
```diff
@@ -0,0 +1,110 @@
...
+	authtypes "github.com/cosmos/cosmos-sdk/x/auth/types"
+
+	"github.com/pokt-network/poktroll/app/volatile"
+)
+
+// newAnteHandlerFn returns an AnteHandler that waives minimum gas/fees for transactions
+// that contain ONLY morse claim messages (i.e. MsgClaimMorseAccount, MsgClaimMorseApplication,
+// and MsgClaimMorseSupplier).
+//
+// TODO_MAINNET(@bryanchriswhite):
```
Olshansk, line 17: Please start using TODO_MAINNET_CRITICAL.

Olshansk (changes requested):
Did a quick glance (to get an idea of things) but now I'm confused: do we not need the "new token" solution anymore?

x/migration/keeper/msg_server_claim_morse_supplier.go
bryanchriswhite: image

I set it up.

x/migration/keeper/msg_server_claim_morse_supplier.go
bryanchriswhite: 1. There is no way to check AFAIK because we cannot get a reference to the tx in this scope (message handler). Typically the ante handler deals with this long before any message handlers are called.
2. What makes you say that there would be a panic? The consequence of this assumption being wrong is NOT a panic, rather it's charging the incorrect gas amount for that key type. For additional context, the gas amount for each key type is its own `auth` module param.

x/migration/keeper/msg_server_claim_morse_supplier.go
bryanchriswhite: I #PUC'd the comment above its declaration; it's needed to track whether the Morse account was already claimed prior to the execution of the rest of the message handler. 

We want to restore the normal gas behavior as much as possible. In this case, that means gas WILL be charged for a **valid** claim message which has already been processed. This mitigates a replay attack on Morse claim messsages, for example.

x/migration/keeper/msg_server_claim_morse_supplier.go
bryanchriswhite: I've also iterated a bit on the ante handler logic to restrict the gas waiver to txs with exactly one secp256k1 signer. This hardens things a bit further.

bryanchriswhite:
Still looking into the failing tests...

app/ante.go
```diff
@@ -0,0 +1,156 @@
...
+
+// freeSigGasConsumer is a signature verification gas consumer that does not
+// consume any gas for signature verification. It is intended to ONLY be applied
+// to txs that consist of ONLY Morse claim messages.
+func freeSigGasConsumer(
+	_ storetypes.GasMeter,
+	_ signing.SignatureV2,
+	_ authtypes.Params,
+) error {
+	return nil
```
Olshansk, line 148: Should there be a TODO here to change it?

If not, #PUC why returning nil is enough.

app/ante.go
```diff
@@ -0,0 +1,156 @@
...
+}
+
+// freeTxFeeChecker is a TxFeeChecker that always returns zero gas fees
```
Olshansk, line 151: Ditto, if we are expecting to iterate here, #PUC

app/ante.go
```diff
@@ -0,0 +1,156 @@
...
+		return false
+	}
+
+	for _, msg := range msgs {
+		msgTypeUrl := cosmostypes.MsgTypeURL(msg)
+		switch msgTypeUrl {
+		case claimMorseAcctMsgTypeUrl,
+			claimMorseAppMsgTypeUrl,
+			claimMorseSupplierMsgTypeUrl:
+			// check the remaining messages...
```
Olshansk, line 128: TODO? Which ones?

Olshansk (approved):
@bryanchriswhite Preliminary approval.

Pushed some edits and left some comments to tend to before merging.

app/ante.go
bryanchriswhite: No todo, this describes the `continue` branch's behavior.

app/ante.go
bryanchriswhite: Nope. It's not clear to me why you ask. :thinking: 

```suggestion
	// Intentionally not consuming any gas; no-op.
	return nil
```

bryanchriswhite:
:rotating_light: PLEASE **DO NOT DELETE THIS BRANCH** UNTIL DOWNSTREAM BRANCHE(S) HAVE BEEN REBASED - THANK YOU! :rotating_light:

Issue #1034, opened 2025-01-20
[Morse->Shannon Migration] Migration module


bryanchriswhite:
## Objective

Enable account (identity and balances) state migration from Morse to Shannon.

## Origin Document

https://www.notion.so/buildwithgrove/Technical-Migration-from-Morse-Shannon-173a36edfff6800689fad43fd15287cd

## Goals

- Enable on-chain commitment to Morse migration state
- Enable self-claiming of Morse account identity
- Enable self-claiming of Morse account balance(s)

## Deliverables

- [x] A PR which scaffolds a new "migration" module
  - #1032 
- [x] A PR which scaffolds a new `MorseClaimableAccount` on-chain persisted map
  - #1068 
- [x] A PR which adds a `poktrolld` CLI command to convert the Morse state export to the Shannon import.
   - #1039 
- [x] PR(s) which scaffold and implement a new ~~`MsgCreateMorseAccountState`~~ `MsgImportMorseClaimableAccounts` message which persists Morse claimable accounts on Shannon
  - ~~#1045~~
  - ~~#1047~~
  - #1072
- [x] PR(s) which scaffold and implement a new ~~`MsgClaimMorsePOKT`~~ `MsgClaimMorseAccount`
  - ~~#1046~~
  - #1075 
  - #1059
- [x] PR(s) which scaffold and implement a new `MsgClaimMorseApplication`
  - #1080
  - #1082
- [x] PR(s) which scaffold and implement a new `MsgClaimMorseSupplier`
  - #1094 
  - #1095
- [ ] PR(s) which implement the claiming CLI
  - #1105
  - #1148 
  - #1153
- [x] PR(s) which implement an E2E test that covers local morse state export through claiming
  - #1133
- [x] A PR which adds documentation
  - #1115

## Non-goals / Non-deliverables

- On-chain validation or Morse import state.
- Supporting applications staked for multiple services (can claim but must choose one).
- **Any** kind of stake/unstaked token redistribution as a result of the claiming.
  - Operators can update their actor stakes/configs on Morse **prior** to migration (i.e. "migration height").
  - Operators can update their actor stakes/configs on Shannon **after** claiming.

## General deliverables

- [ ] **Comments**: Add/update TODOs and comments alongside the source code so it is easier to follow.
- [ ] **Testing**: Add new tests (unit and/or E2E) to the test suite.
- [ ] **Makefile**: Add new targets to the Makefile to make the new functionality easier to use.
[...]

PR #1179, opened 2025-04-07
[Morse -> Shannon Migration] refactor: remove public key from `MorseAccountState`

bryanchriswhite:
## Summary

Remove unneeded `public_key` field from the onchain `MorseAccountState` data structure. This change has the following impacts:

- Significantly reduces the onchain size of the `MorseAccount`s which will have multiple performance benefits.
- Simplifies onchain Morse account/actor claiming business logic.
- Re-establishes the viability of using a pruned snapshot as the starting point for Morse state ETVL operations.

Changes:
- Remove `public_key` from `MorseAccountState` protobuf type
- Add `morse_public_key` to the Morse account/actor claim messages
- Refactor Morse claim messages constructors accordingly
  - Include public key
  - Compute address from public key (simplification)
- Refactor and consolidate Morse claim message validation
- Refactor Morse claim message handlers accordingly

## Issue

- Issue: #1034

## Type of change

- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Sanity Checklist

- [x] For code covered by @oneshot E2E tests, `make test_e2e_oneshot` passes on localnet


proto/pocket/migration/tx.proto
```diff
@@ -97,9 +97,12 @@ message MsgClaimMorseAccount {
   // E.g.: 00f9900606fa3d5c9179fc0c8513078a53a2073e
   string morse_src_address = 2 [(gogoproto.jsontag) = "morse_src_address"];
 
+  // The ed25519 public key of the morse account with morse_src_address.
+  bytes morse_public_key = 3 [(gogoproto.casttype) = "crypto/ed25519.PublicKey"];
+
   // The hex-encoded signature, by the Morse account, of this message (where this field is nil).
   // I.e.: morse_signature = private_key.sign(marshal(MsgClaimMorseAccount{morse_signature: nil, ...}))
-  bytes morse_signature = 3 [(gogoproto.jsontag) = "morse_signature"];
+  bytes morse_signature = 4 [(gogoproto.jsontag) = "morse_signature"];
```
bryanchriswhite, line 105: @okdas is this going to be an issue? We don't have to fix these if so.

proto/pocket/migration/tx.proto
```diff
@@ -131,17 +134,19 @@ message MsgClaimMorseApplication {
 
   // The hex-encoded address of the Morse account whose balance will be claimed.
   // E.g.: 00f9900606fa3d5c9179fc0c8513078a53a2073e
-
   string morse_src_address = 2 [(gogoproto.jsontag) = "morse_src_address"];
 
+  // The ed25519 public key of the morse account with morse_src_address.
+  bytes morse_public_key = 3 [(gogoproto.casttype) = "crypto/ed25519.PublicKey"];
```
bryanchriswhite, line 140: Rename to `morse_src_public_key`

proto/pocket/migration/tx.proto
Olshansk: _tl;dr We ned to do this by the book._

1. There's @red-0ne's alt approach to multiple protobufs
2. This is unlikely an issue because it's not being used
3. All of that said, the risk is not worth it. Please use `reserved` as we're officially in production.

Olshansk (changes requested):
@bryanchriswhite 

1. Leaving one comment: https://github.com/pokt-network/poktroll/pull/1179#discussion_r2032057199
2. Saw this is a draft but review was requested so am confused.

Requesting changes as a signal for you to re-request if it's already ready/.

bryanchriswhite:
> 2. Saw this is a draft but review was requested so am confused.
> 
> Requesting changes as a signal for you to re-request if it's already ready/.

Apologies for the false start. I reverted it back to a draft and will re-request when ready.

x/migration/types/morse_validation.go
```diff
@@ -0,0 +1,70 @@
...
+import (
+	"crypto/ed25519"
+
+	cmted25519 "github.com/cometbft/cometbft/crypto/ed25519"
+	cosmostypes "github.com/cosmos/cosmos-sdk/types"
+)
+
+// morseClaimMessage is an interface that all Morse account/actor claim messages
+// implement which allows for a non-generic implementations of common behavior.
+type morseClaimMessage interface {
```
Olshansk, line 12: A few notes about the design of this:
1. The functions here don't use the keeper
2. The functions do "basic validation"
3. We are introducing an interface for a proto msg

Why don't we follow the same structure we have elsewhere like `func (msg *MsgStakeApplication) ValidateBasic() error {`?

bryanchriswhite: > 1. The functions here don't use the keeper

Correct! Since the public key is now part of the Morse account/actor claim messages (and no longer persisted onchain)

> 2. The functions do "basic validation"

I would define this as "all validation which doesn't require onchain data"; which now applies to Morse signature verification.

> 3. We are introducing an interface for a proto msg

Indeed. Since there are multiple Morse account/actor claim messages, and this is such critical logic, consolidating it requires the use of an interface.

x/migration/types/message_claim_morse_supplier.go
```diff
@@ -38,65 +38,72 @@ func NewMsgClaimMorseSupplier(
...
+// - The supplier service configs are valid.
+// - The morsePublicKey is valid.
+// - The morseSrcAddress matches the public key.
+// - The morseSignature is valid.
 func (msg *MsgClaimMorseSupplier) ValidateBasic() error {
-	if len(msg.MorseSignature) != MorseSignatureLengthBytes {
-		return ErrMorseSupplierClaim.Wrapf(
-			"invalid morse signature length; expected %d, got %d",
-			MorseSignatureLengthBytes, len(msg.MorseSignature),
+	if _, err := sdk.AccAddressFromBech32(msg.GetShannonOwnerAddress()); err != nil {
```
Olshansk, line 48: Should we ensure `OwnerAddress != ""` like below?

x/migration/types/message_claim_morse_application.go
```diff
@@ -36,59 +36,62 @@ func NewMsgClaimMorseApplication(
...
+// - The morsePublicKey is valid.
+// - The morseSrcAddress matches the public key.
+// - The morseSignature is valid.
 func (msg *MsgClaimMorseApplication) ValidateBasic() error {
-	if len(msg.MorseSignature) != MorseSignatureLengthBytes {
-		return ErrMorseApplicationClaim.Wrapf(
-			"invalid morse signature length; expected %d, got %d",
-			MorseSignatureLengthBytes, len(msg.MorseSignature),
+	// Validate the shannonDestAddress is a valid bech32 address.
+	if _, err := sdk.AccAddressFromBech32(msg.GetShannonDestAddress()); err != nil {
```
Olshansk, line 46: Should we check if its empty first?

x/migration/types/morse_validation.go
Olshansk: Update.

I saw how this is used.

Do you see an opportunity to:
1. Enforce a certain type that implements this interface so it's explicit where it's used?
2. #PUC on which messages its used for.

x/migration/types/morse_validation.go
```diff
@@ -0,0 +1,70 @@
+package types
```
Olshansk, line 1: Consider renaming file to `morse_claim_message.go` so its clearer.

x/migration/types/morse_validation.go
```diff
@@ -0,0 +1,70 @@
...
+			msg.GetMorseSrcAddress(),
+			publicKey.Address().String(),
+		)
+	}
+	return nil
+}
+
+// validateMorseSignature validates the morseSignature of the given morseClaimMessage.
+func validateMorseSignature(msg morseClaimMessage) error {
+	if len(msg.GetMorseSignature()) != MorseSignatureLengthBytes {
```
Olshansk, line 49: Can we move `MorseSignatureLengthBytes` here?

proto/pocket/migration/tx.proto
```diff
@@ -131,9 +134,11 @@ message MsgClaimMorseApplication {
 
   // The hex-encoded address of the Morse account whose balance will be claimed.
   // E.g.: 00f9900606fa3d5c9179fc0c8513078a53a2073e
-
   string morse_src_address = 2 [(gogoproto.jsontag) = "morse_src_address"];
 
+  // The ed25519 public key of the morse account with morse_src_address.
+  bytes morse_public_key = 5 [(gogoproto.casttype) = "crypto/ed25519.PublicKey"];
```
Olshansk, line 140: Why do we have `morse_public_key` AND `morse_src_address` if one can be derived from the other?

Feels like we have "two sources of truth"

x/migration/types/morse_validation.go
```diff
@@ -0,0 +1,70 @@
...
+	GetMorsePublicKey() ed25519.PublicKey
+	GetMorseSrcAddress() string
+	GetMorseSignature() []byte
+	ValidateMorseAddress() error
+	ValidateMorseSignature() error
+}
+
+// validateMorseAddress validates that the morseSrcAddress matches
+// the Morse public key of the given Morse claim message.
+func validateMorseAddress(msg morseClaimMessage) error {
```
Olshansk, line 26: I left a comment in `proto/pocket/migration/tx.proto`, but don't fully understand why this is necessary.

proto/pocket/migration/tx.proto
```diff
@@ -187,6 +192,9 @@ message MsgClaimMorseSupplier {
   // See: https://dev.poktroll.com/operate/configs/supplier_staking_config#staking-types.
   string shannon_operator_address = 2 [(cosmos_proto.scalar) = "cosmos.AddressString", (gogoproto.jsontag) = "shannon_operator_address"];
 
+  // The ed25519 public key of the morse account with morse_src_address.
+  bytes morse_public_key = 6 [(gogoproto.casttype) = "crypto/ed25519.PublicKey"];
```
Olshansk, line 196: Can we remove `morse_src_address` if we have this?

proto/pocket/migration/tx.proto
```diff
@@ -187,6 +192,9 @@ message MsgClaimMorseSupplier {
   // See: https://dev.poktroll.com/operate/configs/supplier_staking_config#staking-types.
```
Olshansk, line 192: If we're not ordering the indecies from top to bottom, I highly recommend following the `Next Free Index` comment pattern.

Screenshot 2025-04-10 at 1 35 30 PM

x/migration/module/cmd/collect_morse_accounts.go
```diff
@@ -231,26 +232,44 @@ func collectInputApplicationStakes(inputState *migrationtypes.MorseStateExport,
...
-
-		appStakeAmtUpokt, ok := cosmosmath.NewIntFromString(exportApplication.StakedTokens)
-		if !ok {
-			return ErrMorseExportState.Wrapf("failed to parse application stake amount %q", exportApplication.StakedTokens)
+			logger.Logger.Warn().
+				Str("app_address", appAddr).
+				Msg("no account found for application")
+
+			// DEV_NOTE: If no auth account was found for this application, create a new one.
+			newAppAccount := &migrationtypes.MorseAuthAccount{
```
Olshansk, line 241: `s/newAppAccount/newAppAuthAccount/g` ?

x/migration/module/cmd/collect_morse_accounts.go
```diff
@@ -231,26 +232,44 @@ func collectInputApplicationStakes(inputState *migrationtypes.MorseStateExport,
...
+				)
+			}
 		}
 
-		if err := morseWorkspace.addAppStake(appAddr, appStakeAmtUpokt); err != nil {
-			return fmt.Errorf(
-				"adding application stake amount to account balance of address %q: %w",
-				appAddr, err,
-			)
+		if exportApplication.StakedTokens != "" {
```
Olshansk, line 256: 1. Is there ever a situation where this could be empty? If so, #PUC
2. If not, should we log a warning, error or other?

x/migration/module/cmd/collect_morse_accounts.go
```diff
@@ -270,9 +289,25 @@ func collectInputSupplierStakes(inputState *migrationtypes.MorseStateExport, mor
...
 
-		// DEV_NOTE: An account SHOULD exist for each actor.
 		if !morseWorkspace.hasAccount(supplierAddr) {
-			return ErrMorseExportState.Wrapf("account not found corresponding to supplier with address %q", supplierAddr)
+			logger.Logger.Warn().
+				Str("supplier_address", supplierAddr).
+				Msg("no account found for supplier")
+
+			// DEV_NOTE: If no auth account was found for this supplier, create a new one.
+			newSupplierAccount := &migrationtypes.MorseAuthAccount{
```
Olshansk, line 298: Ditto comment above.

x/migration/module/cmd/collect_morse_accounts.go
```diff
@@ -270,9 +289,25 @@ func collectInputSupplierStakes(inputState *migrationtypes.MorseStateExport, mor
...
+					Address: exportSupplier.Address,
+					Coins:   []cosmostypes.Coin{},
+				},
+			}
+			if _, _, err := morseWorkspace.addAccount(supplierAddr, newSupplierAccount); err != nil {
+				return fmt.Errorf(
+					"adding supplier account to account balance of address %q: %w",
+					supplierAddr, err,
+				)
+			}
```
Olshansk, line 310: Why don't we have the != "" check for staked tokens for a supplier?

pkg/deps/config/client.go
```diff
@@ -0,0 +1,81 @@
+package config
```
Olshansk, line 1: What's the motivation between `client.go` as the name for this file?

Can you think of something better? E.g. even `tx_client.go` would be more detail.

pkg/deps/config/client.go
```diff
@@ -0,0 +1,81 @@
...
+	"fmt"
+
+	"github.com/cosmos/cosmos-sdk/client/flags"
+	"github.com/cosmos/cosmos-sdk/types"
+	"github.com/spf13/cobra"
+
+	"github.com/pokt-network/poktroll/app/volatile"
+	"github.com/pokt-network/poktroll/pkg/client"
+	"github.com/pokt-network/poktroll/pkg/client/tx"
+)
```
Olshansk, line 13: I know @red-0ne had some recent related changes.

Wanted to make sure we're not duplicating logic?

pkg/deps/config/client.go
```diff
@@ -0,0 +1,81 @@
...
+
+// GetTxClientGasAndFeesOptions returns a slice of TxClientOptions which
```
Olshansk, line 15: Can you #PUC with a `why` explaining what happens in option 1 vs option 2 and when?

pkg/deps/config/client.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+
+	gasAdjustment, err := cmd.Flags().GetFloat64(flags.FlagGasAdjustment)
+	if err != nil {
+		return nil, err
+	}
+
+	// The RelayMiner always uses tx simulation to estimate the gas since this
+	// will be variable depending on the tx being sent.
+	// Always use the "auto" gas setting for the RelayMiner.
```
Olshansk, line 61: Why are we calling out the RelayMiner here?

pkg/deps/config/client.go
```diff
@@ -0,0 +1,81 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	gasAdjustment, err := cmd.Flags().GetFloat64(flags.FlagGasAdjustment)
+	if err != nil {
+		return nil, err
+	}
+
+	// The RelayMiner always uses tx simulation to estimate the gas since this
```
Olshansk, line 59: Where does the simulation come in?

pkg/deps/config/client.go
```diff
@@ -0,0 +1,81 @@
...
+	// Always use the "auto" gas setting for the RelayMiner.
+	gasSetting, err := flags.ParseGasSetting("auto")
+	if err != nil {
+		return nil, err
+	}
+
+	// Ensure that the gas prices include upokt
+	for _, gasPrice := range gasPrices {
+		if gasPrice.Denom != volatile.DenomuPOKT {
+			// TODO_TECHDEBT(red-0ne): Allow other gas prices denominations once supported (e.g. mPOKT, POKT)
```
Olshansk, line 70: Can we remove this?

Olshansk (changes requested):
Leaving a handful of comments alongside the changes I pushed: [`d71cf98` (#1179)](https://github.com/pokt-network/poktroll/pull/1179/commits/d71cf98e480d2017845b1e8b4221c9b87e58c077)

cmd/flags/tx_client.go
```diff
@@ -63,6 +67,21 @@ func GetTxClient(ctx context.Context, cmd *cobra.Command) (client.TxClient, erro
 	}
 	deps = depinject.Configs(deps, depinject.Supply(txCtx))
 
+	// Prepare default tx client options.
+	gasAndFeesOptions, err := config.GetTxClientGasAndFeesOptions(cmd)
+	if err != nil {
+		return nil, err
+	}
+
+	defaultTxClientOpts := append(
```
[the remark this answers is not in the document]
bryanchriswhite: Nice - much better name! :sunglasses:

proto/pocket/migration/tx.proto
bryanchriswhite: I chose to leave the address in the messages to keep the transparency in onchain events. While we can derive the address from the public key in our custom events, there are many built-in events that cosmos-sdk will generate which would lose the Morse address as a result.

x/migration/types/morse_validation.go
bryanchriswhite: Given, https://github.com/pokt-network/poktroll/pull/1179#discussion_r2039511580, we need to validate this.

proto/pocket/migration/tx.proto
bryanchriswhite: See https://github.com/pokt-network/poktroll/pull/1179#discussion_r2039511580

x/migration/types/message_claim_morse_supplier.go
bryanchriswhite: No, in fact, the check below shouldn't be there either.

x/migration/types/message_claim_morse_application.go
bryanchriswhite: This will catch both empty and invalid addresses with reasonable error messages. I think it's a bit redundant to add a check for the empty case.

proto/pocket/migration/tx.proto
bryanchriswhite: This is something that I would argue we should automate. I'm happy to do it and/or open an issue for it in the meantime. I think it should be pretty straightforward given that we've already done this sort of thing previously for ensuring the presence of the `stable_marshaler_all` protobuf option on all protobufs (see: [tools/scripts/protocheck/cmd](https://github.com/pokt-network/poktroll/blob/main/tools/scripts/protocheck/cmd/unstable.go)).

x/migration/module/cmd/collect_morse_accounts.go
bryanchriswhite: > 1. Is there ever a situation where this could be empty? If so, #PUC

It depends on Morse protocol details, so I'm not sure and haven't looked at the code to answer this question yet. Perhaps adding a warning log and running a recent snapshot would be sufficient to answer this question for practical purposes.

> 2. If not, should we log a warning, error or other?

I think this case is going to be problematic because of the constraints that we added with respect to account/actor claiming. I.e. unstaked, application, and supplier accounts can ONLY be claimed using their respective claim message, which currently precludes the possibility of an account which is staked as multiple actors.

Wdyt should be the correct behavior here?

x/migration/module/cmd/collect_morse_accounts.go
bryanchriswhite: This, I can confirm, definitely does happen in snapshot data. It seems like there are several Morse applications and suppliers which have no corresponding auth account; i.e. no unstaked balances. As of the current state of this branch, these accounts would be eligible to claim on Shannon as their respective actor type, but they will not have any unstaked balance. For practical purposes, this means that they will require intervention from an external account on Shannon in order to do anything (unstake, restake, etc.); however, this may also be the case for these accounts on Morse as well.

x/migration/module/cmd/collect_morse_accounts.go
bryanchriswhite: Nice catch - thank you!

pkg/deps/config/client.go
bryanchriswhite: `tx_client_options.go`

pkg/deps/config/client.go
bryanchriswhite: This derived from what @red-0ne wrote, it's mostly an extraction in the name of reusability (which is needed, not just nice to have, in this case).

proto/pocket/migration/morse_onchain.proto
```diff
@@ -29,8 +29,12 @@ message MorseClaimableAccount {
   // The hex-encoded address of the Morse account whose balance will be claimed.
   string morse_src_address = 2 [(gogoproto.jsontag) = "morse_src_address"];
 
+  // This field was momentarily used to hold the Morse public key; however, an
+  // optimization eliminates the need for this. Reserved for backwards compatibility.
+  //
   // The ed25519 public key of the account.
-  bytes public_key = 4 [(gogoproto.casttype) = "crypto/ed25519.PublicKey"];
+  // bytes public_key = 4 [(gogoproto.casttype) = "crypto/ed25519.PublicKey"];
+  reserved 4;
```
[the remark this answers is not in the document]
bryanchriswhite: I would like to keep it. :sweat_smile:

proto/pocket/migration/tx.proto
```diff
@@ -176,6 +193,8 @@ message MsgClaimMorseApplicationResponse {
 
   // The application which was staked as a result of the claim.
   application.Application application = 5 [(gogoproto.jsontag) = "application"];
+
+  // Next free index: 6
```
red-0ne, line 197: I'm not a big fan of out of order indexes and relying on comments to identify the next one. I believe it's a bad dx.

We're already having them out of order before even having Morse claiming records onchain.

I'm in favor of fully opting for proto migration if it's necessary.

proto/pocket/migration/tx.proto
bryanchriswhite: - #1193

red-0ne (commented):
Left a few messages but no major blocker.

We're getting close to merging :+1:

pkg/deps/config/tx_client_options.go
```diff
@@ -59,7 +78,7 @@ func GetTxClientGasAndFeesOptions(cmd *cobra.Command) ([]client.TxClientOption,
 	// The RelayMiner always uses tx simulation to estimate the gas since this
 	// will be variable depending on the tx being sent.
 	// Always use the "auto" gas setting for the RelayMiner.
```
bryanchriswhite, line 80:
```suggestion
```

proto/pocket/migration/tx.proto
bryanchriswhite: Can we move this discussion to an issue or something? I don't think it should block this PR.

proto/pocket/migration/tx.proto
Olshansk: 1. Can you #PUC for each `morse_public_key` with that explanation?
2. Add that explanation next to the validation function as well.
3. Please bias to bullet points.

Olshansk: > there are many built-in events that cosmos-sdk

Can you provide an example?

e2e/tests/update_params_test.go
```diff
@@ -15,14 +15,14 @@ import (
...
 
-	"github.com/pokt-network/poktroll/api/pocket/application"
-	"github.com/pokt-network/poktroll/api/pocket/gateway"
-	"github.com/pokt-network/poktroll/api/pocket/proof"
-	"github.com/pokt-network/poktroll/api/pocket/service"
-	"github.com/pokt-network/poktroll/api/pocket/session"
-	"github.com/pokt-network/poktroll/api/pocket/shared"
-	"github.com/pokt-network/poktroll/api/pocket/supplier"
-	"github.com/pokt-network/poktroll/api/pocket/tokenomics"
+///	"github.com/pokt-network/poktroll/api/pocket/application"
```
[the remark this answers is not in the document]
Olshansk: Please remove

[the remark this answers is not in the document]
bryanchriswhite: Thanks for catching this - some automate process did this. I'm not sure why. :thinking:

pkg/deps/config/tx_client_options.go
```diff
@@ -0,0 +1,101 @@
...
+		return nil, err
+	}
+	gasAdjustment, err := cmd.Flags().GetFloat64(flags.FlagGasAdjustment)
+	if err != nil {
+		return nil, err
+	}
+
+	// The RelayMiner always uses tx simulation to estimate the gas since this
+	// will be variable depending on the tx being sent.
+	// Always use the "auto" gas setting for the RelayMiner.
```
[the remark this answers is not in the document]
bryanchriswhite: Meant to delete. :sweat_smile:

proto/pocket/migration/tx.proto
Olshansk: After we ship everything, we're going to have a lot of time to automate things. As long as it's after we ship everything, no objections.

Screenshot 2025-04-11 at 9 49 31 AM

proto/pocket/migration/tx.proto
red-0ne: That's what we should do.

Marking the issue as resolved.

proto/pocket/migration/tx.proto
```diff
@@ -94,7 +100,7 @@ message MsgClaimMorseAccount {
   // This account is liable for any fees incurred by violating the constraints of Morse
   // account/actor claim message fee waiving; the tx contains ONE OR MORE Morse account/actor
   // claim messages AND has EXACTLY ONE signer.
-  string shannon_signing_address = 4 [(cosmos_proto.scalar) = "cosmos.AddressString", (gogoproto.jsontag) = "shannon_signing_address"];
+  string shannon_signing_address = 5 [(cosmos_proto.scalar) = "cosmos.AddressString", (gogoproto.jsontag) = "shannon_signing_address"];
```
Olshansk, line 103: We should not be changing the index of an existing type/field.

proto/pocket/migration/tx.proto
```diff
@@ -135,17 +148,19 @@ message MsgClaimMorseApplication {
   // This account is liable for any fees incurred by violating the constraints of Morse
   // account/actor claim message fee waiving; the tx contains ONE OR MORE Morse account/actor
   // claim messages AND has EXACTLY ONE signer.
-  string shannon_signing_address = 5 [(cosmos_proto.scalar) = "cosmos.AddressString", (gogoproto.jsontag) = "shannon_signing_address"];
+  string shannon_signing_address = 6 [(cosmos_proto.scalar) = "cosmos.AddressString", (gogoproto.jsontag) = "shannon_signing_address"];
```
Olshansk, line 151: Ditto

proto/pocket/migration/tx.proto
Olshansk: Moving conversation to discord.

x/migration/module/cmd/collect_morse_accounts.go
Olshansk: So in some cases you can :
- Definitely confirm something does / does not happen
- Not sure if/why it does / does not happen

My ask is to add a TODO_MAINNET_MIGRATION in both cases with the detail and #PUC.

x/migration/module/cmd/collect_morse_accounts.go
```diff
@@ -231,26 +232,48 @@ func collectInputApplicationStakes(inputState *migrationtypes.MorseStateExport,
...
+					appAddr, err,
+				)
+			}
+
+			morseWorkspace.accumulatedTotalAppStake = morseWorkspace.accumulatedTotalAppStake.Add(appStakeAmtUpokt)
+			morseWorkspace.numApplications++
+		} else {
+			logger.Logger.Warn().
+				Str("app_address", appAddr).
+				Msg("account stakes as BOTH an application AND a supplier; this account will be unclaimable")
```
Olshansk, line 274: Can you #PUC error messages OR add a comment explaining why `StakedTokens != ""` is a result of staking both application AND a supplier?

It's not very clear to me.

Ditto below.

x/migration/module/cmd/collect_morse_accounts.go
Olshansk: #PUC with a TODO_MAINNET_MIGRATION explaining what leads to this and we'll figure it out later.

Olshansk (changes requested):
@bryanchriswhite Left a few comments through. We can follow up here or on discord.

proto/pocket/migration/tx.proto
bryanchriswhite: I think there may have been some reversion during a merge conflict resolution. :thinking:

proto/pocket/migration/morse_offchain.proto
```diff
@@ -102,7 +102,7 @@ message MorseApplication {
   bool   jailed = 3[(gogoproto.jsontag) = "jailed", (gogoproto.moretags) = "yaml:\"jailed\""];
   int32  status = 4 [(gogoproto.jsontag) = "status", (gogoproto.moretags) = "yaml:\"status\""];
   // The string representation of the BigInt amount of upokt.
-  string staked_tokens = 6 [(gogoproto.jsontag) = "tokens"];
+  string staked_tokens = 6 [(gogoproto.jsontag) = "staked_tokens"];
```
bryanchriswhite, line 105: This fixes a bug in collect morse accounts

bryanchriswhite:
`make test_e2e_migration_fixture` confirmed passing locally.

Olshansk (approved):
Love how much code we deleted in the last one!

PR #1189, opened 2025-04-08
[WIP][Docs] Some documentation updates while reviewing PATH docs

Olshansk:
Documentation updates made while reviewing https://github.com/buildwithgrove/path/pull/189

## Type of change

- [ ] Bug fix
- [ ] Code health or cleanup
- [x] Documentation
- [ ] Other (specify)


docusaurus/docs/operate/faq/1_account_faq.md
```diff
@@ -0,0 +1,45 @@
+---
+sidebar_position: 2
+title: Account FAQ
+---
+
+## Why am I getting an `account sequence mismatch` error when trying to submit a `bank send` transaction?
+
+When trying to submit a `bank send` transaction like so:
```
red-0ne, line 8: This usually happens when you try to send multiple transactions with the same account on the same block.

Olshansk: ACK your statement but need a more actionable comment.

- Do you suggest we update the message?
- Remove this section?

Sending multiple transactions with the same account on the same block is going to be a VERY common operation by MANY users so we have to get ahead of it.

Issue #1206, opened 2025-04-15
[Loclanet] Remove cluster name checks - Add notes on network realy readiness


RawthiL:
## Objective

During `make localnet_up` the local cluster is expected to be named `kind-kind`, if this is not true, the process fails. 

The documentation is not clear that after doing `make localnet_up` the network will be ready to send relays. 

## Origin Document

Cluster Name: Cluster naming should not be mandatory for all devs. Localnet should be deployed in any cluster name.

Localnet readiness: Despite the general health of the localnet and that it is producing blocks, there is mandatory step: `make acc_initialize_pubkeys`. This is required regardless you are following the full guide or just need parts of it running.

## Goals

- Remove `kind-kind` cluster name requirement
- Add note stating that `make acc_initialize_pubkeys` is mandatory for relaying near the `make localnet_up` command.

**Creator**: @RawthiL

PR #1207, opened 2025-04-15
[Loclanet] Remove cluster name checks - Add notes on network realy readiness

RawthiL:
## Summary

PR solving 
https://github.com/pokt-network/poktroll/issues/1206
(copied below)

## Objective

During make localnet_up the local cluster is expected to be named kind-kind, if this is not true, the process fails.

The documentation is not clear that after doing make localnet_up the network will be ready to send relays.

## Origin Document

Cluster Name: Cluster naming should not be mandatory for all devs. Localnet should be deployed in any cluster name.

Localnet readiness: Despite the general health of the localnet and that it is producing blocks, there is mandatory step: make acc_initialize_pubkeys. This is required regardless you are following the full guide or just need parts of it running.

## Goals
- Remove kind-kind cluster name requirement
- Add note stating that make acc_initialize_pubkeys is mandatory for relaying near the make localnet_up command.

### DIFF
--- a/docusaurus/docs/develop/developer_guide/walkthrough.md
+++ b/docusaurus/docs/develop/developer_guide/walkthrough.md
@@ -164,7 +164,7 @@ Create a new `k8s` cluster for your LocalNet:
 kind create cluster
 ```
 
-If you use `k8s` for other things, you may need to switch your context as well:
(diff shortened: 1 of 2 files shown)


Olshansk (approved):
Ty!

PR #1204, opened 2025-04-15
[Upgrade] Upgrade transactions for `v.0.1.2`

Olshansk:
Validated like so:

Generated like so:

```bash
 ./tools/scripts/upgrades/prepare_upgrade_tx.sh v0.1.2
Downloading checksum file from https://github.com/pokt-network/poktroll/releases/download/v0.1.2/release_checksum...
Created tools/scripts/upgrades/upgrade_tx_v0.1.2_alpha.json
Created tools/scripts/upgrades/upgrade_tx_v0.1.2_beta.json
Created tools/scripts/upgrades/upgrade_tx_v0.1.2_local.json
Created tools/scripts/upgrades/upgrade_tx_v0.1.2_main.json
All upgrade transaction files created successfully.
```

```bash
jq -r '.body.messages[0].plan.info | fromjson | .binaries[]' tools/scripts/upgrades/upgrade_tx_v0.1.2_main.json | while IFS= read -r url; do
  go-getter "$url" .
done
2025/04/15 14:56:07 success!
2025/04/15 14:56:11 success!
2025/04/15 14:56:15 success!
2025/04/15 14:56:19 success!
```

Ref: https://github.com/pokt-network/poktroll/releases/tag/v0.1.2

PR #1213, opened 2025-04-15
[QuickPR] skip checksum in upgrade TXs

okdas:
https://github.com/pokt-network/poktroll/pull/1204/files#r2045636235

image


Olshansk (commented):
Trusting this works.

PR #1226, opened 2025-04-18
[Perf] Implement Memory-Efficient Iterator Pattern for Large Data Sets

red-0ne:
## Summary

Replace direct access to bulk collections with iterators to optimize memory usage and performance when dealing with large data sets.

### Primary Changes:
- Implemented RecordIterator interface in shared types for efficient iteration
- Replaced GetAllSuppliers with GetAllSuppliersIterator to avoid loading all suppliers in memory
- Replaced ClaimsIterator with RecordIterator for retrieving expiring claims
- Updated tokenomics module to use the new iterator pattern

### Secondary changes:
- Added proper error handling for iterator values
- Fixed supplier processing in EndBlockers to use iterators
- Added unit tests for the new iterator functionality
- Updated supplier keeper mocks in tests to support iterator pattern

## Issue

image
image

## Type of change

- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Sanity Checklist

- [ ] For code covered by @oneshot E2E tests, `make test_e2e_oneshot` passes on localnet


x/supplier/keeper/prune_supplier_service_config_history.go
```diff
@@ -64,7 +77,7 @@ func (k Keeper) EndBlockerPruneSupplierServiceConfigHistory(
 		// Update the supplier's service config history with the pruned list.
 		supplier.ServiceConfigHistory = retainedServiceConfigs
 
-		k.SetSupplier(ctx, supplier)
+		k.SetSupplier(ctx, *supplier)
```
Olshansk, line 80: Two lines above this, we are modifying a field from the pointer.

I don't think it should be an issue, but it does raise a flag that we should potentially make a copy of the supplier at the very top.

Avoid potential value/pointer concurrency issues.

x/supplier/keeper/prune_supplier_service_config_history.go
```diff
@@ -25,7 +25,20 @@ func (k Keeper) EndBlockerPruneSupplierServiceConfigHistory(
...
+		supplier, err := allSuppliersIterator.Value()
+		if err != nil {
+			logger.Error(fmt.Sprintf("could not get supplier from iterator: %v", err))
+			return 0, err
+		}
+		if supplier == nil {
+			logger.Error(fmt.Sprintf("unexpected nil supplier in iterator at %s", allSuppliersIterator.Key()))
+			continue
+		}
+
```
Olshansk, line 41:
```suggestion
		supplierValue, err := allSuppliersIterator.Value()
		if err != nil {
			logger.Error(fmt.Sprintf("could not get supplier from iterator: %v", err))
			return 0, err
		}
		if supplierValue == nil {
			logger.Error(fmt.Sprintf("unexpected nil supplier in iterator at %s", allSuppliersIterator.Key()))
			continue
		}
		// Make a copy of the supplier to avoid pointer related concurrency issues
		supplier := *supplierValue

```

Olshansk: See my comment below for reference

x/supplier/keeper/activate_services.go
```diff
@@ -37,7 +37,21 @@ func (k Keeper) BeginBlockerActivateSupplierServices(
...
+		supplier, err := allSuppliersIterator.Value()
+		if err != nil {
+			logger.Error(fmt.Sprintf("could not get supplier from iterator: %v", err))
+			return 0, err
+		}
+
+		if supplier == nil {
+			logger.Error(fmt.Sprintf("unexpected nil supplier in iterator at %s", allSuppliersIterator.Key()))
+			continue
+		}
```
Olshansk, line 53: Ditto my comment in x/supplier/keeper/prune_supplier_service_config_history.go

x/session/keeper/session_hydrator.go
```diff
@@ -180,27 +180,38 @@ func (k Keeper) hydrateSessionSuppliers(ctx context.Context, sh *sessionHydrator
...
+	for ; allSuppliersIterator.Valid(); allSuppliersIterator.Next() {
+		supplier, err := allSuppliersIterator.Value()
+		if err != nil {
+			logger.Error(fmt.Sprintf("could not get supplier from iterator: %v", err))
+			return err
+		}
+		if supplier == nil {
+			logger.Error(fmt.Sprintf("unexpected nil supplier in iterator at %s", allSuppliersIterator.Key()))
+			continue
+		}
```
Olshansk, line 205: Ditto my comment in x/supplier/keeper/prune_supplier_service_config_history.go

x/tokenomics/keeper/settle_pending_claims.go
```diff
@@ -47,7 +47,10 @@ func (k Keeper) SettlePendingClaims(ctx cosmostypes.Context) (
 	logger.Debug("settling expiring claims")
 	numExpiringClaims := 0
 	for ; expiringClaimsIterator.Valid(); expiringClaimsIterator.Next() {
-		claim := expiringClaimsIterator.Value()
+		claim, iterErr := expiringClaimsIterator.Value()
+		if iterErr != nil {
+			return settledResults, expiredResults, iterErr
+		}
 		numExpiringClaims++
```
Olshansk, line 54: Do we not need to check if `claim` is nil, just in case?

x/shared/types/record_iterator.go
```diff
@@ -0,0 +1,96 @@
...
+// It combines:
+// - A low-level Cosmos SDK store iterator
+// - A function to convert raw bytes into typed objects
+type recordIterator[T any] struct {
+	storeIter         storetypes.Iterator
+	deserializeRecord DataRecordAccessor[T]
+}
+
+// Next advances the iterator to the next position.
+func (ri *recordIterator[T]) Next() {
```
Olshansk, line 52: I'm beginning to think that we could (should?) update `Next` like so:

```go
// Next advances the iterator to the next non-nil position.
// It keeps advancing until it finds a valid record or reaches the end.
func (ri *recordIterator[T]) Next() {
    var zero T
    
    // First, advance to the next position
    ri.storeIter.Next()
    
    // Continue advancing until we find a non-nil value or reach the end
    for ri.storeIter.Valid() {
        // Attempt to get the current record
        rawBytes := ri.storeIter.Value()
        value, err := ri.deserializeRecord(rawBytes)
        
        // If we got a valid value (not zero/nil and no error), we're done
        if err == nil && value != zero {
            return
        }
        
        // Otherwise, move to the next position and try again
        ri.storeIter.Next()
    }
}
```

Wdyt?

x/shared/types/record_iterator.go
```diff
@@ -0,0 +1,96 @@
...
+	ri.storeIter.Next()
+}
+
+// Value returns the current record.
+// Process:
+// 1. Gets raw bytes from the store iterator
+// 2. Deserializes bytes into a typed object using the deserializer function
+func (ri *recordIterator[T]) Value() (T, error) {
```
Olshansk, line 60: If we do the above, we can modify this function to ALWAYS return a value and skip nils altogether.

Potentialy, we could have :

`.Next()`
`.NextNonNil()`
`.Pointer()`
.`.Value()`

To keep everything open to the user.

x/supplier/keeper/unbond_suppliers.go
```diff
@@ -26,13 +26,25 @@ func (k Keeper) EndBlockerUnbondSuppliers(ctx context.Context) (numUnbondedSuppl
...
+	for ; allSuppliersIterator.Valid(); allSuppliersIterator.Next() {
+		supplier, err := allSuppliersIterator.Value()
+		if err != nil {
+			logger.Error(fmt.Sprintf("could not get supplier from iterator: %v", err))
+			return numUnbondedSuppliers, err
+		}
+		if supplier == nil {
+			logger.Error(fmt.Sprintf("unexpected nil supplier in iterator at %s", allSuppliersIterator.Key()))
+			continue
+		}
```
Olshansk, line 40: Ditto

x/supplier/keeper/supplier_test.go
```diff
@@ -100,6 +100,24 @@ func TestSupplier_GetAll(t *testing.T) {
 	)
 }
 
+func TestSupplier_GetAllSuppliersIterator(t *testing.T) {
```
Olshansk, line 103: Can you add a similar test for claims please?

Olshansk (changes requested):
@red-0ne I pushed some changes myself but also leaving comments to get your feedback.

x/supplier/keeper/prune_supplier_service_config_history.go
red-0ne: Will make the iterator return a value instead of a pointer. It will actually be closer to what `GetSupplier/GetAllSuppliers` return

x/shared/types/record_iterator.go
red-0ne: I agree—returning values instead of pointers keeps the change minimal and matches the previous behaviour.

I’d also like to stay as close as possible to store.Iterator, letting DataRecordAccessor[T] decide whether the caller gets a value or a copy. That keeps the iterator thin, low‑maintenance, and familiar, while still giving consumers flexibility.

I considered adding a DataRecordFilter option, but I think that’s outside the scope of this PR. I’ll leave a TODO_CONSIDERATION for it.

x/shared/types/record_iterator.go
red-0ne: If we go this route, I'd like to add a `DataRecordFilter` which would be handy on the data indexing PR.

I think it's outside the scope of this PR, but would love to read your take one it.

x/tokenomics/keeper/settle_pending_claims.go
red-0ne: As mentioned in other comments, I will revert to return values instead of pointers.

A nil positive check will trigger an error as we should not expect a nil claim stored onchain.

x/shared/types/record_iterator.go
Olshansk: #PUC w/ a TODO + explanation + link to this thread.

x/shared/types/record_iterator.go
Olshansk: #PUC w/ a TODO + explanation + link to this thread.

Olshansk (approved):
Approving but please add the appropriate TODOs

PR #1228, opened 2025-04-18
fix: add upgrade flag to install script

commoddity:
## 🌿 Summary

Improve installation flexibility by adding an upgrade flag to the install script and removing the strict requirement for `pocketd` to be installed via Go.

### 🌱 Primary Changes:
- Added `--upgrade` flag to `install.sh` to allow force reinstallation of the latest `pocketd` version
- Removed requirement for `pocketd` to be installed in a specific Go bin location, now only checks PATH availability

### 🍃 Secondary changes:
- Updated installation messages to be more informative about upgrade options
- Made `checks.mk` executable (mode change from 644 to 755)

## 🛠️ Type of change

- [ ] Code health or cleanup

## 🤯 Sanity Checklist

- [ ] I have updated the GitHub Issue 'assignees', 'reviewers', 'labels', 'project', 'iteration' and 'milestone'


scripts/install.sh
```diff
@@ -1,9 +1,22 @@
 #!/usr/bin/env bash
 
 # This script installs the pocketd binary if not already installed.
+# Use with -u or --upgrade flag to force reinstallation of the latest version.
```
Olshansk, line 4:
```suggestion
# Use with -u or --upgrade flag to force reinstallation of the latest version by removing the existing binary first.
```

Olshansk (approved):

Issue #1236, opened 2025-04-22
[Morse -> Shannon Migration] Morse Account Claimer Token


bryanchriswhite:
## Objective

Unblock Morse account claiming by working around the fact that cosmos-sdk requires an onchain account state for a given account **prior** to the attempted validation a transaction signed by that account.

### Background

Given:

1. The need to generate new cryptographic keys (see [“Why can’t this be done in Shannon genesis”](https://dev.poktroll.com/operate/morse_migration/introduction#why-cant-this-be-done-in-shannon-genesis))
2. Cosmos-sdk **doesn’t support** transactions signed by accounts which have no corresponding on-chain state (`auth` module sequence number)

In order to claim a Morse account to any given Shannon account, that Shannon account must first be “created” onchain. The simplest way to do this is using existing and basic functionality of cosmos-sdk: **to send token(s) to the new account address.**

### Proposal

1. **New token (MACT)**

    Create a new “MACT” (Morse Account Claimer Token) token, which is intended specifically and only for this purpose: to facilitate the creation of onchain account state for new Shannon accounts which will be used to claim Morse accounts.

2. **Mainnet MACT faucet**

    In order for Morse account claimers to “fund” their new Shannon accounts with MACT, there MUST be a centralized but permissionless source of it in Mainnet. To this end, existing cosmos-sdk faucet tooling may be applied (reused). In addition to the deployment and maintenence of this faucet infrastructure, genesis will need to include distribution(s) of MACT to (at least) PNF.

    It should only be necessary for each Shannon account to only ever receive 1 MACT (i.e. the amount doesn’t matter) as this is sufficient for creating the onchain account sequence number required for subsequent Morse account claiming.

### Design Constraint/Effect Causal Flowchart

```mermaid
flowchart

nka(new key algo):::constraint
nacc(**user** MUST generate new account):::effect
nasn(no initial onchain account sequence number):::constraint
noregen(can't use regensis):::constraint

claim(MUST claim Morse accounts):::effect

nka --> nacc
nacc -->|user-initiated == cannot be predicted/pre-computed| noregen

noregen --> claim
noregen --> nasn

cacc("MUST 'create' onchain account (additional Tx)"):::effect

nasn --> cacc

nt("new 'Morse Account Claimer Token' (MACT)")
f(MACT faucet)

cacc --> nt
cacc --> f
f -.-> nt

mac(Morse account/actor claim protocol)

claim --> mac

classDef constraint color:#f00,stroke:#f00
[...]

PR #1242, opened 2025-04-23
[Draft][WIP] Ante to create base accounts

Olshansk:
Untested proposal / suggesiton for #1236

### DIFF
--- a/app/ante.go
+++ b/app/ante.go
@@ -4,34 +4,119 @@ import (
 	storetypes "cosmossdk.io/store/types"
 	"github.com/cosmos/cosmos-sdk/crypto/keys/secp256k1"
 	cosmostypes "github.com/cosmos/cosmos-sdk/types"
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
 	"github.com/cosmos/cosmos-sdk/types/tx/signing"
 	"github.com/cosmos/cosmos-sdk/x/auth/ante"
+	authkeeper "github.com/cosmos/cosmos-sdk/x/auth/keeper"
 	authsigning "github.com/cosmos/cosmos-sdk/x/auth/signing"
 	authtypes "github.com/cosmos/cosmos-sdk/x/auth/types"
 
 	"github.com/pokt-network/poktroll/app/volatile"
 )
 
-// newAnteHandlerFn returns an AnteHandler that waives minimum gas/fees for transactions
-// that contain ONLY morse claim messages.
-// I.e. MsgClaimMorseAccount, MsgClaimMorseApplication and MsgClaimMorseSupplier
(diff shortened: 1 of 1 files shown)


bryanchriswhite:
Thanks @Olshansk, great idea - thanks for taking a stab! :raised_hands: :hocho: 

I was previously able to demonstrate that the ante handler is down-stack from where the auth module returns the kind of error seen below, which rules out this class of solutions, unfortunately. :confused:

I took a few minutes just now to test this branch specifically and was able to reproduce this behavior. Additionally, I confirmed that the stack never reaches any of this code prior to the error (using `gdb`).

```
rpc error: code = NotFound desc = rpc error: code = NotFound desc = account pokt1m2j7gun57jqpc0mcynzs0tcjz3p66zw34799yt not found: key not found
```

[asciicast](https://asciinema.org/a/RIUOHeOlKRAKtYPwmzRxuQpjy)
> `minimum-gas-fee: 0.000000001upokt`; key `newbie2` has no onchain state

The error goes away as soon as I do a ` pocketd tx bank send faucet $(make pocketd_addr ACC_NAME=newbie2) 1upokt --fees=1upokt --gas=auto` and retry the claim.

bryanchriswhite:
I'm recalling that we also need to account for the account query which happens unless you pass `--offline`.

So now I've also double-checked using `--offline=true`, which then requires that BOTH `--account-number` and `--sequence`, for which no valid values exist. Setting them to an arbitrary number (e.g. `0`) yields the same error.

Despite eliminating the account query with `--offline`, the error can traced to (via onchain breakpoints) to [cosmos-sdk/x/auth/keeper/grpc_query.go:78](https://github.com/cosmos/cosmos-sdk/blob/v0.50.13/x/auth/keeper/grpc_query.go#L78). This seems to imply that there's a grpc query happening at some point for some reason, but it's not yet clear to me where that's coming from. I'm quite confident that the `--offline`, `--sequence` and `--account-number` flags are hooked up correctly because they are validated and will cause the error message to change when invalid values are used (e.g. one or both are missing or negative).

bryanchriswhite:
UPDATE: upon further inspection, `offline` doesn't seem to be set quite right after all - still investigating.

bryanchriswhite:
Deepest error yet, I've had to do some trickery with the fee granter:
```
signature verification failed; please verify account number (23) and chain-id (pocket): (unable to verify single signer signature): unauthorized: error during ABCI check tx [/home/bwhite/go/pkg/mod/cosmossdk.io/errors@v1.0.1/errors.go:155]
```

bryanchriswhite:
You can see [here](https://github.com/cosmos/cosmos-sdk/blob/main/client/tx/tx.go#L325) that cosmos-sdk is using indeed serializing the account number as part of the tx bytes to be signed.

PR #1338, opened 2025-05-16
[Morse -> Shannon Migration] chore: respect migrated unbonding period

bryanchriswhite:
## Summary

- ...

## Issue

- TBD

## Type of change

- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Sanity Checklist


x/migration/keeper/msg_server_claim_morse_application.go
```diff
@@ -137,12 +138,68 @@ func (k msgServer) ClaimMorseApplication(ctx context.Context, msg *migrationtype
...
+	if !morseClaimableAccount.UnstakingCompletionTime.IsZero() {
+		unbondedApp := &apptypes.Application{
+			Address:                 shannonAccAddr.String(),
+			Stake:                   &postClaimAppStake,
+			UnstakeSessionEndHeight: uint64(previousSessionEnd),
+			// Services:             (intentionally omitted, no services were staked),
+			// ServiceConfigHistory: (intentionally omitted, no services were staked),
+		}
+
+		durationUntilUnstake := morseClaimableAccount.UnstakingCompletionTime.Sub(time.Now())
```
bryanchriswhite, line 173:
```suggestion
		durationUntilUnbondingEnd := morseClaimableAccount.UnstakingCompletionTime.Sub(time.Now())
```

x/migration/keeper/msg_server_claim_morse_application.go
```diff
@@ -137,12 +138,68 @@ func (k msgServer) ClaimMorseApplication(ctx context.Context, msg *migrationtype
...
+		if durationUntilUnstake.Seconds() <= 0 {
+			// Emit an event which signals that the claimed Morse supplier's unbonding
+			// period began on Morse, and ended while waiting to be claimed.
+			morseSupplierUnbondingEndEvent := apptypes.EventApplicationUnbondingEnd{
+				Application:        unbondedApp,
+				Reason:             apptypes.ApplicationUnbondingReason_APPLICATION_UNBONDING_REASON_ELECTIVE,
```
bryanchriswhite, line 179: @Reviewier, there is no `APPLICATION_UNBONDING_REASON_UNSPECIFIED`. It could be added EITHER:
A) **Preserving backwards compatibility**: 
    - Using the next available enum index (not 0)
    - Against convention
B) **Following convention**:
    - Shifting the enum values to set UNSPECIFIED to 0
    - Not backwards compatible for existing consumers of migration protobuf types (assuming any exist)

proto/pocket/migration/morse_offchain.proto
```diff
@@ -133,6 +137,9 @@ message MorseValidator {
   // The string representation of the BigInt amount of upokt.
   string staked_tokens = 7 [(gogoproto.jsontag) = "tokens"];
 
+  // TODO_IN_THIS_COMMIT: comment...
+  google.protobuf.Timestamp UnstakingCompletionTime = 8 [(gogoproto.nullable) = false, (gogoproto.stdtime) = true, (gogoproto.jsontag) = "unstaking_time", (gogoproto.moretags) = "yaml:\"unstaking_time\""];
```
Olshansk, line 141: Can we call it `UnbondingTimestamp`?

x/migration/types/morse_claimable_account.go
```diff
@@ -1,7 +1,44 @@
...
+	// Retrieve the estimated block duration for the current chain from a lookup table.
+	// DEV_NOTE: This is an offchain config value; i.e. not queryable.
+	estimatedBlockDuration := int64(volatile.EstimatedBlockDurationByChainId[sdkCtx.ChainID()])
+
+	// TODO_IN_THIS_COMMIT: comment...
+	// ... return early if unstaking is already complete...
+	durationUntilUnstakeCompletion := int64(time.Since(m.UnstakingCompletionTime))
+	if durationUntilUnstakeCompletion <= 0 {
+		// The unstaking completion time has already elapsed.
+		return -1
```
Olshansk, line 32: Make sure the explanation for -1 is in the comments

x/migration/types/morse_claimable_account.go
```diff
@@ -1,7 +1,44 @@
...
+func (m *MorseClaimableAccount) GetEstimatedUnbondingEndHeight(ctx context.Context) int64 {
+	sdkCtx := cosmostypes.UnwrapSDKContext(ctx)
+
+	// Retrieve the estimated block duration for the current chain from a lookup table.
+	// DEV_NOTE: This is an offchain config value; i.e. not queryable.
+	estimatedBlockDuration := int64(volatile.EstimatedBlockDurationByChainId[sdkCtx.ChainID()])
+
+	// TODO_IN_THIS_COMMIT: comment...
+	// ... return early if unstaking is already complete...
+	durationUntilUnstakeCompletion := int64(time.Since(m.UnstakingCompletionTime))
```
Olshansk, line 29: Why isn't this `time.Until` instead of `time.Since`?

x/migration/keeper/msg_server_claim_morse_application.go
```diff
@@ -137,12 +138,68 @@ func (k msgServer) ClaimMorseApplication(ctx context.Context, msg *migrationtype
 		morseClaimableAccount,
 	)
 
+	claimedAppStake := morseClaimableAccount.GetApplicationStake()
+	sharedParams := k.sharedKeeper.GetParams(ctx)
+	sessionEndHeight := sharedtypes.GetSessionEndHeight(&sharedParams, sdkCtx.BlockHeight())
+	claimedUnstakedBalance := morseClaimableAccount.GetUnstakedBalance()
+
+	currentSessionStart := sharedtypes.GetSessionStartHeight(&sharedParams, sdkCtx.BlockHeight())
```
Olshansk, line 146:
```suggestion
	currentSessionStartHeight := sharedtypes.GetSessionStartHeight(&sharedParams, sdkCtx.BlockHeight())
```

x/migration/keeper/msg_server_claim_morse_application.go
```diff
@@ -137,12 +138,68 @@ func (k msgServer) ClaimMorseApplication(ctx context.Context, msg *migrationtype
...
+	previousSessionEnd := sharedtypes.GetSessionEndHeight(&sharedParams, currentSessionStart-1)
```
Olshansk, line 147:
```suggestion
	previousSessionEndHeight := sharedtypes.GetSessionEndHeight(&sharedParams, currentSessionStart-1)
```

x/migration/keeper/msg_server_claim_morse_application.go
```diff
@@ -137,12 +138,68 @@ func (k msgServer) ClaimMorseApplication(ctx context.Context, msg *migrationtype
...
+
 	// Query for any existing application stake prior to staking.
 	preClaimAppStake := cosmostypes.NewInt64Coin(volatile.DenomuPOKT, 0)
 	foundApp, isFound := k.appKeeper.GetApplication(ctx, shannonAccAddr.String())
 	if isFound {
 		preClaimAppStake = *foundApp.Stake
 	}
+	postClaimAppStake := preClaimAppStake.Add(morseClaimableAccount.GetApplicationStake())
+
+	// TODO_IN_THIS_COMMIT: comment...
```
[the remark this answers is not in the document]
Olshansk: Leaving one comment, but let's tend to all of thes.

x/migration/keeper/msg_server_claim_morse_application.go
```diff
@@ -137,12 +138,68 @@ func (k msgServer) ClaimMorseApplication(ctx context.Context, msg *migrationtype
...
+	// ... application began unbonding on Morse and the unbonding period has since elapsed...
+	// ... both stake and unstatked balance were minted to the shannonDestAddr account...
+	// ... no application was staked...
+	// - use a lookup table to get the est. block times per network...
+	// - use the shared params and the block time to calculate the unstake session end height...
+	// - emit an unbonding start event...
+	if !morseClaimableAccount.UnstakingCompletionTime.IsZero() {
```
Olshansk, line 164: Shouldn't this be less than or equal to zero (given the -1 return value in the function)?

x/migration/keeper/msg_server_claim_morse_supplier.go
```diff
@@ -187,19 +188,21 @@ func (k msgServer) ClaimMorseSupplier(
 	postClaimSupplierStake := preClaimSupplierStake.Add(morseClaimableAccount.GetSupplierStake())
 	minStake := k.supplierKeeper.GetParams(ctx).MinStake
 
+	currentSessionStart := sharedtypes.GetSessionStartHeight(&sharedParams, sdkCtx.BlockHeight())
```
Olshansk, line 191: Please add `height` as a postfix here and below

x/migration/keeper/msg_server_claim_morse_supplier.go
```diff
@@ -187,19 +188,21 @@ func (k msgServer) ClaimMorseSupplier(
...
+	previousSessionEnd := sharedtypes.GetSessionEndHeight(&sharedParams, currentSessionStart-1)
+
+	// Construct in case it is already or will become unbonded immediately (i.e. below min stake).
```
Olshansk, line 194: It's a bit confusing we're constructing an unbonded supplier even though we don't need it yet.

I feel like this needs to be restructured or renamed.

Even if we need to make object copies later.

x/migration/module/cmd/types.go
```diff
@@ -147,39 +146,59 @@ func (miw *morseImportWorkspace) addUnstakedBalance(addr string, amount cosmosma
 // addSupplierStake does two things:
 // - Adds the given amount to the corresponding Morse account balances in the morseWorkspace
 // - Sets the MorseOutputAddress if the given outputAddr is not nil
-func (miw *morseImportWorkspace) addSupplierStake(
-	addr string,
-	amount cosmosmath.Int,
-	outputAddr cometcrypto.Address,
-) error {
+func (miw *morseImportWorkspace) addSupplierStake(morseValidator *migrationtypes.MorseValidator) error {
```
Olshansk, line 149: Thoughts on `s/morseValidator/morseSupplier/g` or `morseStakedNode`?

Olshansk: My preference is for `morseSupplier`

x/migration/keeper/msg_server_claim_morse_supplier.go
```diff
@@ -269,28 +317,58 @@ func (k msgServer) ClaimMorseSupplier(
 		return nil, err
 	}
 
-	claimedSupplierStake := morseClaimableAccount.GetSupplierStake()
-	claimedUnstakedBalance := morseClaimableAccount.GetUnstakedBalance()
+	events := []cosmostypes.Msg{
```
Olshansk, line 320: 1. Please create an empty slice + #PUC why we need it.
2. Create the event separately
3. Append it.

x/migration/keeper/msg_server_claim_morse_supplier.go
```diff
@@ -269,28 +317,58 @@ func (k msgServer) ClaimMorseSupplier(
...
-	event := migrationtypes.EventMorseSupplierClaimed{
-		MorseNodeAddress:     msg.GetMorseNodeAddress(),
-		MorseOutputAddress:   morseClaimableAccount.GetMorseOutputAddress(),
-		ClaimSignerType:      claimSignerType,
-		ClaimedBalance:       claimedUnstakedBalance,
-		ClaimedSupplierStake: claimedSupplierStake,
-		SessionEndHeight:     sessionEndHeight,
-		Supplier:             supplier,
+	// If the claimed supplier is still unbonding, emit an unbonding begin event.
+	if !morseClaimableAccount.UnstakingCompletionTime.IsZero() {
```
Olshansk, line 334: Would love to see some tests.

Per the comments in the file where this function is defined and how it's used, I find it hard to grok.

Olshansk: Consider adding ANOTHER helper that wraps `UnstakingCompletionTime` called something like `ShouldStartAutoUnstakingMorseSupplier` for clarity.

x/migration/keeper/msg_server_claim_morse_supplier.go
```diff
@@ -251,10 +254,55 @@ func (k msgServer) ClaimMorseSupplier(
...
+	// ... supplier began unbonding on Morse and the unbonding period has since elapsed...
+	// ... unstatked balance was minted to the shannonSignerAddr account...
+	// ... stake was minted to the shannonOwnerAddr account...
+	// ... no Supplier was staked...
+	// - use a lookup table to get the est. block times per network...
+	// - use the shared params and the block time to calculate the unstake session end height...
+	// - emit an unbonding start event...
+	if !morseClaimableAccount.UnstakingCompletionTime.IsZero() {
+		durationUntilUnstake := time.Until(morseClaimableAccount.UnstakingCompletionTime)
+		if durationUntilUnstake.Seconds() <= 0 {
```
Olshansk, line 274: Anyway you can consolidate these three lines into a semantic helper inside of morseClaimableAccount?

Olshansk (changes requested):
Partial preliminary review.

app/upgrades/types.go
```diff
@@ -41,12 +38,12 @@ const (
 // NetworkAuthzGranteeAddress is a map of network names (i.e chain-id) to their
 // respective authorization (i.e. PNF/DAO) addresses.
 var NetworkAuthzGranteeAddress = map[string]string{
-	AlphaTestNetChainId: AlphaTestNetPnfAddress,
-	BetaTestNetChainId:  BetaTestNetPnfAddress,
+	volatile.AlphaTestNetChainId: AlphaTestNetPnfAddress,
```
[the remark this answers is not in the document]
bryanchriswhite: _TL;DR, I propose renaming this package to `pocket`._

The package name `volatile` isn't very meaningful, nor was it intended to be. IIRC, the package was created to avoid a dependency cycle. At the time we went with `volatile`, with the expectation that the constant would find a new, more appropriate home over time:

```go
// Package volatile is intended to host go module members which are need to be
// re-organized or removed in future work.
package volatile
```
> https://github.com/pokt-network/poktroll/blob/main/app/volatile/constants.go#L1

I added the chain IDs there to, once again, avoid a dependency cycle. After thinking about it for a few minutes, it seems to me like the right thing to do here would be to give this package a proper name (e.g. `pocket`).

proto/pocket/migration/morse_offchain.proto
bryanchriswhite: This is copy/paste from Morse. If we change it to anything, I would argue that it should be to match the JSON tag, **which we CANNOT change** since that's what it will be called in the JSON that's output from the Morse CLI.

```suggestion
  google.protobuf.Timestamp unstaking_time = 8 [(gogoproto.nullable) = false, (gogoproto.stdtime) = true, (gogoproto.jsontag) = "unstaking_time", (gogoproto.moretags) = "yaml:\"unstaking_time\""];
```

x/migration/keeper/msg_server_claim_morse_application.go
bryanchriswhite: I factored out some methods so that this is easier to read:

```go
	if morseClaimableAccount.IsUnbonding() {
		if morseClaimableAccount.SecondsUntilUnbonded() <= 0 {
```

app/volatile/networks.go
```diff
@@ -0,0 +1,28 @@
...
+	AlphaTestNetChainId: AlphaTestNetEstimatedBlockDuration,
+	BetaTestNetChainId:  BetaTestNetEstimatedBlockDuration,
+	MainNetChainId:      MainNetEstimatedBlockDuration,
+}
+
+// TODO_IN_THIS_COMMIT: comment...
+// ... specified in the config.toml; i.e. per validator, offchain ...
+// ... times can be / were inferred by averaging the duration between recent consecutive blocks...
+const (
+	AlphaTestNetEstimatedBlockDuration = time.Minute
```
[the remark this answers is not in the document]
bryanchriswhite: **Please see: https://github.com/pokt-network/pocket-network-genesis/pull/14**

x/migration/keeper/msg_server_claim_morse_supplier.go
bryanchriswhite: There are multiple sibling conditional branches which depend on this variable. I consolidated it into a single instance here to reduce boilerplate.

x/migration/keeper/msg_server_claim_morse_supplier.go
bryanchriswhite: Same as https://github.com/pokt-network/poktroll/pull/1338#discussion_r2095161799

x/migration/types/morse_claimable_account.go
```diff
@@ -1,7 +1,74 @@
...
+)
+
 // IsClaimed returns true if the MorseClaimableAccount has been claimed;
 // i.e. ShannonDestAddress is not empty OR the ClaimedAtHeight is greater than 0.
 func (m *MorseClaimableAccount) IsClaimed() bool {
 	return m.ShannonDestAddress != "" || m.ClaimedAtHeight > 0
 }
+
+// IsUnbonding indicates that the MorseClaimableAccount began unbonding on Morse
+// but its unbonding peroid has NOT yet elapsed.
```
Olshansk, line 20:
```suggestion
// but its unbonding peroid has NOT yet elapsed at the time that the Morse snapshot was taken.
```

x/migration/types/morse_claimable_account.go
```diff
@@ -1,7 +1,74 @@
...
+func (m *MorseClaimableAccount) IsUnbonding() bool {
+	// DEV_NOTE: The UnstakingTime field is a time.Time type, which has a zero value of "0001-01-01T00:00:00Z" when printed as an ISO8601 string.
+	// See: https://pkg.go.dev/time#Time.IsZero
+	return !m.UnstakingTime.IsZero()
+}
+
+// HasUnbonded indicates that the MorseClaimableAccount began unbonding on Morse
+// and the unbonding period has elapsed.
```
Olshansk, line 28: Can you #PUC when this will happen?

E.g. Supplier was claimed > 21 days after it started unbonding.

x/migration/types/morse_claimable_account.go
```diff
@@ -1,7 +1,74 @@
...
+// SecondsUntilUnbonded returns the number of seconds until the MorseClaimableAccount's
+// unbonding period will elapse.
+func (m *MorseClaimableAccount) SecondsUntilUnbonded() int64 {
+	return int64(time.Until(m.UnstakingTime).Seconds())
+}
+
+// GetEstimatedUnbondingEndHeight returns the estimated block height at which the
+// MorseClaimableAccount's unbonding period will end. The estimation process includes:
+//
+// - Calculating the remaining time until unstaking is complete.
```
Olshansk, line 42:
```suggestion
// GetEstimatedUnbondingEndHeight returns the estimated block height at which the
// MorseClaimableAccount's unbonding period will end.
// 
// The estimation process includes:
// - Calculating the remaining time until unstaking is complete.
```

x/migration/types/morse_claimable_account.go
```diff
@@ -1,7 +1,74 @@
...
+//
+// Returns:
+//
+// - The estimated block height when unbonding will end.
+// - -1 if unstaking is already complete.
+func (m *MorseClaimableAccount) GetEstimatedUnbondingEndHeight(ctx context.Context) int64 {
+	sdkCtx := cosmostypes.UnwrapSDKContext(ctx)
+
+	// Retrieve the estimated block duration for the current chain from a lookup table.
+	// DEV_NOTE: This is an offchain config value; i.e. not queryable.
```
Olshansk, line 53: Great comment

x/migration/types/morse_claimable_account.go
```diff
@@ -1,7 +1,74 @@
...
+	estimatedBlockDuration := int64(pocket.EstimatedBlockDurationByChainId[sdkCtx.ChainID()])
+
+	// Check if unstaking is complete:
+	//   - Calculate the remaining duration until unstaking.
+	//   - If the duration is zero or negative, the unstaking period has elapsed.
+	//   - Return -1 to indicate that unbonding is complete.
+	durationUntilUnstakeCompletion := int64(time.Until(m.UnstakingTime))
+	if durationUntilUnstakeCompletion <= 0 {
+		return -1
```
Olshansk, line 62: Optional: Consider creating a constant=-1 with a semantic name.

Olshansk (changes requested):
A few comments but looking pretty good!

Just tests remaining after this, right?

x/migration/types/morse_claimable_account.go
bryanchriswhite: I added an `isUnbonded` return arg instead.

Olshansk (changes requested):
@bryanchriswhite I'm still in the process of reviewing this here: https://github.com/pokt-network/poktroll/pull/1353/files

There's a lot of important logic here, so I want to really focus & understand it.

Let's upgrade alpha & beta tomorrow WITHOUT this change. It'll be a fast follow by EOW.

Olshansk (approved):
Prelininary approval w/ some edits so we can merge this in.

Made some edits in #1353.

It's a lot of code with nuanced lines, so I only have partial confidence in the overall correctness. 

I'm sure further testing (by us or users) will uncover bugs or edge cases.

Issue #1343, opened 2025-05-20
[Tooling] migration CLI: case-insensitive Morse hex addresses


Olshansk:
## Objective

Simplify the Morse -> Shannon migration process for the community

## Origin Document

https://discord.com/channels/553741558869131266/1234943674903953529/1374424092975038504

![Image](https://github.com/user-attachments/assets/e2b13844-95c4-49bb-8eae-48de76605338)

## Goals

- Reduce unnecessary cognitive overhead / support during the migration
- Avoid misc uppercase/lowercase discussions during the migration

## Deliverables

A PR that updates all migration tooling/docs to:
- [ ] Explain that we "auto uppercase" morse addresses
- [ ] Ensure query and transaction operations all default to "auto uppercasing" morse address
- [ ] Ensure tests & docs are updated accordingly

## Non-goals / Non-deliverables

- Treating the wrong casing as an error

---

**Creator**: @Olshansk 
**Co-Owners**: @bryanchriswhite @jorgecuesta

PR #1349, opened 2025-05-21
[Tooling] fix: normalize all Morse hex address cases to uppercase

bryanchriswhite:
## Summary

Normalize all Morse hex addresses usages to upper case so that different cases of the same string are treated as the same address.

## Issue

- #1343

## Type of change

- [x] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Sanity Checklist

- [x] For small changes: `make go_develop_and_test` and `make test_e2e`

### DIFF
--- a/pkg/encoding/hex.go
+++ b/pkg/encoding/hex.go
@@ -16,3 +16,9 @@ func NormalizeTxHashHex(txHash string) string {
 func TxHashBytesToNormalizedHex(txHash []byte) string {
 	return NormalizeTxHashHex(fmt.Sprintf("%x", txHash))
 }
+
+// NormalizeMorseHexAddress defines canonical and unambiguous representation for a
+// morse address hexadecimal string; upper-case.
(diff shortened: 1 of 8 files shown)


Olshansk (approved):
Going to approve to get things moving but do have a couple of questions:

1. Why not upper case when we're uploading/generating the state?
2. Why not also do this in the CLI?

bryanchriswhite:
> 1. Why not upper case when we're uploading/generating the state?
The Morse state export uses uppercase hex addresses, this is the source of truth for the canonical normalization.

> 2. Why not also do this in the CLI?
1. We cannot handle this on the CLI-side for the single query (`QueryMorseClaimableAccount`) because it uses autoCLI.
2. Handling it in all message handlers makes this fix client-agnostic.

Olshansk:
> We cannot handle this on the CLI-side for the single query (QueryMorseClaimableAccount) because it uses autoCLI.

Why is this?

PR #1429, opened 2025-06-04
[Recovery] Add upbit recovery list

Olshansk:
Source: https://docs.google.com/spreadsheets/d/1MAV9F2L3PjFCiZ4EClTjT3QZLZQScECIOPMwbqpxo8o/edit?gid=0#gid=0


x/migration/recovery/recovery_allowlist.go
```diff
@@ -2114,6 +2120,20 @@ var knownAppStakesAllowlist = []string{
 	"381ADD8CF814F45B8BF6DEA56AC185C2358B8BF2",
 }
 
+var upbitAllowlist = []string{
```
ArtSabintsev, line 2123: this is not only upbit. It's exchanges and DAO and Bridge.

Olshansk: How about now?

Issue #1454, opened 2025-06-10
[CLI] Add `auto-fee` to `pocketd`


Olshansk:
## Objective

Simplify the lives of `RelayMiners` and other `pocketd` users by introducing a `--auto-fee` flag that automatically calculates transaction fees based on gas simulation across all pocketd transaction commands.

## Origin Document

- #1447
- Specifically, see the changes implemented here as a reference: https://github.com/pokt-network/poktroll/pull/1447/files

This issue stems from the implementation of a new auto-fee feature that provides convenient fee automation while respecting user intent through a clear priority hierarchy.

## Goals

- Set the right pattern for other custom flags in the future
- Ensure consistent flag behavior across all CLI commands  
- Ensure the `--auto-fee` feature works correctly across all transaction commands (RelayMiner and module txs)
- Validate that the priority hierarchy is properly respected in all scenarios
- Confirm that warning messages are helpful and appear at appropriate times

## Deliverables

- [ ] Introduce a `--auto-fee` that hides the logic of `--fees` or `--gas` in the `pocketd` CLI
- [ ] Test the `--auto-fee` flag with a RelayMiner to ensure that it works E2E
- [ ] Test the `auto-fee` functionality across all transaction command scenarios
  - [ ] **Basic auto-fee usage**: `pocketd tx send ... --auto-fee`
  - [ ] **Auto-fee with custom gas adjustment**: `pocketd tx send ... --auto-fee --gas-adjustment 1.5`
  - [ ] **Explicit fees override**: `pocketd tx send ... --auto-fee --fees 1000upokt`
  - [ ] **Manual gas calculation override**: `pocketd tx send ... --auto-fee --gas 200000 --gas-prices 0.1upokt`
- [ ] Verify the priority hierarchy is correctly implemented and enforced
  - [ ] **Priority 1**: Explicit `--fees` flag (highest priority, overrides everything)
  - [ ] **Priority 2**: Manual `--gas` (not "auto") + `--gas-prices` combination
  - [ ] **Priority 3**: `--auto-fee` logic (when enabled and no explicit fees/gas set)
  - [ ] **Priority 4**: `--gas-adjustment` (always respected, defaults to 1.5 for auto-fee)
- [ ] Set default values
  - [ ] Gas adjustment: `1.5`
  - [ ] Gas Price: `0.0001upokt`
- [ ] Ensure the `--auto-fee` flag doesn't appear in irrelevant command help output
  - [ ] Should appear in transaction command help
[...]

PR #1492, opened 2025-06-13
[Tooling] Add `--auto-fee` flag

bryanchriswhite:
## Note @reviewer :rotating_light: 

This PR is intended to be reviewed on a per-commit basis. This should simplify the review process as well as facilitate executing on an alternative proposal (to only include the relayminer CLI changes).

## Summary

1. Add support for the `--auto-fee` flag, which sets `--gas`, `--gas-adjustment`, and `--gas-prices` to hard-coded constants. 
  1. The application of the flag logic is consolidated in the [root commands persistent-pre-run](https://github.com/pokt-network/poktroll/pull/1492/files#diff-4f1c1c9f6bfd6a15bb42db97708a40d9af358bd89e58c38c1ddf647919a479c7R91).
  2. It only has an effect if the flag is registered on the current command.
2. Register the flag on the relayminer CLI commands (i.e. `start` and `relay`)
3. Register the flag on all TX subcommands for all modules*

_*This is a non-trivial endeavor; this branch implementing the PoC design described below._

### AutoCLI Convention Modifications

This PoC defines conventions for selectively integrating TX RPC service methods with AutoCLI:

1. **If a method is too complex for AutoCLI** (e.g. requires custom flags or behavior),
   → its `RpcCommandOptions` entry is **commented out**.
   → The command is defined manually and **excluded** from AutoCLI generation.
   → _This is consistent with our current methodology._

3. **If a method appears in `RpcCommandOptions` with `Skip: true`**,
   → it is **not generated by AutoCLI**,
   → but is still parsed and eligible for **custom command enhancement** (e.g. via `EnhanceCustomCommand = true`).
   → _This would be the new convention - currently only effects application transfer._

5. **If a method appears in `RpcCommandOptions` without `Skip: true`**,
   → it **will be generated** by AutoCLI,
   → and **will not** receive custom command logic.
   → *This is typically unintended if the method also requires customization.*

> Commenting is used to completely exclude a method from both parsing and generation.
> `Skip: true` allows the method metadata to be declared, but generation to be suppressed.

### Technical Approach

1. Each module which includes ANY TX command MUST have a tx.go which contains a method which constructs and returns the TX command for that module.
2. Each module calls `AddModuleAutoCLICommands()` to generate (skipped) AutoCLI TX commands with the custom logic (i.e. `--auto-fee` flag registration).

### Pros

This allows us to **conventionally express the CLI** (in AutoCLI terms), even when falling back to manual Cobra command registration.

### Cons

- It's possible to unintentionally omit a skip; e.g., when scaffolding w/ `ignite`
- Still a bit unconventional.

## Issue

- Issue_or_PR: #1454 

## Type of change

- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Sanity Checklist

### DIFF
--- a/cmd/autocli.go
+++ b/cmd/autocli.go
@@ -1,0 +1,117 @@
+package cmd
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"cosmossdk.io/api/cosmos/autocli/v1"
+	"cosmossdk.io/core/appmodule"
+	"github.com/spf13/cobra"
+
+	"github.com/pokt-network/poktroll/cmd/flags"
+)
+
+// TODO_IN_THIS_COMMIT: update comment... copied from cosmossdk.io/api/cosmos/autocli/util.go
+// TopLevelCmd creates a new top-level command with the provided name and
+// description. The command will have DisableFlagParsing set to false and
+// SuggestionsMinimumDistance set to 2.
+func TopLevelCmd(ctx context.Context, use, short string) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:                        use,
+		Short:                      short,
+		DisableFlagParsing:         false,
+		SuggestionsMinimumDistance: 2,
+		RunE:                       validateCmd,
+	}
+	cmd.SetContext(ctx)
+
+	return cmd
+}
+
+// TODO_IN_THIS_COMMIT: update comment... copied from cosmossdk.io/api/cosmos/autocli/validate.go
+// validateCmd returns unknown command error or Help display if help flag set
+func validateCmd(cmd *cobra.Command, args []string) error {
+	var unknownCmd string
+	var skipNext bool
+
+	for _, arg := range args {
+		// search for help flag
+		if arg == "--help" || arg == "-h" {
+			return cmd.Help()
+		}
+
+		// check if the current arg is a flag
+		switch {
+		case len(arg) > 0 && (arg[0] == '-'):
+			// the next arg should be skipped if the current arg is a
+			// flag and does not use "=" to assign the flag's value
+			if !strings.Contains(arg, "=") {
+				skipNext = true
+			} else {
+				skipNext = false
+			}
+		case skipNext:
+			// skip current arg
+			skipNext = false
+		case unknownCmd == "":
+			// unknown command found
+			// continue searching for help flag
+			unknownCmd = arg
+		}
+	}
+
+	// return the help screen if no unknown command is found
+	if unknownCmd != "" {
+		err := fmt.Sprintf("unknown command \"%s\" for \"%s\"", unknownCmd, cmd.CalledAs())
+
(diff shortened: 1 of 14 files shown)


bryanchriswhite:
## Investigation Notes

### AutoCLI Integration _(TX only)_

| Module      | has tx.go | has tx autoCLI methods |
| ----------- | --------- | ---------------------- |
| application | yes       | yes (`transfer`)       |
| gateway     | yes       | no                     |
| migration   | yes       | no                     |
| proof       | yes       | no                     |
| service     | yes       | no*                    |
| session     | no        | no                     |
| supplier    | yes       | no                     |
| tokenomics  | no        | no                     |

### `EnhanceCustomCommand` Semantics and Integration Constraints

The `EnhanceCustomCommand` flag in AutoCLI controls how modules reconcile **custom-defined TX commands** with those generated automatically from gRPC service methods:

* ✅ **If `EnhanceCustomCommand = true`**:

  * Existing custom commands will be **augmented** with any missing RPC methods from the associated gRPC service.
  * This allows *hybrid command sets* where some methods are custom and others are auto-generated.

* ❌ **If `EnhanceCustomCommand = false`**:

  * **No AutoCLI commands will be generated** for a gRPC service if *any* method from that service already has a custom command.
  * This mode prevents ambiguous or conflicting command registration.

#### Supported: Merging Distinct Service Methods

It is fully supported and encouraged to merge **distinct gRPC service methods** from multiple services into a **single module-level TX command**, regardless of whether they are custom or AutoCLI-defined.

For example:

* `application tx` includes a custom `transfer` command from `MsgServiceA`.
* AutoCLI can still generate commands from `MsgServiceB`, and both sets of commands can coexist under the same `application tx` parent.

This pattern is **expressed cleanly** using multiple `RpcServiceOptions`, one per service.

#### Not Supported: Defining the Same Method Twice

It is **not** supported to define the **same gRPC method** in both:

* An AutoCLI `RpcCommandOptions` entry, and
* A manually constructed `*cobra.Command`.

This would result in **duplicate registration** of the same method, and AutoCLI will either panic or silently skip the duplicate, depending on implementation. To avoid this:

> When a method requires custom flags or logic, it must be fully handled via a manually defined Cobra command.
> The corresponding `RpcCommandOptions` entry should be commented out or omitted to document the intent but skip auto-generation.


### Misc. Notes

* `GovProposal` is a field on `RpcCommandOptions` that signals AutoCLI to generate a **governance proposal transaction** instead of a standard message broadcast.

  * When `gov_proposal: true`, AutoCLI wraps the message in a governance proposal and generates a `tx submit-proposal` style command.
  * When `false` (default), the message is included in a normal signed transaction and broadcast directly.
  * This field **only affects TX commands** — it has no effect on query commands.

> Useful for parameter changes and other `MsgService` methods that must go through on-chain governance.

Issue #1455, opened 2025-06-10
[RelayMiner] RelayMiner memory consumption


Olshansk:
> [!NOTE]
> Posting this on behalf of @easy2stake

## Pick One or more

- [x] Can't service relays
- [x] Can't scale my relay miner
- [ ] Can't deploy my relay miner
- [ ] Can't run my relay miner
- [x] [Other] Relay miner enters unknown "limbo" state causing errors and slashing

## Priority

- [] P0 (ASAP)
- [x] P1 (one day)
- [x] P2 (1-2 weeks)
- [ ] P3 (2-6 weeks)

## What? Why?

Relay miner becomes unstable and enters a "limbo" state. During this time:

- It stops serving relays
- `smt_store` bloats to ~100GB
- Metrics show only errors
- Eventually, `smt_store` is released and normal operation resumes

This cyclical instability causes missing proofs, which leads to **slashing**.

Logs indicate 60–100 slashed events per supplier on the affected list, with each slash corresponding to roughly 1 upokt lost.

## Morse Analogy

In Morse, I never encountered this type of `smt_store` growth or a relay miner freezing in a "half-alive" state.

If there were issues, they were usually recoverable via restarts or showed up in logs more clearly.

## Supporting Materials

To reoin logs, run:

```bash
cat relayminer.log.part-* > relayminer.log.tar.gz
```

[relayminer1.txt](https://github.com/user-attachments/files/20663650/relayminer1.txt)
[relayminer2.txt](https://github.com/user-attachments/files/20663649/relayminer2.txt)

<img width="596" alt="Image" src="https://github.com/user-attachments/assets/0a8aeef8-4f3b-4d99-87d7-051d1d3669dd" />
<img width="935" alt="Image" src="https://github.com/user-attachments/assets/87ed86f4-dd25-496f-b827-3299c6dbcf66" />
<img width="1716" alt="Image" src="https://github.com/user-attachments/assets/10703217-49ea-426c-8d9a-bd9d94e59a43" />

---

**Creator**: @Olshansk 
**Co-Owners**: @easy2stake @red-0ne

easy2stake:
Adding some more data here.
We're dealing with a memory leak. I'm posting the server memory graph + OOM message logs.

<img width="1715" alt="Image" src="https://github.com/user-attachments/assets/080470b6-9eea-43c6-a20b-3eaa57abb425" />

```
[...]

Issue #1448, opened 2025-06-08
[RelayMiner] Possible RelayMiner Memory Leak


Ajes1337:
## Pick One or more

- [x] Can't service relays
- [ ] Can't scale my relay miner
- [ ] Can't deploy my relay miner
- [ ] Can't run my relay miner
- [ ] [Other]

## Priority

- [ ] P0 (ASAP)
- [x] P1 (one day)
- [ ] P2 (1-2 weeks)
- [ ] P3 (2-6 weeks)

## What? Why?

after 5-6 hours, relayminer has used up ~32gb memory, servicing 100k relays every 15min.

@okdas asked me to submit this issue, I have attached requested pprof snapshot

## Details

Your supplier's address is: `pokt1fpugdgaqxn5rarsqpuelg94n5jlryttz7uwmlu` (beta-testnet)

The command you use to start your relay miner is:

```bash
pocketd \
  relayminer start \
  --grpc-insecure=true \
  --log_level=info \
  --config=/home/ajes/relayminer.yaml \
  --chain-id=pocket-beta
```

Your Supplier's onchain configs are:

```yaml
owner_address: pokt1rvx0krv2ftwasprr88vxv0vcy05za64tgn3340
operator_address: pokt1fpugdgaqxn5rarsqpuelg94n5jlryttz7uwmlu
stake_amount: 104042000000upokt
default_rev_share_percent:
  pokt1rvx0krv2ftwasprr88vxv0vcy05za64tgn3340: 95
  pokt1fpugdgaqxn5rarsqpuelg94n5jlryttz7uwmlu: 5
services:
  - service_id: "eth"
    endpoints:
      - publicly_exposed_url: https://rm0001beta.kalorius.tech
        rpc_type: JSON_RPC
  - service_id: "scroll"
    endpoints:
      - publicly_exposed_url: https://rm0001beta.kalorius.tech
        rpc_type: JSON_RPC
  - service_id: "moonriver"
    endpoints:
      - publicly_exposed_url: https://rm0001beta.kalorius.tech
        rpc_type: JSON_RPC
  - service_id: "kaia"
    endpoints:
      - publicly_exposed_url: https://rm0001beta.kalorius.tech
        rpc_type: JSON_RPC
  - service_id: "evmos"
    endpoints:
      - publicly_exposed_url: https://rm0001beta.kalorius.tech
        rpc_type: JSON_RPC
  - service_id: "ink"
    endpoints:
[...]

PR #1498, opened 2025-06-15
[RelayMiner] Refactor ReplayClient to fix memory leaks and improve event handling

red-0ne:
## Summary

Refactors the `ReplayClient` event handling mechanism to eliminate memory leaks and improve connection stability.

### Primary Changes:
- Completely rewrote `goPublishEvents` with a simpler, more robust event processing loop
- Removed complex nested closures and callbacks that were causing goroutine leaks
- Added proper connection retry handling with logging and clean context management

### Secondary changes:
- Added `CheckRelayRewardEligibility` to verify relays can still receive rewards
- Improved session tree cleanup with consolidated `deleteSession` method
- Added helper functions for claim and session SMT creation

Screenshot_20250616_013454

## Issue

- #1455 
- #1448 

## Type of change

- [x] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Sanity Checklist

- [x] I have updated the GitHub Issue Metadata: `assignees`, `reviewers`, `labels`, `project`, `iteration` and `milestone`
- [x] For small changes: `make go_develop_and_test` and `make test_e2e`


pkg/relayer/session/session.go
```diff
@@ -81,6 +82,15 @@ type relayerSessionsManager struct {
...
 	bankQueryClient client.BankQueryClient
+
+	// stopping is a boolean that indicates whether the relayerSessionsManager is
+	// in the process of a graceful shutting down.
+	// During normal operation, context cancellations (e.g. deadlines) are treated
+	// as failures that should cause failing session trees to be deleted.
+	// When stopping=true, these same cancellations are expected as part of shutdown
+	// and should not trigger deletion, ensuring session trees are properly persisted
+	// for recovery after restart.
+	stopping bool
```
Olshansk, line 93: I kind of feel like this should be an `atomic.bool`

```bash
import "sync/atomic"

stopping atomic.Bool
```

The concurrency scares me given that it's "just a bool"

pkg/relayer/session/session.go
```diff
@@ -81,6 +82,15 @@ type relayerSessionsManager struct {
...
 	// bankQueryClient is used to query for the bank module parameters.
 	bankQueryClient client.BankQueryClient
+
+	// stopping is a boolean that indicates whether the relayerSessionsManager is
+	// in the process of a graceful shutting down.
+	// During normal operation, context cancellations (e.g. deadlines) are treated
+	// as failures that should cause failing session trees to be deleted.
+	// When stopping=true, these same cancellations are expected as part of shutdown
+	// and should not trigger deletion, ensuring session trees are properly persisted
+	// for recovery after restart.
```
Olshansk, line 92:
```suggestion
	// stopping indicates whether the relayerSessionsManager is in the process of graceful shutdown.
	//
	// Why it exists:
	// - During normal operation, context cancellations (e.g., deadlines) are treated as session failures.
	// - These failures trigger cleanup: session trees are deleted.
	//
	// What changes when stopping = true:
	// - Context cancellations during shutdown are expected.
	// - These should NOT trigger deletion.
	// - This ensures session trees are persisted for recovery after restart.
```

Olshansk: Don't get lazy with making comments readable.

If you're not using AI to clean up your comments, you're doing it wrong ;)

pkg/relayer/session/session.go
```diff
@@ -197,6 +207,12 @@ func (rs *relayerSessionsManager) Start(ctx context.Context) error {
 //
 // This ensures no data is lost during shutdown and resources are properly cleaned up.
 func (rs *relayerSessionsManager) Stop() {
+	// Set the stopping flag to true to distinguish between normal errors and shutdown operations.
+	// During shutdown, context cancellations are expected and should not trigger session tree deletion.
+	// This ensures session trees are properly persisted for recovery after restart, rather than
+	// being interpreted as failures requiring cleanup.
```
Olshansk, line 213:
```suggestion
	// Mark the manager as stopping to prevent misinterpreting shutdown cancellations as failures.
	//
	// This ensures:
	// - Session trees are not deleted during shutdown.
	// - Data is preserved for recovery on the next startup.
	rs.stopping = true // TODO_IN_THIS_PR: rs.stopping.Store(true) ???
```

pkg/relayer/relay_authenticator/relay_verifier.go
```diff
@@ -91,6 +91,39 @@ func (ra *relayAuthenticator) VerifyRelayRequest(
 	return ErrRelayAuthenticatorInvalidSessionSupplier
 }
 
+// CheckRelayRewardEligibility verifies the relay's session hasn't expired for reward
```
Olshansk, line 94: #PUC so the community can understand what's happening with the windows here too.

pkg/relayer/relay_authenticator/relay_verifier.go
```diff
@@ -91,6 +91,39 @@ func (ra *relayAuthenticator) VerifyRelayRequest(
...
+		sharedParams,
+		relayRequest.Meta.SessionHeader.GetSessionEndBlockHeight(),
+	)
+
+	// If current height is equal or greater than the claim window opening height,
+	// the relay is no longer eligible for rewards as the session has expired
+	// for reward purposes
+	if currentHeight >= sessionClaimOpenHeight {
+		return ErrRelayAuthenticatorInvalidSession.Wrapf(
+			"session expired, must be before claim window open height (%d), but current height is (%d)",
```
Olshansk, line 118: #PUL with the patterns I showed elsewhere

pkg/relayer/relay_authenticator/relay_verifier.go
```diff
@@ -91,6 +91,39 @@ func (ra *relayAuthenticator) VerifyRelayRequest(
...
+			sessionClaimOpenHeight,
+			currentHeight,
+		)
+	}
+
```
Olshansk, line 123: Add a probabilistic debug log on success

pkg/relayer/session/proof.go
```diff
@@ -114,15 +114,22 @@ func (rs *relayerSessionsManager) waitForEarliestSubmitProofsHeightAndGeneratePr
...
 	// use a block query client to populate the block client replay observable at the time
 	// of block client construction. This check and failure branch can be removed once this
 	// is implemented.
 	if proofsWindowOpenBlock == nil {
-		logger.Warn().Msg("failed to observe earliest proof commit height offset seed block height")
-		failedSubmitProofsSessionsCh <- sessionTrees
+		// Only treat this as a failure if we're not in the process of shutting down.
+		// During normal operation (stopping==false), this is an unexpected error that should be handled.
+		// During shutdown (stopping=true), block observation failures are expected and should be ignored
+		// to prevent unnecessary session tree deletion and ensure proper persistence.
```
Olshansk, line 127:
```suggestion
		// Ignore this failure during shutdown:
		// - When `stopping == true`, context cancellations and observable failures are expected.
		// - Avoid interpreting them as session failures, to ensure session trees are persisted.
		//
		// In normal operation (`stopping == false`), this is a critical error and must be handled.
```

pkg/relayer/session/proof.go
```diff
@@ -114,15 +114,22 @@ func (rs *relayerSessionsManager) waitForEarliestSubmitProofsHeightAndGeneratePr
...
+		if !rs.stopping {
+			logger.Warn().Msg("failed to observe earliest proof commit height offset seed block height")
```
Olshansk, line 129: #PUL so it doesn't leave the reader asking for more.

pkg/relayer/session/proof.go
```diff
@@ -114,15 +114,22 @@ func (rs *relayerSessionsManager) waitForEarliestSubmitProofsHeightAndGeneratePr
 	proofsWindowOpenBlock := rs.waitForBlock(ctx, proofWindowOpenHeight)
 	// TODO_MAINNET: If a relayminer is cold-started with persisted but unproven ("late")
```
Olshansk, line 115:
```go
// TODO_MAINNET: Handle cold-start edge case for "late" sessions.
//
// Problem:
// - If RelayMiner starts with persisted but unproven ("late") sessions,
//    the corresponding proofsWindowOpenBlock may never be observed.
// - A "late" session is:
//     • Unclaimed
//     • Earliest commit height has already passed (i.e., it's in the past)
// - In this scenario, the block observable won't emit the expected signal.
//
// Planned fix:
// - Use a block query client to retroactively populate the observable
//   during block client construction.
//
// Note:
// - Once this is implemented, this failure check and branch can be removed.
```

Olshansk (changes requested):
@red-0ne 

We put a lot of effort into communication in discord.

If we put HALF that effort into better communication in logs, it'll 1000x our comms w/ suppliers.

Let's do it

pkg/relayer/session/session.go
red-0ne: It is under lock, but atomic bool is a good practice.

pkg/relayer/session/proof.go
red-0ne: This comment is no longer applicable. Cold starts are already handled.

pkg/client/events/replay_client.go
```diff
@@ -212,128 +184,111 @@ func (rClient *replayClient[T]) Close() {
...
+// goPublishEvents establishes and maintains an events subscription by
+// - Establishing EventsBytes subscription to events matching the query string
+// - Processes incoming events by decoding them from bytes to the generic type T
+// - Retries connection on failures up to connRetryLimit times with predefined delay
+// - Handles context cancellation to prevent resource leaks
+// - Cleans up connections when errors occur or context is canceled
+//
+// The method runs in a continuous loop until the context is canceled or retry limit is exceeded.
+func (rClient *replayClient[T]) goPublishEvents(ctx context.Context, publishCh chan<- T) {
+	numRetries := 0
```
bryanchriswhite, line 196: Does `numRetries` get reset after a successful retry? It seems like `numRetries` could accumulate between multiple intermittent errors otherwise, resulting in unexpected behavior.

pkg/relayer/proxy/relay_meter.go
```diff
@@ -296,8 +296,11 @@ func (rmtr *ProxyRelayMeter) forEachNewBlockFn(ctx context.Context, block client
 
 	// Delete the relay meters that correspond to settled sessions.
 	for _, sessionRelayMeter := range rmtr.sessionToRelayMeterMap {
-		sessionEndHeight := sessionRelayMeter.sessionHeader.GetSessionEndBlockHeight()
-		sessionClaimOpenHeight := sessionEndHeight + int64(sharedParams.GetClaimWindowOpenOffsetBlocks())
+
```
bryanchriswhite, line 299:
```suggestion
```

pkg/relayer/proxy/sync.go
```diff
@@ -246,6 +248,33 @@ func (server *relayMinerHTTPServer) serveSyncRequest(
 
 	relayer.RelayResponseSizeBytes.With("service_id", serviceId).Observe(float64(relay.Res.Size()))
 
+	// Verify relay reward eligibility a SECOND time AFTER completing the backend request.
```
bryanchriswhite, line 251: :fire:

pkg/relayer/session/session.go
```diff
@@ -81,6 +83,18 @@ type relayerSessionsManager struct {
 
 	// bankQueryClient is used to query for the bank module parameters.
 	bankQueryClient client.BankQueryClient
+
+	// stopping indicates whether the relayerSessionsManager is in the process of graceful shutdown.
```
bryanchriswhite, line 87: :ok_hand:

pkg/relayer/session/session.go
```diff
@@ -653,6 +667,34 @@ func (rs *relayerSessionsManager) deleteSessionTrees(
 	)
 }
 
+// deleteSession deletes the session tree from the relayerSessions and
+// removes it from the disk store.
+func (rs *relayerSessionsManager) deleteSession(sessionTree relayer.SessionTree) {
```
bryanchriswhite, line 672: I think we should EITHER:
- rename this to `#deleteSessionTree()` _(my preference)_
- rename `#deleteSessionTrees()` to `#deleteSessions()`

```suggestion
// deleteSessionTree deletes the session tree from the relayerSessions and
// removes it from the disk store.
func (rs *relayerSessionsManager) deleteSessionTree(sessionTree relayer.SessionTree) {
```

pkg/relayer/session/session.go
```diff
@@ -653,6 +667,34 @@ func (rs *relayerSessionsManager) deleteSessionTrees(
...
+
+	sessionHeader := sessionTree.GetSessionHeader()
+	logger := rs.logger.With(
+		"session_id", sessionHeader.GetSessionId(),
+		"application_address", sessionHeader.GetApplicationAddress(),
+		"service_id", sessionHeader.GetServiceId(),
+		"supplier_operator_address", sessionTree.GetSupplierOperatorAddress(),
+	)
+
+	if err := sessionTree.Delete(); err != nil {
```
bryanchriswhite, line 683: #PUC what is the effect of this?

pkg/relayer/session/session.go
```diff
@@ -653,6 +667,34 @@ func (rs *relayerSessionsManager) deleteSessionTrees(
...
+		logger.Error().Err(err).Msg("failed to delete session tree")
+	}
+
+	sessionSMT := sessionSMTFromSessionTree(sessionTree)
+	if err := rs.deletePersistedSessionTree(sessionSMT); err != nil {
```
bryanchriswhite, line 688: #PUC what is the effect of these?

pkg/relayer/session/session.go
```diff
@@ -676,3 +718,23 @@ func (rs *relayerSessionsManager) supplierSessionsToClaim(
 
 	return sessionsToClaimObs
 }
+
+// claimFromSessionTree creates a claim object from the given SessionTree.
```
bryanchriswhite, line 722:
```suggestion
// claimFromSessionTree returns a claim object from the given SessionTree.
```

bryanchriswhite: Is there a reason not to promote this to a `relayer.SessionTree` method?

pkg/relayer/session/session.go
```diff
@@ -676,3 +718,23 @@ func (rs *relayerSessionsManager) supplierSessionsToClaim(
...
+func claimFromSessionTree(sessionTree relayer.SessionTree) prooftypes.Claim {
+	return prooftypes.Claim{
+		SupplierOperatorAddress: sessionTree.GetSupplierOperatorAddress(),
+		SessionHeader:           sessionTree.GetSessionHeader(),
+		RootHash:                sessionTree.GetClaimRoot(),
+	}
+}
+
+// sessionSMTFromSessionTree creates a SessionSMT from the given SessionTree.
+func sessionSMTFromSessionTree(
```
bryanchriswhite, line 732: Same question as above...

bryanchriswhite (commented):

I just had a few comments/questions. :rocket:

pkg/client/events/replay_client.go
red-0ne: Good catch!

pkg/relayer/session/session.go
```diff
@@ -659,23 +667,35 @@ func (rs *relayerSessionsManager) deleteSessionTrees(
...
 	}
 
+	// Delete the persisted session tree metadata from the disk store.
+	// This is necessary to ensure that the session is not restored on the next startup.
 	sessionSMT := sessionSMTFromSessionTree(sessionTree)
 	if err := rs.deletePersistedSessionTree(sessionSMT); err != nil {
 		rs.logger.Error().
 			Err(err).
-			Str("session_id", sessionId).
 			Msg("failed to delete persisted session tree metadata")
```
Olshansk, line 694: Please add more details and use emojis.

pkg/relayer/session/session.go
```diff
@@ -659,23 +667,35 @@ func (rs *relayerSessionsManager) deleteSessionTrees(
...
 	}
+
+	logger.ProbabilisticDebugInfo(polylog.ProbabilisticDebugInfoProb).
+		Msg("deleted session tree from relayerSessions and disk store")
```
Olshansk, line 698: Please add more details and use emojis.

pkg/relayer/session/sessiontree.go
```diff
@@ -340,6 +340,10 @@ func (st *sessionTree) GetClaimRoot() []byte {
 func (st *sessionTree) Delete() error {
 	st.sessionMu.Lock()
 	defer st.sessionMu.Unlock()
+	// Set treeStore to nil after function execution to prevent potential double-close operations.
+	// This ensures that if another part of the code attempts to close  this treeStore
+	// after deletion, it will safely handle the nil reference instead of causing a
+	// panic by operating on an already closed kvstore instance.
```
Olshansk, line 346:
```suggestion
	// After deletion, set treeStore to nil to:
	// - Prevent double-close operations.
	// - Avoid panics from future use of a closed kvstore instance.
	// - Signal that the treeStore is no longer valid.
```

Olshansk (approved):

PR #1545, opened 2025-06-23
[RM] Improve some logging

Olshansk:
100% Claude Code generated based on user feedback

Screenshot 2025-06-23 at 1 44 38 PM


red-0ne (approved):
PTAL at the change request I proposed.

Otherwise PTAL. Preemptively approving.

PR #1604, opened 2025-06-30
[State bloat] use the relay response payload hash as the SMST value

bryanchriswhite:
## Summary

This PR attempts to address the proof related state bloat at its root (or rather the leaves).

This reduces the size of onchain proofs in places that cannot be impacted by dehydrating or pruning events; e.g., archival TXs (going forward). 

- Add the relay response payload hash to the relay response data structure
- Set the relay response payload to nil before inserting as an SMST value
- Update relayminer test fixtures and fixture generator

## Issue

- #1497 

## Type of change

Select one or more from the following:

- [ ] New feature, functionality or library
- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Sanity Checklist

- [x] I have updated the GitHub Issue Metadata: `assignees`, `reviewers`, `labels`, `project`, `iteration` and `milestone`
- [ ] For docs: `make docusaurus_start`
- [x] For small changes: `make go_develop_and_test` and `make test_e2e`
- [ ] For major changes: `devnet-test-e2e` label to run E2E tests in CI
- [ ] For migration changes: `make test_e2e_oneshot`
- [ ] 'TODO's, configurations and other docs

PR #1620, opened 2025-07-02
[Review] Review of #1604: `[State bloat] use the relay response payload hash as the SMST value`

Olshansk:
Review to accompany comments in #1604


pkg/relayer/miner/gen/gen_fixtures.go
```diff
@@ -146,13 +146,14 @@ func genRandomizedMinedRelayFixtures(
...
 
 			randBz := make([]byte, randLength)
-			relayResponsePayloadHash := protocol.GetRelayHashFromBytes(randBz)
-
 			if _, err := rand.Read(randBz); err != nil {
 				errCh <- err
 				return
 			}
 
+			// Compute the relay hash using the random bytes as the relay bytes.
```
bryanchriswhite, line 154: Nice catch - thank you! :sweat_smile:

PR #1633, opened 2025-07-03
[RelayMiner] feat: relay miner routes to multiple backend URLs based on RPC Type

commoddity:
## 🌿 Summary

Enhances the Relay Miner to dynamically route incoming requests to different backend URLs based on the RPC type specified in the request header, enabling greater flexibility in handling protocol-specific relays.

**NOTE: This PR specifically intends to be backward-compatible by introducing RPC-type specific service configs in addition to the default service config.**

**This means the YAML structure of the Relay Miner config is unchanged so the new functionality may be added in addition to existing configs but does not require NRs to change their config structure.**

### 🌱 Primary Changes:
- Introduced support for multiple service configurations within a single supplier via new `rpc_type_service_configs` YAML field.
- Modified proxy request handling logic to route requests by inspecting the `RPC-Type` header and selecting the corresponding service configuration.
- Added `RPCType` enum with validation logic to support known RPC types (e.g., `json_rpc`, `rest`).

### 🍃 Secondary changes:
- Refactored config structures to distinguish between `ServiceConfig` and `RPCTypeServiceConfigs`.
- Updated tests to validate new config parsing and proxy behavior under varied RPC-type scenarios.
- Renamed usages of `ServiceConfig` to `ServiceConfig` across the codebase for alignment.

## 🛠️ Type of change

## 🤯 Sanity Checklist

- [x] I have updated the GitHub Issue 'assignees', 'reviewers', 'labels', 'project', 'iteration' and 'milestone'
- [x] For code, I have run 'make test_all'
- [x] I added TODOs where applicable


pkg/relayer/proxy/http_server.go
```diff
@@ -143,7 +143,7 @@ func (server *relayMinerHTTPServer) Ping(ctx context.Context) error {
 	for _, supplierCfg := range server.serverConfig.SupplierConfigsMap {
 		c := &http.Client{Timeout: 2 * time.Second}
 
-		backendUrl := *supplierCfg.ServiceConfig.BackendUrl
+		backendUrl := *supplierCfg.DefaultServiceConfig.BackendUrl
```
red-0ne, line 146: We should ping all the configs not only the default one.

pkg/relayer/proxy/sync.go
```diff
@@ -129,28 +129,33 @@ func (server *relayMinerHTTPServer) serveSyncRequest(
...
+
 	if serviceConfig == nil {
 		return relayRequest, ErrRelayerProxyServiceEndpointNotHandled.Wrapf(
 			"service %q not configured",
 			serviceId,
 		)
 	}
 
+	// If the RPC-Type header is set, use the RPC type-specific service config.
+	rpcType := config.RPCType(request.Header.Get("RPC-Type"))
```
red-0ne, line 152: Shouldn't the header key be `Rpc-Type`?

pkg/relayer/proxy/sync.go
```diff
@@ -129,28 +129,33 @@ func (server *relayMinerHTTPServer) serveSyncRequest(
...
+	if rpcType != "" {
```
red-0ne, line 153: We should add a log when we're using the default `serviceConfig`

pkg/relayer/proxy/async.go
```diff
@@ -49,12 +50,29 @@ func (server *relayMinerHTTPServer) handleAsyncConnection(
...
+
+	if serviceConfig == nil {
+		return ErrRelayerProxyServiceEndpointNotHandled.Wrapf(
+			"service %q not configured",
+			serviceId,
+		)
+	}
+
+	// If the RPC-Type header is set, use the RPC type-specific service config.
+	rpcType := config.RPCType(request.Header.Get("RPC-Type"))
```
red-0ne, line 65: This is where it's a bit foggy. Is there a case where we've already determined that it's a websockets connection but still need to look for the `RPCType`?

I believe that at some point `websocket` should only be an `RPCType` instead of a dedicated service.

pkg/relayer/config/types.go
```diff
@@ -179,7 +202,7 @@ type RelayMinerSupplierConfig struct {
 // RelayMinerSupplierServiceConfig is the structure resulting from parsing the supplier
 // service sub-section of the RelayMiner config file.
 type RelayMinerSupplierServiceConfig struct {
-	// BackendUrl is the URL of the service that relays will be proxied to.
+	// BackendUrl is the default URL of the service that relays will be proxied to.
```
red-0ne, line 205: Why `default`. Shouldn't it be the "only` URL given a `RelayMinerSupplierServiceConfig` ?

red-0ne (commented):
Great addition!

Thanks for keeping the config backward compatible.

Uniqueness of the RPC configs is guaranteed :+1: 

Left a few comments but no major blockers.

pkg/relayer/proxy/sync.go
commoddity: I've updated the logger used in the `servceSyncRequest` method with a `service_config_type` field that will be either "default" or the "rpc type" string.

pkg/relayer/proxy/async.go
commoddity: Based on our discussion in Discord I've updated the PR to use the `Rpc-Type: websocket` header to determine whether to use the websocket connection method vs the regular sync method. 

Please see this commit for details: https://github.com/pokt-network/poktroll/pull/1633/commits/6e6e462d4e89068bf0b80780ada529765f0b1a3f

pkg/relayer/proxy/sync.go
```diff
@@ -166,26 +166,47 @@ func (server *relayMinerHTTPServer) serveSyncRequest(
...
+	// If the 'Rpc-Type' header is set, use the RPC type specific service config.
+	rpcType := config.RPCType(request.Header.Get(rpcTypeHeader))
+	if rpcType != "" {
+		// If the RPC type is set for the service, use the RPC type specific service config.
+		if rpcTypeServiceConfig, ok := supplierConfig.RPCTypeServiceConfigs[rpcType]; ok {
+			serviceConfig = rpcTypeServiceConfig
+			// Update the service config type log to indicate the RPC type.
+			serviceConfigTypeLog = string(rpcType)
+		}
+	}
```
Olshansk, line 199: Can you extract this into its own helper function?

pkg/relayer/proxy/sync.go
```diff
@@ -166,26 +166,47 @@ func (server *relayMinerHTTPServer) serveSyncRequest(
...
+		return relayRequest, ErrRelayerProxyServiceEndpointNotHandled.Wrapf(
+			"service %q not configured",
+			serviceId,
+		)
 	}
 
+	// Initialize the service config to the default service config.
+	serviceConfig := supplierConfig.DefaultServiceConfig
+	// Set a string that will be logged to indicate the service config type.
+	serviceConfigTypeLog := "default"
```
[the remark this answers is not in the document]
Olshansk: As a second point. Can you think of a way to avoid this variable altogether?

pkg/relayer/config/supplier_hydrator.go
```diff
@@ -18,40 +18,76 @@ func (supplierConfig *RelayMinerSupplierConfig) HydrateSupplier(
...
+		return err
+	}
+	supplierConfig.DefaultServiceConfig = defaultServiceConfig
+
+	// Hydrate the RPC type-specific service configs (if any)
+	supplierConfig.RPCTypeServiceConfigs = make(
+		map[RPCType]*RelayMinerSupplierServiceConfig,
+		len(yamlSupplierConfig.RPCTypeServiceConfigs),
+	)
+	for rpcType, serviceConfig := range yamlSupplierConfig.RPCTypeServiceConfigs {
```
Olshansk, line 33: #PUC what this loop is doing.

Olshansk (changes requested):
Biggest issue I see right now is the fact that we're changing configs, not updating docs, and some naming discrepencies.

pkg/relayer/config/types.go
```diff
@@ -66,13 +83,14 @@ type YAMLRelayMinerMetricsConfig struct {
...
 type YAMLRelayMinerSupplierConfig struct {
-	ListenUrl             string                              `yaml:"listen_url"`
-	ServiceConfig         YAMLRelayMinerSupplierServiceConfig `yaml:"service_config"`
-	ServiceId             string                              `yaml:"service_id"`
-	SigningKeyNames       []string                            `yaml:"signing_key_names"`
-	RequestTimeoutSeconds uint64                              `yaml:"request_timeout_seconds"`
-	MaxBodySize           string                              `yaml:"max_body_size"`
-	XForwardedHostLookup  bool                                `yaml:"x_forwarded_host_lookup"`
+	ListenUrl             string                                          `yaml:"listen_url"`
+	DefaultServiceConfig  YAMLRelayMinerSupplierServiceConfig             `yaml:"service_config"`
```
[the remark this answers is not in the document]
commoddity: > We renamed the variable but not the yaml.
> 
> We need to do one of two things: `DefaultServiceConfig` vs `service_config`

WDYT about just keeping the variable called `ServiceConfig` then? The other ones are already called `RPCTypeServiceConfigs` so it feels implied they are not the defaults. 

IMO the semantics of `DefaultServiceConfig` vs `ServiceConfig` is not crucial enough to justify forcing all suppliers to change their configs. Simplest solution would be to just leave the var name in sync with the YAML, but open to discuss. CC @red-0ne

pkg/relayer/config/types.go
```diff
@@ -163,14 +181,22 @@ type RelayMinerMetricsConfig struct {
...
 	// ServiceId is the serviceId corresponding to the current configuration.
 	ServiceId string
+
 	// ServerType is the transport protocol used by the supplier, it must match the
 	// type of the relay miner server it is associated with.
-
 	ServerType RelayMinerServerType
-	// ServiceConfig is the config of the service that relays will be proxied to.
+
+	// ServiceConfig is the default config of the service that relays will be proxied to.
```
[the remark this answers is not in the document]
commoddity: See above comment.

proto/pocket/shared/service.proto
```diff
@@ -72,6 +72,7 @@ enum RPCType {
   WEBSOCKET = 2; // WebSocket
   JSON_RPC = 3; // JSON-RPC
   REST = 4; // REST
+  HYBRID = 5; // Hybrid (REST + JSON-RPC)
```
Olshansk, line 75: I feel like this isn't "forward looking enough."

- If it's explicitly `REST + JSON-RPC`, let's name it `JSON_RPC_AND_REST`
- if it's "ambiguous", then let's call it `HYBRID` or `AMBIGUOUS`

pkg/relayer/proxy/sync.go
```diff
@@ -375,6 +389,48 @@ func (server *relayMinerHTTPServer) serveSyncRequest(
 	return relayRequest, nil
 }
 
+// serviceConfigTypeDefault indicates that the service config being used is
+// the default service config, as opposed to an RPC-type specific config.
```
Olshansk, line 393:
```suggestion
// the default service config, as opposed to a service-specific config.
```

pkg/relayer/proxy/sync.go
```diff
@@ -375,6 +389,48 @@ func (server *relayMinerHTTPServer) serveSyncRequest(
...
+			)
+		}
+
+		// If the header is successfully parsed, use the RPC type specific service config.
+		rpcType := sharedtypes.RPCType(rpcTypeInt)
+		if rpcTypeServiceConfig, ok := supplierConfig.RPCTypeServiceConfigs[rpcType]; ok {
+			return rpcTypeServiceConfig, fmt.Sprintf("%s_SERVICE_CONFIG", rpcType.String()), nil
+		}
+	}
+
```
Olshansk, line 429: Please add an error line here with some details about the request (service-id, rpc-type, etc...) explaining we need to default.

It'll be crucial in debugging in the future.

pkg/relayer/config/types.go
```diff
@@ -66,13 +70,14 @@ type YAMLRelayMinerMetricsConfig struct {
...
-	ListenUrl             string                              `yaml:"listen_url"`
-	ServiceConfig         YAMLRelayMinerSupplierServiceConfig `yaml:"service_config"`
-	ServiceId             string                              `yaml:"service_id"`
-	SigningKeyNames       []string                            `yaml:"signing_key_names"`
-	RequestTimeoutSeconds uint64                              `yaml:"request_timeout_seconds"`
-	MaxBodySize           string                              `yaml:"max_body_size"`
-	XForwardedHostLookup  bool                                `yaml:"x_forwarded_host_lookup"`
+	ListenUrl             string                                         `yaml:"listen_url"`
+	ServiceConfig         YAMLRelayMinerSupplierServiceConfig            `yaml:"service_config"`
+	RPCTypeServiceConfigs map[string]YAMLRelayMinerSupplierServiceConfig `yaml:"rpc_type_service_configs"`
```
Olshansk, line 75: Please update the docs with the new configuration: https://dev.poktroll.com/operate/configs/relayminer_config?_highlight=relayminer

pkg/relayer/config/types.go
```diff
@@ -66,13 +70,14 @@ type YAMLRelayMinerMetricsConfig struct {
...
 type YAMLRelayMinerSupplierConfig struct {
-	ListenUrl             string                              `yaml:"listen_url"`
-	ServiceConfig         YAMLRelayMinerSupplierServiceConfig `yaml:"service_config"`
-	ServiceId             string                              `yaml:"service_id"`
-	SigningKeyNames       []string                            `yaml:"signing_key_names"`
-	RequestTimeoutSeconds uint64                              `yaml:"request_timeout_seconds"`
-	MaxBodySize           string                              `yaml:"max_body_size"`
-	XForwardedHostLookup  bool                                `yaml:"x_forwarded_host_lookup"`
+	ListenUrl             string                                         `yaml:"listen_url"`
+	ServiceConfig         YAMLRelayMinerSupplierServiceConfig            `yaml:"service_config"`
```
Olshansk, line 74: Please update the docs explaining this will be the default config.

Add a TODO_TECHDEBT (in the code AND in the docs as an admonition) to rename it to DefaultService Config.

pkg/relayer/config/types.go
```diff
@@ -163,15 +168,31 @@ type RelayMinerMetricsConfig struct {
...
+
 	// ServerType is the transport protocol used by the supplier, it must match the
 	// type of the relay miner server it is associated with.
-
 	ServerType RelayMinerServerType
-	// ServiceConfig is the config of the service that relays will be proxied to.
-	// Other supplier types may embed other fields in the future. eg. "https" may
-	// embed a TLS config.
+
+	// [REQUIRED] ServiceConfig is the default config of the service that relays will be
```
Olshansk, line 176: Per my comment above, have the TODO here and keep the optional/required details to the docs.

pkg/relayer/proxy/async.go
```diff
@@ -8,6 +8,7 @@ import (
 
 	"github.com/pokt-network/poktroll/pkg/polylog"
 	proxyws "github.com/pokt-network/poktroll/pkg/relayer/proxy/websockets"
+	"github.com/pokt-network/poktroll/x/shared/types"
```
Olshansk, line 11: Please run `make go_lint`

pkg/relayer/proxy/async.go
```diff
@@ -22,6 +23,26 @@ func (server *relayMinerHTTPServer) handleAsyncConnection(
 	serviceId := request.Header.Get("Target-Service-Id")
 	appAddress := request.Header.Get("App-Address")
 
+	// Determine the supplier's service configuration.
+	supplierConfig, ok := server.serverConfig.SupplierConfigsMap[serviceId]
+	if !ok {
+		return ErrRelayerProxyServiceEndpointNotHandled
```
Olshansk, line 29: Please add some logs (here and below) on errors.

pkg/relayer/proxy/sync.go
```diff
@@ -375,6 +389,48 @@ func (server *relayMinerHTTPServer) serveSyncRequest(
...
+	request *http.Request,
+) (
+	serviceConfig *config.RelayMinerSupplierServiceConfig,
+	serviceConfigTypeLog string,
+	err error,
+) {
+	// If the following are true:
+	// 	- The RPC type is set for the service
+	// 	- The RPC type specific service config is available
+	// Then, use the RPC type specific service config.
```
Olshansk, line 409:
```suggestion
	// If the following are true:
	// 	- The RPC-type is set for the service
	// 	- The RPC-type service-specific config is available
	// Then, use the RPC-type service-specific config.
	// Otherwise, use the default service config.
```

pkg/relayer/proxy/sync.go
```diff
@@ -375,6 +389,48 @@ func (server *relayMinerHTTPServer) serveSyncRequest(
 	return relayRequest, nil
 }
 
+// serviceConfigTypeDefault indicates that the service config being used is
+// the default service config, as opposed to an RPC-type specific config.
+const serviceConfigTypeDefault = "DEFAULT_SERVICE_CONFIG"
```
Olshansk, line 394: We're ALWAYS returning the default in the function below.

Is this necessary? Not really sure of the intention anymore.

pkg/relayer/proxy/http_server.go
```diff
@@ -141,40 +147,63 @@ func (server *relayMinerHTTPServer) Stop(ctx context.Context) error {
...
-				"failed to ping backend %q for serviceId %q: received status code %d",
-				backendUrl.String(), supplierCfg.ServiceId, resp.StatusCode,
-			)
+	}
+
+	return nil
+}
+
+// testBackendURL tests the connectivity of a backend URL for a given service ID.
+func (server *relayMinerHTTPServer) testBackendURL(backendUrl *url.URL, serviceId string) error {
```
Olshansk, line 173: s/test/ping?

pkg/relayer/proxy/http_server.go
```diff
@@ -141,40 +147,63 @@ func (server *relayMinerHTTPServer) Stop(ctx context.Context) error {
...
+	// Initialize the HTTP client for the backend URL.
+	c := &http.Client{Timeout: 2 * time.Second}
```
Olshansk, line 175: move into a local constant

pkg/relayer/proxy/http_server.go
```diff
@@ -141,40 +147,63 @@ func (server *relayMinerHTTPServer) Stop(ctx context.Context) error {
...
+			backendUrl.String(),
+			backendUrl.Scheme,
+		)
+
+		if backendUrl.Scheme == "ws" {
+			backendUrl.Scheme = "http"
+		} else {
+			backendUrl.Scheme = "https"
 		}
+	}
```
Olshansk, line 191: 1. Can this be moved into a helper?
2. Could/should this logic be reused elsewhere in the codebase?

pkg/relayer/config/supplier_hydrator.go
```diff
@@ -18,40 +23,116 @@ func (supplierConfig *RelayMinerSupplierConfig) HydrateSupplier(
...
 
-	backendUrl, err := url.Parse(yamlSupplierConfig.ServiceConfig.BackendUrl)
+	// Hydrate the default service config
+	defaultServiceConfig, err := supplierConfig.hydrateServiceConfig(yamlSupplierConfig.ServiceConfig)
+	if err != nil {
+		return err
+	}
+	supplierConfig.ServiceConfig = defaultServiceConfig
+
+	// Hydrate the RPC type-specific service configs (if any)
```
Olshansk, line 33:
```suggestion
	// Hydrate the RPC-type service-specific configs (if any)
```

pkg/relayer/config/supplier_hydrator.go
```diff
@@ -18,40 +23,116 @@ func (supplierConfig *RelayMinerSupplierConfig) HydrateSupplier(
...
+	supplierConfig.RPCTypeServiceConfigs = make(
+		map[types.RPCType]*RelayMinerSupplierServiceConfig,
+		len(yamlSupplierConfig.RPCTypeServiceConfigs),
+	)
+
+	// Loop through the RPC type-specific service configs and hydrate them.
```
Olshansk, line 39: Please use `RPC-type service-specificconfigs` here and below

pkg/relayer/config/supplier_hydrator.go
```diff
@@ -18,40 +23,116 @@ func (supplierConfig *RelayMinerSupplierConfig) HydrateSupplier(
...
+		if err != nil {
+			return err
+		}
+
+		rpcTypeServiceConfig, err := supplierConfig.hydrateServiceConfig(serviceConfig)
+		if err != nil {
+			return err
+		}
+
+		supplierConfig.RPCTypeServiceConfigs[rpcType] = rpcTypeServiceConfig
```
Olshansk, line 53: Please add some debug logging on what is being configured when the RelayMiner starts.

Make it useful. Have emojis.

Everyone in the ecosystem looks at them@

pkg/relayer/config/supplier_hydrator.go
```diff
@@ -18,40 +23,116 @@ func (supplierConfig *RelayMinerSupplierConfig) HydrateSupplier(
...
+	}
+
+	return nil
+}
+
+// getRPCTypeFromConfig converts the string RPC type to the
+// types.RPCType enum and performs validation.
+//
+// eg. "rest" -> types.RPCType_REST
```
Olshansk, line 62: Move this function and the one below closer to the `.pb.go` files, where they are compiled and defined.

Olshansk (changes requested):
Posted a fair number of changes, but we're close!

Should be g2g after this round.

pkg/relayer/config/types.go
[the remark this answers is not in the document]
red-0ne: Leaving it as `ServiceConfig` is fine to me. Maybe just mentioning in a comment that it's for backward compatibility.

pkg/relayer/proxy/sync.go
commoddity: The default is not always returned. If the request uses the RPC-type service-specific config then it returns:
` fmt.Sprintf("%s_SERVICE_CONFIG", rpcType.String())`

I've moved this to its own var with a comment to make it clearer.

pkg/relayer/proxy/http_server.go
commoddity: 1. Done
2. Not as far as I know; I don't believe we use this anywhere else. A search of the codebase for `== "wss"` & `== "ws"` confirms this.

pkg/relayer/config/supplier_hydrator.go
commoddity: Done, I've added better logging.

pkg/relayer/proxy/sync.go
commoddity: This is not technically an error. It just indicates either that the `Rpc-Type` header was provided with an invalid value (should not happen as this is handled by PATH) **or** the service is not configured with RPC type service configs.

I've added logs to explain all of these cases, including emojis.

Olshansk (approved):
@commoddity I squashed and merged my review here: https://github.com/pokt-network/poktroll/pull/1669

LGTM otherwise. Let's merge it in!

Issue #1642, opened 2025-07-04
[Critical] Prevent Breaking POKTscan & Pocketdex by Removing `settlement_result` Without Transition Plan


Alann27:
## Objective

Request additional time before the removal of the `settlement_result` from the `EventClaimSettled` event, so we can find an alternative method to retrieve the required data.

---

## Origin Document

If the upgrade that [removes the `settlement_result`](https://github.com/pokt-network/poktroll/pull/1606) from `EventClaimSettled` is released **without allowing us (POKTscan) time to adapt**, it will break both **Pocketdex** and **POKTscan**.

We currently rely on the `settlement_result`—specifically the `mod_to_acct_transfer` field—to populate the following dashboards:

- [Operator Staking](https://poktscan.com/tools/operator)
- [Owner Staking](https://poktscan.com/tools/staking)

<p float="left">
  <img width="100%" alt="Operator Staking Dashboard" src="https://github.com/user-attachments/assets/7ab1c127-db46-4b84-a1cc-0b8e76baa109" />
  <img width="100%" alt="Owner Staking Dashboard" src="https://github.com/user-attachments/assets/fd182072-3f22-4763-82cf-afc633536c24" />
</p>

We also planned to use the `mint` and `burn` values from the `settlement_result` to populate and update [POKT.money](https://pokt.money).

You can see how we're currently indexing `settlement_result` in Pocketdex:

- [Line 339](https://github.com/pokt-network/pocketdex/blob/main/src/mappings/pocket/relays.ts#L339)
- [Line 893](https://github.com/pokt-network/pocketdex/blob/main/src/mappings/pocket/relays.ts#L893)

---

We kindly request more time to determine a reliable alternative for accessing this data and to implement the necessary changes without breaking critical infrastructure.

**Creator**: @Alann27  
**Co-Owners**: @jorgecuesta

Olshansk:
@jorgecuesta Sorry for the lack of communication here. We'll definitely do a better job on it in the future.

However, we need to deploy this to MainNet as soon as Monday to ensure that block size doesn't get out of hand.

Two questions to get started:
1. How much time do you need? I'll need to broadcast a message to all exchanges that the release will be delayed.
2. Do you have a plan to access this data through other means?

<img width="1726" height="251" alt="Image" src="https://github.com/user-attachments/assets/68b31bcf-2f1e-43d2-969c-798145e7e6ef" />

jorgecuesta:
@Olshansk 

[...]

PR #1606, opened 2025-06-30
feat: Remove settlement result from EventClaimSettled

red-0ne:
## Summary

Remove the `settlement_result` field from `EventClaimSettled` to reduce blockchain event bloat and improve performance.

<details>

<summary> :point_left:  Example of a `settlement_result` for a single claim</summary>

```json
{
  "claim": {
    "supplier_operator_address": "pokt1096tssdy6h2kupt8ps36zdnfse8tnypxlqkgjp",
    "session_header": {
      "application_address": "pokt1vvwzl7lru3w4v4mx7rsljttky5yej2x872naeg",
      "service_id": "near",
      "session_id": "0ab5b6ed62ca2a2d0d49b9ffd174725273e4eddb0c3cfb0baaef2f78a8e08793",
      "session_start_block_height": "161051",
      "session_end_block_height": "161100"
    },
    "root_hash": "CbnGvQhtSEeeOssL/NaNWLpvlDZi6gWwOTdJKTbPv0sAAAAAAAGJNAAAAAAAAAAU",
    "proof_validation_status": "VALIDATED"
  },
  "mints": [
    {
      "op_reason": "TLM_RELAY_BURN_EQUALS_MINT_SUPPLIER_STAKE_MINT",
      "DestinationModule": "supplier",
      "coin": { "denom": "upokt", "amount": "3370" }
    },
    {
      "op_reason": "TLM_GLOBAL_MINT_INFLATION",
      "DestinationModule": "tokenomics",
      "coin": { "denom": "upokt", "amount": "1" }
    }
  ],
  "burns": [
    {
      "op_reason": "TLM_RELAY_BURN_EQUALS_MINT_APPLICATION_STAKE_BURN",
      "DestinationModule": "application",
      "coin": { "denom": "upokt", "amount": "3370" }
    }
  ],
  "mod_to_mod_transfers": [
    {
      "op_reason": "TLM_GLOBAL_MINT_REIMBURSEMENT_REQUEST_ESCROW_MODULE_TRANSFER",
      "SenderModule": "tokenomics",
      "RecipientModule": "supplier",
      "coin": { "denom": "upokt", "amount": "0" }
    },
    {
      "op_reason": "TLM_GLOBAL_MINT_REIMBURSEMENT_REQUEST_ESCROW_MODULE_TRANSFER",
      "SenderModule": "application",
      "RecipientModule": "tokenomics",
      "coin": { "denom": "upokt", "amount": "1" }
    }
  ],
  "mod_to_acct_transfers": [
[...]

PR #1643, opened 2025-07-05
Revert "feat: Remove settlement result from EventClaimSettled"

Olshansk:
Reverts pokt-network/poktroll#1606

See the discussino here: #1642

We'll revert it.

Replace the v0.1.26 release.

Deploy it.

Remove it prior to v0.1.27 and align on something different.


Olshansk:
@jorgecuesta I'll re-revert this revert after Monday's release.

After that, we can work on a long-term solution.

👍

Olshansk:
Unit tests passed.

See the comment here regarding why I'm merging with the failed E2E tests in CI: https://github.com/pokt-network/poktroll/pull/1646#issuecomment-3043248212

PR #1647, opened 2025-07-07
Remove settlement result from `EventClaimSettled` in `v0.127` release.

Olshansk:
Reverts pokt-network/poktroll#1643

This is a revert of a revert per the discussion here: #1642.


Olshansk:
Reverts pokt-network/poktroll#1643

This is a revert of a revert per the discussion here: #1642.

PR #1660, opened 2025-07-10
[WIP] Events to query block txs

Olshansk:
How to get the burn of an app
1. Query all blocks between [H1, H2]
2. Filter for all `EventClaimSettled` events
3. Look in `EventClaimSettled.Claim.SessionHeader.ApplicationAddress` for the address you're looking for


Olshansk:
_@fredteumer This is the way - regardless of other parametesr._

---

This is the `EventClaimSettled`, so no matter:
- How parameters change
- How many TLMs we enable/diaable
- Etc

This is the ultimate result for any `(supplier, application, service, session)`.

Here is a snippet from the code:

Screenshot 2025-07-10 at 3 49 43 PM

If you need a (verbose) reminder of the math, here is the link: https://dev.poktroll.com/protocol/tokenomics/intro_to_mint_and_burn#mathematic-representation
Screenshot 2025-07-10 at 3 51 04 PM

Olshansk:
@fredteumer Lmk if it makes sense now.

If the comments here are useful AND the scripts are helpful, lmk if you think we should add more links/details/notes to the docs.

PR #1728, opened 2025-08-12
[SessionRollover] Reproduce and Investigate Session Rollover locally

red-0ne:
## Summary

Sets up a local environment to reproduce and investigate **session rollover issues**.
The goal is to simulate mainnet-like conditions and observe system behavior under load.

## Test Initial Configuration

1. **Block duration:** 30 s (to mimic mainnet) we try to observe ~30 s RTT/latency.
2. **Blocks per session:** 10 (can be increased to 60 for closer mainnet simulation).
3. **Proof submissions:** Always submit proofs to put sustained load on **Claim & Proof** (C&P) processing.
4. **Test duration:** 3 sessions, ensuring ongoing `Relay` serving while C&P workflows are active.
5. **Application scaling:** No need to scale; keep the same application count. Adjust values to introduce occasional timeouts and stress the system:

   * **Initial value**
   * **Increment value**
   * **Max value**

## How to Run

1. Start localnet:

   ```bash
   make localnet_up
   ```
2. Run the stress test:

   ```bash
   make test_load_relays_stress_localnet_single_supplier
   ```


Olshansk:
@bryanchriswhite When you pick up the RelayMiner work, this might be a good PR to productionize to enable faster local load testing.

Issue #1816, opened 2025-10-03
[Morse Recovery] Ledger Recovery


ArtSabintsev:
We need to recover the balance of the following ledger address

`19e18bc48a1662c60e94939bbedabb7c988094b5`

PR #1818, opened 2025-10-04
Fixing issue 1815

Olshansk:
Fixing issue #1816

### DIFF
--- a/pkg/deps/config/suppliers.go
+++ b/pkg/deps/config/suppliers.go
@@ -7,6 +7,7 @@ import (
 	"net/url"
 
 	"cosmossdk.io/depinject"
+	cometclient "github.com/cometbft/cometbft/rpc/client"
 	sdkclient "github.com/cosmos/cosmos-sdk/client"
 	cosmosflags "github.com/cosmos/cosmos-sdk/client/flags"
 	"github.com/cosmos/cosmos-sdk/client/grpc/cmtservice"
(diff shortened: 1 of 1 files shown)


pkg/deps/config/tx_client.go
```diff
@@ -47,10 +47,12 @@ func GetTxClientFromFlags(
 	}
 
 	// Construct dependencies for the tx client
+	// IMPORTANT: NewSupplyCometClientFn must come before NewSupplyBlockClientFn
+	// because BlockClient depends on CometClient via EventsReplayClient
 	deps, err := SupplyConfig(ctx, cmd, []SupplierFn{
 		NewSupplyLoggerFromCtx(ctx),
-		NewSupplyBlockClientFn(queryNodeRPCUrl),
 		NewSupplyCometClientFn(queryNodeRPCUrl),
+		NewSupplyBlockClientFn(queryNodeRPCUrl),
```
red-0ne, line 55: Confirming that this is the only needed change to fix the dependency issue.

pkg/client/block/client.go
```diff
@@ -106,7 +106,7 @@ type blockReplayClient struct {
 	onStartQueryClient client.BlockQueryClient
 
 	// cometClient is the CometBFT client used to get ABCI info for chain version.
-	cometClient cometclient.Client
+	cometClient *cometclienthttp.HTTP
```
red-0ne, line 109: We should keep the dependency of `cometclient.Client` which is the interface that `cometclienthttp.HTTP` implements. This helps us to mock it properly

PR #1785, opened 2025-09-22
[RelayMiner] Add RelayMiner performance optimization and load testing infrastructure

red-0ne:
## Summary

Adds comprehensive performance improvements and load testing capabilities for the RelayMiner to support high-throughput relay processing and benchmarking.

### Primary Changes:
- Implemented instruction-level timing metrics in `RelayMiner` proxy for granular performance analysis
- Added static nginx server and wrk2 infrastructure for baseline performance testing
- Created relay load testing tool that generates proper `RelayRequest` data for realistic benchmarks
- Increased observable channel buffer size from 500 to 10,000 to prevent blocking under load spikes
- Extended consensus timeouts to 30s for more stable load testing sessions

### Secondary changes:
- Added new service configuration and supplier setup for static response testing
- Updated Tiltfile configuration with nginx static server and wrk2 testing tools
- Added make targets for baseline and relay load testing with configurable parameters

## Issue:

Performance optimization and load testing infrastructure for RelayMiner throughput analysis

## Type of change

Select one or more from the following:

- [x] New feature, functionality or library
- [ ] Bug fix
- [x] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Sanity Checklist

- [ ] I have updated the GitHub Issue Metadata: `assignees`, `reviewers`, `labels`, `project`, `iteration` and `milestone`
- [ ] For docs: `make docusaurus_start`
- [ ] For small changes: `make go_develop_and_test` and `make test_e2e`
- [ ] For major changes: `devnet-test-e2e` label to run E2E tests in CI
- [ ] For migration changes: `make test_e2e_oneshot`
- [ ] 'TODO's, configurations and other docs

🤖 Generated with [Claude Code](https://claude.ai/code)

PR #1819, opened 2025-10-04
RM: mining supervisor, per-service eager validation, xsync fast paths

jorgecuesta:
This branch is build on top of https://github.com/pokt-network/poktroll/pull/1785 so should be review/merge only after that one is on.

## Summary

Introduce an in-process **mining supervisor** to run delayed validations after writing the client response, adopt **per-service eager validation** (with a global default that services can override), and switch hot maps/queues to **xsync** for lower latency and contention.

### Primary Changes

- **Mining Supervisor (new)**
  - Bounded queue + worker pool; non-blocking enqueue (or short timeout) with optional **drop-oldest**.
  - Non-blocking downstream send; rate-limited warnings when dropping under backpressure.
  - Session fast-path cache (`xsync.Map[string]*SessionCache`) tracking `{EndHeight, Rewardable}`.
  - Instruction timings for enqueue/evict/timeout wait, processing, session upsert, downstream send.

- **Per-service eager validation**
  - Handler reads the **service config first** and decides:
    - If `enable_eager_relay_request_validation` is **true** for that service: do rate-limit snapshot + signature/session verification **before** the backend call.
    - If **false**: write the response fast, then publish to the mining supervisor for delayed validation + accounting.
  - A **global default** exists and services may **override** it.

- **Hot path concurrency via xsync**
  - Replace RWMutex maps with `xsync.Map`, and fan-in buffering with `xsync.MPMCQueue` where appropriate.

### Secondary Changes

- **Lightweight performance tracker**
  - Per-request tracker with `Start/Finish/Span/Flush`, nested spans, buffered emission, and a `service_id` label for instruction timings.

## Issue

- The latency added before and after the backend call was spiking the response times of the RM.

## Type of change

- [ ] Bug fix  
- [x] Code health or cleanup  
- [x] Documentation  
- [ ] Other (specify)

## Sanity Checklist

- [x] I have updated the GitHub Issue Metadata: `assignees`, `reviewers`, `labels`, `project`, `iteration` and `milestone`
- [ ] For major changes: add `devnet-test-e2e` label to run E2E tests in CI

## Configuration (YAML)

```yaml
# Global default; services can override per entry
default_enable_eager_relay_request_validation: true

mining_supervisor:
  queue_size: 2000
  workers: 8
  enqueue_timeout_ms: 50
  drop_policy: "drop-oldest"          # or "drop-new"
  gauge_sample_interval_ms: 100
  drop_log_interval_ms: 5

suppliers:
  - service_id: static
    listen_url: http://0.0.0.0:8545
    enable_eager_relay_request_validation: false   # per-service override
    service_config:
      backend_url: http://localhost:8548/
      forward_pocket_headers: false

  - service_id: anvil
    listen_url: http://0.0.0.0:8545
    # inherits global default (true in this example)
    service_config:
      backend_url: http://localhost:8547
      forward_pocket_headers: false
```

## Performance snapshot

In this snapshot can clearly observe how the latency spike a bit up on a 300rps, the added latency is not that much (~5ms) considering than this is doing eager validation (check eager-on screenshot). That spike will be solved with a PR from @red-0ne https://github.com/pokt-network/poktroll/pull/1810 which I personally already set on top of this one, getting as result the latency on spike on less that ~1ms.

eager-on

image

## Risk / Compatibility

- **Mining Supervisor** is a new, internal component behind configuration; no on-chain or external API changes.
- Per-service eager validation is config-driven (global default with service-level override).
- Metrics only add new series/labels; existing names remain.

---

### Enqueue & backpressure behavior
- **Default**: non-blocking enqueue; if full and `drop-oldest` is enabled, evict one and retry once.
- **Timeout path**: with `enqueue_timeout_ms > 0`, wait briefly before deciding to evict or drop.
- **Metrics**:  
  - `mining_queue_enqueued_total{service_id}`  
  - `mining_queue_dropped_total{service_id,reason}` with reasons like `full`, `evicted`, `timeout`, `timeout_full_after_evict`, `stopped_or_closed`, `context`
  - `mining_queue_len`
- **Logs**: rate-limited warning when dropping due to backpressure (reason, queue len/cap).

### Session cache semantics (fast path in handler)
- `xsync.Map[string]*SessionCache` with:
  - `EndHeight int64` (updated up, never down)
  - `Rewardable atomic.Bool` (starts true, can only be set to false)
- API:
  - `GetSessionEntry`  
  - `MarkSessionAsKnown(sessionID, endHeight)`  
  - `MarkSessionAsNonRewardable(sessionID)` (monotonic downgrade)
  - `PruneOutdatedKnownSessions` on new block events (endHeight+1 < head ⇒ delete)
- Effect: subsequent requests in the same session can skip unnecessary checks or be fast-rejected if already non-rewardable.

### Performance tracker (instruction timings)
- Per-request tracker on `context.Context` with buffered spans; minimal overhead.
- Labels: `instruction`, `service_id` (defaults to `unknown`, later upgradable).
- Typical handler usage:
  - `ctx, tracker, flush := relayer.EnsurePerfBuffered(ctx, serviceId)` (after service is known)  
  - `defer flush()`  
  - `defer relayer.Span(ctx, relayer.InstructionHTTPClientDo)()` around hot segments
- Supervisor uses the same tracker when the handler passes context into `Publish`.

---

## Handler workflow

```mermaid
flowchart TD
    A[HTTP request received] --> B[Parse and basic validate]
    B -->|error| E1[Return error]
    B --> C[Resolve service config]
    C --> D{Eager validation enabled for service}
    D -->|No| F[Check session cache and mark known if needed]
    D -->|Yes| E[Eager: rate limit snapshot and signature verify]
    E -->|fail| E1
    E -->|pass| F
    F --> G[Setup context deadline and write deadline]
    G --> H[Build backend request]
    H --> I[Call backend]
    I -->|error| E1
    I --> J[Serialize backend response]
    J --> K[Send response to client]
    K --> L{Eager enabled}
    L -->|No| M[Publish to mining supervisor]
    L -->|Yes| N[Final eligibility snapshot]
    N -->|eligible| O[Forward to miner for rewards]
    N -->|ineligible| P[Do not reward]
    M --> R[Return 200 OK]
    O --> R
    P --> R
```

## Mining Supervisor

```mermaid
flowchart TD
    S1[Publish relay] --> S2{Queue has space}
    S2 -->|Yes| S3[Enqueue and record metric]
    S2 -->|No| S4{Drop policy is drop-oldest}
    S4 -->|Yes| S5[Evict one item]
    S5 --> S6{Retry enqueue}
    S6 -->|Success| S3
    S6 -->|Fail| S7[Drop and record metric]
    S4 -->|No| S7

    subgraph Workers
        W1[Dequeue relay] --> W2[Over-service snapshot]
        W2 --> W3[Signature and session verify]
        W3 -->|invalid| W8[Mark non-rewardable and revert optimistic accounting]
        W3 -->|valid| W4[Update session cache]
        W4 --> W5{Rewardable}
        W5 -->|Yes| W6{Downstream channel available}
        W6 -->|Yes| W7[Send downstream]
        W6 -->|No| W9[Drop and mark downstream closed]
        W5 -->|No| W8
    end
```


jorgecuesta:
### ⚡ Benchmark: xsync.Map vs RWMutex for map[string]int64 under read-heavy load (with/without pruning)

This benchmark contrasts a classic map+RWMutex against xsync.Map[string,int64] for read-heavy and mixed loads, including a periodic “prune/touch all keys” pass to simulate a maintenance task.

<details>
  <summary>
    <strong>sessions_bench_test.go</strong>
  </summary>

```go
// sessions_bench_test.go
package demo

import (
    "context"
    "fmt"
    "math/rand"
    "runtime"
    "sync"
    "sync/atomic"
    "testing"
    "time"

    "github.com/puzpuzpuz/xsync/v4"
)

// --- configuration ---
const (
    keyCount   = 10_000 // number of session keys preloaded
    pruneEvery = 250 * time.Microsecond
    rngSalt    = 0x9e3779b97f4a7c15
)

var goroutineSeq uint64 // unique id per RunParallel worker

// ------------------------------------

func makeKeys(n int) []string {
    keys := make([]string, n)
    for i := 0; i < n; i++ {
        keys[i] = fmt.Sprintf("sess-%08d", i)
    }
    return keys
}

// ===== Implementation A: map + RWMutex =====

type rwMap struct {
    mu sync.RWMutex
    m  map[string]int64
}

func newRWMap(keys []string) *rwMap {
    m := make(map[string]int64, len(keys))
    for i, k := range keys {
        m[k] = int64(i)
    }
    return &rwMap{m: m}
}

func (s *rwMap) Load(k string) (int64, bool) {
    s.mu.RLock()
    v, ok := s.m[k]
    s.mu.RUnlock()
    return v, ok
}

func (s *rwMap) Store(k string, v int64) {
    s.mu.Lock()
    s.m[k] = v
    s.mu.Unlock()
}

// PruneTouchAll simulates a pruning pass by touching every key under write lock.
// This is intentionally write-locked to demonstrate reader blocking under contention.
func (s *rwMap) PruneTouchAll() {
    s.mu.Lock()
    for k, v := range s.m {
        s.m[k] = v + 1
        _ = k
    }
    s.mu.Unlock()
}

// ===== Implementation B: xsync.Map =====

type xMap struct {
    m *xsync.Map[string, int64]
}

func newXMap(keys []string) *xMap {
    m := xsync.NewMap[string, int64]()
    for i, k := range keys {
        m.Store(k, int64(i))
    }
    return &xMap{m: m}
}

func (s *xMap) Load(k string) (int64, bool) {
    return s.m.Load(k)
}

func (s *xMap) Store(k string, v int64) {
    s.m.Store(k, v)
}

// PruneTouchAll simulates a pruning pass by ranging without blocking readers.
func (s *xMap) PruneTouchAll() {
    s.m.Range(func(k string, v int64) bool {
        s.m.Store(k, v+1)
        return true
    })
}

// ===== shared harness =====

type kvStore interface {
    Load(string) (int64, bool)
    Store(string, int64)
    PruneTouchAll()
}

func runMixed(b *testing.B, store kvStore, keys []string, readPct int, withPrune bool) {
    ctx, cancel := context.WithCancel(context.Background())
    var wg sync.WaitGroup
    if withPrune {
        wg.Add(1)
        go func() {
            defer wg.Done()
            t := time.NewTicker(pruneEvery)
            defer t.Stop()
            for {
                select {
                case <-ctx.Done():
                    return
                case <-t.C:
                    store.PruneTouchAll()
                }
            }
        }()
    }

    b.ResetTimer()
    b.RunParallel(func(pb *testing.PB) {
        // unique per-goroutine seed
        id := atomic.AddUint64(&goroutineSeq, 1)
        seed := int64(uint64(time.Now().UnixNano()) ^
                (rngSalt * id) ^
                uint64(runtime.GOMAXPROCS(0)))
        r := rand.New(rand.NewSource(seed))

        n := len(keys)
        for pb.Next() {
            k := keys[r.Intn(n)]
            if r.Intn(100) < readPct {
                _, _ = store.Load(k)
            } else {
                store.Store(k, int64(r.Int31()))
            }
        }
    })

    cancel()
    wg.Wait()
    b.StopTimer()
}

// ===== Bench suites =====

func Benchmark_RWMutex_Read100(b *testing.B) {
    keys := makeKeys(keyCount)
    runMixed(b, newRWMap(keys), keys, 100, false)
}

func Benchmark_XSync_Read100(b *testing.B) {
    keys := makeKeys(keyCount)
    runMixed(b, newXMap(keys), keys, 100, false)
}

func Benchmark_RWMutex_Read99Write1(b *testing.B) {
    keys := makeKeys(keyCount)
    runMixed(b, newRWMap(keys), keys, 99, false)
}

func Benchmark_XSync_Read99Write1(b *testing.B) {
    keys := makeKeys(keyCount)
    runMixed(b, newXMap(keys), keys, 99, false)
}

func Benchmark_RWMutex_Read75Write25(b *testing.B) {
    keys := makeKeys(keyCount)
    runMixed(b, newRWMap(keys), keys, 75, false)
}

func Benchmark_XSync_Read75Write25(b *testing.B) {
    keys := makeKeys(keyCount)
    runMixed(b, newXMap(keys), keys, 75, false)
}

// --- With periodic "prune" (touch-all) to show reader stalls vs. xsync's range ---

func Benchmark_RWMutex_Read99Write1_Prune(b *testing.B) {
    keys := makeKeys(keyCount)
    runMixed(b, newRWMap(keys), keys, 99, true)
}

func Benchmark_XSync_Read99Write1_Prune(b *testing.B) {
    keys := makeKeys(keyCount)
    runMixed(b, newXMap(keys), keys, 99, true)
}
```
</details>

#### Observed on AMD Ryzen 9 5950X (Linux, Go bench -32):

```bash
Read-only:         RWMutex ~28.98 ns/op   vs   XSync ~1.946 ns/op   → ~14.9× faster
99% reads / 1% writes:
                   RWMutex ~61.91 ns/op   vs   XSync ~2.265 ns/op   → ~27.3× faster
75% reads / 25% writes:
                   RWMutex ~53.06 ns/op   vs   XSync ~7.491 ns/op   → ~7.1× faster
99/1 with prune:   RWMutex ~103.1 ns/op   vs   XSync ~2.442 ns/op   → ~42.2× faster
```

#### What this tells us

- xsync.Map removes global lock contention in read-heavy paths and during “touch-all” maintenance, which is exactly our session-cache and “known sessions” use case.
- The speedups hold across mixes; the biggest wins are when a periodic write/maintenance (prune) would otherwise block readers.

pkg/client/query/sessionquerier.go
```diff
@@ -3,7 +3,7 @@ package query
 import (
```
Olshansk: The changes in this file seem to be unsafe & incomplete.

Can you:
1. Delete unnecessary commented out things
2. Either add back the mutexes or use xsync
3. Explain (in the comments in the code, not the github response) why we don't need locks here?

pkg/relayer/metrics.go
```diff
@@ -26,6 +35,12 @@ const (
 	delayedRelayRequestRateLimitingCheckTotal  = "delayed_relay_request_rate_limiting_check_total"
 	blockHeightCurrent                         = "block_height_current"
 	instructionTimeSeconds                     = "instruction_time_seconds"
+
+	// TODO: implement metrics for the following:
```
Olshansk: Do we still need this TODO?

pkg/cache/memory/kvcache.go
```diff
@@ -43,102 +40,92 @@ func NewKeyValueCache[T any](opts ...KeyValueCacheOptionFn) (*keyValueCache[T],
...
 // Set adds or updates the value in the cache for the given key.
 func (c *keyValueCache[T]) Set(key string, value T) {
-	c.valuesMu.Lock()
-	defer c.valuesMu.Unlock()
+	c.values.Store(key, cacheValue[T]{value: value, cachedAt: time.Now()})
 
-	c.values[key] = cacheValue[T]{
-		value:    value,
-		cachedAt: time.Now(),
+	if c.config.maxKeys > 0 && int64(c.values.Size()) > c.config.maxKeys {
```
Olshansk: Why do we need to check for `maxKeys > 0`?

Not worried about the performance, just confused.
- Could (should?) we remove it?
- Do we need to #PUC?

pkg/cache/memory/kvcache.go
```diff
@@ -2,22 +2,19 @@ package memory
...
 
 // keyValueCache provides a concurrency-safe in-memory key/value cache implementation.
 type keyValueCache[T any] struct {
 	config keyValueCacheConfig
-
-	// valuesMu is used to protect values AND valueHistories from concurrent access.
-	valuesMu sync.RWMutex
 	// values holds the cached values.
-	values map[string]cacheValue[T]
+	values *xsync.Map[string, cacheValue[T]]
```
Olshansk: Could (should?) we create a TODO (or ticket) to move all map+mutex pair into xsync?

If so, can you create a ticket and leave a TODO somehwere in the code?

pkg/relayer/config/types.go
```diff
@@ -105,19 +109,33 @@ type YAMLRelayMinerPprofConfig struct {
 	Addr    string `yaml:"addr,omitempty"`
 }
 
+// YAMLMiningSupervisorConfig configures the in-process supervisor queue & workers.
+type YAMLMiningSupervisorConfig struct {
```
Olshansk: Please update the docs. This config isn't in there.

pkg/relayer/instruction_timer.go
```diff
@@ -1,75 +0,0 @@
-package relayer
```
Olshansk: Why did we delete this?

pkg/relayer/proxy/mining_supervisor.go
```diff
@@ -0,0 +1,445 @@
+package proxy
```
Olshansk: Can you rename the file to `relay_minin_supervisor.go`?

pkg/relayer/proxy/relay_meter.go
```diff
@@ -187,73 +220,59 @@ func (rmtr *ProxyRelayMeter) IsOverServicing(
...
 
 	// Create a context-specific logger to avoid concurrent access issues
 	logger := rmtr.logger.With(
 		"method", "SetNonApplicableRelayReward",
-		"session_id", reqMeta.GetSessionHeader().GetSessionId(),
+		"session_id", sessionId,
 	)
 
-	sessionRelayMeter, ok := rmtr.sessionToRelayMeterMap[reqMeta.GetSessionHeader().GetSessionId()]
+	// just fetch via xsync
```
Olshansk: Remove? I don't fully understand what this contributes.

pkg/relayer/proxy/mining_supervisor.go
```diff
@@ -0,0 +1,445 @@
...
+	cfg *config.MiningSupervisorConfig,
+	relayMeter relayer.RelayMeter,
+	relayAuthenticator relayer.RelayAuthenticator,
+) *RelayMiningSupervisor {
+	if downstream == nil {
+		// TECH_DEBT: should it panic?
+		panic("RelayMiningSupervisor: downstream channel must not be nil")
+	}
+	if cfg.QueueSize <= 0 {
+		cfg.QueueSize = 10_000
```
Olshansk: If we need "magic numbers" or "magic constants" (e.g. defaults), let's move it into a local constant.

Olshansk (changes requested):
@jorgecuesta  Note that I'm still not done.

Screenshot 2025-10-14 at 6 48 55 PM

---
Overall, it's looking good but I'll really get into the "meat of things" tomorrow.

---

Review info:
1. First review was "preliminary". 
2. Second review is me building context and providing NITs.
3. I anticipate 1 or 2 more reviews given the size, context and importance of this.

---

In terms of next steps, going to outline it to move fast.

1. Can you PTAL at #1837 and make any changes you need/want to it. 
2. Once #1837 looks good, please squash & merge it into this PR.
3. After (2), please look at the comments I have in this PR and either tend, close or comment.
4. In parallel to (1,2,3), I'll start looking at `pkg/relayer/performance_tracker.go` and `pkg/relayer/proxy/mining_supervisor.go` in more detail to understand it.
5. Once all of the above are done, I'll tap in @red-0ne to make sure his changes are not overrriden by the changes in this PR.

--

Would appreciate if you can priortize it ASAP to make sure it's part of the beta release, which we have a cutoff for this friday.

pkg/cache/memory/kvcache.go
```diff
@@ -43,102 +40,92 @@ func NewKeyValueCache[T any](opts ...KeyValueCacheOptionFn) (*keyValueCache[T],
...
 
 // Get retrieves the value from the cache with the given key.
 func (c *keyValueCache[T]) Get(key string) (T, bool) {
 	var zero T
-	c.valuesMu.RLock()
-	defer c.valuesMu.RUnlock()
-
-	cachedValue, exists := c.values[key]
-	if !exists {
+	v, ok := c.values.Load(key)
```
red-0ne:
```suggestion
	cachedValue, exists := c.values.Load(key)
```

pkg/cache/memory/kvcache.go
```diff
@@ -43,102 +40,92 @@ func NewKeyValueCache[T any](opts ...KeyValueCacheOptionFn) (*keyValueCache[T],
...
+	if !ok {
 		return zero, false
 	}
-
-	isCacheValueExpired := time.Since(cachedValue.cachedAt) > c.config.ttl
```
red-0ne: Any reason we're removing intermediate variables? I believe it's a net positive for readability

pkg/cache/memory/kvcache.go
```diff
@@ -43,102 +40,92 @@ func NewKeyValueCache[T any](opts ...KeyValueCacheOptionFn) (*keyValueCache[T],
...
+	c.values = xsync.NewMap[string, cacheValue[T]]()
 }
 
 // evictKey removes one key/value pair from the cache, to make space for a new one,
 // according to the configured eviction policy.
 func (c *keyValueCache[T]) evictKey() {
-	isMaxKeysConfigured := c.config.maxKeys > 0
-	cacheMaxKeysReached := int64(len(c.values)) > c.config.maxKeys
-	if !isMaxKeysConfigured || !cacheMaxKeysReached {
+	if c.config.maxKeys <= 0 || int64(c.values.Size()) <= c.config.maxKeys {
```
red-0ne: I prefer keeping the intermediate variables. I believe it's a net positive for readability.

pkg/cache/memory/kvcache.go
```diff
@@ -43,102 +40,92 @@ func NewKeyValueCache[T any](opts ...KeyValueCacheOptionFn) (*keyValueCache[T],
...
+		return
+	}
+
+	now := time.Now()
+
+	// 1) Prefer to evict any TTL-expired entry (cheap scan, remove one).
+	var expiredKey string
+	c.values.Range(func(k string, v cacheValue[T]) bool {
```
red-0ne: I think going TTL first might have a negative impact on eviction performance.

If no expired entry is found we would have scanned the entire cache for nothing.

This is especially true for update intensive scenarios.

pkg/client/query/sessionquerier.go
```diff
@@ -86,10 +86,6 @@ func (sessq *sessionQuerier) GetSession(
 		return session, nil
 	}
 
-	// Use mutex to prevent multiple concurrent cache updates
-	sessq.sessionsMutex.Lock()
```
red-0ne: This mechanism is necessary to prevent the **thundering herd** problem, which can degrade performance even more than the lock overhead itself by triggering multiple redundant queries.

In a highly concurrent scenario, we can easily encounter the following sequence:

```
p1: CheckCache->QuerySession------------------------->SessionFetched->PopulateCache
p2: -------------------------CheckCache->QuerySession
```

As a result, the full node is queried twice unnecessarily.
This occurs because, during the time it takes for `p1` to complete its query and populate the cache, `p2` still sees the cache as empty and issues its own query as well.

pkg/relayer/config/relayminer_configs_reader.go
```diff
@@ -19,6 +19,17 @@ var DefaultRequestTimeoutDuration time.Duration = time.Duration(DefaultRequestTi
 // DefaultMaxBodySize defines the default maximum HTTP body size as a string, used as a fallback if unspecified.
 const DefaultMaxBodySize = "20MB"
 
+const DefaultEagerRelayRequestValidation = false
+
+// --- Defaults of Mining Supervisor ---
+
+const DefaultMSQueueSize = uint64(10000)
```
red-0ne: What about replacing `MS` by `MiningSupervisor`?

pkg/relayer/proxy/relay_meter.go
```diff
@@ -187,73 +220,59 @@ func (rmtr *ProxyRelayMeter) IsOverServicing(
...
-	relayCost, err := getSingleRelayCostCoin(sharedParams, &service)
-	if err != nil {
-		logger.Warn().Msgf(
-			"[Non critical] Unable to calculate relay cost in session %s. Application may be rate limited more than intended: %v",
-			reqMeta.GetSessionHeader().GetSessionId(),
-			err,
-		)
-	}
+	// keep your existing lock for the math below
+	rmtr.relayMeterMu.Lock()
```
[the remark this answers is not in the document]
red-0ne: @Olshansk, this is just the lock that was at the top of the function.

Which means that now we're locking less code.

pkg/relayer/proxy/relay_meter.go
```diff
@@ -187,73 +220,59 @@ func (rmtr *ProxyRelayMeter) IsOverServicing(
...
 	// Create a context-specific logger to avoid concurrent access issues
 	logger := rmtr.logger.With(
 		"method", "SetNonApplicableRelayReward",
-		"session_id", reqMeta.GetSessionHeader().GetSessionId(),
+		"session_id", sessionId,
 	)
 
-	sessionRelayMeter, ok := rmtr.sessionToRelayMeterMap[reqMeta.GetSessionHeader().GetSessionId()]
+	// just fetch via xsync
+	sRelayMeter, ok := rmtr.sessionToRelayMeterMap.Load(sessionId)
```
red-0ne: Please keep the `sessionRelayMeter` name.

pkg/relayer/proxy/relay_meter.go
```diff
@@ -69,12 +70,17 @@ type ProxyRelayMeter struct {
 	blockQuerier       client.BlockClient
 
 	logger polylog.Logger
+
+	// Per-session relay cost memoization:
```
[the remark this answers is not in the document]
red-0ne: Why not embed the cost within the `sessionRelayMeter` struct? I think this will greatly reduce complexity.

pkg/relayer/proxy/relay_meter.go
```diff
@@ -270,105 +289,87 @@ func (rmtr *ProxyRelayMeter) AllowOverServicing() bool {
...
+		consumedCoin:  cosmostypes.NewInt64Coin(pocket.DenomuPOKT, 0),
+		maxCoin:       supplierAppStake,
+		sessionHeader: reqMeta.SessionHeader,
+	}
 
-		rmtr.relayMeterMu.Lock()
-		rmtr.sessionToRelayMeterMap[sessionId] = relayMeter
-		rmtr.relayMeterMu.Unlock()
+	// Try to publish; if someone beat us, reuse theirs.
+	if existing, loaded := rmtr.sessionToRelayMeterMap.LoadOrStore(sessionId, relayMeter); loaded {
```
red-0ne: `LoadOrStore` returns the newly created value if an entry does not exist.

Why not do the following?

```suggestion
	relayMeter = rmtr.sessionToRelayMeterMap.LoadOrStore(sessionId, relayMeter)
```

pkg/relayer/proxy/relay_meter.go
```diff
@@ -270,105 +289,87 @@ func (rmtr *ProxyRelayMeter) AllowOverServicing() bool {
...
-		// Application stake is guaranteed to be up-to-date as long as the cache is
-		// invalidated at each new block.
-		app, err := rmtr.applicationQuerier.GetApplication(ctx, appAddress)
-		if err != nil {
-			return nil, err
-		}
+	// Build a new entry (same logic you already have)
+	appAddress := reqMeta.GetSessionHeader().GetApplicationAddress()
 
-		// In order to prevent over-servicing, the protocol must split the application's stake
```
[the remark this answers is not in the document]
red-0ne: @Olshansk , all of the `relayMeter` logic is moved below and `xsync` is substituting to the locks.

localnet/pocketd/config/relayminer_config_localnet_vscode.yaml
```diff
@@ -25,6 +27,13 @@ pocket_node:
   query_node_rpc_url: tcp://localhost:26657
   query_node_grpc_url: tcp://localhost:9090
   tx_node_rpc_url: tcp://localhost:26657
+mining_supervisor:
```
red-0ne: Why are those values different from the defaults?

jorgecuesta:
> 1. Can you PTAL at #1837 and make any changes you need/want to it.

https://github.com/pokt-network/poktroll/pull/1837

PR #1834, opened 2025-10-14
[RelayMiner] Add WAL-based backup for in-memory session trees

red-0ne:
## Summary

Simplifies session tree storage architecture by replacing Pebble KVStore with an in-memory SimpleMap backed by a write-ahead log (WAL) for durability and crash recovery.

### Primary Changes:
- Implement write-ahead log (WAL) system for mined relays persistence in `mined_relays_persistence.go`
- Replace Pebble KVStore with SimpleMap + WAL architecture in session trees
- Add `Close()` method to `SessionTree` interface for proper resource cleanup

### Secondary changes:
- Remove storage mode complexity (SimpleMap, Pebble in-memory, Pebble disk variants)
- Update session manager to use unified WAL-based persistence approach
- Simplify session tree import/export logic by removing mode-specific handling

## Issue:

Addresses the need for crash recovery in session trees while maintaining in-memory performance benefits. The WAL ensures that mined relays are not lost during process restarts or crashes.

## Type of change

- [x] New feature, functionality or library
- [ ] Bug fix
- [x] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Sanity Checklist

- [x] I have updated the GitHub Issue Metadata: `assignees`, `reviewers`, `labels`, `project`, `iteration` and `milestone`
- [ ] For docs: `make docusaurus_start`
- [ ] For small changes: `make go_develop_and_test` and `make test_e2e`
- [ ] For major changes: `devnet-test-e2e` label to run E2E tests in CI
- [ ] For migration changes: `make test_e2e_oneshot`
- [ ] 'TODO's, configurations and other docs

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1838, opened 2025-10-15
[Review] Review of #1834: in-memory session tree WAL

Olshansk:
Review of #1834

### DIFF
--- a/pkg/relayer/session/mined_relays_persistence.go
+++ b/pkg/relayer/session/mined_relays_persistence.go
@@ -1,12 +1,16 @@
-// Context:
+package session
+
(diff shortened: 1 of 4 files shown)


pkg/relayer/session/mined_relays_persistence.go
```diff
@@ -191,6 +223,13 @@ func (wal *minedRelaysWriteAheadLog) flushBufferToDisk() error {
 		return err
 	}
 
+	// Sync to disk to ensure durability (fsync). This is critical for crash recovery.
```
Olshansk: Note: this was added by AI

PR #1840, opened 2025-10-16
[pocketd] Add flags to filter onchain suppliers via `--owner` and `--operator`

Olshansk:
Add flags to `pocketd` to filter onchain suppliers via `--owner` and `--operator`.

Screenshot 2025-10-16 at 12 53 01 PM

### DIFF
--- a/.github/workflows/release-artifacts.yml
+++ b/.github/workflows/release-artifacts.yml
@@ -24,9 +24,13 @@ permissions:
 jobs:
   build-artifacts:
     runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        cgo_variant: ['nocgo', 'cgo']
+
+    # TODO_INVESTIGATE: CGO matrix builds disabled - https://github.com/pokt-network/poktroll/discussions/1822
+    # To re-enable: Uncomment the strategy block below to build both CGO and non-CGO variants
+    # strategy:
+    #   matrix:
+    #     cgo_variant: ['nocgo', 'cgo']
+
     steps:
       - uses: actions/checkout@v4
         with:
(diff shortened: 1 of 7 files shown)


x/supplier/keeper/query_supplier.go
```diff
@@ -31,11 +32,23 @@ func (k Keeper) AllSuppliers(
...
+	hasOwnerAddress := req.GetOwnerAddress() != ""
+
+	// Strategy 1: Direct lookup by operator_address (most efficient)
+	if hasOperatorAddress && !hasServiceId && !hasOwnerAddress {
+		return k.getSupplierByOperator(ctx, logger, req)
+	}
+
+	// Strategy 2: Filter by service_id (using service index)
+	if hasServiceId {
 		return k.getAllServiceSuppliers(ctx, logger, req)
```
red-0ne: If `getSupplierByOperator` handles all queries that have `operator` as filter (as mentioned above).

This `getAllServiceSuppliers` could handle any query that have `service` or `service AND owner` as filters.

It would use the `service` index as an initial filter, and apply the `operator` check at each iteration.

Olshansk (commented):
@red-0ne I looked through your comments and don't fully understand the recommended changes.

I feel like what we have is what you're suggesting.

--- EDIT ---

Would you be able to put together "an optimization" PR if I were to just merge this in?

IMO it's already a net improvement over what we have.

Olshansk:
@red-0ne I'm going to merge this in as is and we can iterate later.

1. This is a net improvement over what we have
2. This does not impact onchain scalability (just heavy RPC calls by users)
3. Everything is 🟢 
4. If I'm missing an imporant optimization, can you open up a PR on main later>

PR #1845, opened 2025-10-17
[Infra] Turning off automatic E2E tests in CI

Olshansk:
- E2E test sin CI are costing $2K/month
- This turns off the automatic tests when the `devnet-test-e2e` label is added


fredteumer (approved):
approved. hope we can get this out ASAP this week

PR #1951, opened 2026-05-27
feat(app): add in-process file streamer for KV state archival

jorgecuesta:
## Summary

Opt-in in-process ABCI streaming listener that writes per-block protobuf files. NO-OP unless `[streaming.file] write-dir` is set in `app.toml`.

Workaround for the SDK's `[streaming.<svc>] plugin = "..."` path, which requires an external go-plugin binary that poktroll does not ship.

## Output

Per height H, written via `.partial` + atomic rename:

- `{prefix}block-{H}-meta` — varint-prefixed: `RequestFinalizeBlock`, `ResponseFinalizeBlock`, `ResponseCommit`
- `{prefix}block-{H}-data` — varint-prefixed: `StoreKVPair` per Set/Delete

## Activation

```toml
[streaming.file]
write-dir         = "/path"
keys              = ["*"]    # ["*"] = all stores; [] = none; or explicit list
prefix            = ""       # no path separators or ".."
fsync             = true
output-metadata   = true
stop-node-on-err  = true
```

Unknown store names in `keys` fail at startup with the full list of available stores.

The listener is appended to any pre-existing `StreamingManager` listeners; `stop-node-on-err` resolves to the strictest setting across listeners.

## Type of change

## Sanity Checklist

- [x] `go build ./cmd/pocketd` clean
- [x] `go test ./app/` (8 unit tests, all pass)
- [x] `go vet ./app/...` clean
- [x] `golangci-lint run ./app/...` 0 issues
- [x] Zero behavior change when feature disabled

### DIFF
--- a/app/streaming_file.go
+++ b/app/streaming_file.go
@@ -1,0 +1,404 @@
+package app
+
+// In-process ABCI streaming listener writing per-block protobuf files.
+// Workaround for the cosmos-sdk `[streaming.<svc>] plugin = "..."` path
+// requiring an external go-plugin binary that poktroll does not ship.
+//
+// Per-block (height H), files are written via .partial + atomic rename:
+//   {prefix}block-{H}-meta — varint-prefixed: RequestFinalizeBlock,
+//                            ResponseFinalizeBlock, ResponseCommit.
+//   {prefix}block-{H}-data — varint-prefixed: StoreKVPair per Set/Delete.
+//
+// The meta file is the completion marker: data is renamed to its final
+// name before meta is. A crash mid-block leaves at most a `.partial`
+// (and possibly a final data file with no meta), both of which downstream
+// consumers must treat as incomplete.
+//
+// No-op unless `[streaming.file] write-dir` is set in app.toml.
+
+import (
+	"context"
+	"encoding/binary"
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"sort"
+	"strings"
+	"sync"
+
+	storetypes "cosmossdk.io/store/types"
+	abci "github.com/cometbft/cometbft/abci/types"
+	"github.com/cosmos/cosmos-sdk/baseapp"
+	servertypes "github.com/cosmos/cosmos-sdk/server/types"
+	"github.com/cosmos/gogoproto/proto"
+	"github.com/spf13/cast"
+)
+
+const (
+	streamingFileWriteDir   = "streaming.file.write-dir"
+	streamingFileKeys       = "streaming.file.keys"
+	streamingFilePrefix     = "streaming.file.prefix"
+	streamingFileFsync      = "streaming.file.fsync"
+	streamingFileStopOnErr  = "streaming.file.stop-node-on-err"
+	streamingFileOutputMeta = "streaming.file.output-metadata"
+)
+
+type FileStreamer struct {
+	mu         sync.Mutex
+	writeDir   string
+	prefix     string
+	fsync      bool
+	outputMeta bool
+	stopOnErr  bool
+	// hasListeners indicates whether any KV store is being listened on. When
+	// false the data file is skipped entirely (avoids 0-byte inode churn in
+	// meta-only mode).
+	hasListeners bool
+
+	// Per-block state, set in ListenFinalizeBlock, consumed in ListenCommit.
+	// `prepared` gates the data write so a failed FinalizeBlock can't leave
+	// an orphan data file when stop-node-on-err is false.
+	curHeight          int64
+	curMetaFile        *os.File
+	curMetaPartialPath string
+	curMetaFinalPath   string
+	prepared           bool
+}
+
+// NewFileStreamer returns (nil, nil) when `streaming.file.write-dir` is unset.
+// `hasListeners` MUST be set by the caller via RegisterInProcessFileStreamer
+// based on whether any store keys are being exposed.
+func NewFileStreamer(appOpts servertypes.AppOptions) (*FileStreamer, error) {
+	writeDir := cast.ToString(appOpts.Get(streamingFileWriteDir))
+	if writeDir == "" {
+		return nil, nil
+	}
+	if err := os.MkdirAll(writeDir, 0o755); err != nil {
+		return nil, fmt.Errorf("streaming.file: cannot create write-dir %q: %w", writeDir, err)
+	}
+	prefix := cast.ToString(appOpts.Get(streamingFilePrefix))
+	if strings.ContainsAny(prefix, "/\\") || strings.Contains(prefix, "..") {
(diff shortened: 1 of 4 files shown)


oten91 (commented):
Reviewed the in-process file streamer end-to-end, tracing the wiring through SDK v0.53.7 `baseapp` and the `cosmossdk.io/store` listener path.

**Verdict: mergeable.** Consensus-safe (write-only listener, no keeper cache, no state mutation → zero AppHash impact; the change set is sorted by the store itself in `rootmulti/store.go` `PopStateCache`, so the unsorted `selectExposedKeys` is harmless). Disabled path is a true no-op. Atomic `.partial`+rename, prefix path-traversal guard, and error wrapping all look correct, and the 8 tests cover the happy paths well.

One expectation mismatch worth addressing before relying on this operationally (MED, inline on the `StopNodeOnErr` merge), one stale doc comment (MED, inline), plus a few durability nits (inline). Two findings that don't map to a single line:

- **`keys=[]` + `output-metadata=true` still writes a 0-byte `block-{H}-data` every block** (inode churn for what the doc calls "meta-only mode"). Consider skipping the data file when no stores are exposed.
- **Test gap:** error/cleanup paths (write/rename failure leaving no orphan final file) and the `RegisterInProcessFileStreamer` append + OR-merge logic are untested. Happy paths are well covered.

Details inline.

app/streaming_file.go
```diff
@@ -0,0 +1,306 @@
...
+	stopOnErr := true
+	if raw := appOpts.Get(streamingFileStopOnErr); raw != nil {
+		stopOnErr = cast.ToBool(raw)
+	}
+
+	// Merge: append our listener to any already-registered ones; resolve
+	// stop-node-on-err to the strictest setting (logical OR).
+	mgr := app.StreamingManager()
+	mgr.ABCIListeners = append(mgr.ABCIListeners, streamer)
+	mgr.StopNodeOnErr = mgr.StopNodeOnErr || stopOnErr
```
oten91: **MED — `stop-node-on-err` is a no-op for this listener in SDK v0.53.7.**

`baseapp/abci.go` invokes ABCI listeners at `:876-878` (FinalizeBlock) and `:962-964` (Commit) and only does `logger.Error(...)` on failure — it never reads `StreamingManager.StopNodeOnErr`. That field is consulted **only** by the go-plugin gRPC wrapper (`cosmossdk.io/store/streaming/abci/grpc.go:30,47`), which this PR bypasses by appending the raw `*FileStreamer` directly to `ABCIListeners`.

Consequence: the default `true` and this OR-merge do nothing. An archival I/O error (disk full, EIO) silently drops that block from the stream (gap in the archive) while the node keeps running. The PR description's claim that "`stop-node-on-err` resolves to the strictest setting across listeners" is misleading for the in-process path.

Fix options:
1. Wrap the listener to honor it yourself: capture `stopOnErr`, and in `ListenFinalizeBlock`/`ListenCommit` do `if err != nil && f.stopOnErr { panic(err) }` instead of returning, OR
2. Drop the option and document streaming as best-effort / non-halting so operators don't get false confidence.

app/streaming_file.go
```diff
@@ -0,0 +1,306 @@
...
+	if _, err = w.Write(hdr[:n]); err != nil {
+		return err
+	}
+	_, err = w.Write(buf)
+	return err
+}
+
+// RegisterInProcessFileStreamer wires the FileStreamer into the BaseApp.
+// `streaming.file.keys`: ["*"] exposes all stores; an explicit list exposes
+// only those (unknown names silently ignored); empty or unset = no stores
```
oten91: **MED — doc comment contradicts the code.** This says unknown keys are "silently ignored," but `selectExposedKeys` returns an error on unknown store names (which matches the PR body and `TestSelectExposedKeys`). Update the comment to say unknown names fail at startup.

app/streaming_file.go
```diff
@@ -0,0 +1,306 @@
...
+				f.cleanupMeta()
+				return fmt.Errorf("streaming.file: fsync meta h=%d: %w", f.curHeight, err)
+			}
+		}
+		if err := f.curMetaFile.Close(); err != nil {
+			_ = os.Remove(f.curMetaPartialPath)
+			f.resetMetaState()
+			return fmt.Errorf("streaming.file: close meta h=%d: %w", f.curHeight, err)
+		}
+		if err := os.Rename(f.curMetaPartialPath, f.curMetaFinalPath); err != nil {
```
oten91, line 261: **LOW — meta is renamed to its final name before the data file is written.** If the data write fails mid-`ListenCommit` (and per the `StopNodeOnErr` finding, that failure is now non-halting and merely logged), you're left with a final `block-{H}-meta` and no `block-{H}-data` — an inconsistent pair for downstream consumers. Consider writing+renaming the data file first and treating the meta rename as the completion marker, or rename both only after both are fsynced.

app/streaming_file.go
```diff
@@ -0,0 +1,306 @@
...
+			df.Close()
+			_ = os.Remove(dataPartial)
+			return fmt.Errorf("streaming.file: fsync data h=%d: %w", f.curHeight, err)
+		}
+	}
+	if err := df.Close(); err != nil {
+		_ = os.Remove(dataPartial)
+		return fmt.Errorf("streaming.file: close data h=%d: %w", f.curHeight, err)
+	}
+	if err := os.Rename(dataPartial, dataFinal); err != nil {
```
oten91: **LOW — no parent-directory fsync after rename.** With `fsync=true` the file contents are synced before close, but the directory entry created by `os.Rename` is not durable until the parent dir is fsynced. A crash right after the rename can lose the final filename even though `fsync` was requested. If the `fsync` flag is meant to imply durability, open the write-dir and `Sync()` it after the renames.

app/streaming_file.go
```diff
@@ -0,0 +1,306 @@
...
+	f.prepared = false
+
+	if !f.outputMeta {
+		f.prepared = true
+		return nil
+	}
+
+	finalPath := f.metaPath(req.Height)
+	partialPath := finalPath + ".partial"
+	mf, err := os.OpenFile(partialPath, os.O_CREATE|os.O_WRONLY|os.O_TRUNC, 0o644)
```
oten91, line 151: **LOW — orphan `*.partial` after a crash.** If the process dies between here (meta `.partial` opened/written in `ListenFinalizeBlock`) and the rename in `ListenCommit`, the `.partial` is never cleaned up on restart. Consider sweeping stale `{prefix}block-*.partial` in `NewFileStreamer` at startup.

oten91:
overall good, but let me know about these minor things

jorgecuesta:
> overall good, but let me know about these minor things

I will take all of them since I want this to be clean and strong enough to trust about the results we will consume for the indexer.

oten91 (approved):

